PR #31, opened 2020-01-15
register enums

candleshine:


candleshine:
This saves us some work later for having to register these enums on a per-model basis

PR #102, opened 2020-02-11
cord-api-41 implements create fileNode or Dir

gowthamrodda:


gowthamrodda:
As we have DTO's from @CarsonF  now, I think it needs to be rewritten.

gowthamrodda:
For now, able to upload to s3 and move uploaded file to original location. So creation is done without test cases. Will handle tests in smaller tickets.

src/components/file/files-s3-bucket.factory.ts
```diff
@@ -7,7 +7,12 @@ export const FilesBucketToken = Symbol('FilesBucket');
 export const FilesBucketFactory = {
   provide: FilesBucketToken,
   useFactory: (s3: S3, config: ConfigService) => {
-    return new S3Bucket(s3, config.files.bucket, {});
+    this.s3 = new S3({ apiVersion: '2006-03-01' });
+    return new S3Bucket(s3, config.files.bucket, {
+      accessKeyId: config.aws.accessKeyId,
+      secretAccessKey: config.aws.accessKeyId,
+      region: config.aws.region,
+    });
```
CarsonF: S3 configures automatically just like the `aws-cli` does. There's no need to explicitly pull in env vars.

gowthamrodda: ya, to my surprise its working.

PR #129, opened 2020-02-20
cord-api-115 : Refactor User service to use generic queries 

ibohalsinghkoijam:


candleshine:
generic create method verifies a user's token.  A user can't verify itself before it is created... too weird.  so I declare that the user create cannot use the abstract method

Issue #138, opened 2020-02-24
Typescript Strict Mode


CarsonF:
The following should be applied to `tsconfig.json` (`compilerOptions`):
```json
    "strict": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
```

After this any errors need to be fixed.

There were lots of errors with the previous code style, which is why we held off. The new code should be fine.
I'm hoping we are close enough now to move forward with this and maybe just fix the remaining errors.

richardgirges:
I'll give this one a go 👍

PR #149, opened 2020-02-25
TypeScript strict mode (PR #1)

richardgirges:
This is the first of three or four PRs related to #138 

I'm splitting the work up by `tsconfig.json` property. One PR for each new property introduced, so that it's clear which code changes are related to which `tsconfig.json` property. 

This PR revolves around the addition of `strict: true` to `tsconfig.json`. 

## Breaking Changes
* **[Breaking Change]**: Older APIs, such as the `internship.service`, will now return an error instead of returning a `null` value when an `update` operation fails.

* **[Breaking Change]**: Dev Environment
  * The entire PR is a breaking change for local dev environments if you have work-in-progress. Anyone in the middle of existing work may need to tweak some things if they suddenly see errors in their branch.
  * I would recommend that all developers take the following actions upon this PR being merged:
    * Merge the latest master branch into your local dev branch that you're working out of
    * Disable `yarn start:dev` if you have it running
    * Delete the `dist` directory (it sometimes caches unwanted types)
    * Restart the TypeScript server (VSCode users can do this via CMD + P menu)
    * Restart `yarn start:dev`

## Noteworthy Changes
Please read ALL of the below changes before merging!!

* The line `// @ts-ignore FIXME: unclear how to get around the TS error here` was added in four different parts of the codebase where I couldn't figure out how to resolve the type errors. Perhaps another engineer can take a crack at fixing those type errors.
* Added `strictPropertyInitialization: false` to the `tsconfig.json`. This brought our strict mode errors down from 600+ errors to ~100 errors. This seemed like a fair compromise, as we would otherwise have to instantiate default values for _every_ property in _every_ class that we've defined. Perhaps that's something we want to do at a later date?
* Added `skipLibCheck: true` to the `tsconfig.json`. This ensures that some node_module dependencies that we have no control over won't break our build when they contain internal TS errors
* Added `yarn type-check` command for CLI-based type-checking. This is the only command that is capable of checking _both_ e2e tests and the `src` dir at the same time.
  * **NOTE:** I would highly recommend adding this to our CI build process, as it will ensure no type errors are deployed before attempting to run a build.
* Added missing `@types/` dependency: `@types/triple-beam`
* Mocked an untyped dependency: `any-observable` in the new `typings.d.ts` file. This file will most likely grow as our list of dependencies grow
* Various `?` operators added throughout the code (mostly in tests) when referencing chained objects that have undefined/null properties on the chain
* A few areas of the codebase I've had to cast certain types using the `as` keyword. This was mostly when there was inferred type changes, i.e. from `string` type to `keyof` type. I believe it was safe to use the `as` keyword in those cases but happy to discuss further.


src/core/logger/formatters.ts
```diff
@@ -7,6 +7,13 @@ import { MESSAGE } from 'triple-beam';
 import { inspect } from 'util';
 import { format } from 'winston';
 
+interface Trace {
+  getFunctionName(): string;
+  getFileName(): string;
+  getLineNumber(): number;
+  getColumnNumber(): number;
+}
```
CarsonF: Pull types from library
```
yarn add -D @types/stack-trace
```

src/typings.d.ts
```diff
@@ -0,0 +1,2 @@
+// These are typings that do not have any type definitions and are therefore stubbed so that TS is happy
+declare module 'any-observable';
```
CarsonF: Bleh. This is going away in the next version of cypher-query-builder. I was hoping to avoid dealing with it. It's needed because of our `Transaction` class right?

src/core/logger/nest-logger-adapter.service.ts
```diff
@@ -54,8 +54,7 @@ export class NestLoggerAdapterService implements INestLogger, OnModuleInit {
   }
 
   private mapName(context?: string) {
-    return context
-      ? NestLoggerAdapterService.nameMap[context] || context
-      : 'nest';
+    // @ts-ignore FIXME: unclear how to get around the TS error here
+    return context ? NestLoggerAdapterService.nameMap[context] || context : 'nest';
```
CarsonF: I believe we just need to be explicit with the type definition of `nameMap` to loosen the keys. 

```tsx
private static nameMap: Record<string, string> = {
  NestFactory: 'nest:factory',
  ...
};
```

src/core/database/transaction.ts
```diff
@@ -136,8 +136,9 @@ export class Transaction implements QueryConnection {
    * Streaming is not supported on transactions
    * This observable will immediately emit an error.
    */
+  // @ts-ignore FIXME: unclear how to get around the TS error here
   stream<R = any>(query: Query): Observable<Dictionary<R>> {
```
CarsonF: Is this the `any-observable` problem?

src/core/database/property-updater.service.ts
```diff
@@ -65,7 +65,7 @@ export class PropertyUpdaterService {
     session: ISession;
     object: TObject;
     key: Key;
-    value: UnwrapSecured<TObject[Key]>;
+    value?: UnwrapSecured<TObject[Key]>;
```
CarsonF: I don't think this is right. If there's no value then you shouldn't be calling `updateProperty`. Thoughts?

src/components/organization/organization.service.ts
```diff
@@ -220,12 +220,12 @@ export class OrganizationService {
             count(org) as orgCount
           `,
           {
-            token: session.token,
+            token: session?.token,
```
CarsonF: I think `checkAllOrgs` should just require `session`. The db query doesn't make sense without it.

test/education.e2e-spec.ts
```diff
@@ -30,15 +30,15 @@ describe('Education e2e', () => {
...
+    expect(education?.id).toBeDefined();
   });
 
   it('read one education by id', async () => {
-    const education = await createEducation(app, { userId: user.id });
+    const education = await createEducation(app, { userId: user?.id });
 
     try {
-      const { education: actual } = await app.graphql.query(
+      const result = await app.graphql.query(
```
CarsonF: `result` should not be nullable...

src/common/luxon.neo4j.ts
```diff
@@ -31,6 +31,7 @@ DateTime.prototype.toNeo4JDateTime = function(this: DateTime) {
   );
 };
 
+// @ts-ignore FIXME: unclear how to get around the TS error here
```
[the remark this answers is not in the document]
richardgirges:
```
Element implicitly has an 'any' type because expression of type 'unique symbol' can't be used to index type 'DateTime'.
  Property '["util".inspect.custom]' does not exist on type 'DateTime'.
```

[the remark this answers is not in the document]
CarsonF: Ok TS is basically saying `DateTime` doesn't have that property. This is JS magic and doesn't change the external API, so I think it's fine to `as any` here.
```ts
(DateTime.prototype as any)[inspect.custom] = function() {};
```

src/common/secured-property.ts
```diff
@@ -19,7 +19,7 @@ export type UnwrapSecured<T> = T extends Secured<infer P> ? P : T;
 export const isSecured = <T>(value: T | Secured<T>): value is Secured<T> =>
   isObject(value);
 
-export const unwrapSecured = <T>(value: T | Secured<T>): T =>
+export const unwrapSecured = <T>(value: T | Secured<T>): T | undefined =>
```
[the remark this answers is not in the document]
richardgirges: Got it, it looks like the problem here is line 11 - 15:
```
export interface Secured<T> {
  readonly value?: T;
  readonly canRead: boolean;
  readonly canEdit: boolean;
}
```

`readonly value?: T` is implying that `value` can be undefined. How should we approach that? I can set `value` to not be undefinable but it will probably cause some downstream effects in the property-updater.service. Lemme know your thoughts

src/core/database/cypher.factory.ts
```diff
@@ -41,6 +41,7 @@ export const CypherFactory: FactoryProvider<Connection> = {
       const session: Session | null = origSession.call(conn);
       if (session) {
         const origRun = session.run;
+        // @ts-ignore FIXME: unclear how to get around the TS error here
```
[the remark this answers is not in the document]
CarsonF: Oh gag. This is going to be tricky. Let me think about this more...maybe we just continue to ignore it for now.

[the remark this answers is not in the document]
richardgirges: Sounds good!

src/core/database/transaction.ts
CarsonF: Ok yeah it is. Let's just replace the any-observable import with rxjs.
```diff
-import Observable from 'any-observable';
+import { Observable } from 'rxjs';
```
That should fix the need to explicitly define the `subscriber` parameter too.
And we won't need the `declare module` stub.

src/components/file/files-s3-bucket.factory.ts
```diff
@@ -7,7 +7,7 @@ export const FilesBucketToken = Symbol('FilesBucket');
 export const FilesBucketFactory = {
   provide: FilesBucketToken,
   useFactory: (s3: S3, config: ConfigService) => {
-    return new S3Bucket(s3, config.files.bucket, {});
+    return new S3Bucket(s3, config.files.bucket || '', {});
```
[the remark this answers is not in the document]
richardgirges: Got it! Changing now.

src/core/logger/formatters.ts
richardgirges: Got it! Fixed locally, will push up soon.

test/education.e2e-spec.ts
CarsonF: I mean there shouldn't be a need for the optional chaining on this variable

src/core/logger/nest-logger-adapter.service.ts
richardgirges: Perfect, works :)

src/core/database/transaction.ts
richardgirges: Yeah, `any-observable` gave me some grief here. I stubbed it in `typings.d.ts`, even exported it as type `any` at one point. We may need to define it further to eliminate the error being thrown here. It's complaining that `Observable` cannot be used as a type here.

src/components/organization/organization.service.ts
richardgirges: Makes perfect sense, updating now.

test/education.e2e-spec.ts
richardgirges: The real issue here was in `create-graphql-client.ts`, where the query is typed to return a Promise resolving to potentially `null` or `undefined` (see snippet below). 

Thoughts on changing this to always assume a payload is returned?

```
export interface GraphQLTestClient {
  query: (
    query: DocumentNode,
    variables?: { [name: string]: any },
  ) => Promise<Record<string, any> | null | undefined>;
  mutate: (
    mutation: DocumentNode,
    variables?: { [name: string]: any },
  ) => Promise<Record<string, any> | null | undefined>;
  authToken: string;
}
```

src/typings.d.ts
richardgirges: Yep that's the only place it's being used

src/common/secured-property.ts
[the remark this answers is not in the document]
CarsonF: Hmm ok yeah I see. Ok I'm fine leaving it as you've changed it. I think it does make the most sense in the context of the codebase. 👍

src/common/secured-property.ts
```diff
@@ -48,7 +48,7 @@ SecuredProperty.descriptionFor = (value: string) => stripIndent`
 @ObjectType({
   description: SecuredProperty.descriptionFor('a string'),
 })
-export abstract class SecuredString extends SecuredProperty<string>(
+export abstract class SecuredString extends SecuredProperty<string | null>(
```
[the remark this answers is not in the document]
richardgirges: Yep you're right, fixing now.

package.json
```diff
@@ -20,7 +20,8 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --forceExit --config ./test/jest-e2e.json"
+    "test:e2e": "jest --forceExit --config ./test/jest-e2e.json",
+    "type-check": "tsc"
```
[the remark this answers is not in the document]
richardgirges: Done!

src/core/logger/abstract-logger.ts
```diff
@@ -55,7 +55,7 @@ export abstract class AbstractLogger implements ILogger {
         ? { message: msgOrContext }
         : msgOrContext),
       ...(typeof levelOrEntry === 'string'
-        ? { level: levelOrEntry }
+        ? { level: levelOrEntry, message: '' }
```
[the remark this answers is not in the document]
richardgirges: Good eye! Fixing now.

test/education.e2e-spec.ts
CarsonF: Ah. Ok yeah let's remove the nullable output from that interface. The `return` statements in the implementation will just need a non-null assertion.

src/core/database/property-updater.service.ts
richardgirges: Going to look into this one after addressing https://github.com/SeedCompany/cord-api-v3/pull/149#discussion_r384056579

src/core/logger/nest-logger-adapter.service.ts
richardgirges: Perfect!

src/common/luxon.neo4j.ts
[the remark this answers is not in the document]
richardgirges: Great! Updated and pushed.

src/core/database/cypher.factory.ts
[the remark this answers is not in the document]
CarsonF: Ok figured a good step. Just need to return the result of `.call` instead of `.catch`.
```ts
const origRun = session.run;
session.run = function(this: never, origStatement, parameters, conf) {
  const statement = stripIndent(origStatement);
  logger.debug('\n' + statement, parameters);
  const result = origRun.call(session, statement, parameters, conf);
  result.catch(e => {
    throw jestSkipFileInExceptionSource(e, __filename);
  });
  return result;
};
```

Would you mind applying that?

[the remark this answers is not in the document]
CarsonF: Wait now I'm second guessing myself. Would that `catch` callback be applied if it's not returned? Does catch mutate the original promise or return a new one?

src/core/database/transaction.ts
CarsonF: Did you see my second comment?

test/education.e2e-spec.ts
richardgirges: Sounds good, will do

src/core/database/transaction.ts
richardgirges: Sorry about that, GH page was cached. Switching to rxjs worked!

src/common/secured-property.ts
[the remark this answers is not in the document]
richardgirges: Sounds good!

src/core/database/transaction.ts
CarsonF: No worries. Happens to me all the time. GitHub updates some messages but misses others. I refresh all the time out of paranoia lol

src/common/secured-property.ts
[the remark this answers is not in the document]
richardgirges: @CarsonF so ran into some odd issues here. The `internship-engagement.service` is setting a lot of securedstring values to null intentionally. i.e.:

```
        response.intern = {
          displayFirstName: result.records[0].get('displayFirstName'),
          displayLastName: result.records[0].get('displayLastName'),
          realFirstName: result.records[0].get('realFirstName'),
          realLastName: result.records[0].get('realLastName'),
          email: result.records[0].get('email'),
          id: result.records[0].get('userId'),
          createdAt: DateTime.local(),
          timezone: {
            value: null,
            canRead: true,
            canEdit: true,
          },
          phone: {
            value: null,
            canRead: true,
            canEdit: true,
          },
          bio: {
            value: null,
            canRead: true,
            canEdit: true,
          },
        };
```

easiest fix I can think of is instead of setting value to `null`, set it to `undefined`. At the moment, these are the errors I'm seeing on every `value: null` line:

```
(property) SecuredPropertyClass<string>.value?: string | undefined
Type 'null' is not assignable to type 'string | undefined'.ts(2322)
secured-property.ts(32, 14): The expected type comes from property 'value' which is declared here on type 'SecuredString'
```

src/core/database/cypher.factory.ts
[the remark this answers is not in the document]
richardgirges: Now that I'm thinking through it more, I believe `throw` inside of a `catch` will result in an uncaught exception. What if we `await` the result before returning it, capture the error with `try / catch` it then throw inside of the `try / catch` statement?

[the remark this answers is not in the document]
richardgirges: Actually I think this might be okay...testing now.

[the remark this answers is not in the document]
CarsonF: We can't use async await here because the run result is a promise with an additional subscribe method. It's a clever way to support both promise and observable usage. But wrapping it is more annoying.

We don't currently use the observable API so if push comes to shove I'm fine ignoring it. I would like the logic inside the catch to be applied for clarity with errors in jest output.

src/core/database/property-updater.service.ts
richardgirges: Ah okay so I'm running into an odd issue where doing an `== undefined` assertion doesn't seem to work, most likely because we're doing an `await` call on `this.updateProperty` so TS is assuming that since `await` runs on an asynchronous thread, the value of `changes[prop]` could have changed by the time the `await` is executed. This is my faulty code as of right now:

```
      if (
        changes[prop] == null || changes[prop] == undefined ||
        unwrapSecured(object[prop]) === changes[prop]
      ) {
        continue;
      }
```

but I'm still getting an error on the `value: changes[prop]` line:
```
Type 'UnwrapSecured<TObject[keyof TObject]> | undefined' is not assignable to type 'UnwrapSecured<TObject[keyof TObject]>'.
  Type 'undefined' is not assignable to type 'UnwrapSecured<TObject[keyof TObject]>'
```

src/core/database/cypher.factory.ts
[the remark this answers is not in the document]
richardgirges: I applied your update, intentionally broke a `session.run` query and tested it, and I'm not seeing any issues with the catches, so I think we should be okay! TS Errors are gone. :D

src/core/database/property-updater.service.ts
CarsonF: FYI `foo == null` is the same as `foo === null || foo === undefined`. So the second _or_ condition is redundant. (I know you were probably just trying stuff).

Ok so it's because `Secured['value']` is `T | undefined`. To fix you can wrap in `NonNullable` to strip the `undefined`.
```ts
NonNullable<UnwrapSecured<TObject[keyof TObject]>>
```

BUTTTT thinking through it more, I'm not so sure we should be requiring a value. If some properties are nullable then I think they should be able to be "reset" to `null` explicitly. Lemme talk to Michael and get his thoughts.

src/common/secured-property.ts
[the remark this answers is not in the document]
CarsonF: Those are stubbed responses? I would just change them to `value: undefined` for now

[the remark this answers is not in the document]
richardgirges: Yep, stubbed responses. Just pushed up the update

src/core/database/property-updater.service.ts
richardgirges: I completely blanked on the double equal operator, so used to just using `===` without thought LOL.

And sounds good! As of right now, there are no type errors and the tests are passing. I think this is the only outstanding issue, lemme know how to move forward after you talk to Michael.

CarsonF: Ok after talking on slack let's just leave this for now. We can validate not null is what we want in API validation

richardgirges:
Thanks @CarsonF ! I totally agree. The optional chaining syntax is repetitive and noisy in the tests. I switched to using definitive type assertions and cleaned out all of the optional chaining syntaxes. Let me know how it looks now! It seems a lot cleaner.

richardgirges:
Happy to be a part of it! 🤓

PR #154, opened 2020-02-28
cord-api-130

gowthamrodda:
this has a user attached to file with createdby relationship.

### DIFF
--- a/src/components/file/file.service.ts
+++ b/src/components/file/file.service.ts
@@ -2,7 +2,7 @@ import { BadRequestException, Inject, Injectable } from '@nestjs/common';
 import { Connection } from 'cypher-query-builder';
 import { generate } from 'shortid';
 import { NotImplementedError } from '../../common';
-import { ISession } from '../auth';
+import { ISession, AuthService } from '../auth';
 import {
   CreateFileInput,
   Directory,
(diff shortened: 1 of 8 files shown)


src/components/file/file.service.ts
```diff
@@ -103,25 +107,50 @@ export class FileService {
...
+      let user: User | null;
+      const userSession: ISession = await this.authService.decodeAndVerifyToken(session.token);
+      user = (userSession.userId) ? await this.userService.readOne(userSession.userId, session) : null;
+
       const result = await this.db
         .query()
         .raw(
           `
-        MATCH (token:Token {active: true, value: $token})
+        MATCH
```
candleshine: There is a create abstraction in property-updater service see education service for how it is used.

src/components/file/file.service.ts
```diff
@@ -103,25 +107,50 @@ export class FileService {
...
+        <-[:token {active: true}]-
+        (requestingUser:User {
+          active: true,
+          id: $requestingUserId
+        })
         CREATE
-            (file:FileNode { id: $id, type: $type, name: $name })
+          (file:FileNode {
+            id: $id,
+            type: $type,
```
candleshine: these values should not be properties on the node.  If you look at the database schema.jpg, you will see that these are property nodes.  The create function in property-updater service will do this for you.  you just have to pass in the key/values into the function

test/file.e2e-spec.ts
```diff
@@ -1,25 +1,75 @@
...
     app = await createTestApp();
   });
 
-  it('read file node by id', async () => {
+  afterAll(async () => {
+    await app.close();
+  });
+
+  it('Create FileNode', async () => {
     const token = await createSession(app);
```
candleshine: I re ordered these so Create happens before Read.  token is not used by this function

src/components/file/file.service.ts
```diff
@@ -103,25 +107,50 @@ export class FileService {
   ): Promise<File> {
     try {
       // TODO find a better way to check if object exists in s3 and move
+      let user: User | null;
+      const userSession: ISession = await this.authService.decodeAndVerifyToken(session.token);
+      user = (userSession.userId) ? await this.userService.readOne(userSession.userId, session) : null;
```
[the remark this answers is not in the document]
gowthamrodda: to pull user id from provided token

src/components/file/file.service.ts
gowthamrodda: done

PR #186, opened 2020-03-03
email check done.

aarontopstar:
======Query=========  
query{  
checkEmail(input:{  
email: "test@test.com"  
})  
}
==========Result========  
{  
"data": {  
"checkEmail": false  
}  
}
At this time, I checked email existence on session condition. I can change it whatever our API needs.


michaelmarshall:
Email address nodes are unique in our db. Because of this we only need to check the nodes themselves and not a path through any ACL or User node. So the query needs to be simpler than what you have. Something like `MATCH (email:Email {value: $email}) RETURN email.value`. You don't even need to check the token. Its just a simple lookup.

Issue #142, opened 2020-02-24
Switch all Enums to Pascal Case


richardgirges:
In order to avoid interoperability issues when executing server-side queries and mutations (this mostly happens in e2e tests), we need to switch all enums to have matching key and value. 

Action item: switch all enums to have the exact same key and value in pascal case (e.g. `BibleStories`). This will be a breaking change and should be documented as such in the forthcoming PR. 

See here for more information: https://typegraphql.ml/docs/enums.html#interoperability

michaelmarshall:
closing, old.

PR #202, opened 2020-03-07
Partnerships DTO, service, resolvers, e2e tests refactor

### LINKED ISSUE #140: Refactor partnerships
The Partnerships service needs to be caught up to current conventions:

* Resolvers need to be updated with the latest naming, input, and output conventions
* Service needs to be updated to utilize user sessions, ACLs, generics, logger, and eliminate the use of session DB driver
* Update DTOs to follow the latest conventions
* Switch to module-based approach for loading partnerships API
* Update e2e tests to follow all newly implemented conventions in the GQL API

## Dev Notes
(From meeting w/ Daniel)
  - in partnerships, pay close attention to organization
  -  organization needs to be a link as organization node
  - org id gets passed in, then translated into a node (lookup)
  - side note: organization link is also on the partnership and on the budget

richardgirges:
Closes #140 

## Breaking Changes
Basically all mutations and most queries for Partnerships have a different input/output footprint now.

## API Changes
* PartnershipModule
* Changed all Partnership-specific enum values to have the same key/value. This is a precursor update to #142 
* Refactored Partnership DTOs to follow the latest conventions
* Refactored Partnership Services to follow the latest conventions
* Refactored Partnership Resolvers to follow the latest conventions
* Various updates to the list generic, to properly handle enum/string arrays
* Added `canCreatePartnership: true, canReadPartnerships: true`, to user's ACL when creating a new user

## E2E Changes
The partnership.e2e tests are now re-enabled and follow the latest conventions (createPartnership utility, partnership fragment, etc)


src/components/partnership/dto/create-partnership.dto.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@InputType()
+export class CreatePartnership {
+  @Field(() => PartnershipAgreementStatus)
+  readonly agreementStatus: PartnershipAgreementStatus;
+
+  @Field(() => PartnershipAgreementStatus)
+  readonly mouStatus: PartnershipAgreementStatus;
+
+  @DateTimeField({ nullable: true })
```
CarsonF:
```suggestion
  @DateField({ nullable: true })
```

src/components/partnership/dto/create-partnership.dto.ts
```diff
@@ -0,0 +1,43 @@
...
+  readonly mouStart: DateTime | null;
+
+  @DateTimeField({ nullable: true })
```
CarsonF:
```suggestion
  @DateField({ nullable: true })
```

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+@ObjectType()
+export class Partnership extends Resource {
+  @Field()
+  readonly agreementStatus: SecuredPartnershipAgreementStatus;
+
+  @Field()
+  readonly mouStatus: SecuredPartnershipAgreementStatus;
+
+  @DateTimeField()
+  readonly mouStart: SecuredDateTime;
```
CarsonF:
```suggestion
  @Field()
  readonly mouStart: SecuredDate;
```
Since `SecuredDate` has the custom Field decorator we don't need it here.

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  @DateTimeField()
+  readonly mouEnd: SecuredDateTime;
```
CarsonF:
```suggestion
  @Field()
  readonly mouEnd: SecuredDate;
```

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  @Field(() => Organization, { nullable: true })
+  readonly organization: Organization | null;
```
CarsonF:
```suggestion
  @Field(() => Organization)
  readonly organization: Organization;
```

Organization is required to create a Partnership so it's not nullable.

src/components/partnership/dto/update-partnership.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+  readonly id: string;
+
+  @Field(() => PartnershipAgreementStatus, { nullable: true })
+  readonly agreementStatus?: PartnershipAgreementStatus;
+
+  @Field(() => PartnershipAgreementStatus, { nullable: true })
+  readonly mouStatus?: PartnershipAgreementStatus;
+
+  @DateTimeField({ nullable: true })
+  readonly mouStart?: DateTime;
```
CarsonF:
```suggestion
  @DateField({ nullable: true })
  readonly mouStart?: DateTime;
```

src/components/partnership/dto/update-partnership.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @DateTimeField({ nullable: true })
```
CarsonF:
```suggestion
  @DateField({ nullable: true })
```

src/components/partnership/dto/update-partnership.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+  readonly mouEnd?: DateTime;
+
+  @Field(() => ID, { nullable: true })
+  readonly organizationId?: string;
```
CarsonF:
```suggestion
```
Partnership's Organization cannot be changed. That's more like an entirely new partnership.

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Organization } from '../../organization';
+import { PartnershipAgreementStatus } from './partnership-agreement-status.enum';
+import { PartnershipType } from './partnership-type.enum';
+
+@ObjectType({
+  description: SecuredProperty.descriptionFor('a partnership agreement status'),
+})
+export abstract class SecuredPartnershipAgreementStatus extends SecuredProperty<
+  PartnershipAgreementStatus
+>(GraphQLString) {}
```
CarsonF: Does this output the schema with the correct value type?
i.e.:
```graphql
type SecuredPartnershipAgreementStatus {
  value: PartnershipAgreementStatus
  ...
}
```

src/components/partnership/dto/partnership-agreement-status.enum.ts
```diff
@@ -1,9 +1,9 @@
 import { registerEnumType } from 'type-graphql';
 
 export enum PartnershipAgreementStatus {
-  NotAttached = 'na',
-  AwaitingSignature = 'as',
-  Signed = 's',
+  NonAttached = 'NonAttached',
```
CarsonF:
```suggestion
  NotAttached = 'NotAttached',
```

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@ObjectType()
+export class Partnership extends Resource {
+  @Field()
+  readonly agreementStatus: SecuredPartnershipAgreementStatus;
+
+  @Field()
+  readonly mouStatus: SecuredPartnershipAgreementStatus;
```
CarsonF: Both of these should default to `NotAttached` on partnership create

src/app.module.ts
```diff
@@ -15,8 +15,7 @@ import { InternshipEngagementService } from './components/internship-engagement/
 import { LanguageModule } from './components/language';
 import { LocationModule } from './components/location';
 import { OrganizationModule } from './components/organization';
-import { PartnershipResolver } from './components/partnership/partnership.resolver';
-import { PartnershipService } from './components/partnership/partnership.service';
+import { PartnershipModule } from './components/partnership/partnership.module';
```
CarsonF:
```suggestion
import { PartnershipModule } from './components/partnership';
```

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly mouStart: SecuredDateTime;
+
+  @DateTimeField()
+  readonly mouEnd: SecuredDateTime;
+
+  @Field(() => Organization, { nullable: true })
+  readonly organization: Organization | null;
+
+  @Field(() => [PartnershipType])
+  readonly types: PartnershipType[];
```
CarsonF: I think this needs to be secured as well, so that the frontend can show it's readonly. This is a list but it can be treated as a "scalar/single" value for this case.

src/components/partnership/dto/partnership.dto.ts
richardgirges: 10-4. I'm now setting the default value of `NotAttached` in the partnership's service create method.

src/components/partnership/dto/partnership.dto.ts
richardgirges: I shamelessly lifted this approach from the Education service, and now upon further inspection I can see that it's returning a `value: String` in GQL. 

Here's my current approach:
```typescript
export abstract class SecuredPartnershipAgreementStatus extends SecuredProperty<
  PartnershipAgreementStatus
>(PartnershipAgreementStatus) {}
```

However, this is throwing an error: 
```
Argument of type 'typeof PartnershipAgreementStatus' is not assignable to parameter of type 'GraphQLScalarType | ClassType<PartnershipAgreementStatus> | AbstractClassType<PartnershipAgreementStatus>'.
  Type 'typeof PartnershipAgreementStatus' is not assignable to type 'AbstractClassType<PartnershipAgreementStatus>'.
    Property 'prototype' is missing in type 'typeof PartnershipAgreementStatus' but required in type '{ prototype: PartnershipAgreementStatus; }'.
```

It looks like `SecuredProperty` is expecting a value that can be a class, interface, or scalar. Enums don't seem to match any of these constraints. See `SecuredProperty` definition here:

```typescript
export function SecuredProperty<T>(
  ValueClass: ClassType<T> | AbstractClassType<T> | GraphQLScalarType
) {
  @ObjectType({ isAbstract: true, implements: [Readable, Editable] })
  abstract class SecuredPropertyClass
    implements Readable, Editable, Secured<T> {
    @Field(() => ValueClass, { nullable: true })
    readonly value?: T;
    @Field()
    readonly canRead: boolean;
    @Field()
    readonly canEdit: boolean;
  }

  return SecuredPropertyClass;
}
```

Any thoughts on how we can make `SecuredProperty` abstractly accept enum values while not compromising type safety in TS?

src/components/partnership/dto/partnership.dto.ts
richardgirges: Got it! I'm running into a similar issue here as I am with `SecuredPartnershipAgreementStatus` - can't seem to easily create a `SecuredProperty` that returns a secured `value` of `PartnershipType[]` - is that the right approach?

src/components/partnership/dto/partnership.dto.ts
CarsonF: Ok it looks like type-graphql emits schema correctly when passing the enum in, so it's just TS that doesn't like it.

My only thought is to add `object` to the acceptable input types.

src/components/partnership/dto/partnership.dto.ts
CarsonF: Bleh. Yeah that is the right output. Sounds like for this case we should just forgo the `SecuredProperty` abstraction. See `SecuredDate` for implementation without using abstraction.

src/components/partnership/dto/partnership.dto.ts
richardgirges: Perfect! That should do it!

src/components/partnership/dto/partnership.dto.ts
richardgirges: Okay, I took your advice but went a slightly different direction and created a new abstract class: `SecuredPropertyList`. It takes the same `T` types as `SecuredProperty`, only they're all arrays. Check it out in `secured-property.ts`: 

```typescript

export function SecuredPropertyList<T>(
  ValueClassList:
    | Array<ClassType<T>>
    | Array<AbstractClassType<T>>
    | GraphQLScalarType[]
    | object[]
) {
  @ObjectType({ isAbstract: true, implements: [Readable, Editable] })
  abstract class SecuredPropertyClass
    implements Readable, Editable, Secured<T> {
    @Field(() => ValueClassList, { nullable: true })
    readonly value?: T;
    @Field()
    readonly canRead: boolean;
    @Field()
    readonly canEdit: boolean;
  }

  return SecuredPropertyClass;
}
```

Seems to get us what we need and allows us to have more secured lists of different types in the future. What do you think of this approach? 

If this is too much abstraction for something that you believe will only ever happen once or twice, I could implement it locally in the partnership.dto instead following the `SecuredDate` pattern - lemme know!

richardgirges: Now that I'm looking at this more, I'm thinking I could easily just add array types to SecuredProperty as well. Lemme know your thoughts on either approach!

CarsonF: I think this is a good direction. Realizing now we'll have more of these properties so this will be useful.

One tweak though, still pass in a single value/type and map it to a list in this abstraction.
```typescript
  abstract class SecuredPropertyClass
    implements Readable, Editable, Secured<T[]> {
    @Field(() => [ValueClass], { nullable: true })
    readonly value?: T[];
  }
```

I also wonder if the hidden value should be null/undefined or an empty list? I could see both ways making sense...

richardgirges: Perfect, just made those changes!

Generally, I'm a fan of empty non-nullable lists in GQL. But given this is access-related, I can see a distinction between null/undefined vs an empty list. Empty list = there is no data. null/undefined = you don't have access. What do you think?

CarsonF: Changes look good!

> Generally, I'm a fan of empty non-nullable lists in GQL. But given this is access-related, I can see a distinction between null/undefined vs an empty list. Empty list = there is no data. null/undefined = you don't have access. What do you think?

Well no because `canRead` indicates whether it's empty with access or not accessible.
I think we should probably be consistent anyways and we do the same thing for `SecuredList`. So let's make it non-nullable.

src/components/partnership/dto/partnership.dto.ts
```diff
@@ -20,8 +20,8 @@ export abstract class SecuredPartnershipAgreementStatus extends SecuredProperty<
   description: SecuredProperty.descriptionFor('a list of partnership types'),
 })
 export abstract class SecuredPartnershipTypes extends SecuredPropertyList<
-  PartnershipType[]
->([PartnershipType]) {}
+  PartnershipType
+>(PartnershipType) {}
```
CarsonF: Shouldn't need to explicitly type generic FYI

richardgirges: Just fixed this in the last commit

src/components/partnership/dto/partnership.dto.ts
richardgirges: Got it! Makes sense, just pushed up the update making it non-nullable

CarsonF: So doing means the `SecuredProperty.descriptionFor` is incorrect for this case.
Can you create a `SecuredPropertyList.descriptionFor` to use instead?
Sorry I know this is a lot of back and forth ☹️

richardgirges: Haha no worries at all man. Just pushed up a `SecuredPropertyList.descriptionFor` - lemme know how that looks!

src/components/partnership/dto/create-partnership.dto.ts
```diff
@@ -0,0 +1,43 @@
...
+  readonly mouStart: DateTime | null;
+
+  @DateField({ nullable: true })
+  readonly mouEnd: DateTime | null;
+
+  @Field(() => ID)
+  readonly organizationId: string;
+
+  @Field(() => [PartnershipType])
+  readonly types: PartnershipType[];
```
CarsonF: I'm not sure if I missed this before, all of these should be optional (with `?:`) except `organizationId`.

src/components/partnership/dto/partnership.dto.ts
CarsonF: Looks great!

src/components/partnership/dto/create-partnership.dto.ts
richardgirges: Done!

CarsonF: Decorators need `{ nullable: true }` too 😬

richardgirges: Whoops! 😆 Okay should be good now. Just fixed the type error that was failing in CI as well

PR #208, opened 2020-03-09
Forgot/Reset Password

aarontopstar:
Completed reset password.
 - 2 endpoints
 - AWS configuration and SES config.
 At this time, I tested this endpoint with my AWS account info.

### DIFF
--- a/src/components/auth/auth.dto.ts
+++ b/src/components/auth/auth.dto.ts
@@ -34,3 +34,12 @@ export class LoginOutput {
 
   // TODO Global Permissions
 }
+
+@InputType()
+export abstract class ResetPasswordInput {
+  @Field()
+  readonly token: string;
+
+  @Field()
+  readonly password: string;
+}
--- a/src/components/auth/auth.resolver.ts
+++ b/src/components/auth/auth.resolver.ts
@@ -1,6 +1,11 @@
 import { Args, Mutation, Resolver } from '@nestjs/graphql';
 import { UserService } from '../user';
-import { CreateSessionOutput, LoginInput, LoginOutput } from './auth.dto';
+import {
+  CreateSessionOutput,
+  LoginInput,
+  LoginOutput,
+  ResetPasswordInput,
+} from './auth.dto';
 import { AuthService } from './auth.service';
 import { ISession, Session } from './session';
 
@@ -45,6 +50,24 @@ export class AuthResolver {
   })
   async logout(@Session() session: ISession): Promise<boolean> {
     await this.authService.logout(session.token);
+    return true;
+  }
+
+  @Mutation(() => Boolean, {
+    description: 'Forgot password; send password reset email',
+  })
+  async forgotPassword(@Args('email') email: string): Promise<boolean> {
+    await this.authService.forgotPassword(email);
+    return true;
+  }
+
+  @Mutation(() => Boolean, {
+    description: 'Reset Password',
+  })
+  async resetPassword(
+    @Args('input') input: ResetPasswordInput
+  ): Promise<boolean> {
+    await this.authService.resetPassword(input);
     return true;
   }
 }
--- a/src/components/auth/auth.service.ts
+++ b/src/components/auth/auth.service.ts
@@ -1,9 +1,10 @@
 import { Injectable, UnauthorizedException } from '@nestjs/common';
 import * as argon2 from 'argon2';
+import { SES } from 'aws-sdk';
 import { sign, verify } from 'jsonwebtoken';
 import { DateTime } from 'luxon';
 import { ConfigService, DatabaseService, ILogger, Logger } from '../../core';
-import { LoginInput } from './auth.dto';
+import { LoginInput, ResetPasswordInput } from './auth.dto';
 import { ISession } from './session';
 
 interface JwtPayload {
@@ -15,6 +16,7 @@ export class AuthService {
   constructor(
     private readonly db: DatabaseService,
     private readonly config: ConfigService,
+    private readonly ses: SES,
     @Logger('auth:service') private readonly logger: ILogger
   ) {}
 
@@ -181,6 +183,112 @@ export class AuthService {
     };
     this.logger.debug('Created session', session);
     return session;
+  }
+
+  async forgotPassword(email: string): Promise<void> {
+    const result = await this.db
+      .query()
+      .raw(
+        `
+        MATCH
+        (email:EmailAddress {
+          value: $email
+        })
+        RETURN
+        email.value as email
+        `,
+        {
+          email: email,
+        }
+      )
+      .first();
+
+    if (!result) {
+      throw new Error('Could not find email on database');
+    }
+
+    const token = this.encodeJWT();
+    await this.db
+      .query()
+      .raw(
+        `
+      CREATE(et:EmailToken{value:$value, token: $token, createdOn:datetime()})
+      RETURN et as emailToken
+      `,
+        {
+          value: email,
+          token: token,
+        }
+      )
+      .first();
+    const params = {
+      Destination: { ToAddresses: [email] },
+      Message: {
+        Body: {
+          Html: {
+            Charset: 'UTF-8',
+            Data: `<html><body><p>This is your secret login code:</p>
+                          <a href="${this.config.resetPasswordURL}?token=${token}">Go to Login</a></body></html>`,
+          },
+          Text: {
+            Charset: 'UTF-8',
+            Data: `${this.config.resetPasswordURL}?token=${token}`,
+          },
+        },
+        Subject: {
+          Charset: 'UTF-8',
+          Data: 'Forget Password - CORD Field',
+        },
+      },
+      Source: this.config.emailForm,
+    };
+
+    await this.ses.sendEmail(params).promise();
+  }
+
+  async resetPassword(input: ResetPasswordInput): Promise<void> {
+    const { token, password } = input;
+    const checkDate = new Date();
+
+    const result = await this.db
+      .query()
+      .raw(
+        `
+        MATCH(emailToken: EmailToken{token: $token})
+        RETURN emailToken.value as email, emailToken.token as token, emailToken.createdOn as createdOn
+        `,
+        {
+          token: token,
+        }
+      )
+      .first();
+
+    if (!result) {
+      throw new Error('Could not find token on database');
+    } else {
+      if (
+        Math.abs(
+          (checkDate.getTime() - Date.parse(result.createdOn)) / (1000 * 3600)
+        ) > 24
+      ) {
+        throw new Error('token has been expired');
+      }
+      await this.db
+        .query()
+        .raw(
+          `
+          MATCH(e:EmailToken{token: $token})
+          DELETE e WITH * OPTIONAL MATCH(:EmailAddress{active: true, value:$email})<-[:email{active:true}]-(user:User{active:true})-[:password{active:true}]->(password:Property{active:true})
+          SET password.value=$password return password
+          `,
+          {
+            token: token,
+            email: result.email,
+            password: password,
+          }
+        )
+        .first();
+    }
   }
 
   private encodeJWT() {
--- a/src/components/auth/auth.service.ts
+++ b/src/components/auth/auth.service.ts
@@ -204,7 +204,8 @@ export class AuthService {
       .first();
 
     if (!result) {
-      throw new Error('Could not find email on database');
+      this.logger.warning('Email not found; Skipping reset email', { email });
+      return;
     }
 
     const token = this.encodeJWT();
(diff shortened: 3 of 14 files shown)


src/components/auth/auth.service.ts
```diff
@@ -184,6 +188,93 @@ export class AuthService {
...
+            Charset: 'UTF-8',
+            Data: `http://localhost:3333/auth/reset?token=${token}`,
+          },
+        },
+        Subject: {
+          Charset: 'UTF-8',
+          Data: 'Forget Password',
+        },
+      },
+      Source: this.env.string('SOURCE_EMAIL').optional('core-field'),
```
CarsonF: `EnvironmentService` is meant to be a private helper service for `ConfigService` (as noted in the class docs). This is also why it's not exported from the `CoreModule`.

Can you move this to the `ConfigService`? Something like:
```ts
class ConfigService {
  emailFrom = this.env.string('EMAIL_FROM').optional('noreply@cordfield.com');
}
```
The reason for this is to prevent using "magic strings" all over the codebase and centralizes it all in one spot.

src/core/aws-ses.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Injectable()
+export class SesService {
+  private readonly ses: any;
+  constructor(private readonly env: EnvironmentService) {
+    AWS.config.update({
+      accessKeyId: this.env.string('AWS_ACCESS_KEY').optional('cord-field'),
+      secretAccessKey: this.env.string('AWS_SECRET_KEY').optional('cord-field'),
+      region: this.env.string('AWS_REGION').optional('cord-field'),
+    });
```
CarsonF: We don't want to be changing a global singleton in this specific service (`update()`).

Also, _access key id_ and _secret access key_ will automatically be pulled from the environment or a number of different places automatically. So we don't want to limit ourselves to only env vars.

I believe the default region is also determined in a similar way, but if not that should be moved to `ConfigService` just like `emailFrom` and defaulted to `us-east-2`. That still can be passed directly into the `SES` constructor instead of the global `update()` call.

src/core/aws-ses.service.ts
```diff
@@ -0,0 +1,28 @@
+import { Injectable } from '@nestjs/common';
+import * as AWS from 'aws-sdk';
+import { EnvironmentService } from './config/environment.service';
+
+@Injectable()
+export class SesService {
```
CarsonF: `SES` can be made injectable with a factory so there's no need for a wrapper service.
Take a look at [`aws-s3.factory.ts`](https://github.com/SeedCompany/cord-api-v3/blob/aba8071027962ea21df4b6a7e97d845e2c2d4f8d/src/core/aws-s3.factory.ts)
This can be referenced in the `providers` & `exports` of `CoreModule` allowing `SES` to be injected into any service.

src/core/aws-ses.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    return this.ses.sendEmail(params, (err: any, data: any) => {
+      if (err) {
+        console.log(err, err.stack);
+        return false;
+      } else {
+        console.log(data);
+        return true;
+      }
+    });
+  }
```
CarsonF: Most AWS methods support both callbacks and promises to maintain backwards compatibility. We want to only use the promises. Also FYI the returned boolean in the callback will not make the outer `sendEmail` method return a boolean.

Instead do:
```ts
await ses.sendEmail(params).promise();
```
I don't think a wrapping method is needed either and can just be called with the directly injected SES.

Side note: We may in the future decide we want an `EmailService` with specific methods for each type of email sent.

src/components/auth/auth.service.ts
```diff
@@ -184,6 +188,93 @@ export class AuthService {
...
+    await this.db
+      .query()
+      .raw(
+        `
+      CREATE(et:EmailToken{value:$value, token: $token, createdOn:datetime()})
+      RETURN et as emailToken
+      `,
+        {
+          $value: email,
+          $token: token,
```
CarsonF: I don't believe these keys are supposed to be prefixed with `$`

src/components/auth/auth.service.ts
```diff
@@ -184,6 +188,93 @@ export class AuthService {
     return session;
   }
 
+  async forget(input: UserEmailInput): Promise<boolean> {
```
CarsonF:
```suggestion
  async forgotPassword(email: string): Promise<void> {
```
There's only ever one parameter needed here, `email`, so I would avoid creating an input type for it and just have the argument directly on the query.

We also don't want to return booleans that indicate success. These either work or throw an error. GraphQL doesn't have the concept of a `void` return so in those cases we `return true` in the resolver. But that's a graphql transport detail which is why it goes there and not in the service.

src/components/auth/auth.service.ts
```diff
@@ -184,6 +188,93 @@ export class AuthService {
...
+      )
+      .first();
+    const params = {
+      Destination: { ToAddresses: [email] },
+      Message: {
+        Body: {
+          Html: {
+            Charset: 'UTF-8',
+            Data: `<html><body><p>This is your secret login code:</p>
+                          <a href="http://localhost:3333/auth/reset?token=${token}">Go to Login</a></body></html>`,
```
CarsonF: We'll want to have the frontend url be in the config service so it can different deployment environments reference their own UI. Something like `resetPasswordUrl`

src/components/auth/auth.dto.ts
```diff
@@ -34,3 +34,12 @@ export class LoginOutput {
 
   // TODO Global Permissions
 }
+
+@InputType()
+export abstract class ResetInput {
```
CarsonF:
```suggestion
export abstract class ResetPasswordInput {
```

src/components/auth/auth.resolver.ts
```diff
@@ -47,4 +52,19 @@ export class AuthResolver {
     await this.authService.logout(session.token);
     return true;
   }
+
+  @Mutation(() => Boolean, {
+    description: 'Forget Password',
+  })
+  async forget(@Args('input') input: UserEmailInput): Promise<boolean> {
```
CarsonF:
```suggestion
  async forgotPassword(@Args('email') email: string): Promise<boolean> {
```

src/components/auth/auth.resolver.ts
```diff
@@ -47,4 +52,19 @@ export class AuthResolver {
     await this.authService.logout(session.token);
     return true;
   }
+
+  @Mutation(() => Boolean, {
+    description: 'Forget Password',
```
CarsonF:
```suggestion
    description: 'Forgot password; send password reset email',
```

src/components/auth/auth.resolver.ts
```diff
@@ -47,4 +52,19 @@ export class AuthResolver {
...
+  })
+  async forget(@Args('input') input: UserEmailInput): Promise<boolean> {
+    if (await this.userService.checkEmail(input)) return false;
+    return await this.authService.forget(input);
+  }
+
+  @Mutation(() => Boolean, {
+    description: 'Reset Password',
+  })
+  async reset(@Args('input') input: ResetInput): Promise<boolean> {
```
CarsonF:
```suggestion
  async resetPassword(@Args('input') input: ResetInput): Promise<boolean> {
```

src/components/auth/auth.module.ts
```diff
@@ -7,7 +9,13 @@ import { SessionPipe } from './session';
 @Global()
 @Module({
   imports: [forwardRef(() => UserModule)],
-  providers: [AuthResolver, AuthService, SessionPipe],
+  providers: [
+    AuthResolver,
+    AuthService,
+    SessionPipe,
+    ConfigService,
+    EnvironmentService,
```
CarsonF:
```suggestion
```
`EnvironmentService` isn't used and `ConfigService` is global.

src/components/auth/auth.service.ts
```diff
@@ -16,6 +18,7 @@ export class AuthService {
   constructor(
     private readonly db: Connection,
     private readonly config: ConfigService,
+    @Inject(AwsSESFactory.provide) private readonly ses: SES,
```
CarsonF:
```suggestion
    private readonly ses: SES,
```

src/components/auth/auth.service.ts
```diff
@@ -184,6 +187,92 @@ export class AuthService {
...
+          Charset: 'UTF-8',
+          Data: 'Forget Password',
+        },
+      },
+      Source: this.config.emailForm,
+    };
+
+    await this.ses.sendEmail(params).promise();
+
+    return true;
```
CarsonF:
```suggestion
```

src/core/config/config.service.ts
```diff
@@ -16,6 +16,14 @@ export class ConfigService {
 
   jwtKey = this.env.string('JWT_AUTH_KEY').optional('cord-field');
 
+  emailForm = this.env.string('EMAIL_FORM').optional('core-field');
```
CarsonF:
```suggestion
  emailFrom = this.env.string('EMAIL_FROM').optional('noreply@cordfield.com');
```

src/core/config/config.service.ts
```diff
@@ -16,6 +16,14 @@ export class ConfigService {
...
+
+  awsRegion = this.env.string('AWS_REGION').optional('core-field');
```
CarsonF:
```suggestion
  awsRegion = this.env.string('AWS_REGION').optional('us-east-2');
```

src/core/config/config.service.ts
```diff
@@ -16,6 +16,14 @@ export class ConfigService {
...
+
+  resetPasswordURL = this.env
+    .string('RESET_PASSWORD_URL')
+    .optional('core-field');
```
CarsonF:
```suggestion
    .optional('https://cordfield.com/login/reset-password/');
```

src/components/auth/auth.service.ts
```diff
@@ -184,6 +187,92 @@ export class AuthService {
...
+      RETURN et as emailToken
+      `,
+        {
+          value: email,
+          token: token,
+        }
+      )
+      .first();
+    const params = {
+      Destination: { ToAddresses: ['leopard3551@gmail.com'] },
```
CarsonF: Shouldn't this be the email passed in?

src/components/auth/auth.service.ts
```diff
@@ -184,6 +187,92 @@ export class AuthService {
...
+                          <a href="${this.config.resetPasswordURL}?token=${token}">Go to Login</a></body></html>`,
+          },
+          Text: {
+            Charset: 'UTF-8',
+            Data: `${this.config.resetPasswordURL}?token=${token}`,
+          },
+        },
+        Subject: {
+          Charset: 'UTF-8',
+          Data: 'Forget Password',
```
CarsonF:
```suggestion
          Data: 'Forgot Password - CORD Field',
```

src/components/auth/auth.service.ts
```diff
@@ -184,6 +187,92 @@ export class AuthService {
...
+      ) {
+        return false;
+      }
+      await this.db
+        .query()
+        .raw(
+          `
+          MATCH(e:EmailToken{token: $token})
+          DELETE e WITH * OPTIONAL MATCH(:EmailAddress{active: true, value:"test@test.com"})<-[:email{active:true}]-(user:User{active:true})-[:password{active:true}]->(password:Property{active:true})
+          SET password.value=$password return password
```
CarsonF: Looks like the email address is hardcoded here. Shouldn't it be grabbed from somewhere?

src/components/auth/auth.resolver.ts
```diff
@@ -47,4 +52,21 @@ export class AuthResolver {
     await this.authService.logout(session.token);
     return true;
   }
+
+  @Mutation(() => Boolean, {
+    description: 'Forgot password; send password reset email',
+  })
+  async forgetPassword(@Args('email') email: string): Promise<boolean> {
```
CarsonF:
```suggestion
  async forgotPassword(@Args('email') email: string): Promise<boolean> {
```

src/components/auth/auth.service.ts
```diff
@@ -184,6 +187,92 @@ export class AuthService {
     return session;
   }
 
+  async forget(email: string): Promise<boolean> {
```
CarsonF:
```suggestion
  async forgotPassword(email: string): Promise<boolean> {
```

src/components/auth/auth.resolver.ts
```diff
@@ -47,4 +52,21 @@ export class AuthResolver {
...
+    if (await this.userService.checkEmail(email)) return false;
```
CarsonF: It would be better if this check was in `AuthService`. We want all business logic in services. Resolvers are only the transport medium.

src/components/auth/auth.dto.ts
```diff
@@ -34,3 +34,12 @@ export class LoginOutput {
...
   // TODO Global Permissions
 }
+
+@InputType()
+export abstract class ResetPasswordInput {
+  @Field()
+  token: string;
+
+  @Field()
+  password: string;
```
CarsonF:
```suggestion
  @Field()
  readonly token: string;

  @Field()
  readonly password: string;
```

This ensure data is correctly processed and silly things are not done with passing `input` around; it is also consistent with all other inputs.

CarsonF:
Thanks @aarontopstar!

PR #209, opened 2020-03-09
Rename Property Updater

### LINKED ISSUE #196: Rename the PropertyUpdaterService class
Rename the PropertyUpdaterService class to something more appropriate to its new, wider scope.

richardgirges:
Closes #196 

In this PR, the Property Updater is now known as `DatabaseService`, and the original `DatabaseService` is now known as `DeprecatedDBService`.

## Breaking Changes
None, but this is a conflict-prone PR

## API Updates
* Renamed the original `DatabaseService` to `DeprecatedDBService` (we should create another task to completely eliminate the use of this service later)
* Renamed `PropertyUpdater` to `DatabaseService`
* Changed all instances of `DeprecatedDBService`'s `this.db` to `this.deprecatedDb`
* Changed all instances of `DatabaseService`'s `this.propertyUpdater` to `this.databaseService`
* Changed all instances of `Connection`'s `this.db` to `this.dbConn`
* All relevant files renamed respectively (e.g. `property-updater.service.ts` is now `database.service.ts`)


richardgirges:
Sounds like a great idea to me! I just pushed up this change. 

1. Added `query()` method to `DatabaseService`
2. Renamed `DatabaseService` instances from `this.dbService` to `this.db`
3. Switched all instances of `dbConn.query()` to `this.db.query()`

Lemme know how this all looks!

PR #213, opened 2020-03-10
Production build updates

richardgirges:
Closes SeedCompany/cord-infra#8

## Breaking Changes
None

## Infrastructure Updates
The Dockerfile has been updated with various changes:
* All environment variables that are needed by the cord-api server are now listed in the Dockerfile
* We are no longer doing blind `yarn install` commands - instead we are doing deterministic dependency installations, therefore speeding up the build process
  * See [here](https://hackernoon.com/using-yarn-with-docker-c116ad289d56) for more info on deterministic yarn builds in Docker
* The cord-api code is now being placed in a more conventional `/opt/cord-api` location on the Docker container FS
* **[IMPORTANT]** We are now running the `nestjs` build command (`yarn build`). This will create a compiled version of the server within the `dist` folder
* **[IMPORTANT]** We are now running the `node dist/main.js` command to run the server.

## Deploying
This Dockerfile should be good to go, ready for testing in Production. Simply pass in the production-specific environment variables into Fargate and you should be good to go! Let me know if you run into any issues.

Here is an example `docker run` command for this Dockerfile:
```bash
docker run -e NEO4J_URL=bolt://host.docker.internal -e NEO4J_USERNAME=neo4j -e NEO4J_PASSWORD=asdf -e JWT_AUTH_KEY=asdf -e FILES_S3_BUCKET=asdf -p 3000:3000 -t -i cord_docker_image
```

For production, you'll want to pass in the `NODE_ENV` variable as well:
```bash
docker run -e NODE_ENV=production -e NEO4J_URL=bolt://host.docker.internal -e NEO4J_USERNAME=neo4j -e NEO4J_PASSWORD=asdf -e JWT_AUTH_KEY=asdf -e FILES_S3_BUCKET=asdf -p 3000:3000 -t -i cord_docker_image
```


Dockerfile
```diff
@@ -1,14 +1,31 @@
...
 WORKDIR /usr/src/app
 
-COPY . .
+ENV NODE_ENV=development \
+    PORT=3000 \
+    NEO4J_URL= \
+    NEO4J_USERNAME= \
+    NEO4J_PASSWORD= \
+    JWT_AUTH_KEY= \
+    FILES_S3_BUCKET=
```
CarsonF: Shouldn't env vars be defined per env?

Dockerfile
```diff
@@ -1,14 +1,31 @@
...
+RUN mkdir -p /opt/cord-api && cd /opt/cord-api && ln -s /tmp/node_modules
+COPY . /opt/cord-api
+WORKDIR /opt/cord-api
+
+RUN yarn build
 
 RUN apk del build-dependencies
 
-EXPOSE 3000
+EXPOSE $PORT
```
CarsonF: Why not standard port 80? It can be mapped when running right?

richardgirges:
> There's `.dockerignore` in master now as of this past weekend.
> It's got `node_modules` listed there so docker will not pull in the local folder on your machine.
> This should remove the need to use a temp folder.
> 
> But I do agree that `package.json` and `yarn.lock` should be listed separately so docker can cache that independently.

So the real reason we use the `tmp` folder here isn't to avoid `node_modules` being copied in. Docker is smart enough to cache your `RUN` commands deterministically. So when Docker sees that the `yarn.lock` and `package.json` file are the only files associated with the Docker image's `tmp` folder, it will automatically cache the `RUN` command associated with the `tmp` folder and only bust that cache when `yarn.lock` or `package.json` are updated. This is a trick I learned in the article linked to this PR titled "Using yarn with docker"

Dockerfile
richardgirges: We seem to be using port 3000 by default, I didn't want to mess up any of Dmitiry's deployments because he's already mapping to the container's port 3000 in Fargate. Happy to change this to 80 if Dimitry is open to updating the port mappings in Fargate first!

Dockerfile
richardgirges: The env vars here are simply to show that these are the environment variables required in order for this Docker container to run. Technically they are defined on a per-env level so I could remove this. What do you think?

Dockerfile
CarsonF: @chaurov What do you think? Can you change? Port 80 is the standard web server port.

CarsonF:
Hmm why can't yarn just be ran in the project root and that's what's cached at that layer? Then the rest of the src/etc files copied in after that?

Dockerfile
CarsonF: Ah they are defaults. Alright that works for me.

richardgirges:
> Hmm why can't yarn just be ran in the project root and that's what's cached at that layer? Then the rest of the src/etc files copied in after that?

Gotchaaa, let me test that out. I _believe_ that Docker will automatically re-run the `yarn install` command every time any code change in `src` happens if we do it that way. But I'm gonna run some quick tests to find out and report back!

richardgirges:
@CarsonF you were right! Ran some tests, changed the src folder contents randomly, but the `yarn` command stayed cached despite that. Just updated the Dockerfile. This seems much cleaner :D

Dockerfile
richardgirges: Sweet!

CarsonF:
One more thing, we have two `WORKDIR` actions. Can we consolidate to one?

richardgirges:
Didn't even notice that `WORKDIR` at the top of the file 😆 

Dockerfile updated to be DRY-compliant! 🎉

Dockerfile
chaurov: I can change, if needed

CarsonF:
Ok I made a bunch of changes. I rebased this on `master` branch and will merge into that. We can merge into `integration` from `master`. Basically `master` & `integration` mirrors `develop` & `master` in our repos respectively.

# Dockerfile multi-stages

We now use a handful of stages to keep dockerfile DRY and utilize its caching layers.
But more importantly it allows `dev` & `production` targets.
- `dev` target can run server & tests. (No linting)
- `production` target (default) removes tests and dev dependencies
  - This reduces the image size by 60% 🤘 (255MB down from 630MB)
- Removed unnecessary build tools now that argon2 v0.26 has prebuilt binaries for docker
- Ignored more files from docker like ops config, linters, dev notes, local files, generated files

# yarn docker:build
This is a helper script to build docker images
- It builds for both `dev` and `production` targets
- TAG env var can be used to specify one or more tags separated by
    comma or space instead of the default `latest` tag
  ```
  TAG=1c2b34,latest yarn docker:build
  ```
  These only apply to the `production` target. `dev` will always be tagged as `dev`

- If AWS info can be identified both the tags and targets will be mirrored to the ECR url as well
  - Info can be provided via `AWS_ACCOUNT` and `AWS_DEFAULT_REGION`
  - Or the AWS cli will be used to determine both.
  -  `AWS_PROFILE` can also be used to select another configured profile with the aws cli

For example:
Given `TAG=foo,1234 yarn docker:build` and assuming aws is configured the following images and tags will be built:
```
cord-api-v3:dev
123456789012.dkr.ecr.us-east-2.amazonaws.com/cord-api-v3:dev
cord-api-v3:foo
123456789012.dkr.ecr.us-east-2.amazonaws.com/cord-api-v3:foo
cord-api-v3:1234
123456789012.dkr.ecr.us-east-2.amazonaws.com/cord-api-v3:1234
```

The ECR urls are what is pushed to ECR, and the other ones can be used to reference locally without the long annoying ECR url.

# docker-compose
`docker-compose.yml` has been updated to run neo4j instance local to the service so no other setup is necessary.
After checking out the repo the next command can be `docker-compose up` and you'll have a running server & db.

# buildspec.yml

I didn't touch buildspec but it needs a ton of work. I've had the most success when it's built by CDK. This allows codebuild to reference or be configured with the connection points around aws without having to specify those values directly.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ADD https://raw.githubusercontent.com/eficode/wait-for/96511d65c6578d4866591283fdec6e2fba7e6770/wait-for /usr/local/bin/wait-for
+RUN chmod +x /usr/local/bin/wait-for
+
+ENV NODE_ENV=development \
+    NEO4J_URL= \
+    NEO4J_USERNAME= \
+    NEO4J_PASSWORD= \
+    JWT_AUTH_KEY= \
+    FILES_S3_BUCKET= \
+    PORT=80
```
richardgirges: You can drop this line since we're exposing port `80` raw in the line below

CarsonF: App still needs to be configured on that port 😉

richardgirges: Touche!

PR #226, opened 2020-03-13
Switch to using the cypher query builder for the createUser service

### LINKED ISSUE #200: CreateUser optional properties incorrectly required by DB query
`phone`, `timezone`, `bio` are optional properties of create user.
The DB query should not require them to be set.

richardgirges:
Closes #200 

In this PR:
* UserService.create switched to utilize the cypher query builder for the DB insertion
* Removed some debug code from the DB service that got merged in a prior PR

I attempted to use the `db.createNode` generic. This turned out to be a pretty big waste of time that required too much refactoring. Due to the access control props we're setting on the user's base node, I think it justifies utilizing the query builder. Perhaps we can revisit this later if the `db.createNode` proves to be less flexible in other services.

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -1,5 +1,6 @@
 import { Injectable, NotFoundException } from '@nestjs/common';
 import * as argon2 from 'argon2';
+import { node, relation } from 'cypher-query-builder';
 import { generate } from 'shortid';
 import { DatabaseService, ILogger, Logger, OnIndex } from '../../core';
 import { ISession } from '../auth';
@@ -166,25 +167,19 @@ export class UserService {
       throw new Error('Password is required when creating a new user');
     }
 
-    // ensure token doesn't have any users attached to it
-    // await this.logout(session.token);
-
+    const id = generate();
     const pash = await argon2.hash(input.password);
-    /** CREATE USER
-     * get the token, then create the user with minimum properties
-     * create an ACL node and ensure the user can edit their own properties
-     */
-    const result = await this.db
+
+    await this.db
       .query()
-      .raw(
-        `
-        MATCH (token:Token {active: true, value: $token})
-        CREATE
-          (user:User {
-            id: $id,
+      .matchNode('token', 'Token', { active: true, value: session.token })
+      .create([
+        [
+          node('user', 'User', {
+            id,
             active: true,
-            createdAt: datetime(),
-            createdByUserId: "system",
+            createdAt: 'datetime()',
+            createdByUserId: 'system',
             canCreateFileNode: true,
             canCreateOrg: true,
             canReadOrgs: true,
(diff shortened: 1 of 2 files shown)


src/components/user/user.service.ts
```diff
@@ -209,58 +207,131 @@ export class UserService {
...
+          relation('out', '', 'email', {
+            active: true,
+            createdAt: 'datetime()',
+          }),
+          node('email', 'EmailAddress:Property', {
             active: true,
-            value: $email,
-            createdAt: datetime()
+            value: input.email,
+            createdAt: 'datetime()',
```
CarsonF: This is interpreted as a literal string with the query builder.
```suggestion
            createdAt: DateTime.local().toNeo4JDateTime(),
```
Feel free to make variable to not repeat yourself.

src/components/user/user.service.ts
```diff
@@ -209,58 +207,131 @@ export class UserService {
...
+          node('user'),
+          relation('out', '', 'displayLastName', {
+            active: true,
+            createdAt: 'datetime()',
+          }),
+          node('displayLastName', 'Property', {
+            active: true,
+            value: input.displayLastName,
+          }),
+        ],
```
CarsonF: Most of these property queries are the same. How about a helper function?
```ts
const property = (prop: string, value: any) => {
  if (!value) {
    return [];
  }
  return [
    node('user'),
    relation('out', '', prop, {
      active: true,
      createdAt: DateTime.local().toNeo4JDateTime(),
    }),
    node(prop, 'Property', {
      active: true,
      value,
    }),
  ];
};

.create([
  ...
  property('phone', input.phone),
  ...
])
```

richardgirges: Love this idea. Should this be an enclosed method you think? Or a private method in the db service class?

src/components/user/user.service.ts
richardgirges: Changes applied!

src/components/user/user.service.ts
richardgirges: Changes applied!

src/components/user/user.service.ts
```diff
@@ -227,22 +229,39 @@ export class UserService {
...
+    const property = (prop: string, value: any) => {
+      if (!value) {
+        return [];
+      }
+
+      return [
+        node('user'),
+        relation('out', '', prop, {
+          active: true,
+          createdAt,
```
CarsonF: This can reference the variable defined below?

richardgirges: const vars get hoisted. But I can see how that looks a little weird, I'll move it to the top

richardgirges: Okay! latest updates pushed!

CarsonF: Yeah you're right. But I agree it looks better now, thanks!

PR #229, opened 2020-03-13
Cord api 214

candleshine:


src/components/user/education/education.service.ts
```diff
@@ -64,11 +64,11 @@ export class EducationService {
       });
     } catch (e) {
       console.log(e);
-      this.logger.error(`Could not create education for user ${input.userId}`);
+      this.logger.error(`Could not create education for user`);
```
CarsonF: It would still be helpful to have `{ id, userId: input.userId }` in the context here. We would only get one of the two log statements (success/failure) so the important case is missing info.

candleshine: oh I thought you didn't want that in here?

CarsonF: Not interpolated into the message.
```diff
-      this.logger.error(`Could not create education for user ${input.userId}`);
+      this.logger.error(`Could not create education for user`, { id, userId: input.userId });
```

Issue #178, opened 2020-03-02
CRUD + List for Ceremony using latest generics and schema


michaelmarshall:
Use the user, org, and language service + resolvers as an example.

richardgirges:
This is code complete but E2E tests aren't written yet. E2E tests depend on Engagements CRUD since ceremonies are created by engagements. I will resume work on this task after Engagements CRUD is merged in.

michaelmarshall:
closing, duplicate.

PR #273, opened 2020-03-17
Refactor create generic to use query builder

richardgirges:
I apologize for sending a PR that has no direct issue. I was originally working on #178 and found myself working on this refactor as a precursor to that task.

## Breaking Changes
None

## API Updates
* The `DatabaseService.createNode` method (and all dependent methods) are now refactored to utilize the cypher query builder
* All dependent methods of `createNode` switched to `private`
* Eliminated `try / catch` in `createProperty` method that was swallowing DateTime errors which resulted in empty values for all DateTime fields in Neo4J
* ~Created `massagePropertyValue` private method for massaging values before sending them to Neo4J on update and create.~
  * This has been removed now that we have the `ParamaterTransformer`
* Eliminated casting of all values to strings in Neo4J. This resolves an issue where arrays were being saved as comma-delimited strings. We are now inserting and receiving raw JS arrays from Neo4J

### DIFF
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -113,7 +113,7 @@ export class PartnershipService {
         canEdit: !!result.canEditMouEnd,
       },
       types: {
-        value: result.types ? result.types.split(',') : [],
+        value: result.types ? result.types : [],
         canRead: !!result.canReadTypes,
         canEdit: !!result.canEditTypes,
       },
@@ -232,7 +232,7 @@ export class PartnershipService {
       changes: {
         ...input,
         // TODO: propertyService.update and propertyService.createNode appear to handle array types differently...
-        types: (input.types ? input.types.join(',') : undefined) as any,
+        types: (input.types ? input.types : undefined) as any,
       },
       nodevar: 'partnership',
     });
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -231,7 +231,6 @@ export class PartnershipService {
       props: ['agreementStatus', 'mouStatus', 'mouStart', 'mouEnd', 'types'],
       changes: {
         ...input,
-        // TODO: propertyService.update and propertyService.createNode appear to handle array types differently...
         types: (input.types ? input.types : undefined) as any,
       },
       nodevar: 'partnership',
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -113,7 +113,7 @@ export class PartnershipService {
         canEdit: !!result.canEditMouEnd,
       },
       types: {
-        value: result.types ? result.types : [],
+        value: result.types ?? [],
         canRead: !!result.canReadTypes,
         canEdit: !!result.canEditTypes,
       },
@@ -231,7 +231,7 @@ export class PartnershipService {
       props: ['agreementStatus', 'mouStatus', 'mouStart', 'mouEnd', 'types'],
       changes: {
         ...input,
-        types: (input.types ? input.types : undefined) as any,
+        types: input.types as any,
       },
       nodevar: 'partnership',
     });
--- a/src/components/user/education/education.service.ts
+++ b/src/components/user/education/education.service.ts
@@ -87,7 +87,7 @@ export class EducationService {
     await this.db
       .query()
       .raw(query, {
-        userId: session.userId,
+        userId,
         id,
       })
       .first();
--- a/src/components/user/education/education.service.ts
+++ b/src/components/user/education/education.service.ts
@@ -43,7 +43,10 @@ export class EducationService {
     };
   }
 
-  async create(input: CreateEducation, session: ISession): Promise<Education> {
+  async create(
+    { userId, ...input }: CreateEducation,
+    session: ISession
+  ): Promise<Education> {
     const id = generate();
     const acls = {
       canReadDegree: true,
@@ -66,12 +69,12 @@ export class EducationService {
       console.log(e);
       this.logger.error(`Could not create education for user `, {
         id,
-        userId: input.userId,
+        userId,
       });
       throw new Error('Could not create education');
     }
 
-    this.logger.info(`Created user education`, { id, userId: input.userId });
+    this.logger.info(`Created user education`, { id, userId });
 
     // connect the Education to the User.
     const query = `
--- a/src/components/user/unavailability/unavailability.service.ts
+++ b/src/components/user/unavailability/unavailability.service.ts
@@ -62,7 +62,7 @@ export class UnavailabilityService {
     await this.db
       .query()
       .raw(query, {
-        userId: session.userId,
+        userId,
         id,
       })
       .first();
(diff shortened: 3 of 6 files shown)


richardgirges:
@CarsonF  PR has been updated to be caught up with #281 and is ready for review/merge

src/core/database/database.service.ts
```diff
@@ -572,30 +572,29 @@ export class DatabaseService {
...
-            value: input[key as keyof TObject] as string,
-            id: input.id!,
-          });
-        })
-    );
+
+    for (const k in input) {
+      if (k === 'id' || k === 'userId') {
+        continue;
+      }
```
CarsonF: Shouldn't these just not be passed in? Or maybe `id` is special. But def not `userId` other relationships will be named differently (`languageId`, `projectId`, etc.)

src/core/database/database.service.ts
```diff
@@ -608,114 +607,126 @@ export class DatabaseService {
...
-        ${key}.value, rel
-      `;
-
-    try {
-      await this.db
-        .query()
-        .raw(query, {
-          token: session.token,
-          requestingUserId: session.userId,
+          createdAt: DateTime.local().toNeo4JDateTime(),
```
CarsonF:
```suggestion
          createdAt: DateTime.local(),
```

src/core/database/database.service.ts
richardgirges: I left `id` in there and eliminated all instances where `userId` was being passed in 👍

src/components/partnership/partnership.service.ts
```diff
@@ -231,8 +231,7 @@ export class PartnershipService {
       props: ['agreementStatus', 'mouStatus', 'mouStart', 'mouEnd', 'types'],
       changes: {
         ...input,
-        // TODO: propertyService.update and propertyService.createNode appear to handle array types differently...
-        types: (input.types ? input.types.join(',') : undefined) as any,
+        types: (input.types ? input.types : undefined) as any,
```
CarsonF:
```suggestion
        types: input.types as any,
```

src/components/partnership/partnership.service.ts
```diff
@@ -113,7 +113,7 @@ export class PartnershipService {
         canEdit: !!result.canEditMouEnd,
       },
       types: {
-        value: result.types ? result.types.split(',') : [],
+        value: result.types ? result.types : [],
```
CarsonF:
```suggestion
        value: result.types ?? [],
```

src/components/user/education/education.service.ts
```diff
@@ -43,7 +43,10 @@ export class EducationService {
     };
   }
 
-  async create(input: CreateEducation, session: ISession): Promise<Education> {
+  async create(
+    { userId, ...input }: CreateEducation,
+    session: ISession
+  ): Promise<Education> {
```
CarsonF: How does the education node get connected to the user now?

src/components/user/unavailability/unavailability.service.ts
```diff
@@ -18,7 +18,7 @@ export class UnavailabilityService {
   ) {}
 
   async create(
-    input: CreateUnavailability,
+    { userId, ...input }: CreateUnavailability,
```
CarsonF: Same question

src/core/database/database.service.ts
```diff
@@ -400,7 +400,7 @@ export class DatabaseService {
         const list = typeof prop === 'object' ? prop.list : false;
 
         if (list) {
-          const value = row[propName] ? row[propName].split(',') : [];
+          const value = row[propName] ? row[propName] : [];
```
CarsonF:
```suggestion
          const value = row[propName] ?? [];
```

src/core/database/database.service.ts
```diff
@@ -567,35 +567,34 @@ export class DatabaseService {
     aclEditProp,
   }: {
     session: ISession;
-    input: { [Key in keyof TObject]?: UnwrapSecured<TObject[Key]> };
+    input: { [Key in keyof TObject]?: any };
```
CarsonF: `id` could be pulled out explicitly here

src/core/database/database.service.ts
```diff
@@ -604,117 +603,126 @@ export class DatabaseService {
...
+        })
+        .first();
+
+      // If the user doesn't have permission to perform the create action...
+      if (!aclResult || !aclResult.editProp) {
         throw new ForbiddenError(`${aclEditProp} missing or false`);
       }
-      this.logger.error(`${e} create Node error`);
+
+      this.logger.error(`DatabaseService create error`, {
```
CarsonF: The loggers are always scoped to a subject/service/namespace so prefixing isn't necessary.

src/components/user/education/education.service.ts
richardgirges: @CarsonF the education node gets connected to the user just a few lines deeper in this function:

```
    // connect the Education to the User.
    const query = `
      MATCH (user: User {id: $userId, active: true}),
        (education:Education {id: $id, active: true})
      CREATE (user)-[:education {active: true, createdAt: datetime()}]->(education)
      RETURN  education.id as id
      `;

    await this.db
      .query()
      .raw(query, {
        userId: session.userId,
        id,
      })
      .first();

    return await this.readOne(id, session);
```

src/components/user/unavailability/unavailability.service.ts
richardgirges: Same approach as the Education service, just a few lines deeper in this function.

src/core/database/database.service.ts
richardgirges: Nice! I've never seen this operator before 😆

CarsonF: Same as `||` (which I see you used in other lines) it's just stricter.
`||` chooses RHS on falsy values
`??` chooses RHS on null or undefined values

src/components/user/education/education.service.ts
CarsonF: Doh! I see though that that is incorrectly using the session user Id. We want to attach to the user specified not the user making the call.

src/components/user/unavailability/unavailability.service.ts
CarsonF: Same response here 🤣

src/core/database/database.service.ts
richardgirges: Do you mean update `createNode` to take a separate `id` argument? Or destructure `input` so that `id` is pulled out of the argument?

CarsonF: destructure so you don't have to have the if condition in the loop

src/components/user/education/education.service.ts
richardgirges: Good catch! Fixed!

src/components/user/unavailability/unavailability.service.ts
richardgirges: Fixed!

src/core/database/database.service.ts
richardgirges: Updated!

Issue #172, opened 2020-03-02
DTOs for Team


michaelmarshall:
The language service DTOs are a good example to use.

CarsonF:
Let's rename this object `ProjectMember`. It was previously `TeamMember` but there could be multiple teams in this app, so Project is more explicit to the relation.

Ok so this follows the same concept as _Partnerships_.
```
Project <-> Partnership <-> Organization
```
```
Project <-> ProjectMember <-> User
```
`ProjectMember` should have the following properties
- `user: SecuredUser` - required, not updatable, `userId: string` on create
- `roles: SecuredRoles (SecuredPropertyList)` - optional, updatable
- `modifiedAt: DateTime`

`editable` and `description` are not needed.
Follow the pattern of Partnerships for everything else.

PR #282, opened 2020-03-18
Project Member DTO & Service

aarontopstar:
##PR for #172 - ProjectMember

*Project <-> ProjectMember <-> User

- Properties
user: SecuredUser
roles: SecuredRoles (SecuredPropertyList)
modifiedAt: DateTime


src/components/project-member/dto/project-member.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  description: SecuredPropertyList.descriptionFor(
+    'a role of projectMember types'
+  ),
+})
+export abstract class SecuredRoles extends SecuredPropertyList(Role) {}
+
+@ObjectType({
+  implements: [Resource],
+})
+export class ProjectMember extends Resource {
```
CarsonF: `modifiedAt` can still be on this object. It's just read-only which is why it's not on the inputs.

src/components/user/user.service.ts
```diff
@@ -286,6 +286,8 @@ export class UserService {
             canReadRegion: true,
             canCreateCountry: true,
             canReadCountry: true,
+            canCreateProjectMember: true,
+            canReadProjectMembers: true,
```
CarsonF: @michaelmarshall I'm not up to date on this code, but I believe this should be determined by the project not globally. Members will always be created in the context of a project and we'll check `project.team.canCreate/Read` (API) for access.

src/components/project/project.module.ts
```diff
@@ -1,10 +1,12 @@
 import { Module } from '@nestjs/common';
+import { ProjectMemberModule } from '../project-member';
 import { InternshipProjectResolver } from './internship-project.resolver';
 import { ProjectResolver } from './project.resolver';
 import { ProjectService } from './project.service';
 import { TranslationProjectResolver } from './translation-project.resolver';
 
 @Module({
+  imports: [ProjectMemberModule],
```
CarsonF: We'll want to `export` this module here as well

src/components/project/project.service.ts
```diff
@@ -243,6 +249,29 @@ export class ProjectService {
     throw new NotImplementedException();
   }
 
+  async listProjectMembers(
```
CarsonF:
```suggestion
  async team(
```

src/components/project/internship-project.resolver.ts
```diff
@@ -25,4 +29,20 @@ export class InternshipProjectResolver {
   ): Promise<SecuredInternshipEngagementList> {
     return this.projects.listEngagements(project, input, session);
   }
+
+  @ResolveProperty(() => SecuredProjectMemberList, {
+    description: 'projectMembers by project',
+  })
+  async projectMembers(
```
CarsonF:
```suggestion
  async team(
```

PR #285, opened 2020-03-18
168 fix types and stub service

candleshine:
some budget record stubs need adding


src/components/budget/dto/create-budget.dto.ts
```diff
@@ -1,25 +1,29 @@
...
 import { MinLength, ValidateNested } from 'class-validator';
-import { Field, InputType, ObjectType } from 'type-graphql';
+import { Field, ID, InputType, ObjectType } from 'type-graphql';
 import { Budget, BudgetStatus } from '../budget';
 
 @InputType()
 export abstract class CreateBudget {
   @Field({ nullable: true })
   @MinLength(2)
   readonly status: BudgetStatus;
```
CarsonF: I think status is just Pending to start with, no need to specify here

src/components/budget/dto/update-budget.dto.ts
```diff
@@ -1,16 +1,16 @@
...
+import { Budget, BudgetRecord } from './budget';
 
 @InputType()
 export abstract class UpdateBudget {
   @Field(() => ID)
   readonly id: string;
 
   @Field()
-  readonly status: SecuredBudgetStatus;
+  readonly status: BudgetStatus;
```
CarsonF:
```ts
@Field(() => BudgetStatus, { nullable: true })
readonly status?: BudgetStatus;
```

candleshine:
need to fix resolver

src/components/budget/dto/create-budget.dto.ts
candleshine: that's fine

PR #292, opened 2020-03-23
createNode infer generic with type param

CarsonF:
Previously the generic, `TObject`, was not inferred with `createNode`. So this generic was meaningless without explicitly defining it (`createNode<Foo>(...)`).

I changed it to pass in `type` so it can automatically be inferred.
We also use this `type` we also default the `baseNodeLabel` and `aclEditProp` (unless `false`)

---

I also abstracted the session matching query boilerplate to its own function

### DIFF
--- a/src/components/organization/organization.service.ts
+++ b/src/components/organization/organization.service.ts
@@ -31,9 +31,9 @@ export class OrganizationService {
     try {
       await this.db.createNode({
         session,
+        type: Organization.classType,
         input: { id, ...input },
         acls,
-        baseNodeLabel: 'Organization',
         aclEditProp: 'canCreateOrg',
       });
     } catch {
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -46,6 +46,34 @@ export type DbValue = Many<
   string | number | boolean | DateTime | null | undefined
 >;
 
+export const matchSession = (
+  session: ISession,
+  {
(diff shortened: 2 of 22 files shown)


src/components/organization/organization.service.ts
```diff
@@ -31,9 +31,9 @@ export class OrganizationService {
     try {
       await this.db.createNode({
         session,
+        type: Organization.classType,
         input: { id, ...input },
         acls,
-        baseNodeLabel: 'Organization',
         aclEditProp: 'canCreateOrg',
```
richardgirges: Should we maybe rename this edit prop across the board to `canCreateOrganization` for consistency's sake?

src/core/database/database.service.ts
```diff
@@ -28,6 +29,51 @@ interface ReadPropertyResult {
...
+  Record<string, DbValue>;
+
+export type ACLs = Record<string, boolean>;
+
+/** A value that con be passed into the db */
+export type DbValue = Many<
+  string | number | boolean | DateTime | null | undefined
+>;
+
+export const matchSession = (
```
richardgirges: Nice!

src/components/organization/organization.service.ts
CarsonF: Probably. It was referenced in a few different places. I didn't want to change all of them here.

PR #296, opened 2020-03-24
Engagement CRUD Service

aarontopstar:
Status: WIP
Please don't merge.


src/components/engagement/engagement.service.ts
```diff
@@ -15,8 +26,274 @@ import {
...
+
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canReadLastSuspendedAt:ACL {canReadLastSuspendedAt: true})-[:toNode]->(languageEngagement)-[:lastSuspendedAt {active: true}]->(lastSuspendedAt:Property {active: true})
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canEditLastSuspendedAt:ACL {canEditLastSuspendedAt: true})-[:toNode]->(languageEngagement)
+
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canReadLastReactivatedAt:ACL {canReadLastReactivatedAt: true})-[:toNode]->(languageEngagement)-[:lastReactivatedAt {active: true}]->(lastReactivatedAt:Property {active: true})
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canEditLastReactivatedAt:ACL {canEditLastReactivatedAt: true})-[:toNode]->(languageEngagement)
+
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canReadStatusModifiedAt:ACL {canReadStatusModifiedAt: true})-[:toNode]->(languageEngagement)-[:statusModifiedAt {active: true}]->(statusModifiedAt:Property {active: true})
+        WITH * OPTIONAL MATCH (requestingUser)<-[:member]-(canEditStatusModifiedAt:ACL {canEditStatusModifiedAt: true})-[:toNode]->(languageEngagement)
```
CarsonF: - initialEndDate
- lastSuspendedAt
- lastReactivatedAt
- statusModifiedAt
- modifiedAt
Are all calculated fields. The user will never be able to write them directly. So they don't need the canEdit* checks.
Also it might make sense if the canRead* were just from the status field since those are dates based on that.

src/components/engagement/engagement.service.ts
```diff
@@ -35,16 +312,202 @@ export class EngagementService {
...
+      await this.db.createNode({
+        session,
+        input: {
+          id: ceremonyId,
+          ...inputForCeremony,
+        },
+        acls: aclsForCeremony,
+        baseNodeLabel: 'Ceremony',
+        aclEditProp: 'canCreateCeremony',
+      });
```
CarsonF: Move to `CeremonyService.create`

src/components/user/user.service.ts
```diff
@@ -293,6 +293,12 @@ export class UserService {
             canReadCountry: true,
             canCreateProjectMember: true,
             canReadProjectMembers: true,
+            canCreateLanguageEngagement: true,
+            canReadLanguageEngagements: true,
+            canCreateInternshipEngagement: true,
+            canReadInternshipEngagements: true,
+            canCreateCeremony: true,
```
CarsonF: Ceremonies aren't created by the user so they don't need an ACL check

src/components/user/user.service.ts
```diff
@@ -293,6 +293,12 @@ export class UserService {
             canReadCountry: true,
             canCreateProjectMember: true,
             canReadProjectMembers: true,
+            canCreateLanguageEngagement: true,
+            canReadLanguageEngagements: true,
+            canCreateInternshipEngagement: true,
+            canReadInternshipEngagements: true,
```
CarsonF: I believe these could be combined to just `canCreateEngagement`, `canReadEngagements`

src/components/engagement/engagement.service.ts
```diff
@@ -15,8 +26,274 @@ import {
...
+    if (!result) {
+      throw new NotFoundException(
+        'Could not find langage engagement or internship engagement'
+      );
+    }
+
+    let language: Language = RedactedLanguage;
+    if (result.canReadLanguage) {
+      language = await this.languageService.readOne(result.languageId, session);
+    }
```
CarsonF: If the user cannot read it, the value should be `undefined`. There's no need for `RedactedLanguage`.

src/components/engagement/engagement.service.ts
```diff
@@ -15,54 +26,823 @@ import {
...
+
+    const ceremony = await this.ceremonyService.readOne(
+      result.ceremonyId,
+      session
+    );
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    let language: Language = {} as Language;
+    if (result.languageId) {
+      language = await this.languageService.readOne(result.languageId, session);
+    }
```
CarsonF: `language` should not be an empty object. It should be a fully flushed out `Language` or `undefined`.
```suggestion
    const language = result.languageId
      ? await this.languageService.readOne(result.languageId, session)
      : undefined;
```

Same goes for the other 3 properties below.

src/components/engagement/engagement.service.ts
```diff
@@ -15,54 +26,823 @@ import {
...
+      internUser = await this.userService.readOne(result.internUserId, session);
+    }
+    if (result.mentorUserId) {
+      mentorUser = await this.userService.readOne(result.mentorUserId, session);
+    }
+    if (result.countryOfOriginId) {
+      countryOfOrigin = (await this.locationService.readOne(
+        result.countryOfOriginId,
+        session
+      )) as Country;
```
CarsonF: Looks like `LocationService` has a `readOneCountry` which would remove the need to typecast.

src/components/engagement/engagement.service.ts
```diff
@@ -15,54 +26,823 @@ import {
...
+    session: ISession
   ): Promise<SecuredProductList> {
-    throw new NotImplementedException();
+    const result = await this.products.list(
+      {
+        ...input,
+        filter: {
+          ...input.filter,
+        },
+      },
```
CarsonF:
```suggestion
      input,
```

src/components/engagement/engagement.service.ts
```diff
@@ -15,54 +26,823 @@ import {
...
+    );
+
+    return {
+      items,
+      total: items.length,
+      hasMore: false,
+    };
   }
 
   async listProducts(
```
CarsonF: This doesn't filter by engagement yet or check for permissions

src/components/engagement/engagement.service.ts
```diff
@@ -15,54 +26,823 @@ import {
...
+          projectId: projectId,
+          ceremonyId: ceremony.id,
+          id: id,
+        })
+        .first();
+
+      const res = (await this.readOne(id, session)) as LanguageEngagement;
+      return res;
+    } catch (e) {
+      this.logger.error(e);
```
CarsonF: Unfortunately types cannot be specified in catch statements, so `e` is always `any`.
Nonetheless, errors currently cannot be passed directly to log statements. 

```suggestion
      this.logger.error('Failed to create language engagement', { exception: e });
```

Same for the rest of the calls in this file.

src/components/engagement/engagement.service.ts
aarontopstar: 1. It is filtered by `internship`, `language`, or `engagement` . 
So if user entered {filter: {type: "internship"}}, then it is getting only interhshipengagement. 
And when {filter: {type: "language"}}, only languageengagement. When {filter: {type: "engagement or something"}}, both internship and language. 

2. Permission means user permission?

CarsonF: 1. No it's needs to be a list of products for the specific engagement given.
2. `canRead` and `canCreate` should not be hardcoded here https://github.com/SeedCompany/cord-api-v3/pull/296/files#diff-80606fe792a29f08bb910c50d206873cR500-R501

aarontopstar: I thought it is for list function. I got it.

PR #344, opened 2020-04-03
Add DB constraints to language service

ibohalsinghkoijam:


michaelmarshall:
@candleshine The solution is to use transactions, which will auto-rollback when one query fails. Carson implemented this in the database service but I haven't used it yet. You'll need to do a knowledge transfer with him.

PR #346, opened 2020-04-03
Authorization Service

### LINKED ISSUE #181: Authorization Service - Create Security Group + Permission
Admin's need a way to grant permissions to other users.

richardgirges:
Closes #181 

@michaelmarshall figured this is a good stopping point for review. The only things missing at the moment are:
- e2e tests for the authorization.service's `list` methods
- tying `session.token` to the authorization.service's `list` methods. Need some guidance on how the active session is supposed to affect querying in the following methods:
  - `listSecurityGroupsUserIsMemberOf`
  - `listSecurityGroupsUserIsAdminOf`
  - `listPermissionsInSecurityGroup`

I left comments in each service method explaining my queries - let me know if any of them look off.


richardgirges:
It looks like there are some failing unit tests that I'm not too familiar with (type-check failures in those spec files). I'm going to take a crack at fixing these

CarsonF:
user.service.spec.ts is failing due to the signature change @michaelmarshall made to `create` and adding `createAndLogin`

src/components/authorization/authorization.service.ts
```diff
@@ -0,0 +1,662 @@
...
+      result.userId !== request.userId ||
+      result.baseNodeId !== request.baseNodeId
+    ) {
+      return false;
+    }
+
+    return true;
+  }
+
+  async deleteSecurityGroup(id: string, session: ISession): Promise<void> {
```
michaelmarshall, line 525: The delete function doesn't check to see if the requesting user is an admin of the security group.

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,141 @@
...
+      .query()
+      .match([[node('sg', 'RootSecurityGroup')]])
+      .return('sg')
+      .first();
+
+    if (result) {
+      return true;
+    } else {
+      return false;
     }
```
CarsonF, line 38:
```suggestion
    return !!result;
```

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,141 @@
...
-
-    // ORGS
-    for (let i = 0; i < totalOrgs; i++) {
-      // this.orgService.create({
-      //   name: 'org_' + generate(),
-      // });
+  async rootAdminSecurityGroupExists(): Promise<boolean> {
+    const result = await this.db
+      .query()
+      .match([[node('sg', 'RootSecurityGroup')]])
```
CarsonF, line 30:
```suggestion
      .match(node('sg', 'RootSecurityGroup'))
```
Might be easier to read

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,141 @@
...
 
-  async consistencyCheck(): Promise<AdminOutputDto> {
-    const response = new AdminOutputDto();
-    let totalNodes = 0;
-    const session = this.db.driver.session();
-    const result = await session.run('MATCH (n) RETURN count(n) as count', {});
-    totalNodes = result.records[0].get('count');
-    for (let i = 0; i < totalNodes; i++) {
-      // todo
+  async createRootAdminSecurityGroup(): Promise<boolean> {
```
CarsonF, line 65: We don't want to return booleans to indicate success. Throw an error if there is a failure.

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,141 @@
...
   }
-  async deleteAllData(): Promise<AdminOutputDto> {
-    const response = new AdminOutputDto();
-    const session = this.db.driver.session();
-    await session.run('MATCH (n) DETACH DELETE n');
-    session.close();
-    response.success = true;
-    return response;
+
+  async createRootAdminUser(): Promise<boolean> {
```
CarsonF, line 86: Same here

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,141 @@
...
   }
 
-  async removeAllConstraintsAndIndexes(): Promise<AdminOutputDto> {
-    const response = new AdminOutputDto();
-    const session = this.db.driver.session();
-    await session.run('CALL apoc.schema.assert({}, {})');
-    session.close();
-    response.success = true;
-    return response;
+  async mergeRootAdminUserToSecurityGroup(): Promise<boolean> {
```
CarsonF, line 107: Same here

src/components/admin/root-security-group.ts
```diff
@@ -0,0 +1,6 @@
+export class RootSecurityGroup {
+  canReadOrgs = true;
+  canListUsers = true;
+  canCreateUser = true;
+  canCreateSecurityGroup = true;
+}
```
CarsonF, line 6: We are trying to avoid using classes for DTOs / value objects. This keeps logic in services and allows us to spread properties around without fear of loosing an unspoken requirement of the object prototype.

If both type and default value is needed here's the pattern we use in this rest of the codebase.
```suggestion
export class RootSecurityGroup {
  static defaultValue = new RootSecurityGroup();
  protected constructor() {}

  canReadOrgs = true;
  canListUsers = true;
  canCreateUser = true;
  canCreateSecurityGroup = true;
}
```

If the type isn't really needed just make it a constant:
```suggestion
export const RootSecurityGroup = {
  canReadOrgs: true,
  canListUsers: true,
  canCreateUser: true,
  canCreateSecurityGroup: true,
};
```

src/components/auth/index.ts
```diff
@@ -1,2 +0,0 @@
-export * from './auth.service';
-export * from './auth.module';
```
CarsonF, line 2: Let's keep using index files for exports.
They help convey which exports are "public" vs private to the module.
If anything, to be consistent with the rest of the codebase.

src/components/authorization/authorization.service.ts
```diff
@@ -0,0 +1,662 @@
...
+              active: true,
+              value: session.token,
+            }),
+            relation('in', '', 'token', {
+              active: true,
+            }),
+            node('requestingUser', 'User', {
+              active: true,
+              id: session.userId,
+            }),
```
CarsonF, line 541:
```suggestion
            ...matchSession(session),
```
Just a thought.

src/components/authorization/authorization.service.ts
richardgirges: Done

src/components/admin/admin.service.ts
richardgirges: Done

src/components/admin/admin.service.ts
richardgirges: Done

src/components/admin/admin.service.ts
richardgirges: Done

src/components/auth/index.ts
richardgirges: I believe that this was a change made by Michael based on a prior discussion w/ Daniel regarding circular dependency issues in node that arise from using index files for batch exports. Happy to bring it back though if the consensus was to resume the use of index files?

src/components/admin/root-security-group.ts
richardgirges: 10-4, going with the `const` approach

src/components/authorization/authorization.service.ts
richardgirges: Good call - done!

src/components/auth/index.ts
CarsonF: That was an edge case. Using the index.ts files was just the symptom. We shouldn't have a problem.

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,131 @@
...
+  async onApplicationBootstrap(): Promise<void> {
+    if (!(await this.rootAdminSecurityGroupExists())) {
+      await this.createRootAdminSecurityGroup();
+    }
+
+    if (!(await this.doesRootAdminUserAlreadyExist())) {
+      await this.createRootAdminUser();
+    }
+
+    await this.mergeRootAdminUserToSecurityGroup();
```
CarsonF, line 24: This is being called on every test suite

src/components/admin/admin.service.ts
CarsonF:
```suggestion
  async createRootAdminSecurityGroup(): Promise<void> {
```

src/components/admin/admin.service.ts
```diff
@@ -1,151 +1,131 @@
...
-      //   'token - replace me later',
-      // );
-    }
+  async createRootAdminSecurityGroup(): Promise<boolean> {
+    const result = await this.db
+      .query()
+      .create([
+        [node('sg', ['RootSecurityGroup', 'SecurityGroup'], RootSecurityGroup)],
+      ])
+      .first();
```
CarsonF, line 67: This query is failing in tests

PR #354, opened 2020-04-07
(WIP)Create DTOs and function creating workflow

aarontopstar:
# Workflow API Implementation


src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+      throw new NotFoundException('could not find a state');
+    }
+
+    return {
+      id: result.id,
+      value: result.value
+    }
+  }
+  // only 1 workflow per base node, only 1 base node per workflow. WF doesn't need a name, just id
+  // when the workflow node is created, the :CurrentState node is created.
```
michaelmarshall, line 63: Let's delete this comment. We'll eventually enable multiple workflows on a node.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+      this.logger.warning('Failed to delete state', {
+        exception: e,
+      });
+
+      throw e;
+    }
+  };
+
+  // attachSecurityGroupToState
+  async attachSecurityGroup(session: ISession, input: GroupState): Promise<void>{
```
michaelmarshall, line 444: the requesting user needs to be an admin of the base node and an admin of the security group in order to attached a security group to a state.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not attach security group to state', {
+        exception: e
+      });
+      throw e;
+    }
+  };
+
+  // removeSecurityGroupFromState
+  async removeSecurityGroup(session: ISession, input: GroupState): Promise<void>{
```
michaelmarshall, line 491: same as comment above.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+      this.logger.warning('could not remove security group from state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // we are using security groups as notification groups for now
+  // attachNotificationGroupToState
+  async attachNotificationGroup(session: ISession, input: GroupState): Promise<void>{
```
michaelmarshall, line 533: for this kind of addition, let's just require the requesting user to be an admin of the base node. They don't need to be an admin of the security group.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not attach security group to state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // removeNotificationGroupFromState
+  async removeNotificationGroup(session: ISession, input: GroupState): Promise<void>{
```
michaelmarshall, line 581: same as above.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not remove security group from state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // changeCurrentStateInWorkflow
+  async changeCurrentState(session: ISession, input: ChangeState): Promise<CommentState>{
```
michaelmarshall, line 622: for this the user needs to have write permission on the currentState identifier. the videos will explain this. the query will look something like: `(token)-(user)-(securityGroup)-(permission)-(baseNode)-(currentState)` where the permission node has a value of `currentState` and `canWrite = true`.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+      });
+      throw e;
+    }
+    throw new NotImplementedException();
+  }
+
+  // creates a relationship from one state to another
+  // later we can create an abstracted function that creates and attaches a state to another state
+  // addPossibleStateToState
+  async addPossibleState(session: ISession, input: PossibleState): Promise<void>{
```
michaelmarshall, line 643: user needs to be admin of base node and admin of sg.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not add possible state to state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // removePossibleStateFromState
+  async removePossibleState(session: ISession, input: PossibleState): Promise<void>{
```
michaelmarshall, line 686: same as above.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // there will be more than one required field relationship between a state node and a base node.
+  // this is so each required field can be queried without inspecting the property name in app code.
+  // addRequiredFieldToState
+  async addField(session: ISession, input: RequiredField): Promise<void>{
```
michaelmarshall, line 723: the name of this relationship is literally: `:requiredField` it has a property named `value` which is set to the identifier of the field that is required, like `name`. So the relationship object would like: `-[:requiredField {value: "name"}]->`

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not add field to state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // listAllRequiredFieldsInAState
+  async listFields(session: ISession, stateId: string): Promise<RequiredFieldListOutput>{
```
michaelmarshall, line 795: The frontend needs this info to render the form correctly so let's make it available to all users who can access the fields. no need to check for any permission except that they have read access to each field. That will be security groups and permission nodes.

michaelmarshall: also should change the name to listRequiredFields.

src/components/workflow/workflow.service.ts
michaelmarshall: should change the name to addRequiredField

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,866 @@
...
+    } catch (e) {
+      this.logger.warning('could not list fields', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // removeRequiredFieldFromState
+  async removeField(session: ISession, input: RequiredField): Promise<void>{
```
michaelmarshall, line 835: change to removeRequiredField.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,871 @@
...
+            }),
+            relation('in', '', 'token', {
+              active: true
+            }),
+            node('user'),
+            relation('in', '', 'admin', {
+              active: true
+            }),
+            node('baseNode'),
+            relation('out', '', `${input.propertyName}`),
```
michaelmarshall, line 741: the relationship type should be `requiredProperty` and the `value` property should be the `${input.propertyName}`

michaelmarshall: relation('out', '', `requiredProperty`, {value: ${input.propertyName}),

or something like that.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,944 @@
...
+          [
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
+            }),
+            node('sg', 'SecurityGroup'),
+            relation('out', '', 'permission', {
+              write: true
+            }),
+            node('permission', 'Permission'),
```
michaelmarshall, line 667: You need to specify that the permission node has a value of the property you are asking about and that the canWrite property is true. 

Something like:

`node('permission', 'Permission', {value: currentState, canWrite: true})`

Otherwise, you aren't checking if the user has permission to write. Look at the schema for Security Groups and Permissions.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1033 @@
...
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
+            }),
+            node('sg', 'SecurityGroup'),
+            relation('out', '', 'permission', {
+              write: true
+            }),
+            node('permission', 'Permission', {
+              canWrite: true,
```
michaelmarshall, line 743: I was wrong, the schema shows the property conferring write permission is called `write`

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1033 @@
...
+        .query()
+        .match([
+          [
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
+            }),
+            node('sg', 'SecurityGroup'),
+            relation('out', '', 'permission', {
+              write: true
```
michaelmarshall, line 740: there shouldn't be a property on this relationship

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1033 @@
...
+        .set({
+          values: {
+            'oldRel.active': false
+          }
+        })
+        .merge([
+          [
+            node('baseNode'),
+            relation('out', '', `${currentStateAndWorkflow.stateIdentifier}`),
+            node('state', 'State', {
```
michaelmarshall, line 763: this doesn't point to a :State node. it points to the currentState node which is just a :Property node. Change the 'State' to 'Property'

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1033 @@
...
+        ])
+        .set({
+          values: {
+            'oldRel.active': false
+          }
+        })
+        .merge([
+          [
+            node('baseNode'),
+            relation('out', '', `${currentStateAndWorkflow.stateIdentifier}`),
```
michaelmarshall, line 762: Need to add an active=true property.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1033 @@
...
+
+    } catch (e) {
+      this.logger.warning('could not remove security group from state', {
+        exception: e
+      });
+      throw e;
+    }
+  }
+
+  // changeCurrentStateInWorkflow
```
michaelmarshall, line 653: since this is a multi-query function, please add comments before each query so other devs know why each query is being run.

src/components/workflow/workflow.service.ts
michaelmarshall: line 748 isn't in the schema

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1038 @@
...
+        .match([
+          [
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
+            }),
+            node('sg', 'SecurityGroup'),
+            relation('out', '', 'permission'),
+            node('permission', 'Permission', {
+              read: true
```
michaelmarshall, line 223: should be `write: true`

michaelmarshall: the user needs to have write permission to do an update.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1038 @@
...
+              active: true,
+              value: session.token
+            }),
+            relation('in', '', 'token', {
+              active: true,
+            }),
+            node('requestingUser', 'User', {
+              id: session.userId
+            }),
+            relation('in', '', 'member'),
```
michaelmarshall, line 623: The requesting user needs to be admin on the BaseNode, that's it. They don't need to be a member of the notification's security group.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1038 @@
...
+  async changeCurrentState(session: ISession, input: ChangeCurrentState): Promise<void>{
+    try{
+      // get current state and workflow
+      const workflow = await this.db
+        .query()
+        .match([
+          [
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
```
michaelmarshall, line 657: We don't need to check for admin to read a property. You can remove the admin line.

src/components/workflow/workflow.service.ts
```diff
@@ -0,0 +1,1038 @@
...
+      
+
+      // validate the new state is a legal nextPossibleState on the current state
+      const possibleState = await this.db
+        .query()
+        .match([
+          [
+            ...matchSession(session),
+            relation('in', '', 'member', {
+              admin: true
```
michaelmarshall, line 693: same as above. remove this line.

src/components/workflow/workflow.service.ts
michaelmarshall: I'm not sure what I was thinking, but I was wrong on this one. The requesting user needs to be an admin on the BaseNode and that's it. Security groups don't come into play here. Sorry for that. 

The clause for the createWorkflow is an example for this function as well:

```
        .match([
          [
            node('token', 'Token', {
              active: true,
              value: session.token
            }),
            relation('in', '', 'token', {
              active: true
            }),
            node('user', 'User'),
            relation('in', '', 'admin', {
              active: true
            }),
            node('baseNode', 'BaseNode', {
              id: input.baseNodeId
            }),
          ],
        ])
```

src/components/workflow/workflow.service.ts
michaelmarshall: lol, Same as above! Sorry Aaron. My bad:)

PR #361, opened 2020-04-07
Use @seedcompany/eslint-plugin

CarsonF:
Uses our new `@seedcompany/eslint-plugin` to centralize and simplify our lint standards and eslint configuration.

I also removed the `eslint-plugin-local-rules` setup as well. Custom rules can be created in our eslint-plugin repo, even if they are not used by default.

Other rules can still be added/overridden in local configs as needed.
For example, we have our own logger here so we are enforcing `no-console`. This could be upstreamed to the `nestjs` config if we decide all nestjs projects should use a logger instead of console.

I'm also trying out the new `naming-conventions` rule here that I wasn't ready to apply everywhere.

Excluding `yarn.lock`: 
```
70 insertions(+), 382 deletions(-)
```


richardgirges:
So much cleaner!

PR #378, opened 2020-04-10
Generic function to validate a property is a singleton relationship

ibohalsinghkoijam:


src/components/user/user.service.ts
```diff
@@ -200,6 +200,16 @@ export class UserService {
...
 
+  async isEmailUnique(id: string, session: ISession): Promise<boolean> {
+    const isUnique = await this.db.isPropertyUnique({
+      id,
+      session,
+      baseNodeLabel: 'User',
+      propName: 'email',
+    });
+    return isUnique;
+  }
```
CarsonF, line 211: We need to enforce the user's email is unique before the user is created.

The only other time this info would be helpful is for the frontend to check for the email before submitting. And in that case they don't have a user ID, only an email address.

candleshine: this isn't a ticket for checking Uniqueness or email.  This was for the uniqueness function in db-service.  The email uniqueness was an example michael wrote in the ticket, so it was the tester function that allowed Koijam to run an e2e test through the db-service.  
the utility of checking email uniqueness is probably low.  Email uniqueness is already a DB constraint.

CarsonF: Ah gotcha. So we don't need the resolver then. Just call the service method directly in test.

src/core/database/database.service.ts
```diff
@@ -731,4 +731,41 @@ export class DatabaseService {
...
+          node(propName as string, 'Property', {
+            active: true,
+          }),
+        ],
+      ])
+      .return('count(r) as total')
+      .run();
+
+    const totalNumber = result[0]?.total || 0;
+    const isUnique = totalNumber == 1 ? true : false;
```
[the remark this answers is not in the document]
ibohalsinghkoijam: Changes are done as per @CarsonF comment. Thank @CarsonF

src/core/database/database.service.ts
```diff
@@ -731,4 +731,41 @@ export class DatabaseService {
...
+          }),
+          node(propName as string, 'Property', {
+            active: true,
+          }),
+        ],
+      ])
+      .return('count(r) as total')
+      .run();
+
+    const totalNumber = result[0]?.total || 0;
```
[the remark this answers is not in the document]
ibohalsinghkoijam: Changes are done as per @CarsonF comment.

PR #401, opened 2020-04-17
Cookie Sessions & Initial CORS config

### LINKED ISSUE #399: Cookie session logic
In order to support frontend cookie auth on the backend:
* Create a cookie domain env variable
* Update `createSession` mutation to take a `browser: boolean` argument. If it's true, set cookie with cookie env variable
* Update auth middleware to read either the session cookie _or_ the `Authorization` header
* Set the cookie to never expire
* `httpOnly: true`
* Setup CORS

richardgirges:
Closes #399 

I haven't set up any CORS `Access-Control-Allow` headers yet, figured we'd need to have discussions on what we want to allow first.

Two new environment variables introduced in this PR:
* `SESSION_COOKIE_NAME` - can probably be left alone, defaults to `cordsession`
* `SESSION_COOKIE_DOMAIN` - MUST be set on all deployed environments! Must always have a leading `.` to enable cross-domain cookies. Example value: `.cordfield.com`


src/components/authentication/authentication.resolver.ts
```diff
@@ -13,14 +16,31 @@ import { AuthenticationService } from './authentication.service';
...
+    private readonly userService: UserService,
+    private readonly config: ConfigService
   ) {}
 
   @Mutation(() => CreateSessionOutput, {
     description: 'Create a session',
   })
-  async createSession(): Promise<CreateSessionOutput> {
+  async createSession(
+    @Args('browser') browser: boolean,
```
CarsonF, line 27: Should default to....false?

src/components/authentication/authentication.resolver.ts
```diff
@@ -13,14 +16,31 @@ import { AuthenticationService } from './authentication.service';
...
+      // http cookies must have an expiration in order to persist, so we're setting it to 10 years in the future
+      const expires = DateTime.local().plus({ years: 10 }).toJSDate();
+
+      res.cookie(this.config.sessionCookieName, token, {
+        expires,
+        httpOnly: true,
+        path: '/',
+        domain: this.config.sessionCookieDomain,
+      });
+    }
```
CarsonF, line 42: Needs return so token is excluded

src/core/config/config.service.ts
```diff
@@ -57,6 +58,26 @@ export class ConfigService {
...
+    };
+  }
+
+  @Lazy() get sessionCookieName(): string {
+    return this.env.string('SESSION_COOKIE_NAME').optional('cordsession');
+  }
+
+  @Lazy() get sessionCookieDomain(): string | undefined {
+    return this.env.string('SESSION_COOKIE_DOMAIN').optional(undefined);
+  }
```
CarsonF, line 79: Will you combine these into `session()`?

src/core/config/config.service.ts
```diff
@@ -57,6 +58,26 @@ export class ConfigService {
...
+      credentials: true,
+    };
+  }
+
+  @Lazy() get sessionCookieName(): string {
+    return this.env.string('SESSION_COOKIE_NAME').optional('cordsession');
+  }
+
+  @Lazy() get sessionCookieDomain(): string | undefined {
+    return this.env.string('SESSION_COOKIE_DOMAIN').optional(undefined);
```
CarsonF, line 78: FYI `optional()` would work - no need to be explicit with the undefined

CarsonF:
I've been trying to not to explicitly call out env names. But I've been thinking about having different defaults for each env or at least dev/deployed.
We could default to w/e for dev and say `.required()` for prod/deployed

src/components/authentication/authentication.resolver.ts
richardgirges: Hmm, one thing I didn't anticipate here is we can't blindly return in this resolver. Because the `CreateSessionOutput` is a required field in graphql. What do you think the best way to go about handling this is?

CarsonF: Yeah we should change that DTO to make that property optional and a description saying it's returned if browser is false. A description on the browser arg would be good too.

richardgirges:
> > I haven't set up any CORS `Access-Control-Allow` headers yet, figured we'd need to have discussions on what we want to allow first.
> 
> Those default to the headers requested. I think it's fine to not restrict it.
> 
> > * `SESSION_COOKIE_DOMAIN` - MUST be set on all deployed environments! Must always have a leading `.` to enable cross-domain cookies. Example value: `.cordfield.com`
> 
> Maybe we should enforce the leading period in code then?
> Is it safe to leave it undefined for localhost?

How would we go about enforcing it in the code? 

It's safe to have it undefined for localhost - it's actually intentional to keep it undefined because we don't want the cookie to be tied to a domain in localhost mode

CarsonF:
> How would we go about enforcing it in the code?

I just mean add it if it's not given

> It's safe to have it undefined for localhost - it's actually intentional to keep it undefined because we don't want the cookie to be tied to a domain in localhost mode

Ok cool. I think it's fine to default to dev presets for now. We do the same thing with `EMAIL_SEND`.

src/components/authentication/authentication.resolver.ts
richardgirges: Sounds good 👍

richardgirges:
> > How would we go about enforcing it in the code?
> 
> I just mean add it if it's not given

Got it! just pushed up this update in the ConfigService. 

Also, just fixed all the test errors that were related to the cookie logic. Appears a lot of tests are currently failing regardless of that in master though.

src/core/config/config.service.ts
```diff
@@ -70,12 +70,25 @@ export class ConfigService {
...
+  } {
+    const cookieName = this.env
+      .string('SESSION_COOKIE_NAME')
+      .optional('cordsession');
+
+    let cookieDomain = this.env.string('SESSION_COOKIE_DOMAIN').optional();
+
+    // prepend a leading "." to the domain if one doesn't exist, to ensure cookies are cross-domain-enabled
+    if (cookieDomain && !cookieDomain.startsWith('.')) {
+      cookieDomain += '.' + cookieDomain;
```
CarsonF, line 85: herp derp
```suggestion
      cookieDomain = '.' + cookieDomain;
```

CarsonF:
Master looks green. Do you need to rebase on latest?

richardgirges:
Weird, locally in master I'm seeing `create budget` test errors, and I'm also seeing them on CI. Let me rebase for good measure.

richardgirges:
Looks like rebase did it. Master is still failing for me locally though (not a huge deal as it doesn't block this PR)

PR #404, opened 2020-04-21
Convert createSession mutation to session query

richardgirges:
* Convert the `createSession` mutation to the `session` query
* Return a nullable `user` object in `session` query, for the currently-logged-in user

### DIFF
--- a/src/components/authentication/authentication.resolver.ts
+++ b/src/components/authentication/authentication.resolver.ts
@@ -1,29 +1,32 @@
-import { Args, Context, Mutation, Resolver } from '@nestjs/graphql';
-import { Response } from 'express';
+import { Args, Context, Mutation, Query, Resolver } from '@nestjs/graphql';
+import { Request, Response } from 'express';
 import { DateTime } from 'luxon';
 import { ISession, Session } from '../../common';
 import { ConfigService } from '../../core';
 import { UserService } from '../user';
 import {
-  CreateSessionOutput,
   LoginInput,
   LoginOutput,
   ResetPasswordInput,
+  SessionOutput,
 } from './authentication.dto';
 import { AuthenticationService } from './authentication.service';
+import { SessionPipe } from './session.pipe';
 
 @Resolver()
 export class AuthenticationResolver {
   constructor(
     private readonly authService: AuthenticationService,
     private readonly userService: UserService,
-    private readonly config: ConfigService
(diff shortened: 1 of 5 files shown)


src/components/authentication/authentication.resolver.ts
```diff
@@ -33,8 +36,15 @@ export class AuthenticationResolver {
       defaultValue: false,
     })
     browser?: boolean
-  ): Promise<CreateSessionOutput> {
-    const token = await this.authService.createToken();
+  ): Promise<SessionOutput> {
+    // TODO: the 'browser' ternary check here isn't necessary but the e2e tests fail without it due to the cookieParser middleware not being loaded in the e2e app module
+    const existingToken =
+      this.sessionPipe.getTokenFromAuthHeader(req) ||
+      (browser ? this.sessionPipe.getTokenFromCookie(req) : null);
```
CarsonF, line 43: Is it just `cookies` not existing on request? Could you just change the impl to this?
```ts
  getTokenFromCookie(req: Request): string | undefined {
    return req.cookies?.[this.config.session.cookieName];
  }
```

test/utility/create-session.ts
```diff
@@ -3,13 +3,13 @@ import { TestApp } from './create-app';
 
 export async function createSession(app: TestApp): Promise<string> {
   const result = await app.graphql.mutate(gql`
```
CarsonF, line 5:
```suggestion
  const result = await app.graphql.query(gql`
```

src/components/authentication/authentication.resolver.ts
richardgirges: good call! this will work

src/components/authentication/authentication.resolver.ts
```diff
@@ -33,8 +36,15 @@ export class AuthenticationResolver {
       defaultValue: false,
     })
     browser?: boolean
-  ): Promise<CreateSessionOutput> {
-    const token = await this.authService.createToken();
+  ): Promise<SessionOutput> {
+    // TODO: the 'browser' ternary check here isn't necessary but the e2e tests fail without it due to the cookieParser middleware not being loaded in the e2e app module
```
CarsonF, line 40:
```suggestion
```

richardgirges: Whoops!

PR #406, opened 2020-04-21
consistency checker for project nodes

gowthamrodda:


michaelmarshall:
@gowthamrodda How is this different then 371?

PR #428, opened 2020-04-29
fixed unique org name issue

aaronkchsu:
Organization name should be unique globally.

### DIFF
--- a/src/components/organization/organization.service.ts
+++ b/src/components/organization/organization.service.ts
@@ -66,13 +66,14 @@ export class OrganizationService {
         .raw(
           `
           MATCH(org:OrgName {value: $name}) return org
-          `, {
-            name: input.name
+          `,
+          {
+            name: input.name,
           }
         )
-        .first()
+        .first();
 
-      if ( checkOrg ) {
+      if (checkOrg) {
         throw new NotFoundException(
           'Organization name should be unique globally.'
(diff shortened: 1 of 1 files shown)


src/components/organization/organization.service.ts
```diff
@@ -61,6 +61,23 @@ export class OrganizationService {
...
+          `, {
+            name: input.name
+          }
+        )
+        .first()
+
+      if ( checkOrg ) {
+        throw new NotFoundException(
+          'Organization name should be unique globally.'
+        );
```
CarsonF, line 79:
```suggestion
        throw new BadRequestException(
          'Organization with that name already exists.'
          'Duplicate',
        );
```

src/components/organization/organization.service.ts
```diff
@@ -61,6 +62,25 @@ export class OrganizationService {
...
+          }
+        )
+        .first();
+
+      if (checkOrg) {
+        throw new BadRequestException(
+          'Organization with that name already exists.',
+          'Duplicate'
+        );
+      }
```
CarsonF, line 82: Actually this will get caught by the catch below and change the error type. So it needs to be moved out of the try/catch

aaronkchsu: Ah, correct.

PR #432, opened 2020-05-01
fixes ceremony consistency with unit tests

gowthamrodda:


src/components/engagement/engagement.service.ts
```diff
@@ -546,7 +546,12 @@ export class EngagementService {
       };
 
       const ceremony = await this.ceremonyService.create(
-        { type: CeremonyType.Dedication },
+        {
+          type: CeremonyType.Dedication,
+          planned: true, // yet to know how to derive this value
+          estimatedDate: CalendarDate.local(), // yet to know how to derive this value
+          actualDate: CalendarDate.local(), // yet to know how to derive this value
```
CarsonF, line 553: Those are nullable. Omit them here.

gowthamrodda: sure, so I am not assigning those values here. Consistency check would check just `type` now.

PR #458, opened 2020-05-14
Refactor project service to use security groups and permissions

ibohalsinghkoijam:


src/components/project/project.service.ts
```diff
@@ -74,105 +75,201 @@ export class ProjectService {
...
-          requestingUserId: session.userId,
-          owningOrgId: session.owningOrgId,
-          id,
-        }
-      )
-      .first();
+        relation('out', '', 'baseNode', { active: true }),
+        node('project'),
+        relation('out', '', 'location', { active: true }),
+        node('location', 'location', { active: true }),
```
gowthamrodda, line 190: does we have `location` label ? I guess its `Country`

ibohalsinghkoijam: Yes, I changed to Country

PR #472, opened 2020-05-20
Better fake data

CarsonF:


test/utility/create-country.ts
```diff
@@ -10,7 +11,7 @@ export async function createCountry(
   input: Partial<CreateCountry> = {}
 ) {
   const country: CreateCountry = {
-    name: 'Country' + generate(),
```
candleshine, line 13: the reason this is here is because there aren't enough fake countries for us to accidentally create countries with the same name in the test suite.

michaelmarshall:
closing. cleaning up stale PRs.

PR #495, opened 2020-05-23
fix internships

gowthamrodda:
Fixes - update and consistency


src/core/database/database.service.ts
```diff
@@ -577,6 +578,22 @@ export class DatabaseService {
     };
   }
 
+  getPropLabel(key: string) {
```
candleshine, line 581: move this to engagement service after the DB-service call.

PR #496, opened 2020-05-25
directory lazy-field for project

aaronkchsu:


CarsonF:
Hey @aaronkchsu we actually want `rootDirectory` to return a `Directory` object. We can fetch the children from that and graphql will handle that nested call.

aaronkchsu:
So we will only return Directory object in this call? 
Then I am not sure about the response structure.
rootDirectory {
     files {
          items {
          }
     }
}

CarsonF:
Yep just return a Directory from the resolver. That is the response structure.

src/components/file/dto/list.ts
```diff
@@ -25,6 +25,12 @@ export abstract class FileFilters {
     nullable: true,
   })
   readonly category?: readonly FileNodeCategory[];
+
+  @Field(() => ID, {
+    description: 'Only childrens in root directory of this project',
+    nullable: true,
+  })
+  readonly projectId?: string;
```
CarsonF, line 33: This isn't needed

PR #498, opened 2020-05-26
(WIP)File Service Update With Parent Field

aaronkchsu:
In this commit, I created a query getting parent directories which is created serially.


CarsonF:
This is waiting on #500 and needs to be rebased on top of once merged.

PR #500, opened 2020-05-28
Files Rewrite

CarsonF:
This rewrites most of the file module. Most of the API was changed. Most of the db queries rewritten. Many workflow holes were plugged. There's only a couple TODOs left and two list methods for getting children/parents.

I've also mapped out defined files (f0da92e3b9a2f070f84ae5151c1d67afef47e115), something we've wanted to implement for a long time. This first usage of this is in Partnerships which now have `mou` and `agreement` files (2970bb36aec0de1d9447d8e4a23ded0dd76853aa).


CarsonF:
Ok I overwrote the changes in master for file queries. They will need to be applied again for the updated queries in `FileRepository`.

PR #507, opened 2020-06-01
list of my projects

gowthamrodda:


src/components/project/project.service.ts
```diff
@@ -377,6 +378,37 @@ export class ProjectService {
       },
     });
 
+    if (filter.mine) {
```
CarsonF, line 381: Filters need to be applied at the database level so pagination is accurate.

test/utility/create-file.ts
```diff
@@ -63,7 +63,6 @@ export async function createFileVersion(
   const file: CreateFileVersionInput = {
     uploadId: input.uploadId,
     parentId: input.parentId,
-    ...input,
```
CarsonF, line 66: Why?

src/components/project/project.service.ts
gowthamrodda: yes done

test/utility/create-file.ts
gowthamrodda: reverted :)

PR #508, opened 2020-06-02
Created sgCreateNode function in db Service

punamasawa777:
Created sgCreateNode Service


src/components/organization/organization.service.ts
```diff
@@ -118,7 +118,7 @@ export class OrganizationService {
     ];
   };
 
-  async create(
+  async creates(
```
candleshine, line 121: did you leave the old version in for a reason?  Please remove this.  If we need the history we can use git history.

PR #509, opened 2020-06-02
Updates of file service for parents and children lazy field

aaronkchsu:


CarsonF:
@aaronkchsu I added tests for expected functionality. Please make them pass.

src/components/file/file.service.ts
```diff
@@ -138,13 +138,18 @@ export class FileService {
...
   ): Promise<FileListOutput> {
     const result = await this.repo.getChildrenById(session, parentId, input);
-    const items = await Promise.all(
+    let items = await Promise.all(
       result.children.map((node) => this.adaptBaseNodeToFileNode(node, session))
     );
+    if (input?.filter?.category)
+      items = items.filter((item) =>
+        input.filter.category!.includes(item.category)
+      );
```
CarsonF, line 147: Filters need to be applied in the query, otherwise `total` and pagination is invalid

src/components/file/file.service.ts
```diff
@@ -138,13 +138,18 @@ export class FileService {
...
+
     return {
       items: items,
       total: result.total,
-      hasMore: result.hasMore,
+      hasMore: false,
```
CarsonF, line 152: This can be calculated from pagination and total, see logic in db service.

test/file.e2e-spec.ts
```diff
@@ -308,27 +308,27 @@ describe('File e2e', () => {
         page: 1,
       });
       expect(firstPage.total).toEqual(expectedTotalChildren);
-      expect(firstPage.hasMore).toBeTruthy();
+      expect(firstPage.hasMore).toBeFalsy();
```
CarsonF, line 311: Other than the `beforeAll` -> `beforeEach`, nothing should be changed here.

CarsonF: Actually `expectedTotalFiles` should include `expectedTotalVideos`. Sorry about that.

src/components/file/file.repository.ts
```diff
@@ -93,20 +93,18 @@ export class FileRepository {
...
+      .limit(input?.count ?? 25);
     const result = await query.run();
-
     const total = result.length === 0 ? 0 : result[0].total;
+    let children = result.map(baseNodeFromResult);
 
-    const children = result.map(baseNodeFromResult);
-
+    if (input?.filter?.type)
+      children = children.filter((item) => item.type === input.filter.type);
```
CarsonF, line 104: This filter also needs to be applied in the query.
You can swap `FileNode` label on line 81 with the filter type if given.

src/components/file/file.repository.ts
```diff
@@ -93,20 +93,18 @@ export class FileRepository {
           createdBy: [{ id: 'createdById' }],
         },
       ])
-      .orderBy([`${input.sort} ${input.order}`])
-      .skip((input.page - 1) * input.count)
-      .limit(input.count);
-
+      .orderBy([`${input?.sort ?? 'name'} ${input?.order ?? 'ASC'}`])
```
CarsonF, line 96: This is ordering by the name _node_. You want to order by the name _value_, if `sort` is _name_.

I did say order didn't on slack. But now I'm thinking we have to at least have one order so results are consistent back otherwise index-based pagination won't work.

Issue #474, opened 2020-05-20
DB service read one function needs to reflect the security group


candleshine:
Now that we know how to read a node with the security group permissions, we need a db-service function to read the node and its properties correctly.  

rewrite the org service read function to use this new db-service function.

PR #512, opened 2020-06-03
Created sgReadOne function in db Service 

punamasawa777:
Created sgReadOne function in db Service #474

### DIFF
--- a/src/components/organization/organization.service.ts
+++ b/src/components/organization/organization.service.ts
@@ -1,11 +1,9 @@
 import {
   BadRequestException,
-  ForbiddenException,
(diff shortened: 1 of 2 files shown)


src/components/organization/organization.service.ts
```diff
@@ -88,54 +86,15 @@ export class OrganizationService {
   }
 
   async readOne(orgId: string, session: ISession): Promise<Organization> {
-    const result = await this.db
```
candleshine, line 91: You can't remove this yet, there are services that have not migrated to the security group model and are still looking for ACL.

Issue #483, opened 2020-05-20
Favorite API - Module, Service, Resolvers, DTOs


michaelmarshall:
The DB schema has been updated to include favorites. A favorite is simply a relationship between a `:User` and a `:BaseNode`. 

Create an API that lets the user add a `:favorite` relationship to anything. CRUD + List.

Get with Michael for a knowledge transfer before starting.

candleshine:
Each favorite has an `id` so that we can list and delete them as necessary.  
I do not see a use case for `Read` or `Update` please correct me if I'm wrong @michaelmarshall 

So just `add(baseNodeId)`can only add to session.userId, `delete(baseNodeId)` session.userId checked to be the owner of the favorite.  list(baseNodeId) can only list own favorites.

michaelmarshall:
@candleshine makes sense. thanks.

PR #513, opened 2020-06-05
Favorite API - Module, Service, Resolvers, DTOs #483

punamasawa777:


src/components/favorites/dto/list-favorite.dto.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ValidateNested } from 'class-validator';
+import {
+  PaginatedList,
+  SecuredList,
+  SortablePaginationInput,
+} from '../../../common';
+import { Favorite } from './favorite';
+
+@InputType()
+export abstract class FavoriteFilters {
```
candleshine, line 12: I can see no use case for this search filter

candleshine: I would search for filters belonging to one User.  Then I might filter it by type of BaseNode.  For Example.  Favorite Locations, Favorite Projects, Favorite Languages.  
These options might be an enum, that is not every existing type.  
The ones I would suggest would be 

- Project
Language
Organization
Location
Film
Story
LiteracyMaterial
User

All that to say, If we require this, It should be a new ticket, not this ticket.  

I would remove the baseNodeId as a filter

PR #531, opened 2020-06-12
root sg member can create node 

gowthamrodda:
A root SG member can create baseNode, as root user is member of RSG, and RSG has permission to Root Permission with read and edit  rights to its baseNode and its properties. 
(reqUser)<-[:member]-[:RSG]-[:Permission]-[:basenode]->(:BaseNode)-[:property]->(:Property)

### DIFF
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -1250,23 +1250,37 @@ export class DatabaseService {
     aclEditProp?: string;
     sgName: string;
   }) {
+    const id = generate();
+    const createdAt = DateTime.local();
+    const nodeName = upperFirst(nodevar);
+    const aclEditPropName = aclEditProp || `canEdit${nodeName}`;
+    const baseNode = nodeName + ':BaseNode';
+    const properties = [];
+    const permissions = [];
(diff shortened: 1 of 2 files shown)


src/core/database/database.service.ts
```diff
@@ -1278,25 +1302,7 @@ export class DatabaseService {
...
-            node('readerSG', 'SecurityGroup', {
-              active: true,
-              createdAt,
-              name: sgName + ' users',
-            }),
-            relation('out', '', 'member', { active: true, createdAt }),
-            node('requestingUser'),
-          ],
-          ...permissions,
+          ...permissionQueries,
```
candleshine, line 1305: This does not work. You still need to create the security group and the permission nodes.  This removes them entirely.  A node that a root user creates still operates like a normal node.  
It is the matchSession call
``` 
     const query = this.db
        .query()
        .match(matchSession(session, { withAclEdit: aclEditPropName }))
```
that cannot fail when the user is a root user.  If this is already true, then leave it alone.

gowthamrodda: Created RSG and Permission nodes.

PR #534, opened 2020-06-16
Reduce number of object created to execute list view test cases

ibohalsinghkoijam:


src/components/budget/budget.service.ts
```diff
@@ -215,10 +215,10 @@ export class BudgetService {
...
 
     if (!result) {
-      this.logger.error(`Could not find budget:  `, {
-        id,
-        userId: session.userId,
-      });
+      // this.logger.error(`Could not find budget:  `, {
+      //   id,
+      //   userId: session.userId,
+      // });
```
CarsonF, line 221: This could be left and changed to a warning. Otherwise delete it; don't leave commented out code.
```suggestion
     this.logger.warning(`Could not find budget`, {
        id,
        userId: session.userId,
      });
```

ibohalsinghkoijam: Thank you @CarsonF

PR #578, opened 2020-06-22
The first budget status 'current' with the project creation

ibohalsinghkoijam:
…he project


CarsonF:
This just changes the default value. This will need to be reverted but it can wait until after I get event handling in.

PR #580, opened 2020-06-22
Bug Org Filter Fixed

punamasawa777:


punamasawa777:
Without Filters -- Org List

Screenshot 2020-06-22 at 11 45 59 PM


And With Filters -- Org List

Screenshot 2020-06-22 at 11 49 03 PM

michaelmarshall:
Thanks @punamasawa777 but it looks like the other org list refactor got in first. I'll close this now.

PR #615, opened 2020-06-26
createPartnership returns nulls for mouStart and mouEnd

ibohalsinghkoijam:


src/components/partnership/partnership.service.ts
```diff
@@ -241,6 +244,25 @@ export class PartnershipService {
         })
         .first();
 
+      const fiscalRange = fiscalYears(input.mouStart, input.mouEnd);
```
michaelmarshall, line 247: please add some comments to explain what and why this code is doing.

michaelmarshall:
thanks for the PR @ibohalsinghkoijam . I'm going to start asking for comments in the code so we can help others understand why we are doing things. I just added one request. I also noticed tests are failing. please check that too.

src/components/partnership/partnership.service.ts
candleshine: this is old code that's been in there the whole time.  It calculates which fiscal years are represented by the mouStart to mouEnd.  IE Dec 1 2020 is in FY 2021

candleshine: added comment

candleshine:
The broken test in project reveals a bug elsewhere in the system.  Not in this ticket.  I'm going to skip that one.

Issue #596, opened 2020-06-24
Sync Project & Partnership Start/End Dates


CarsonF:
- Project mouStart and mouEnd should set the values of all partnership's mouStart and mouEnd that are null.
- Also partnerships start/end should default to project start/end if already set.

CarsonF:
There's more too for updating existing partnership dates based on project status. I can't remember of the top of my head. @sethmcknight feel free to help doc if you can.

CarsonF:
Thinking about it more I think there's a better way. I think really we want to express that:
1. this partnership matches the project's date range 
2. this partnership is a subset of the project's date range. i.e. ends before the project does or starts after the project does

Therefore we could have the null values represent _inheriting_ from the project's range, and actual values be the "overrides".
Then we would need computed properties to determine the current value of the range - the override values or project's values.

Taking this approach means we don't have to figure out all the edge cases in the _assumptions_ of what the user what's do. Which is inherently fragile.

CarsonF:
I've done something similar in my language-api branch.
I have a [calculated `population` field](https://github.com/SeedCompany/cord-api-v3/blob/2134ff8e95822b2d499dc2fed9e68714f8e08485/src/components/language/language.resolver.ts#L64-L82) that uses the [`populationOverride` field](https://github.com/SeedCompany/cord-api-v3/blob/2134ff8e95822b2d499dc2fed9e68714f8e08485/src/components/language/dto/language.dto.ts#L43-L46) if defined or it fallsback to the `ethnoluge.population` field.

We could have the same pattern here:
- `mouStartOverride` & `mouEndOverride` fields
- calculated `mouStart` & `mouEnd` fields which use those override fields or fallback to the project's `mouStart/mouEnd`.

sethmcknight:
> I've done something similar in my language-api branch.
> I have a [calculated `population` field](https://github.com/SeedCompany/cord-api-v3/blob/2134ff8e95822b2d499dc2fed9e68714f8e08485/src/components/language/language.resolver.ts#L64-L82) that uses the [`populationOverride` field](https://github.com/SeedCompany/cord-api-v3/blob/2134ff8e95822b2d499dc2fed9e68714f8e08485/src/components/language/dto/language.dto.ts#L43-L46) if defined or it fallsback to the `ethnoluge.population` field.
> 
> We could have the same pattern here:
> 
> * `mouStartOverride` & `mouEndOverride` fields
[...]

PR #622, opened 2020-06-29
partnership.organization.id is null

ibohalsinghkoijam:
Assumed that partnership's mouStart and mouEnd should default to project start/end as per #596


michaelmarshall:
Spoke with Koijam. He's still working on this.

candleshine:
The order of the optionalMatches makes a difference.  He was able to get the organization ID if the mouStart and mouEnd are higher up.  :magic:  I don't know why it is, but I saw it work.

candleshine:
The order of the optionalMatches makes a difference.  For some reason when you order the mouStart and mouEnd up above the

PR #628, opened 2020-06-29
Trim name fields on input

### LINKED ISSUE #604: Organization >Create > Org can be created with just spaces
**Describe the bug**
Organization >Create > Org can be created with just spaces

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap and execute request.

**Expected Result**
Validation message should be returned as Org name has <=2 chars.

**Actual Result**
Org is created successfully

image

Commit # 835a471

zkhin:
Fixes #604

### DIFF
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -19,3 +19,4 @@ export * from './util';
 export { ISession, Session } from './session';
 export * from './types';
 export * from './validators';
+export * from './name-field';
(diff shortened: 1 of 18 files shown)


zkhin:
> We cannot be this strict on names. Current names include parentheses, periods, hyphens, etc.

Can you let me know the special characters that should be excluded in org name?
If you can write regex then that will be helpful.

zkhin:
> We cannot be this strict on names. Current names include parentheses, periods, hyphens, etc.

Thanks for looking at it. Where could I find a full list of constraints or should it be just beginning with a length of > 2 non-space characters? I might have to also check on how the graphql injection security is working first.

CarsonF:
Yeah I think we just need to trim the value and then the 2+ char length validation is enough

Issue #611, opened 2020-06-26
Make org IDs public 


michaelmarshall:
A user who hasn't created an org needs to be able to create a partnership with that org. Currently org names are public but the IDs aren't able to be used as expected.

michaelmarshall:
to repro this, create a session:
![image](https://user-images.githubusercontent.com/1165460/86051718-7bac6380-ba1b-11ea-8796-011393c3b08f.png)

register a user:
![image](https://user-images.githubusercontent.com/1165460/86051752-85ce6200-ba1b-11ea-84e3-d43808390160.png)

create an org:
![image](https://user-images.githubusercontent.com/1165460/86051782-941c7e00-ba1b-11ea-8ac7-5a6d7d99b4c9.png)

create a project:
![image](https://user-images.githubusercontent.com/1165460/86052530-c1b5f700-ba1c-11ea-9db0-cb14f9b5ba91.png)

logout:
![image](https://user-images.githubusercontent.com/1165460/86052572-d09ca980-ba1c-11ea-80a9-fd3d192af27c.png)

register a new user:
![image](https://user-images.githubusercontent.com/1165460/86052596-db573e80-ba1c-11ea-821b-e1b94f4d2e96.png)

try to create a partnership with the org id and project id the previous user created:
![image](https://user-images.githubusercontent.com/1165460/86052759-19546280-ba1d-11ea-88d2-47618a749015.png)

PR #640, opened 2020-06-30
Issue 611: Make org ids public

zkhin:
https://github.com/SeedCompany/cord-api-v3/issues/611


michaelmarshall:
still not sure why changes like this are working, but they are. need to investigate later.

Issue #574, opened 2020-06-19
Project Initial Budget Status


CarsonF:
The first budget created with the project creation should have a status of _Current_ instead of _Pending_.

Need to pass parameter here to make that happen
https://github.com/SeedCompany/cord-api-v3/blob/67721f328dd20846b400e2ea090aeb1298799a58/src/components/project/project.service.ts#L714

CarsonF:
Only the first budget created for a project should be current. Subsequent budgets should be pending.

CarsonF:
After talking through this more I want to change the logic.
- Budgets should _always_ be created as a pending status
- Budget records are only editable if the budget is pending
- `ProjectService.currentBudget` should filter to the budget that is current (as it's doing now) but then fallback to the first pending budget (new logic).
  - Thus `project.budget` can either be a current budget or a pending one
- The pending budget should be set to active when the project gets set to active

This way isolates budget mutably logic from project logic, and the only connection point is their statuses. This is easier to reason about.

candleshine:
@CarsonF Budget does not have an Active Status.  The enum is 
```
export enum BudgetStatus {
  Pending = 'Pending',
  Current = 'Current',
  Superceded = 'Superceded',
  Rejected = 'Rejected',
}
```

Also there is no field for status in UpdateProject DTO

CarsonF:
Sorry I meant _current_.
Yes project status gets changed through a workflow API, which still needs to be addressed. That specific logic could wait to be done later.

candleshine:
@ibohalsinghkoijam do not change updateProject DTO.

PR #641, opened 2020-07-01
Project Initial Budget Status

ibohalsinghkoijam:


src/components/project/project.service.ts
```diff
@@ -792,6 +805,32 @@ export class ProjectService {
...
+
+    const pendingBudget = budgets.items.find(
+      (b) => b.status === BudgetStatus.Pending
+    );
+    //574 -The pending budget should be set to active i.e Current when the project gets set to active
+    if (
+      (changes.status.includes(ProjectStatus.InDevelopment) ||
+        changes.status.includes(ProjectStatus.Pending)) &&
+      pendingBudget?.status.includes(BudgetStatus.Pending)
+    ) {
```
CarsonF, line 825: I know this will be changing with the event code. But until then, it's when the project becomes _active_. The comment is correct but the logic is wrong.
```suggestion
    //574 -The pending budget should be set to active i.e Current when the project gets set to active
    if (changes.status.includes(ProjectStatus.Active) && pendingBudget) {
```

ibohalsinghkoijam: Thank you @CarsonF for correcting logic.

src/components/budget/budget.service.ts
```diff
@@ -356,6 +356,39 @@ export class BudgetService {
...
+      },
+    ]);
+
+    const readProject = await projectStatusQuery.first();
+    //Budget records are only editable if Project.status not active - ProjectStatus.Active
+    if (readProject?.status.includes('Active')) {
+      throw new BadRequestException(
+        'budget of active project can not be modified '
+      );
+    }
```
CarsonF, line 390: A main point of #574 is to separate budget changes from project status.
Since `updateBudget` only updates the status and we'll be doing that via workflow, I think this API can just go away. And that just leaves this method as a helper for internal use.

Where budgets really change is on their records. `updateRecord` is what needs to check for whether the budget status is _Pending_.

ibohalsinghkoijam: ok, I will change that.

ibohalsinghkoijam: done the changes

Issue #647, opened 2020-07-01
Enforce created engagements match their project types


CarsonF:
- LanguageEngagements can only be created on TranslationProjects
- InternshipEngagements can only be created on InternshipProjects

PR #652, opened 2020-07-01
647 - Check project type that matches with engagement type

zkhin:
https://github.com/SeedCompany/cord-api-v3/issues/647

- LanguageEngagements can only be created on TranslationProjects
- InternshipEngagements can only be created on InternshipProjects

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -25,6 +25,7 @@ import {
   ProductService,
   SecuredProductList,
 } from '../product';
+import { ProjectType } from '../project/dto/type.enum';
 import { UserService } from '../user';
 import {
   CreateInternshipEngagement,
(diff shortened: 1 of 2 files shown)


src/components/engagement/engagement.service.ts
```diff
@@ -1560,4 +1575,13 @@ export class EngagementService {
...
     );
   }
+
+  protected async getProjectTypeById(projectId: string): Promise<string> {
+    const qr = `
+    MATCH (p:Project {id: $projectId, active: true}) RETURN p.type as type
+    `;
+    const results = await this.db.query().raw(qr, { projectId }).first();
+
+    return results?.type;
```
CarsonF, line 1585:
```suggestion
    return results?.type as ProjectType | undefined;
```
This is better and more correct than `string`

zkhin: Yes, I've changed.

src/components/engagement/engagement.service.ts
```diff
@@ -984,6 +977,13 @@ export class EngagementService {
       }
       throw new ServerException('Could not create Internship Engagement');
     }
+    // InternshipEngagements can only be created on InternshipProjects
+    const projectType = await this.getProjectTypeById(projectId);
+
+    if (projectType && projectType !== ProjectType.Internship) {
+      throw new BadRequestException('That Project type is not Intership');
+    }
```
CarsonF, line 985: Why'd you move this? It was fine where it was before. Currently this will give an error back even though the db has been changed.

zkhin: Oh, yes. You are right. There was another confusion lol. I changed again and fixed test case.

Issue #648, opened 2020-07-01
InternshipEngagement.methodologies should return empty array by default instead of error


CarsonF:
![Screen Shot 2020-07-01 at 11 34 47 AM](https://user-images.githubusercontent.com/932566/86269167-fc3fa100-bb8e-11ea-95da-920686559730.png)

API version: 1caff78ebd6db5aae49bd630c1fd8a769f4e120e

PR #658, opened 2020-07-02
648 - InternshipEngagement.methodologies should return empty array by default instead of error

zkhin:
#648

### DIFF
--- a/test/engagement.e2e-spec.ts
+++ b/test/engagement.e2e-spec.ts
@@ -594,4 +594,31 @@ describe('Engagement e2e', () => {
       })
     ).rejects.toThrow('languageId is invalid');
   });
+
+  it('should return empty methodologies array even if it is null', async () => {
+    // Create InternshipEngagement without methodologies
+    const internshipEngagement = await createInternshipEngagement(app, {
+      projectId: project.id,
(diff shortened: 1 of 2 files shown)


zkhin:
> I know this src code will be refactored soon. Can you add this check to a test (existing or new)?

Okay let me check

zkhin:
@CarsonF I added a new test case for checking that it returns an empty array.

test/engagement.e2e-spec.ts
```diff
@@ -491,6 +497,58 @@ describe('Engagement e2e', () => {
...
+      `,
+      {
+        input: {
+          engagement: internshipEngagementInput,
+        },
+      }
+    );
+
+    const internshipEngagement: InternshipEngagement =
+      result.createInternshipEngagement.engagement;
```
CarsonF, line 531:
```suggestion
    // Create InternshipEngagement without methodologies
    const internshipEngagement = await createInternshipEngagement(app, {
      projectId: project.id,
      internId: intern.id,
      methodologies: [],
    });
```

zkhin: @CarsonF  Yes, that's a great idea. I changed

Issue #674, opened 2020-07-06
Project permission bugs


CarsonF:
Created a project, and requested details.
- [ ] name.canRead/Edit is false but the value is still returned
  - [ ] `canRead/Edit` should be true since I created the project & while it's status is in development
  - [ ] `value` should not be returned if `canRead` is false
- [ ] `location`, `mouStart`, `mouEnd`'s `canRead/Edit` should be true since I created the project & while it's status is in development
- [ ] `engagements.canCreate` should be true since I created the project & while it's status is in development

API: https://github.com/SeedCompany/cord-api-v3/commit/8b7caae7f7c03cca1ef71790cd57ead6b9116d15

![Screen Shot 2020-07-06 at 1 07 36 PM](https://user-images.githubusercontent.com/932566/86624904-c719c100-bf89-11ea-8c7b-c97c7744362f.png)

zkhin:
@CarsonF Should I have to return `canRead: true, canEdit: true` even if those (`name`, `location`, `mouStart`, `mouEnd`) fields' values are null since I created the project & while it's status is in development

CarsonF:
Yeah. Talking with @michaelmarshall too he says that we should just always create those fields with a default value of null if omitted from input during the create mutation. That way they are available to query later.

zkhin:
I think if the `value` is null, it doesn't create `Property` node so the permission fields are always `false`
If I try to create Property node with `null` value, it gives an error...

CarsonF:
> I think if the `value` is null, it doesn't create `Property` node so the permission fields are always `false`

Yeah that's my understanding too.

> If I try to create Property node with `null` value, it gives an error...

What error? from cypher? or our code?

zkhin:
Seems it's from cypher.
`Neo.ClientError.Schema.ConstraintValidationFailed`

CarsonF:
That's constraints we've placed. You might need to clear them as they've changed recently (they'll get re-added on server boot).
Michael says `call apoc.schema.assert({},{})` does the trick.

Of course this could be from a constraint we still have. Either way we need it removed.

zkhin:
![image](https://user-images.githubusercontent.com/4212008/86660474-1dd8d680-c02e-11ea-9174-9848f7bb3296.png)
I added this like above but it gives error
`Neo4jError: There is no procedure with the name `apoc.schema.assert` registered for this database instance. Please ensure you've spelled the procedure name 
[...]

PR #686, opened 2020-07-07
Fixes - 674: Project permission bug

zkhin:
Fixes :  https://github.com/SeedCompany/cord-api-v3/issues/674

### DIFF
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -106,9 +106,6 @@ export class ProjectService {
 
   // helper method for defining properties
   property = (prop: string, value: any) => {
-    if (!value) {
-      return [];
-    }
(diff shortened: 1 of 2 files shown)


src/components/project/project.service.ts
```diff
@@ -607,6 +604,7 @@ export class ProjectService {
       modifiedAt: DateTime.local(),
       ...input,
     };
+    const canEdit = createInput.status === ProjectStatus.InDevelopment;
```
CarsonF, line 607: All projects start with their status as in development. Whether the fields can currently be edited is based on this changing `status`, so it probably needs to be done on query. We currently haven't integrated workflow into this permission model so @michaelmarshall might want to chime in here.

Issue #639, opened 2020-06-30
Refactor Language API/DB


CarsonF:
Finish off db query changes needed to match new API 
https://github.com/SeedCompany/cord-api-v3/compare/refactor/language-api

PR #692, opened 2020-07-08
Issue - 639: Refactor/language api

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/639


src/components/language/ethnologue-language/ethnologue-language.service.ts
```diff
@@ -0,0 +1,367 @@
...
+  EthnologueLanguage,
+  UpdateEthnologueLanguage,
+} from '../dto';
+
+@Injectable()
+export class EthnologueLanguageService {
+  constructor(
+    private readonly db: DatabaseService,
+    private readonly config: ConfigService,
+    @Logger('ethnologue:language:service') private readonly logger: ILogger
```
CarsonF, line 28:
```suggestion
    @Logger('language:ethnologue:service') private readonly logger: ILogger
```
This is a namespaced name and we are in the LanguageModule.

src/components/language/ethnologue-language/ethnologue-language.service.ts
```diff
@@ -0,0 +1,367 @@
...
+          'id.value': input.id,
+          'code.value': input.code,
+          'provisionalCode.value': input.provisionalCode,
+          'name.value': input.name,
+          'population.value': input.population,
+        });
+      await query.run();
+    } catch (e) {
+      this.logger.error('e:>>', e);
+    }
```
CarsonF, line 365: Please don't swallow errors like this.
```suggestion
    } catch (e) {
      this.logger.error('update failed', { exception: e });
      throw new ServerException('Failed to update ethnologue language');
    }
```

test/utility/fragments.ts
```diff
@@ -360,17 +385,50 @@ export const languageEngagement = gql`
         beginFiscalYear {
           value
         }
-        ethnologueName {
+        isDialect {
```
CarsonF, line 388: This language object could reuse the language fragment.
```gql
language {
  ...language
}
```
Then add the fragment to the end of the string
```ts
export const languageEngagement = gql`
  fragment languageEngagement on LanguageEngagement {
    ...
  }
  ${language}
`;
```

src/components/language/language.service.ts
```diff
@@ -499,13 +639,62 @@ export class LanguageService {
...
+        }),
+        node('location'),
+      ])
+      .run();
+  }
 
   async removeLocation(
     _languageId: string,
     _locationId: string,
     _session: ISession
```
CarsonF: These were just prefixed with _ to make the linter happy. They can be renamed as soon as they are used.

src/components/language/language.service.ts
```diff
@@ -403,33 +490,97 @@ export class LanguageService {
...
+        node('project', 'Project', {
+          active: true,
+        }),
+      ])
+      .return({
+        project: [{ id: 'id' }],
+      })
+      .run();
+
+    this.logger.info(`Language project lists ${result.length}`);
```
CarsonF, line 508: Don't inject dynamic data into log messages. It makes it harder to track them in production.
```suggestion
    this.logger.debug(`list projects results`, { total: results.length });
```

src/components/language/language.service.ts
```diff
@@ -388,10 +453,32 @@ export class LanguageService {
...
+    );
+
     return {
-      items: [],
-      total: 0,
+      items: items,
+      total: items.length,
       hasMore: false,
       canCreate: true,
       canRead: true,
```
CarsonF: These should come from db.

PR #697, opened 2020-07-08
Fixed Optional properties to null for engagement and project service

punamasawa777:
1. Fixed Optional properties to null for engagement service
2. Added a Test case for the same
                                   
3. Fixed Optional properties to null for project service
4. Added a Test case for the same

5 Password Validation (Min length for password to 6 Character)


src/common/validators/password.validator.ts
```diff
@@ -0,0 +1,21 @@
...
+      constraints: [options],
+      validator: {
+        validate: (value) =>
+          typeof value === 'string' &&
+          isLength(value, { min: 6, max: undefined }),
+        defaultMessage: () => 'Password must have minimum 6 characters',
+      },
+    },
+    validationOptions
+  );
```
CarsonF, line 21: `@MinLength` decorator already exists

michaelmarshall: please update

punamasawa777:
@michaelmarshall , I assumed to have custom text for validation
i will make changes with @MinLength Decorator

Issue #695, opened 2020-07-08
Get all Locations > query locations($input : LocationListInput!) > Input - Filter is mandatory 


iPiyushPatel:
**Describe the bug**
Get all Locations > query locations($input : LocationListInput!) > Input - Filter is mandatory 

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
Location results are returned and Filter input shouldn't be mandatory. 

**Actual Result**
Getting "Expected parameter(s): filter" error.

GraphQL Query ( partnerships ):
![image](https://user-images.githubusercontent.com/5212585/86937324-bc635580-c15c-11ea-97d3-5398f6fcd43f.png)

**Query and Input File**
[locations_FilterError.txt](https://github.com/SeedCompany/cord-api-v3/files/4891569/locations_FilterError.txt)

CordAPIV3# 3dd7156

CarsonF:
It's a DB query error

zkhin:
@CarsonF If the filter input is not provided, should I have to return all locations?

CarsonF:
Exactly.

CarsonF:
Here's an example of a conditional where clause
https://github.com/SeedCompany/cord-api-v3/blob/dd8863a98b01a81e936235c7d9da879ec4a5f8ed/src/components/file/file.repository.ts#L103-L112

reneepe:
@zkhin I'm getting the following error.

![image](https://user-images.githubusercontent.com/38309285/87360759-c9a38a00-c538-11ea-9400-daa8eec56bdd.png)

CarsonF:
@reneepe That's your error. You've defined `$input` variable in your locations query (first line), but then have not provided it in the _query variables_ panel at the bottom.

You don't have to define the query with a variable, you could just hardcode the necessary parameters in the locations input (line 2). That's usually what I do.
For example,
```gql
query {
  location(id: "12asdf34") {
    ...
  }
}
```
Or in this case specifically
```gql
query {
  locations(input: { page: 1, count: 25 }) {
    ...
  }
}
```
All of those parameters are optional though so you could just omit them
```gql
query {
  locations {
    ...
  }
}
```

PR #698, opened 2020-07-08
Issue: 695 - Fix location list filter mandatory

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/695

### DIFF
--- a/src/components/location/location.service.ts
+++ b/src/components/location/location.service.ts
@@ -205,6 +205,7 @@ export class LocationService {
         `
           MATCH (location {active: true})-[:name {active:true }]->(name:Property {active: true})
           WHERE name.value CONTAINS $filter
+          AND ('Zone' IN labels(location) OR 'Region' IN labels(location) OR 'Country' IN labels(location))
           WITH COUNT(name) as total, name, location
           MATCH (location {active: true})-[:name {active:true }]->(name:Property {active: true})
           RETURN total, location.id as id, name.value as name
@@ -215,7 +216,7 @@ export class LocationService {
         {
           skip: (page - 1) * count,
           count,
-          filter: filter.name,
+          filter: filter.name ? filter.name : '',
         }
       )
       .run();
--- a/src/components/location/location.service.ts
+++ b/src/components/location/location.service.ts
@@ -203,20 +203,16 @@ export class LocationService {
       .query()
       .raw(
         `
-          MATCH (location {active: true})-[:name {active:true }]->(name:Property {active: true})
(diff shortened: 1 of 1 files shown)


src/components/location/location.service.ts
```diff
@@ -205,6 +205,7 @@ export class LocationService {
         `
           MATCH (location {active: true})-[:name {active:true }]->(name:Property {active: true})
           WHERE name.value CONTAINS $filter
+          AND ('Zone' IN labels(location) OR 'Region' IN labels(location) OR 'Country' IN labels(location))
```
CarsonF, line 208: I'm not going to harp on this right now, but the base node should have a `Location` label. That way we can select it in the match clause instead of having to filter here.

src/components/location/location.service.ts
```diff
@@ -205,6 +205,7 @@ export class LocationService {
         `
           MATCH (location {active: true})-[:name {active:true }]->(name:Property {active: true})
           WHERE name.value CONTAINS $filter
```
CarsonF, line 207:
```suggestion
          WHERE name.value CONTAINS ''
```
This may work but is it the best?

src/components/location/location.service.ts
```diff
@@ -215,7 +216,7 @@ export class LocationService {
         {
           skip: (page - 1) * count,
           count,
-          filter: filter.name,
+          filter: filter.name ? filter.name : '',
```
CarsonF, line 219: FYI
```suggestion
          filter: filter.name ?? '',
```

src/components/location/location.service.ts
zkhin: If we don't use this clause, it will fetch all nodes which `Property.name` contains `filter` and call `readOne` function. And in this `readOne` function if the label doesn't match with [`Country`, `Region`, `Zone`], it will throw `BadRequest`.. so I added this clause in this query.

michaelmarshall: We will always need labels on match clauses to prevent a full DB scan. Labels are indexed.

michaelmarshall: `UNION` may be worth looking into. You can make 3 separate queries using labels that should provide better performance and do what you need it to.

CarsonF: `UNION` shouldn't be necessary. The base node can be created with all 3 labels, `BaseNode:Location:Country` and then we match for whatever the use case is.

michaelmarshall: that's right, forgot about :Location being a generic

PR #702, opened 2020-07-09
introduced mou override concept to partnerships. removed mouStart/End…

michaelmarshall:
… from update partnerships.
#596

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ lerna-debug.log*
 
 temp.cypher
 temp.graphql
+temp.json
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,3 @@ lerna-debug.log*
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
-
-temp.cypher
-temp.graphql
-temp.json
--- a/src/components/partnership/dto/update-partnership.dto.ts
+++ b/src/components/partnership/dto/update-partnership.dto.ts
@@ -25,10 +25,10 @@ export abstract class UpdatePartnership {
   readonly mouStatus?: PartnershipAgreementStatus;
 
   @DateField({ nullable: true })
-  readonly mouStart?: CalendarDate;
+  readonly mouStartOverride?: CalendarDate;
 
   @DateField({ nullable: true })
-  readonly mouEnd?: CalendarDate;
+  readonly mouEndOverride?: CalendarDate;
 
   @Field(() => [PartnershipType], { nullable: true })
   readonly types?: PartnershipType[];
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -533,24 +533,17 @@ export class PartnershipService {
     let mouEnd = null;
 
     // if user has access to project mou and there is no partnership override
-    if (result.projectMouStart && result.mouStartOveride !== null) {
+    if (result.canReadProjectMouStart && result.canReadMouStartOverride) {
       mouStart = result.mouStartOverride ?? result.projectMouStart;
     }
-    if (result.projectMouEnd && result.mouEndOverride !== null) {
+    if (result.canReadProjectMouEnd && result.canReadMouEndOverride) {
       mouEnd = result.mouEndOverride ?? result.projectMouEnd;
     }
 
-    let canReadMouStart = false;
-
-    if (result.canReadProjectMouStart && result.canReadMouStartOverride) {
-      canReadMouStart = true;
-    }
-
-    let canReadMouEnd = false;
-
-    if (result.canReadProjectMouEnd && result.canEditMouEndOverride) {
-      canReadMouEnd = true;
-    }
+    const canReadMouStart =
+      result.canReadProjectMouStart && result.canReadMouStartOverride;
+    const canReadMouEnd =
+      result.canReadProjectMouEnd && result.canEditMouEndOverride;
 
     return {
       id,
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -426,6 +426,12 @@ export class PartnershipService {
         },
       ],
       projectMouEnd: [{ value: 'projectMouEnd' }],
+      projectMouEndPerm: [
+        {
+          read: 'canReadProjectMouEnd',
+          edit: 'canEditProjectMouEnd',
+        },
+      ],
       partnership: [{ id: 'id', createdAt: 'createdAt' }],
       agreementStatus: [{ value: 'agreementStatus' }],
       canReadAgreementStatus: [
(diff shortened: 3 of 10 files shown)


test/partnership.e2e-spec.ts
```diff
@@ -157,6 +157,51 @@ describe('Partnership e2e', () => {
     );
   });
 
+  it.only('update mou overrides partnership', async () => {
```
CarsonF, line 160: Shouldn't this just be apart of the update test?

src/components/partnership/partnership.service.ts
```diff
@@ -53,7 +53,8 @@ export class PartnershipService {
   // helper method for defining properties
   property = (prop: string, value: any) => {
     if (!value) {
-      return [];
+      // return [];
```
CarsonF, line 56:
```suggestion
```

src/components/partnership/partnership.service.ts
```diff
@@ -473,15 +560,25 @@ export class PartnershipService {
         canEdit: !!result.canEditMouStatus,
       },
       mouStart: {
-        value: result.mouStart,
+        value: mouStart,
         canRead: !!result.canReadMouStart,
         canEdit: !!result.canEditMouStart,
```
CarsonF, line 565: Where are these coming from?

src/components/partnership/partnership.service.ts
```diff
@@ -449,6 +521,21 @@ export class PartnershipService {
...
     }
 
+    let mouStart = null;
+    let mouEnd = null;
+
+    // if user has access to project mou and there is no partnership override
+    if (result.projectMouStart && result.mouStartOveride !== null) {
+      mouStart = result.mouStartOverride ?? result.projectMouStart;
+    } else {
+      this.logger.info(`user doesn't have access or there is no override`);
```
CarsonF, line 531: access to what? Also probably more of a debug message

src/components/partnership/dto/update-partnership.dto.ts
```diff
@@ -25,10 +25,10 @@ export abstract class UpdatePartnership {
   readonly mouStatus?: PartnershipAgreementStatus;
 
   @DateField({ nullable: true })
-  readonly mouStart?: CalendarDate;
+  readonly mouStartOverride?: CalendarDate;
 
   @DateField({ nullable: true })
-  readonly mouEnd?: CalendarDate;
+  readonly mouEndOverride?: CalendarDate;
```
CarsonF, line 31: These same changes should be made in `CreatePartnership` as well.

.gitignore
```diff
@@ -52,3 +52,4 @@ lerna-debug.log*
 
 temp.cypher
 temp.graphql
+temp.json
```
CarsonF, line 55: Is this necessary?

test/partnership.e2e-spec.ts
michaelmarshall: My preference is to keep tests small and just make more tests for each specific permutation.

.gitignore
michaelmarshall: well it helps me. I assume it doesn't bother anyone else. I've found that I'll accidently commit temp files and this just keeps it out of the repo.

test/partnership.e2e-spec.ts
CarsonF: Ok
```suggestion
  it('update mou overrides partnership', async () => {
```

.gitignore
CarsonF: FYI I put those in my global gitignore file so I don't have to keep adding them to each repos ignore.

test/partnership.e2e-spec.ts
michaelmarshall: forgot to remove that!

src/components/partnership/partnership.service.ts
michaelmarshall: I had changed that. Must have not commited.

src/components/partnership/dto/update-partnership.dto.ts
michaelmarshall: Updated

.gitignore
michaelmarshall: and that is a better idea. I'll change it.

src/components/partnership/partnership.service.ts
michaelmarshall: fixed.

src/components/partnership/partnership.service.ts
michaelmarshall: the partnership mou fields. not the overrides. I didn't see a reason to make more complex logic for this field. It behoves whoever does security for the partnership/project to ensure the user can read both the partnership and project mou fields.

CarsonF: This field is calculated, not stored, so it doesn't make sense to have security for it.
`canRead` should be based on the override and project props security.
`canEdit` is always false.

src/components/partnership/partnership.service.ts
```diff
@@ -500,13 +593,20 @@ export class PartnershipService {
   }
 
   async update(input: UpdatePartnership, session: ISession) {
+    // mou start and end are now computed fields and do not get updated directly
```
CarsonF, line 596: Which is why they aren't in the `UpdatePartnership` DTO 😉

michaelmarshall: just helping other devs be alerted to changes they may not be aware of.

src/components/partnership/partnership.service.ts
michaelmarshall: updates now check project mou perms.

src/components/partnership/partnership.service.ts
```diff
@@ -447,6 +529,29 @@ export class PartnershipService {
...
+    }
+    if (result.projectMouEnd && result.mouEndOverride !== null) {
+      mouEnd = result.mouEndOverride ?? result.projectMouEnd;
+    }
+
+    let canReadMouStart = false;
+
+    if (result.canReadProjectMouStart && result.canReadMouStartOverride) {
+      canReadMouStart = true;
+    }
```
CarsonF, line 547: FYI
```suggestion
    const canReadMouStart = result.canReadProjectMouStart && result.canReadMouStartOverride;
```

src/components/partnership/partnership.service.ts
```diff
@@ -447,6 +529,29 @@ export class PartnershipService {
       throw new NotFoundException('Could not find partnership');
     }
 
+    let mouStart = null;
+    let mouEnd = null;
+
+    // if user has access to project mou and there is no partnership override
+    if (result.projectMouStart && result.mouStartOveride !== null) {
```
CarsonF, line 536: It's better to use double equals instead of tripe for comparing to null because that matches null & undefined which are used pretty interchangeably. 

It's a little weird that the checks aren't consistent here either. The first is _truthy_ and the second is not nullable.
Either is probably fine...
```suggestion
    if (result.projectMouStart && result.mouStartOveride) {
```
```suggestion
    if (result.projectMouStart != null && result.mouStartOveride != null) {
```

src/components/partnership/partnership.service.ts
michaelmarshall: lol. I was thinking I had more complexity but then decided against it and never cleaned it up. thanks for the catch.

src/components/partnership/partnership.service.ts
michaelmarshall: I forgot I needed to refactor this now that I'm returning the project permissions. The conditionals are now calculated purely with permissions.

Issue #694, opened 2020-07-08
Get all Partnerships  > query partnerships($input: PartnershipListInput!) > Getting "Cannot return null for non-nullable field Partnership.organization" error when NOT using filter in inputs


iPiyushPatel:
**Describe the bug**
Get all Partnerships  > query partnerships($input: PartnershipListInput!) > Getting "Cannot return null for non-nullable field Partnership.organization" error when NOT using filter in inputs.

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
Partnership results are returned.

**Actual Result**
Getting "Cannot return null for non-nullable field Partnership.organization" error when NOT using filter in inputs. 
See video - https://www.awesomescreenshot.com/video/355807?key=38c1a02e7944048dc94de48541cf5c3d

GraphQL Query ( partnerships ):
![image](https://user-images.githubusercontent.com/5212585/86931820-32b08980-c156-11ea-933d-5ec5f612782d.png)

**Query and Input File**
[partnerships_OrgNullerror.txt](https://github.com/SeedCompany/cord-api-v3/files/4891293/partnerships_OrgNullerror.txt)

CordAPIV3# 3dd7156

PR #709, opened 2020-07-09
Issue - 694: Fix Partnership list filter issue

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/694

### DIFF
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -624,7 +624,7 @@ export class PartnershipService {
         },
       });
 
-      await Promise.all(
+      const items = await Promise.all(
         result.items.map(async (item) => {
           const query = `
               MATCH (partnership:Partnership {id: $id, active: true})
@@ -639,12 +639,13 @@ export class PartnershipService {
             .first();
 
           const org = await this.orgService.readOne(orgId?.id, session);
-          Object.assign(item, {
+          return {
             ...item,
             organization: org,
-          });
+          };
         })
       );
+      result.items = items;
     }
 
     return {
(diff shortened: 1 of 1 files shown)


src/components/partnership/partnership.service.ts
```diff
@@ -621,6 +623,28 @@ export class PartnershipService {
...
+            })
+            .first();
+
+          const org = await this.orgService.readOne(orgId?.id, session);
+          Object.assign(item, {
+            ...item,
+            organization: org,
+          });
+        })
+      );
```
CarsonF, line 647: Mutating like this is not a good pattern. It's better to create a new object and new array with those changes.
```suggestion
      const items = await Promise.all(
        result.items.map(async (item) => {
          const query = `
              MATCH (partnership:Partnership {id: $id, active: true})
                -[:organization {active: true}]->(organization)
              RETURN organization.id as id
            `;
          const orgId = await this.db
            .query()
            .raw(query, {
              id: item.id,
            })
            .first();

          const org = await this.orgService.readOne(orgId?.id, session);
          return {
            ...item,
            organization: org,
          };
        })
      );
```

src/components/partnership/partnership.service.ts
```diff
@@ -578,27 +578,28 @@ export class PartnershipService {
...
           projectId,
-          skip: (page - 1) * count,
-          count,
         })
         .run();
 
+      result.total = projectPartners.length;
+      result.hasMore = count * page < result.total ?? true;
+
+      projectPartners = projectPartners.splice((page - 1) * count, count);
```
CarsonF, line 596: We definitely want to keep the db determining the current page, otherwise it has to send all the results just for us to only use some of them.

Issue #707, opened 2020-07-09
Get All Users > query users($input: UserListInput) > 'total' number of users data looks incorrect (i.e. whooping 78K+) )


iPiyushPatel:
**Describe the bug**
Get All Users > query users($input: UserListInput) > 'total' number of users data looks incorrect (i.e. whooping 78K+) )

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
Users total info is correct.

**Actual Result**
 'total' number of users data looks incorrect (i.e. whooping 78K+) ). See video - https://www.awesomescreenshot.com/video/364100?key=89a1d762a55241cfbf12c87eb5f3085b

Via postman & seeing on UI have different counts for users.

Also i double check while seeing 100 record per page on 60th page and it returned null, so something is wrong here.
![image](https://user-images.githubusercontent.com/5212585/87048829-da42c000-c219-11ea-9d90-c6e2c971ffeb.png)

GraphQL Query: (Users)
![image](https://user-images.githubusercontent.com/5212585/87046823-43750400-c217-11ea-9618-9e0ead24aaaa.png)

**Query and Input File**
[users_bigTotal.txt](https://github.com/SeedCompany/cord-api-v3/files/4897186/users_bigTotal.txt)

CordAPIV3# e5a7ad9

zkhin:
@iPiyushPatel I created over 10K users and tried to reproduce this issue but it works fine on my side. Can you still reproduce on your side?

iPiyushPatel:
Hi @zkhin 

I've looked into this further and compared User Records displayed on Neo4J browser (http://localhost:7474/browser/) & returned via API.
Still I see record count is different on both i.e. Neo4j browser say 368 where as API returns 80647 . I've also tried creating a new one and both increased by +1 :) 

SEE: https://drive.google.com/file/d/13wQwPRfsykjXdNuDBOscf9sfUowOaQb3/view?usp=drivesdk

I've used below query to get data on Neo4J browser: MATCH (n:User) RETURN count(n) as count

Let me know if you still any other information or have any feedback/suggestion for me.

Thanks,

zkhin:
`MATCH (n:User {active: true}) RETURN count(n) as count`
It really matches exactly with the API total and neo4j results on my side.

iPiyushPatel:
Hi @zkhin 
[...]

PR #710, opened 2020-07-09
Issue: 707 - Fix user list total

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/707

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -117,7 +117,7 @@ export class UserService {
     return {
       items: result.items,
       hasMore: result.hasMore,
-      total: result.total,
+      total: result.items.length,
     };
   }
 


src/components/user/user.service.ts
```diff
@@ -117,7 +117,7 @@ export class UserService {
     return {
       items: result.items,
       hasMore: result.hasMore,
-      total: result.total,
+      total: result.items.length,
```
CarsonF, line 120: The `items` is only the current page. The `total` is the total across all pages. Thus you can't use the length of the current page to determine the `total`. If this were the case we wouldn't even ask for it and just use the length of the `items` ourselves.

zkhin: I see... so this issue https://github.com/SeedCompany/cord-api-v3/issues/707 is not issue, right?

CarsonF: If `hasMore` is true then `items` should not be empty. Also page 60 * count of 100 is offset 60,000 with the supposed total of 78436 again the items should not be empty.

Issue #716, opened 2020-07-10
Neo4jError: Variable `engagementId` not defined


CarsonF:
```
Neo4jError: Variable `engagementId` not defined (line 10, column 7 (offset: 1168))
"WHERE engagementId.value CONTAINS $value2"
       ^

    at DatabaseService.list (src/core/database/database.service.ts:759:32)
    at ProductService.list (src/components/product/product.service.ts:102:34)
    at EngagementService.listProducts (src/components/engagement/engagement.service.ts:391:40)
```
8476112b1c4b235f6807309e7ee36a6b381a65c5

zkhin:
@CarsonF Do we create `engagementId` property while creating `product`? 
https://github.com/SeedCompany/cord-api-v3/blob/8476112b1c4b235f6807309e7ee36a6b381a65c5/src/components/engagement/engagement.service.ts#L386-L407

CarsonF:
No it's pre-existing. A product belongs to an engagement.

zkhin:
Okay, then I think we need to rewrite the `listProducts` to fetch all Products which belongs to that engagement.
This issue occurs because it tries to find Products by `engagementId` in filter

CarsonF:
Oh ok so that generic list provides a way to get a list of all products regardless of the engagement they are attached to.

zkhin:
Yes, right. If we don't provide `engagementId` in filter, it will get all products regardless of that engagement.

CarsonF:
I think for now we want to keep both of those functionalities.

zkhin:
If you want to keep current `listProducts` function, then we need to refactor `product.service.ts/list` function. So if there's `engagementId` in filter, we need to write query to fetch products by `engagementId`

zkhin:
@CarsonF Can you let me know what is the `Relationship` between `Product` and `Engagement`? I wonder if we create a relation between `Product` and `LanguageEngagement` while creating `LanguageEngagement`. Is there other mutation that adds product to engagement?

CarsonF:
No products are created after the language engagement is created. There can be 0+ products on a language engagement.
The mutation is `createProduct`.

zkhin:
It will be blocked until we create relation between `Product` and `LanguageEngagement` while creating `Product`

CarsonF:
Please continue forward with that.

PR #725, opened 2020-07-12
Issue: 716 - Refactor product service list and readone

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/716

### DIFF
--- a/src/components/product/product.service.ts
+++ b/src/components/product/product.service.ts
@@ -8,7 +8,7 @@ import { difference } from 'lodash';
 import { DateTime } from 'luxon';
 import { generate } from 'shortid';
 import { ISession } from '../../common';
-import { DatabaseService, ILogger, Logger } from '../../core';
+import { DatabaseService, ILogger, Logger, matchSession } from '../../core';
 import {
   AnyProduct,
   CreateProduct,
@@ -101,7 +101,8 @@ export class ProductService {
       ],
     });
 
-    if (!result || !result.id.value) {
+    if (!result || !result.id) {
+      this.logger.warning(`Could not find product`, { id: id });
       throw new NotFoundException('Could not find product');
     }
 
--- a/src/components/product/product.service.ts
+++ b/src/components/product/product.service.ts
@@ -169,7 +169,15 @@ export class ProductService {
       throw new NotFoundException('Could not find product');
     }
 
-    return result;
+    return {
+      ...result,
+      scriptureReferences: {
+        // TODO
+        canRead: true,
+        canEdit: true,
+        value: [],
+      },
+    };
   }
 
   async update(input: UpdateProduct, session: ISession): Promise<AnyProduct> {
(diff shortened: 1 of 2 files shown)


src/components/product/product.service.ts
```diff
@@ -63,63 +74,116 @@ export class ProductService {
...
-      },
-    });
+      });
+
+      const products = await productQuery.run();
+
+      result.items = await Promise.all(
+        products.map(async (product) => this.readOne(product.id, session))
+      );
+      result.total = result.items.length;
```
michaelmarshall, line 162: We don't want to run more than one DB query per function if we can help it.

src/components/product/product.service.ts
```diff
@@ -63,63 +74,116 @@ export class ProductService {
...
+          { name: 'methodology', secure: false },
+        ],
+        input: {
+          page,
+          count,
+          sort,
+          order,
+          filter,
+        },
+      });
```
michaelmarshall, line 185: We need to not use the old `database.service.ts` file functions. Instead, use the `query.helpers.ts` exports.

src/components/product/product.service.ts
```diff
@@ -63,63 +74,116 @@ export class ProductService {
...
+            active: true,
+            id: $requestingUserId
+          }),
+          (engagement {id: $engagementId, active: true, owningOrgId: $owningOrgId})
+          ,(product:Product {active:true})
+        WITH COUNT(product) as total, engagement, product
+          RETURN total, product.id as id, product.createdAt as createdAt
+          ORDER BY ${sort} ${order}
+          SKIP $skip LIMIT $count
+      `;
```
michaelmarshall, line 147: We need to compose the query using the functions from `query.helpers.ts`. The `organization` service's `list` function is a good example.

src/components/product/product.service.ts
```diff
@@ -63,63 +74,116 @@ export class ProductService {
...
-      approach: result.approach.value,
-      methodology: result.methodology.value,
+      createdAt: result.createdAt,
+      type: result.type,
+      books: result.books,
+      mediums: result.mediums,
+      purposes: result.purposes,
+      approach: result.approach,
+      methodology: result.methodology,
     };
```
michaelmarshall, line 118: This needs to be in the `query.helpers.ts` style. Use the `organization` service's `readOne` function as an example.

zkhin:
@michaelmarshall I think I need to implement `filterByEngagement` in `query.helper.ts` for filtering products by engagementId. Maybe it's similar to `filterByUser` but let me that I can use `:engagement` relationshipType.

src/components/product/product.service.ts
```diff
@@ -23,106 +37,141 @@ import {
...
+    const props = ['mediums', 'purposes', 'methodology'];
+    const query = this.db
+      .query()
+      .call(matchRequestingUser, session)
+      .call(matchUserPermissions, 'Product', id)
+      .call(addAllSecureProperties, ...props)
+      .with([
+        ...props.map(addPropertyCoalesceWithClause),
+        'coalesce(node.id) as id',
+        'coalesce(node.createdAt) as createdAt',
```
CarsonF, line 163: FYI there's no need to `coalesce` if there's not more than one argument in the function call.

src/core/database/query.helpers.ts
```diff
@@ -621,6 +621,22 @@ export function filterByUser(
   ]);
 }
 
+// used to search a specific engagement's relationship to the target base node
+// for example, searching all products a engagement is a part of
+export function filterByEngagement(
```
CarsonF, line 626: This is specific to engagements. It should live in the engagements folder not core.

src/components/product/product.service.ts
```diff
@@ -242,4 +216,109 @@ export class ProductService {
...
+    const query = this.db
+      .query()
+      .call(matchRequestingUser, session)
+      .call(matchUserPermissions, 'Product');
+
+    if (filter.methodology) {
+      query.call(filterByString, label, 'methodology', filter.methodology);
+    } else if (filter.approach) {
+      query.call(filterByString, label, 'approach', filter.approach);
+    } else if (filter.engagementId) {
```
CarsonF, line 238: These filters are not either/or. They all could be applied at once.

src/components/product/product.service.ts
```diff
@@ -23,106 +37,141 @@ import {
...
+    if (!result) {
+      throw new ServerException('failed to create default product');
     }
 
+    const id = result.id;
+
+    // add root admin to new product as an admin
+    await this.db.addRootAdminToBaseNodeAsAdmin(id, 'Product');
+
+    this.logger.debug(`product created, id ${id}`);
```
CarsonF, line 108:
```suggestion
    this.logger.debug(`product created`, { id });
```

PR #744, opened 2020-07-15
Products API Refactor

CarsonF:
- Moved `Film`, `LiteracyMaterial`, and `Story` to top level modules.
- Added `Song` module with stubbed service
- Replaced `Range` module with `Scripture` module
   - We are not going to treat these ranges like entities that need their own service
   - Replaced DTOs to be _book/chapter/verse_ objects. Absolute verse numbers will be a private concept.
- Updated `Product` DTOs
   - Have secured properties
   - `Product` is now an interface with two concretes:
     - `DirectScriptureProduct` - The straight up scripture one with one scripture references
     - `DerivativeScriptureProduct` - All the other types that connect to other resources, like Film, via a new Producible interface
- Removed range code in services and stubbed the new code


candleshine:
> Replaced DTOs to be book/chapter/verse objects. Absolute verse numbers will be a private concept.
when we first discussed this, this was going to be a front end translation of Abstract Concepts (book, chapter, verse) to number

CarsonF:
I couldn't remember. That's what I thought too. Then I talked to Michael and he said we would translate it in the API. I then agreed with him because it allows that concept to be hidden in API and removed or changed when it needs to. It also allows other API consumers to handle that data without having to implement the translation themselves.

@blakekrammes does already have this translation code done on the frontend. Blake could you port that to the API?

blakekrammes:
@CarsonF Should that be a part of this PR?

CarsonF:
No it can come after

blakekrammes:
@candleshine Should I go ahead and start writing a ScriptureReference service to start implementing these changes or are you already stubbing that out?

candleshine:
You should do it.  I have not started anything.

candleshine:
to be specific, it is called the scripture-reference module

CarsonF:
https://github.com/SeedCompany/cord-api-v3/tree/products-again/src/components/scripture

blakekrammes:
I have the code in the module. I can do a pr once this is merged.

PR #754, opened 2020-07-15
Absolute verse number

blakekrammes:


src/components/scripture/scripture.module.ts
```diff
@@ -1,7 +1,137 @@
...
 import { ScriptureReferenceResolver } from './scripture-reference.resolver';
 
+interface Range<T> {
+  start: T;
+  end: T;
+}
+
+type VerseRange = Range<number>;
+
+type ScriptureRange = Range<Required<ScriptureReference>>;
```
CarsonF, line 13: This already exists as a DTO, it can be used instead of defining this.

blakekrammes: Were you just referring to ScriptureRange here? I just used a new interface for VerseRange.

PR #779, opened 2020-07-20
Fixed query user by id bug

punamasawa777:


CarsonF:
I just did it myself with #782  since this was a unique NestJS area.

PR #786, opened 2020-07-21
Fix corrupted files in filesystem bucket

glassblowerscat:
Fixes an issue where files uploaded to the "filesystem bucket" (for dev purposes) could not be opened directly or after download. This turned out to be due to us converting the incoming `Buffer` to a `string` and writing/reading it as such.

- Update `upload()` method of `local-bucket` controller to leave request body as `Buffer` instead of converting to a `string`
- Update `readFile()` and `writeFile()` methods of `filesystem-bucket` to read/write file as `Buffer`
- Update `FakeAwsFile` type to set `Body` property type to `Buffer`
- Update `generateFakeFile()` method in `test/utility/create-file` to generate `Buffer` for the `content` property instead of a `string`

### DIFF
--- a/src/components/file/bucket/filesystem-bucket.ts
+++ b/src/components/file/bucket/filesystem-bucket.ts
@@ -45,7 +45,7 @@ export class FilesystemBucket extends LocalBucket {
       throw new NotFoundException();
     }
     const raw = await this.readFile(key + '.info');
-    const info = JSON.parse(raw);
+    const info = JSON.parse(raw.toString());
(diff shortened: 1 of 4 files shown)


CarsonF:
I think we need to update the size calculation too
https://github.com/SeedCompany/cord-api-v3/blob/fe91dfa2264bda68e607c9fea40431c1f90979f8/src/components/file/bucket/local-bucket.ts#L33
```tsx
      ContentLength: Buffer.byteLength(file.Body),
```

PR #794, opened 2020-07-22
Query refactor2

michaelmarshall:
refactors engagement lists

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -818,6 +818,7 @@ export class EngagementService {
         ...baseNodeMetaProps,
         ...childBaseNodeMetaProps.map((x) => x.returnIdentifier),
         'labels(node) as labels',
+        '__typename',
       ]);
 
     let result;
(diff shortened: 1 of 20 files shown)


src/components/engagement/engagement.service.ts
```diff
@@ -707,7 +715,11 @@ export class EngagementService {
     id: string,
     session: ISession
   ): Promise<LanguageEngagement | InternshipEngagement> {
-    this.logger.debug('readOne', { id, userId: session.userId });
+    this.logger.info('readOne', { id, userId: session.userId });
+
+    if (!id) {
+      throw new NotFoundException('no id given');
+    }
```
CarsonF, line 722: Not necessary

michaelmarshall: It should error out, right? Without that catch it finds the first engagement and returns that.

CarsonF: If id is empty/falsy/not a short id it should give back a validation error before this method is ever called. Is that not what you are experiencing?

PR #823, opened 2020-07-24
refactor listEngagement function

tjgersho:


src/components/project/project.service.ts
```diff
@@ -627,42 +627,23 @@ export class ProjectService {
...
 
-    const retVal: SecuredEngagementList = {
-      total: items.length,
-      hasMore: false,
-      items,
-      canRead: true,
+    return {
+      ...result,
       canCreate: true,
+      canRead: true,
```
michaelmarshall, line 645: You'll need to return the user's permissions on the `engagement` property. Looks like you may need to query readOne to get them. 

One of the reasons for these refactors is to implement all the permissioning concepts. Whenever you see constant return values you'll need to refactor them to get the actual values from the db.

tjgersho: Got it.

tjgersho: Just updated it.

PR #838, opened 2020-07-28
Fixed engagement null Issue

punamasawa777:


src/components/engagement/engagement.service.ts
```diff
@@ -884,6 +884,26 @@ export class EngagementService {
...
+      countryOfOrigin: {
+        value: result.countryOfOrigin.value ? result.countryOfOrigin.value : null,
+        canRead: !!result.countryOfOrigin.canRead,
+        canEdit: !!result.countryOfOrigin.canEdit,
+      },
+      ceremony: {
+        value: result.ceremony.value ? result.ceremony.value : null,
+        canRead: !!result.ceremony.canRead,
+        canEdit: !!result.ceremony.canEdit,
+      },
```
michaelmarshall, line 906: These objects are already being shaped in the with clause in line 802+. Please make these changes in that clause. One way might be to use the `coalesce` function.

candleshine: This isn't being done here because of the shape of these fields.  This is being done here because it needs to handle the case where the requesting user does not have access to the field in question.  For example, the requesting user might be able to access the Internship Engagement, but not the Intern User or the Mentor User.  In those cases, the value needs to be set to NULL.  The Cypher is not doing this correctly for Typescript.  Also you can see in the lines just before this, that Punam is following the pattern already being used for methodologies.

michaelmarshall: yes, methodologies was different because it needed to return an empty array, not null, but you can do what you mentioned in cypher. With optional matches, the intern value is null when the user can't see it, and using a coalesce function you can return false for the canRead/canEdit fields.

michaelmarshall: I'm trying to move all shaping code into the cypher so we can start returning the response object directly. I believe we can do that with methodologies as well but when I tried I wasn't able to figure it out quickly so I moved on. I'm not asking for that now, but in the future I want us to do direct returns from our helper functions.

candleshine: @punamasawa777 here's what michael suggests

candleshine: image

Issue #852, opened 2020-07-28
Implement Change Password


CarsonF:
https://github.com/SeedCompany/cord-api-v3/blob/f3bd4b45e4b44975db656abede6ac1772857db50/src/components/authentication/authentication.service.ts#L258-L264

Old password should match or throw error.

PR #865, opened 2020-07-30
Issue: 852 - Add changing password feature and test case

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/852

### DIFF
--- a/src/components/authentication/authentication.service.ts
+++ b/src/components/authentication/authentication.service.ts
@@ -1,6 +1,7 @@
 import {
   BadRequestException,
   Injectable,
+  NotAcceptableException,
   InternalServerErrorException as ServerException,
   UnauthorizedException as UnauthenticatedException,
 } from '@nestjs/common';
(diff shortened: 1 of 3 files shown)


src/components/authentication/authentication.service.ts
```diff
@@ -256,11 +258,44 @@ export class AuthenticationService {
...
+      .first();
+
+    if (!result || !(await argon2.verify(result.passwordHash, oldPassword))) {
+      throw new UnauthenticatedException('Invalid credentials');
+    }
+    if (oldPassword === newPassword) {
+      throw new NotAcceptableException(
+        'newPassword could not be same with oldPassword'
+      );
+    }
```
CarsonF, line 283: Let's skip this for now. We don't have a good reason to block them changing their password to what it is currently. Esp since they just confirmed they know their old password.

zkhin: Okay

test/project.e2e-spec.ts
```diff
@@ -557,13 +557,13 @@ describe('Project e2e', () => {
...
+      times(numProjectMembers, async (index) => {
+        await createProjectMember(app, {
           userId: memberIds[index],
           projectId,
           roles: [Role.Consultant],
-        })
-      )
+        });
+      })
     );
```
CarsonF, line 567: Removing the `async` modifier could work too. Not sure why this is just failing

PR #875, opened 2020-08-03
849 refactor product service

tjgersho:


src/components/product/dto/list-product.dto.ts
```diff
@@ -24,6 +24,10 @@ export abstract class ProductFilters {
   })
   readonly methodology?: ProductMethodology;
 
+  @Field({
+    description: 'Only products for this engagement',
+    nullable: true,
+  })
```
CarsonF, line 30: I want to keep these id filters private/internal and not expose them through the API. That's why it doesn't have the Field decorator. 

```suggestion
```

Issue #873, opened 2020-07-31
Add Roles to User service


michaelmarshall:
I started branch `873-roles` to add `SecuredRoles` to the user service. The FE needs to be able to write and read back an array of `Roles` on the User object.

michaelmarshall:
![image](https://user-images.githubusercontent.com/1165460/89078054-38f6e780-d349-11ea-8b28-1da678f00f04.png)

`Role` is in the `project-member` service.

michaelmarshall:
Each role needs to be a separate property in the DB. So this field will be an array. We need to be able to create/update/delete a single role at a time, hence the array. We will NOT be storing a string array in a single property node. Each role is a separate property node.

PR #876, opened 2020-08-03
Add roles to UserService

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/873

### DIFF
--- a/src/components/location/dto/location.dto.ts
+++ b/src/components/location/dto/location.dto.ts
@@ -5,7 +5,7 @@ import {
   ObjectType,
 } from '@nestjs/graphql';
 import { Resource, SecuredProperty, SecuredString } from '../../../common';
-import { SecuredUser } from '../../user/dto';
+import { SecuredUser } from '../../user/dto/user.dto';
 
 @InterfaceType()
 export abstract class Place {
@@ -17,10 +17,10 @@ export abstract class Place {
   implements: [Resource, Place],
 })
 export class Zone extends Resource implements Place {
-  @Field()
+  @Field(() => SecuredString)
   readonly name: SecuredString;
 
-  @Field()
+  @Field(() => SecuredUser)
   readonly director: SecuredUser;
 }
 
@@ -33,13 +33,13 @@ export class SecuredZone extends SecuredProperty(Zone) {}
   implements: [Resource, Place],
 })
 export class Region extends Resource implements Place {
-  @Field()
+  @Field(() => SecuredString)
   readonly name: SecuredString;
 
-  @Field()
+  @Field(() => SecuredZone)
   readonly zone: SecuredZone;
 
-  @Field()
+  @Field(() => SecuredUser)
   readonly director: SecuredUser;
 }
 
@@ -52,10 +52,10 @@ export class SecuredRegion extends SecuredProperty(Region) {}
   implements: [Resource, Place],
 })
 export class Country extends Resource implements Place {
-  @Field()
+  @Field(() => SecuredString)
   name: SecuredString;
 
-  @Field()
+  @Field(() => SecuredRegion)
   region: SecuredRegion;
 }
 
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -472,7 +472,7 @@ export class UserService {
       ...result,
       roles: {
         value: result.roles.value || [],
-        canRead: result.roles.canRead,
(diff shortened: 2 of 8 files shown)


michaelmarshall:
created a new user and get this as a response on the `user` query:
image
I believe this is because when the value is null it returns false for read. It should still return true when the permission is true. Since our cypher doesn't account for this case, for now, just return read=true when edit=true.

zkhin:
> created a new user and get this as a response on the `user` query:
> image
> canRead is false but it should be true.
> I believe this is because when the value is null it returns false for read. It should still return true when the permission is true. Since our cypher doesn't account for this case, for now, just return read=true when edit=true.

Okay

michaelmarshall:
awesome. thx.

src/components/location/dto/location.dto.ts
```diff
@@ -52,10 +52,10 @@ export class SecuredRegion extends SecuredProperty(Region) {}
   implements: [Resource, Place],
 })
 export class Country extends Resource implements Place {
-  @Field()
+  @Field(() => SecuredString)
   name: SecuredString;
 
-  @Field()
+  @Field(() => SecuredRegion)
```
CarsonF, line 58: These changes aren't necessary.

src/components/user/user.service.ts
```diff
@@ -174,6 +177,16 @@ export class UserService {
...
 
+  roleProperties = (roles?: Role[]) => {
+    let roleProperties: any[] = [];
+    roles?.forEach(
+      (role) =>
+        (roleProperties = roleProperties.concat(this.property('roles', role)))
+    );
+
+    return roleProperties;
+  };
```
CarsonF, line 188: Usually `.forEach` can be replaced by `.map` since forEach returns void.
I've never had a need to use `.forEach`.

```suggestion
  private roleProperties(roles?: Role[]) {
    return (roles || []).flatMap(role => this.property('roles', role));
  }
```

src/components/user/user.service.ts
```diff
@@ -469,6 +499,23 @@ export class UserService {
...
+        .query()
+        .match([
+          node('user', ['User', 'BaseNode'], {
+            active: true,
+            id: input.id,
+          }),
+        ])
+        .create([...this.roleProperties(newRoles)])
+        .run();
+    }
```
CarsonF, line 516: This doesn't cover removing roles.

src/components/user/user.service.ts
```diff
@@ -433,27 +448,42 @@ export class UserService {
...
     if (!result) {
       throw new NotFoundException('Could not find user');
     }
 
-    return result;
+    const response: any = {
+      ...result,
+      roles: {
+        value: result.roles.value || [],
+        canRead: result.roles.canEdit ?? result.roles.canRead,
```
CarsonF, line 475: This doesn't seem right. Why would canRead come from canEdit first? This prevents a read-only view of this property. Maybe what you meant was `||` instead of `??` which would fallback to canRead if canEdit was false. Currently if canEdit false canRead is never used.

src/components/user/user.service.ts
```diff
@@ -433,27 +448,42 @@ export class UserService {
       .query()
       .call(matchRequestingUser, session)
       .call(matchUserPermissions, 'User', id)
-      .call(addAllSecureProperties, ...props)
+      .call(addAllSecureProperties, ...[...props, 'roles'])
       .with([
         ...props.map(addPropertyCoalesceWithClause),
+        `
+        {
+          value: collect(distinct roles.value),
```
CarsonF, line 456: Doesn't `collect` always return an array? If not maybe it could be wrapped in a `coalesce(..., [])` to prevent the need to map result to response below.

src/components/user/user.service.ts
michaelmarshall: This makes canRead true if canEdit is true. If canEdit is false then canRead may or may not be true so it returns directly. The principle is if canEdit is true, we'll assume that canRead should be true.

CarsonF: Ok this isn't doing what you think it is then. The second value, canRead, will never be used here - only `canEdit`. If canEdit was nullable that would be a different story, but it's not - it's always a boolean.
```tsx
(false ?? true) // false
```

michaelmarshall: that's right. my mistake. this could have been done in the cypher `WITH` clause where `canEdit` might be null.

PR #886, opened 2020-08-04
Refactor user.service roles

zkhin:
We need to discuss more how to delete the user `roles` while updating `user`

### DIFF
--- a/src/components/location/dto/location.dto.ts
+++ b/src/components/location/dto/location.dto.ts
@@ -17,10 +17,10 @@ export abstract class Place {
   implements: [Resource, Place],
 })
 export class Zone extends Resource implements Place {
-  @Field(() => SecuredString)
+  @Field()
   readonly name: SecuredString;
 
-  @Field(() => SecuredUser)
+  @Field()
   readonly director: SecuredUser;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/location/dto/location.dto.ts
```diff
@@ -55,7 +55,7 @@ export class Country extends Resource implements Place {
   @Field(() => SecuredString)
   name: SecuredString;
 
-  @Field(() => SecuredRegion)
```
CarsonF, line 58: I meant all of the changes in this file. Secured fields don't need their types explicitly set because they can be inferred from the property type.

src/components/user/user.service.ts
```diff
@@ -471,8 +465,8 @@ export class UserService {
     const response: any = {
       ...result,
       roles: {
-        value: result.roles.value || [],
-        canRead: result.roles.canEdit ?? result.roles.canRead,
+        value: result.roles.value,
+        canRead: result.roles.canEdit || result.roles.canRead,
```
CarsonF, line 469: I believe this be done in the cypher mapping so we don't have to map here.

zkhin: image
If we don't set `roles` while registering user, the `canRead` will be `false` but that should be `true` if the `canEdit` is `true`

CarsonF: Well that sounds like a different problem in general. Why would one be set but not the other?

Also doesn't this work?
```js
canRead: coalesce(canEdit, canRead, false)
```

zkhin: Yes, that's working. Thx

PR #888, opened 2020-08-04
fix unable to update product

tjgersho:


tjgersho:
@reneepe , It adds the validation for produces parameter when updating the product. So if produces is not the valid ID for producible node(Story, film ...), it will return "not found exception" `Could not find producible node`.

Issue #884, opened 2020-08-04
Engagement.start/end dates need override


CarsonF:
Engagement start/end dates should follow the same pattern as Partnership's. They should have override fields explicitly define the range, otherwise they should fall back to the project's date range.

PR #890, opened 2020-08-05
Add engagement start/end override fields

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/884


src/components/engagement/engagement.service.ts
```diff
@@ -930,6 +943,27 @@ export class EngagementService {
...
+      startDate = result.startDateOverride.value ?? readProject.mouStart.value;
+    }
+    if (readProject.mouEnd.canRead && result.endDateOverride.canRead) {
+      endDate = result.endDateOverride.value ?? readProject.mouEnd.value;
+    }
+
+    const canReadStartDate =
+      readProject.mouStart.canRead && result.startDateOverride.canRead;
+    const canReadEndDate =
+      readProject.mouEnd.canRead && result.endDateOverride.canRead;
```
CarsonF, line 965:
```suggestion
  const canReadStartDate =
    readProject.mouStart.canRead && result.startDateOverride.canRead;
  const startDate = canReadStartDate
    ? result.startDateOverride.value ?? readProject.mouStart.value
    : null;
  const canReadEndDate =
    readProject.mouEnd.canRead && result.endDateOverride.canRead;
  const endDate = canReadEndDate
    ? result.endDateOverride.value ?? readProject.mouEnd.value
    : null;
```

michaelmarshall:
@reneepe You'll want to test the override concept and make sure it is properly overriding the project dates when the overrides are present, and not when the overrides are absent.

PR #893, opened 2020-08-05
Remove ID filters from API

CarsonF:
Those lists can be determined by relationships on the owning objects
`budgets.filters.projectId` -> `project.budgets`

### DIFF
--- a/test/education.e2e-spec.ts
+++ b/test/education.e2e-spec.ts
@@ -111,18 +111,25 @@ describe('Education e2e', () => {
       times(numEducations).map(() => createEducation(app, { userId: user.id }))
     );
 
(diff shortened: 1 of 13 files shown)


test/education.e2e-spec.ts
```diff
@@ -131,7 +131,9 @@ describe('Education e2e', () => {
       }
     );
 
-    expect(educations.items.length).toBeGreaterThanOrEqual(numEducations);
+    expect(result.user.education.items.length).toBeGreaterThanOrEqual(
+      numEducations
+    );
```
CarsonF, line 136: Haha I'm dumb 🤦 . I got that change on one of the suites, but missed it on the rest.

Issue #898, opened 2020-08-06
Update Partnership Permission Error


reneepe:
User receives "You do not have permission to update this file" error message when trying to update a partnership that user just created.

Steps to reproduce:
1. Go to the playground
2. Enter query data as below and in attached document.
3. Press Play

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/4ebca9a3-70a8-4b0c-887d-cac6672e7060)

API: 8bf3b57   
FE: 2e940964     

Expected Results: Partnership updates
Actual Results: Partnership doesn't update due to user permission.

Desktop:
- OS: Windows 10 Pro 1903
- Browser: Chrome 
- Version: 84.0.4147.105

[Update Partnership Error 8.6.20.txt](https://github.com/SeedCompany/cord-api-v3/files/5036130/Update.Partnership.Error.8.6.20.txt)

PR #901, opened 2020-08-06
Fix update partnership permission error

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/898


michaelmarshall:
small change so I'm merging before Renee tests it.

PR #920, opened 2020-08-08
915 fix product service

tjgersho:
If produces is not a valid Producible(Story/Film..), product would not be created while returning validation error.


src/components/partnership/partnership.service.ts
```diff
@@ -168,8 +168,7 @@ export class PartnershipService {
     { organizationId, projectId, ...input }: CreatePartnership,
     session: ISession
   ): Promise<Partnership> {
-    const id = generate();
-    const createdAt = DateTime.local();
+    const createdAt = DateTime.local().toString();
```
CarsonF, line 171: There's no need to convert DateTime's to strings like this. We automatically convert them to Neo4J DateTime objects. Doing this string conversion actually prevents this.

src/components/product/product.service.ts
```diff
@@ -146,6 +146,17 @@ export class ProductService {
...
+      const produce = await this.db
+        .query()
+        .match([node('pr', 'Producible', { id: input.produces, active: true })])
+        .return('pr')
+        .first();
+      if (!produce) {
+        this.logger.warning(`Could not find producible node`, {
+          id: input.produces,
+        });
+        throw new NotFoundException('Could not find producible node');
```
CarsonF, line 158: Can you change this `NotFoundException` import from nestjs/common to ours? This will allow you to pass the field that has the not found exception. Like this:
```suggestion
        throw new NotFoundException('Could not find producible node', 'product.produces');
```

PR #926, opened 2020-08-11
Language read refactor

mikecoll:
Language Readone Refactor


CarsonF (commented):
Lots of commented out code here. That only adds more clutter. git should be used for previous versions of code.

izquierdoalfredo:
It's not finished yet, I will clean code after finish.

CarsonF:
Ok got it. If it's not ready for review please open the PR as a draft.

izquierdoalfredo:
@CarsonF Can you please check this PR?
It's ready for review.

src/components/language/language.service.ts
```diff
@@ -388,65 +381,119 @@ export class LanguageService {
...
       .call(matchRequestingUser, session)
-      .call(matchUserPermissions, 'Language', langId)
-      .call(addAllSecureProperties, ...props)
-      .call(addAllMetaPropertiesOfChildBaseNodes, ...childBaseNodeMetaProps)
-      .with([
-        ...props.map(addPropertyCoalesceWithClause),
-        ...childBaseNodeMetaProps.map(addShapeForChildBaseNodeMetaProperty),
-        ...baseNodeMetaProps.map(addShapeForBaseNodeMetaProperty),
-        'node',
+      .match([node('node', 'Language', { active: true, id: langId })])
```
frankyma, line 398: Nitpick: give a more specific name for 'node', maybe 'languageNod'.  

I'm just reading this PR to understand the new query changes but betting labeling would help a beginner like me follow along

CarsonF: I believe `node` almost always represents the thing you are fetching. We are in the `LanguageService` so it represents a language.

michaelmarshall: its convention used in the `query.helper.ts` file. It must be `node`.

frankyma: got it, thanks guys

PR #932, opened 2020-08-12
implement scripture reference override type

tjgersho:


src/common/secured-property.ts
```diff
@@ -60,11 +60,11 @@ export function SecuredPropertyList<T>(
 ) {
   @ObjectType({ isAbstract: true, implements: [Readable, Editable] })
   abstract class SecuredPropertyListClass
-    implements Readable, Editable, Secured<T[]> {
+    implements Readable, Editable, Secured<T[] | null> {
     @Field(() => [valueClass], {
       nullable: options.isOverride,
     })
-    readonly value: T[];
+    readonly value: T[] | null;
```
CarsonF, line 67: I don't think this change should be made

src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
```diff
@@ -94,7 +94,9 @@ export class SyncBudgetRecordsToFundingPartners
       !partnership.mouStart?.value ||
       !partnership.mouEnd?.value ||
       // TODO: decide whether to move the following funding and type checks to an earlier step.
-      !partnership.types.value.includes(PartnershipType.Funding) || // Partnership is not funding, so do nothing
+
+      (partnership.types.value &&
+        !partnership.types.value.includes(PartnershipType.Funding)) || // Partnership is not funding, so do nothing
       !partnership.types?.value // Partnership Type is not provided, so do nothing.
```
CarsonF, line 100: `partnership.types.value` should always be an array

src/common/secured-property.ts
tjgersho: In this case, how can we assign null to `ScriptureReferencesOverride.value`. Its type is `scriptrueRange[]`, so `can not assign null` error is occurring.

CarsonF: Bleh. You are right. Ok let me make a change that makes it only nullable if that option is true.

CarsonF: Ok pushed. Let me know if it doesn't work.

tjgersho: Works like a charm.

src/components/product/product.service.ts
```diff
@@ -381,11 +380,14 @@ export class ProductService {
           id: produces.p.properties.id,
           createdAt: produces.p.properties.createdAt,
           __typename: (ProducibleType as any)[typeName],
-          ...(await this.getProducibleByType(
-            produces.p.properties.id,
-            typeName,
-            session
-          )),
+          // scriptureReferencesOverride is not null or an empty array.
+          scriptureReferences: !scriptureReferencesOverride.length
```
CarsonF, line 384: We want an empty array here to take precedence over the producible's scripture. null and empty array have to be handled differently - otherwise we wouldn't even both with the nulls 😄

tjgersho:
@CarsonF , I would like to explain what I implemented so I can get any idea if it was a correct logic or not.

1. _ScriptureReferenceOverride.value_ is _ScriptureRange_ array. In the database, a Product have multiple relations named **scriptureReferenceOverride** and each relation is connected with **ScriptureRange** node. ScritpureRange node has **start** and **end** field(ScriptureRange does not have **value** filed like other override node). Also, **start** and **end** can not be null. 
These values are Absolute verses and these are initialized only with valid numeric values which are calculated from input data(book, chapter, verse). 

In which case should `ScriptureReferenceOverride.value` return null?

I thought the only case which `ScriptureReferenceOverride.value` returns null should be no active **scriptureReferenceOverride** relation connected with **ScriptureRange** node in that Product. 

I think I commented wrongly on the code and I should delete that line.
https://github.com/SeedCompany/cord-api-v3/blob/544d30c6cde73f5b54f4de748795dd8fa836d07e/src/components/product/product.service.ts#L383

Empty array totally means null case in my implementation.
https://github.com/SeedCompany/cord-api-v3/blob/544d30c6cde73f5b54f4de748795dd8fa836d07e/src/components/product/product.service.ts#L553-L576
https://github.com/SeedCompany/cord-api-v3/blob/544d30c6cde73f5b54f4de748795dd8fa836d07e/src/components/product/product.service.ts#L354-L359
https://github.com/SeedCompany/cord-api-v3/blob/544d30c6cde73f5b54f4de748795dd8fa836d07e/src/components/product/product.service.ts#L383-L391

Any suggestion for it?

Issue #938, opened 2020-08-13
Partnership.fundingType


sethmcknight:
When a Partner is added to a Partnership as a Managing Partner, users should select the Financial Reporting Type from:
- Funded
- Field Engaged

zkhin:
@sethmcknight 
Can you clarify that I understood your requirements correctly?
- Add new optional `fundingType` field to `Partnership`
- This field can only be set to `Funded` or `FieldEngaged` when the `Partnership.types` includes the `Managing` type.

CarsonF:
Exactly 👍

PR #958, opened 2020-08-14
Add funding type to partnership

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/938

### DIFF
--- a/src/components/partnership/dto/update-partnership.dto.ts
+++ b/src/components/partnership/dto/update-partnership.dto.ts
@@ -35,7 +35,7 @@ export abstract class UpdatePartnership {
   readonly types?: PartnershipType[];
 
   @Field(() => PartnershipFundingType, { nullable: true })
-  readonly fundingType?: PartnershipFundingType;
+  readonly fundingType?: PartnershipFundingType | null;
 }
 
 @InputType()
--- a/src/components/partnership/dto/update-partnership.dto.ts
+++ b/src/components/partnership/dto/update-partnership.dto.ts
@@ -4,6 +4,7 @@ import { ValidateNested } from 'class-validator';
 import { CalendarDate, DateField, IdField } from '../../../common';
 import { CreateDefinedFileVersionInput } from '../../file/dto';
 import { PartnershipAgreementStatus } from './partnership-agreement-status.enum';
+import { PartnershipFundingType } from './partnership-funding-type.enum';
 import { PartnershipType } from './partnership-type.enum';
 import { Partnership } from './partnership.dto';
 
@@ -32,6 +33,9 @@ export abstract class UpdatePartnership {
 
   @Field(() => [PartnershipType], { nullable: true })
   readonly types?: PartnershipType[];
+
+  @Field(() => PartnershipFundingType, { nullable: true })
+  readonly fundingType?: PartnershipFundingType;
 }
 
 @InputType()
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -513,8 +513,8 @@ export class PartnershipService {
     const object = await this.readOne(input.id, session);
 
     this.verifyFundingType(
-      input.fundingType ?? (object.fundingType.value as PartnershipFundingType),
-      input.types ?? (object.types.value as PartnershipType[])
+      input.fundingType ?? object.fundingType.value,
+      input.types ?? object.types.value
     );
 
     const { mou, agreement, ...rest } = input;
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -8,7 +8,12 @@ import { node, Query, relation } from 'cypher-query-builder';
 import { RelationDirection } from 'cypher-query-builder/dist/typings/clauses/relation-pattern';
 import { flatMap, upperFirst } from 'lodash';
 import { DateTime } from 'luxon';
-import { fiscalYears, ISession, NotFoundException } from '../../common';
+import {
+  fiscalYears,
+  InputException,
+  ISession,
+  NotFoundException,
+} from '../../common';
 import {
   addAllMetaPropertiesOfChildBaseNodes,
   addAllSecureProperties,
@@ -33,6 +38,7 @@ import { ProjectService } from '../project/project.service';
 import {
   CreatePartnership,
   Partnership,
+  PartnershipFundingType,
   PartnershipListInput,
   PartnershipListOutput,
   PartnershipType,
@@ -188,9 +194,7 @@ export class PartnershipService {
       throw e;
     }
 
-    if (input.fundingType && !this.canAddFundingType(input.types)) {
-      throw new ServerException('cannot add funding type');
-    }
+    this.verifyFundingType(input.fundingType, input.types);
 
     const mou = await this.files.createDefinedFile(
       `MOU`,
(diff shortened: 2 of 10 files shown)


src/components/partnership/partnership.service.ts
```diff
@@ -188,6 +188,10 @@ export class PartnershipService {
       throw e;
     }
 
+    if (input.fundingType && !this.canAddFundingType(input.types)) {
+      throw new ServerException('cannot add funding type');
+    }
```
CarsonF, line 193: This is not the server's fault, it's the users.
```suggestion
    if (input.fundingType && !this.canAddFundingType(input.types)) {
      throw new InputException('Funding type can only be applied to managing partners', 'partnership.fundingType');
    }
```

src/components/partnership/partnership.service.ts
```diff
@@ -734,4 +753,8 @@ export class PartnershipService {
       node('node', label, { active: true }),
     ]);
   }
+
+  protected canAddFundingType(types: PartnershipType[] | undefined): boolean {
+    return (types || []).includes(PartnershipType.Managing) ? true : false;
```
CarsonF, line 758: This could be simplified
```suggestion
    return types?.includes(PartnershipType.Managing);
```

src/components/partnership/partnership.service.ts
```diff
@@ -491,6 +513,12 @@ export class PartnershipService {
   }
 
   async update(input: UpdatePartnership, session: ISession) {
+    if (input.fundingType && !this.canAddFundingType(input.types)) {
```
CarsonF, line 516: This needs to account for existing types if it's not passed in.

src/components/partnership/partnership.service.ts
CarsonF: Actually you could change the `canAddFundingType` function into a `verifyFundingType(fundingType, types)` function so you don't have to duplicate the exception message in both spots

src/components/partnership/partnership.service.ts
CarsonF: null coalescing would be better too as it would skip the includes check all together.

src/components/partnership/partnership.service.ts
zkhin: Aha got it.

src/components/partnership/partnership.service.ts
```diff
@@ -494,6 +512,11 @@ export class PartnershipService {
     // mou start and end are now computed fields and do not get updated directly
     const object = await this.readOne(input.id, session);
 
+    this.verifyFundingType(
+      input.fundingType ?? (object.fundingType.value as PartnershipFundingType),
+      input.types ?? (object.types.value as PartnershipType[])
```
CarsonF, line 517: Looks like I broke the types here when making the secured list changes. Let me see if I can fix.

zkhin: Okay, thx

CarsonF: Ok I've got that fix ready in #959. I'll merge it as soon as it goes green. Can you rebase after that and remove these typecasts?

zkhin: Okay, I will.

zkhin: I rebased and I can remove typecast for `types`, but not working with `fundingType`

CarsonF: But of course we need to do the same thing for single values lol. #961 up to do that.

CarsonF: Ok it's merged!

src/components/partnership/partnership.service.ts
```diff
@@ -494,6 +512,11 @@ export class PartnershipService {
     // mou start and end are now computed fields and do not get updated directly
     const object = await this.readOne(input.id, session);
 
+    this.verifyFundingType(
+      input.fundingType ?? object.fundingType.value,
```
CarsonF, line 516: If a `fundingType` is not given, but the managing `type` is removed then this will error.
We should handle this edge case.

We should set the `fundingType` to `null` for them if they omit it and the new or existing `types` don't include managing.
If they pass in a `fundingType` and still want to remove the managing type we could either:
- Give an error (only managing can having funding type)
- Set it to null silently for them

I'm leaning towards the former so the API doesn't do something "unexpected". What do you think?

zkhin: That's great.
I will try to follow your steps.

zkhin: What is the best way to set the `fundingType` to `null` while updating properties.
```
Object.assign(input, {
 fundingType: null,
});
```
I tried above code but it's not working because it rejects if the value is null or same with origin value in the `database.service.ts`

CarsonF: Well first you should not modify the `input`. If you need to change a property, create a new object.
```tsx
const changes = {
  ...input,
  fundingType: null,
};
```

Secondly, the database service is probably just ignoring nullable values. In most places we treat `null` & `undefined` the same. Which is why we do checks like `foo == null` instead of `===`. For these updates though, we want to treat `null` as an actual value and `undefined` as an omitted value.
My guess would be this line 
https://github.com/SeedCompany/cord-api-v3/blob/aff9e6a4554574d6c320c487e8cc105672da3397/src/core/database/database.service.ts#L450

src/components/partnership/dto/update-partnership.dto.ts
```diff
@@ -32,6 +33,9 @@ export abstract class UpdatePartnership {
 
   @Field(() => [PartnershipType], { nullable: true })
   readonly types?: PartnershipType[];
+
+  @Field(() => PartnershipFundingType, { nullable: true })
+  readonly fundingType?: PartnershipFundingType;
```
CarsonF, line 38:
```suggestion
  readonly fundingType?: PartnershipFundingType | null;
```
This would fix the need to typecast

src/components/partnership/partnership.service.ts
```diff
@@ -493,15 +511,45 @@ export class PartnershipService {
...
+          changes: {
+            types: [],
+            fundingType: (null as unknown) as PartnershipFundingType,
+          },
+          nodevar: 'partnership',
+        });
+        // Throw User Input Error
+        this.verifyFundingType(input.fundingType, input.types);
+      }
+    }
```
CarsonF, line 542: I don't understand all these changes. Why call validateFundingType twice? Why update types & fundingType separately from the rest of the properties below?

src/components/partnership/dto/update-partnership.dto.ts
zkhin: If I change like above, there's type error in `sgUpdateProperties`
`Type 'PartnershipFundingType | null | undefined' is not assignable to type 'PartnershipFundingType | undefined'`

src/components/partnership/partnership.service.ts
zkhin: My steps:
- If the `fundingType` and `types` are passed, and they are invalid, then set those fields to [], `null` and throw `InputException`
- In other cases, it only set the `fundingType` to `null` following other fields updated.

src/components/partnership/dto/update-partnership.dto.ts
CarsonF: Ah! Ok yeah this is because we are ambiguous with our output types on whether null is allowed when you do have permission to read it. Let me try to fix. I know we've been back and forth on this, but these are good things that we've been putting off.

zkhin: Okay, thanks

src/components/partnership/partnership.service.ts
zkhin: `this.verifyFundingType(input.fundingType, input.types);` is just for throw error

src/components/partnership/dto/update-partnership.dto.ts
CarsonF: Ok I've got https://github.com/SeedCompany/cord-api-v3/pull/975 up. You can rebase and `{ nullable: true }` to `SecuredPartnershipFundingType` after it's merged.

src/components/partnership/partnership.service.ts
CarsonF: If the input is invalid we can just throw there. There's no reason to do a set before for that.

src/components/partnership/dto/update-partnership.dto.ts
zkhin: Okay

PR #962, opened 2020-08-15
refactor create and update readOne

tjgersho:


src/components/story/story.service.ts
```diff
@@ -312,19 +350,24 @@ export class StoryService {
...
-      hasMore: boolean;
-      total: number;
-    } = await runListQuery(query, input, secureProps.includes(input.sort));
+    query.call(addAllSecureProperties, ...secureProps).with(
+      `
+          {
+            ${addBaseNodeMetaPropsWithClause(baseNodeMetaProps)},
+            ${listWithSecureObject(secureProps)}
+          } as node
+        `
```
CarsonF, line 359: Just an FYI we have a [`mapping`](https://github.com/SeedCompany/cord-api-v3/blob/master/src/core/database/mapping.helper.ts) function that could make this easier
```suggestion
      mapping('node', {
        ...addBaseNodeMetaPropsWithClause(baseNodeMetaProps),
        ...listWithSecureObject(secureProps),
      })
```
Those two other functions would need to be replaced with ones that return objects instead of strings.

tjgersho: Got it.

src/core/database/query.helpers.ts
```diff
@@ -995,10 +995,31 @@ export function listWithSecureObject(props: string[]) {
...
 export function addBaseNodeMetaPropsWithClause(props: string[]) {
   return props.map((x) => `${x}: node.${x}`).join(', ');
 }
 
+export function addBaseNodeMetaPropsWithClauseAsObject(props: string[]) {
+  const result: any = {};
+  props.map((x) => {
+    result[x] = `node.${x}`;
+  });
+  return result;
```
CarsonF, line 1020: `.map()` usage is incorrect here because the function has a return type that should be used. If there is not result then a `for..of` loop should be used.

For the case of converting a list to a map/object we have another `mapFromList` helper.
The iteratee should return a tuple of a `[key, value]` that is merged into the resulting object. Like this:
```suggestion
  return mapFromList(props, prop => [prop, `node.${prop}`]);
```

src/core/database/query.helpers.ts
```diff
@@ -995,10 +995,31 @@ export function listWithSecureObject(props: string[]) {
...
+  const result: any = {};
+  props.map((x) => {
+    result[x] = `{
+        value: coalesce(${x}.value, null),
+        canRead: coalesce(${x}ReadPerm.read, false),
+        canEdit: coalesce(${x}EditPerm.edit, false)
+      }
+    `;
+  });
+  return result;
```
CarsonF, line 1008: We already have securedProperty which does most of this. Also same goes for this as my other comment.
https://github.com/SeedCompany/cord-api-v3/blob/6b19c5456e08e972e902cb71a3e58466a7c1ad60/src/core/database/query.helpers.ts#L927-L931

```suggestion
  return mapFromList(props, prop => [prop, mapping(securedProperty(prop)]);
```

tjgersho: You are right. Thanks.

CarsonF: Thank you! 🎉

PR #970, opened 2020-08-17
fix an intermittent bug. needs to be refactored out.

michaelmarshall:


test/project.e2e-spec.ts
```diff
@@ -587,6 +587,12 @@ describe('Project e2e', () => {
         id: project.id,
       }
     );
+
+    // tech debt: this prevents an intermittant bug
+    setTimeout(() => {
+      //
+    });
```
CarsonF, line 594: Does it? Or does it do nothing and you just aren't experiencing the _intermittent_ bug?

michaelmarshall: Well it did on my machine. Not in CI. I'd toggle it many times to check.

michaelmarshall: I'm giving up and skipping it. this is holding up progress.

michaelmarshall: I'll create a ticket to fix it.

Issue #978, opened 2020-08-18
mutation updateLanguage > 'isDialect' and 'leastOfThese' updated to null for the input value 'false'


iPiyushPatel:
**Describe the bug**
mutation updateLanguage > 'isDialect' and 'leastOfThese' updated to null for the input value 'false'.

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request.

**Expected Result**
'isDialect' and 'leastOfThese' should return the same value as requested.

**Actual Result**
'isDialect' and 'leastOfThese' updated to null for the input value 'false'.
See video - https://drive.google.com/file/d/1LLzOU4mZ0IdInavYKEegTgNV29SbFoDZ/view?usp=drivesdk

GraphQL Query ( updateLanguage ):
![image](https://user-images.githubusercontent.com/5212585/90494884-99679080-e161-11ea-91a3-c90815a2320e.png)

**Query and Input File**
[updateLanguage_isDialect_LeastOfthese_NULL.txt](https://github.com/SeedCompany/cord-api-v3/files/5089434/updateLanguage_isDialect_LeastOfthese_NULL.txt)

CordAPIV3# 29ee3b29

I can replicate the same on the 'integration' environment as well.

PR #981, opened 2020-08-18
Refactor language readOne (set property value to null if canRead is false)

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/978

### DIFF
--- a/src/components/language/language.service.ts
+++ b/src/components/language/language.service.ts
@@ -449,11 +449,12 @@ export class LanguageService {
       if (record?.property === 'sensitivity') {
         response[record.property] = record.value;
       } else {
+        const canRead = perms[record.property]
+          ? perms[record.property].canRead
+          : false;
         response[record.property] = {
-          value: record.value || null,
-          canRead: perms[record.property]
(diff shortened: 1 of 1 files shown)


src/components/language/language.service.ts
```diff
@@ -450,7 +450,7 @@ export class LanguageService {
         response[record.property] = record.value;
       } else {
         response[record.property] = {
-          value: record.value || null,
+          value: record.value ?? null,
```
michaelmarshall, line 453: This returns the value if it exists, but we need to only return the value if the user has canRead=true.

zkhin: `false || null` is `null` so it will raise the https://github.com/SeedCompany/cord-api-v3/issues/978 issue.
So if we only want to return value if the `canRead` is `true` then:
```
const canRead = perms[record.property] ? perms[record.property].canRead : false;
response[record.property] = {
   value: canRead ? record.value : null,
   canRead: canRead,
   ...
};
```
How do you think about this change? @michaelmarshall

PR #985, opened 2020-08-18
Quick refactor ethnologue readOne for anonymous user

zkhin:


src/components/language/ethnologue-language/ethnologue-language.service.ts
```diff
@@ -292,11 +284,12 @@ export class EthnologueLanguageService {
           response[record.property] = {};
         }
         if (record?.property) {
+          const canRead = perms[record.property]
+            ? perms[record.property].canRead
+            : false;
```
CarsonF, line 289: FYI this could be simplified
```suggestion
          const canRead = perms[record.property]?.canRead ?? false;
```

PR #988, opened 2020-08-18
Remove `FileNodeCategory` from API

glassblowerscat:
This field was intended to be used by the UI for determining things like the display icon for a file, but it's been decided to handle this on the front end.


CarsonF (dismissed):
I'd rather wait a few days to be sure that we can't think of anything that this is needed for. 
It's not critical either way, so there's no rush.

CarsonF:
@glassblowerscat I think it's safe to assume we aren't going to use this now right? We are heavily invested until mime types on the UI side, right?

glassblowerscat:
> @glassblowerscat I think it's safe to assume we aren't going to use this now right?

Agreed

CarsonF (approved):

Issue #979, opened 2020-08-18
mutation updateLanguage > User can successfully update the language which is created by another user


iPiyushPatel:
**Describe the bug**
mutation updateLanguage > User can successfully update the language which is created by another user.

**Steps to Reproduce**
1. Login with user1
2. Create a Language
3. Register user2
4. Login with user2
5. Update the language created by user1 (As in step 2).

**Expected Result**
Unauthorized error should be thrown for the user updating language which is not created by him.

**Actual Result**
User can successfully update the language which is created by another user.
See video - https://drive.google.com/file/d/1yC2VofEFKmX-EiB06yoOFCgOHjxAtymD/view?usp=drivesdk

GraphQL Query ( createLanguage ):
![image](https://user-images.githubusercontent.com/5212585/90497986-61624c80-e165-11ea-992e-f2117e5e301e.png)

GraphQL Query ( updateLanguage ):
![image](https://user-images.githubusercontent.com/5212585/90498086-81920b80-e165-11ea-8769-7195e6223859.png)

**Query and Input File**
[createLanguage_anotherUserUpdate_ISSUE.txt](https://github.com/SeedCompany/cord-api-v3/files/5089588/createLanguage_anotherUserUpdate_ISSUE.txt)
[updateLanguage_anotherUserUpdate_ISSUE.txt](https://github.com/SeedCompany/cord-api-v3/files/5089590/updateLanguage_anotherUserUpdate_ISSUE.txt)

CordAPIV3# 29ee3b29

I can replicate the same on the 'integration' environment as well.

michaelmarshall:
adding @izquierdoalfredo as he is also working on this. the new refactor isn't working for permissions yet. I'd rather get that part working as soon as possible so it'll help to have more devs thinking about this.

zkhin:
@michaelmarshall @izquierdoalfredo 
Let's suppose the user doesn't have permission for that `language` and if that user tries to update the language, should we have to throw `ForbiddenException` or just skip the updating?
https://github.com/SeedCompany/cord-api-v3/pull/981 will return null values if the user does not have permission.

michaelmarshall:
I would throw.

PR #989, opened 2020-08-19
Check permission before updating language

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/979

### DIFF
--- a/src/components/project/project-member/project-member.service.ts
+++ b/src/components/project/project-member/project-member.service.ts
@@ -50,9 +50,6 @@ export class ProjectMemberService {
 
   // helper method for defining properties
   property = (prop: string, value: any) => {
-    if (!value) {
-      return [];
-    }
     const createdAt = DateTime.local();
     return [
       [
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -481,13 +481,13 @@ export class DatabaseService {
     const update = this.db
       .query()
       .match([matchSession(session)])
-      .optionalMatch([
+      .match([
         node(nodevar, upperFirst(nodevar), {
           id: object.id,
           active: true,
         }),
       ])
-      .optionalMatch([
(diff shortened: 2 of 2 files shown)


src/components/language/language.service.ts
```diff
@@ -200,6 +202,29 @@ export class LanguageService {
...
+      .query()
+      .call(matchRequestingUser, session)
+      .call(matchUserPermissions, label, id)
+      .return('permList, node')
+      .first();
+
+    if (!result) {
+      throw new ForbiddenException(
+        'User does not have permission to read a language'
+      );
```
CarsonF, line 224: Read or update?

zkhin: Yeah, update!

CarsonF: You aren't checking for either of these permissions though. How do you know which one?

CarsonF: Same ^

zkhin: So I have to check the permissions(`canEdit`) fields one by one? I just checked the `permList` and if it's empty array, it throws an exception.

michaelmarshall: @zkhin try these changes on lines 490+ 498. That should do it. Let me know.
image

zkhin: After only these changes, we will get `Could not find object` 404 exceptions if the user does not have permission. Is this fine? And I don't need to check permissions before updating `Language`?

michaelmarshall: yes that is fine for now. after mvp we will refactor to give the FE better errors.

zkhin: Okay, cool

CarsonF (dismissed):
If it's good with @michaelmarshall it's good with me! 👍

Issue #889, opened 2020-08-05
Budget Records array is empty


glassblowerscat:
**Describe the bug**
Querying value of a `Project`'s budget records just returns an empty array

**Steps to Reproduce**
1. Create a new project
2. Run `createPartnership` to create `Funding` partnership between the project and an existing organization
3. Run `updateProject` and provide values for the `mouStart` and `mouEnd` fields (I used `2020-08-04` and `2021-08-04`, respectively)
4. Run the following query and see that `records` is an empty array:

```
query {
  project(id: $id) {
    budget {
      value {
        id
        records {
          id
        }
      }
    }
  }
}
```

**Please give the commit short hash of the API and Front End repo if you are able to**.  
API: `344e5af6`
FE: `456a2384`

**Expected and Actual Results**
Expected: A project with a valid budget, funding partnership, and `mouStart` and `mouEnd` dates should return budget records
Actual: Budget records array is empty

**Platform/Environment**  
Clear report on what the bug can be reproduced. Please complete the following information accordingly.

Desktop:
- OS: MacOS 10.15.4
- Browser: Playground
- Version: N/A

thomas-adev:
@glassblowerscat Can you provide an example of the specific "createPartnership" query you are running along with your query variables.

glassblowerscat:
@thomas-adev Sure! Here's the mutation I just did:

```
mutation {
  createPartnership(input: {partnership: {organizationId: "seedcompanyid", projectId: "XCVtcjJeE", types: Funding}}) {
    partnership {
      id
      types {
        value
      }
    }
  }
}
```

`XCVtcjJeE` is a Translation Project I just created. Here's the return I got back when I sent the mutation:

```
{
  "data": {
    "createPartnership": {
      "partnership": {
        "id": "V_zVbVfWP",
        "types": {
          "value": [
            "Funding"
          ],
          "__typename": "SecuredPartnershipTypes"
        },
        "__typename": "Partnership"
      },
      "__typename": "CreatePartnershipOutput"
    }
  },
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": "2020-08-05T16:59:38.784Z",
      "endTime": "2020-08-05T16:59:39.552Z",
      "duration": 767455522,
      "execution": {
        "resolvers": [
          {
            "path": [
              "createPartnership"
            ],
            "parentType": "Mutation",
            "fieldName": "createPartnership",
            "returnType": "CreatePartnershipOutput!",
[...]

PR #991, opened 2020-08-19
Fix budget records array empty

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/889

### DIFF
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -6,14 +6,9 @@ import {
 } from '@nestjs/common';
 import { node, Query, relation } from 'cypher-query-builder';
 import { RelationDirection } from 'cypher-query-builder/dist/typings/clauses/relation-pattern';
-import { flatMap, upperFirst } from 'lodash';
+import { upperFirst } from 'lodash';
 import { DateTime } from 'luxon';
-import {
-  fiscalYears,
-  InputException,
-  ISession,
-  NotFoundException,
-} from '../../common';
+import { InputException, ISession, NotFoundException } from '../../common';
 import {
   addAllMetaPropertiesOfChildBaseNodes,
   addAllSecureProperties,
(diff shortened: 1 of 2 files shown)


src/components/partnership/partnership.service.ts
```diff
@@ -397,12 +364,6 @@ export class PartnershipService {
       .call(matchUserPermissions, 'Partnership', id)
       .call(addAllSecureProperties, ...props)
       .optionalMatch([
-        node('canReadProject', 'Permission', {
-          property: 'project',
-          read: true,
-          active: true,
-        }),
-        relation('out', '', 'baseNode'),
```
CarsonF, line 405: Was this check doing nothing?

zkhin: Yes, I don't think we need to check the `project` permission because while creating `Partnership`, we don't add `project' permission.

zkhin:
> This is a good start, but it's not the only problem. If a partnership gets updated to funding type, then the array is still empty.

I created a partnership without `Funding` type first. And then I updated Partnership with `Funding` type and query `project` and it returns budget records. Can you try on your side?

src/components/partnership/partnership.service.ts
```diff
@@ -326,38 +321,10 @@ export class PartnershipService {
 
       const partnership = await this.readOne(result.id, session);
 
-      if (!input.types) {
-        return partnership;
-      }
```
CarsonF, line 331: This was (incorrectly) done to handle types being null, when they are defined as an array. Can you handle this? Maybe the DTO gets a default value or maybe a check is done on the DB result to ensure it's an array?

zkhin: How's about this for handling an empty array.
```
if (!input.types || isEmpty(input.types)) {
   return partnership;
}
```

CarsonF: Sorry for the confusion. The early return is wrong. We always want to publish the event. 

It's the fact that we have `types` as an array but it was sometimes `null`. The change you made on the DTO might be enough. If it's not, then we should ensure that it's an array here. But we should always publish the event.

zkhin: Okay got it

zkhin: I think changing on DTO is enough because `readOne` will return `[]` if the `types` is `null`.

PR #993, opened 2020-08-20
Refactor Engagement ReadOne

frankyma:
Following Language ReadOne refactor.  

For the `startDate` and `endDate` values we check if `canRead` is true on a ternary, but since the engagement type doesn't allow a `null` value I had to evaluate to `undefined`.  Please double check my Typescript line 923

```
    const startDate = canReadStartDate
      ? props.startDateOverride ?? project.mouStart.value
      : undefined;
```

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -13,12 +13,6 @@ import { DateTime } from 'luxon';
 import { generate } from 'shortid';
 import { DuplicateException, ISession } from '../../common';
 import {
-  addAllMetaPropertiesOfChildBaseNodes,
-  addAllSecureProperties,
-  addPropertyCoalesceWithClause,
-  addShapeForBaseNodeMetaProperty,
-  addShapeForChildBaseNodeMetaProperty,
-  ChildBaseNodeMetaProperty,
   ConfigService,
   DatabaseService,
   filterByBaseNodeId,
@@ -26,10 +20,16 @@ import {
   Logger,
   matchRequestingUser,
   matchSession,
-  matchUserPermissions,
   matchUserPermissionsForList,
   runListQuery,
 } from '../../core';
+import {
+  DbPropsOfDto,
+  parseBaseNodeProperties,
+  parsePropList,
+  parseSecuredProperties,
+  StandardReadResult,
+} from '../../core/database/results';
 import { CeremonyService } from '../ceremony';
 import { CeremonyType } from '../ceremony/dto/type.enum';
 import { FileService } from '../file';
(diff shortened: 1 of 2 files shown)


src/components/engagement/engagement.service.ts
```diff
@@ -982,12 +945,36 @@ export class EngagementService {
...
-      return (response as unknown) as LanguageEngagement;
-    } else if (result.__typename === 'InternshipEngagement') {
-      return (response as unknown) as InternshipEngagement;
+    if (engagement.__typename === 'LanguageEngagement') {
+      return (engagement as unknown) as LanguageEngagement;
+    } else if (engagement.__typename === 'InternshipEngagement') {
+      return (engagement as unknown) as InternshipEngagement;
     } else {
       throw new NotFoundException('could not find Engagement');
     }
```
CarsonF, line 980: Is this still needed?

src/components/engagement/engagement.service.ts
```diff
@@ -777,7 +777,7 @@ export class EngagementService {
   async readOne(
     id: string,
     session: ISession
-  ): Promise<LanguageEngagement | InternshipEngagement> {
+  ): Promise<LanguageEngagement | InternshipEngagement | any> {
```
CarsonF:
```suggestion
  ): Promise<LanguageEngagement | InternshipEngagement> {
```

src/components/engagement/engagement.service.ts
frankyma: Michael said yes (a few refactors ago) but I don't know.  I'll defer to your expertise

src/components/engagement/engagement.service.ts
frankyma: yep I missed that on the first commit, fixed

src/components/engagement/engagement.service.ts
CarsonF: Typecasting is just to make TS happy. If it can be removed, it should be. As long as the return type of the function doesn't change.

michaelmarshall: I think this is just for TS.

frankyma: removed this typecasting

frankyma:
refactored the engagement dto from `SecuredDate` to `SecuredDateNullable`

src/components/engagement/engagement.service.ts
```diff
@@ -982,15 +945,31 @@ export class EngagementService {
...
         canEdit: false,
       },
+      methodologies: {
+        ...securedProperties.methodologies,
+        value: securedProperties.methodologies.value ?? [],
+      },
+      ceremony: {
+        ...securedProperties.ceremony,
+        value: result.ceremonyId,
+      },
```
CarsonF, line 955: Would be nice if there was a way to handle these "relationship" reference more easily. 🤔

frankyma: 🤔🤔 seems simple enough already, what did you have in mind?

CarsonF: Yeah idk It's good enough for now 👍

PR #996, opened 2020-08-20
973 refactor project readone

izquierdoalfredo:
#973


izquierdoalfredo:
Yes, I will update project
Zayar is working in the location now.

PR #1020, opened 2020-08-21
Lint exception imports

glassblowerscat:
- Un-comment rule in `.eslintrc.yml` that prohibits import of `@nestjs/common`
  Exception classes
- Replace all instances of NestJS Exceptions with our own, EXCEPT
- In `src/core/exception.filter` and `src/core/exception.filter.test`, add
  `eslint-disable` wrapper around imports of NestJS Exceptions. These are
  needed in order to replace the default Exceptions with our own
- Create `NoSessionException` class in `authentication` module to handle specific use case for session handling in Apollo client
- Add `field` and `previous` params to exceptions where appropriate

Also, we have quite a few instances of this kind of thing:

```
} catch (e) {
  this.logger.info(
    'Failed to use existing session token, creating new one.',
    { exception: e }
  );
  throw new ServerException('Could not delete user', exception);
}
```

Sometimes we were naming this error `e`, as above, but sometimes `exception`, so we could use property value shorthand. While I was updating the `ServerException`s, I changed every instance I encountered to `exception`, for consistency.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -35,29 +35,28 @@ rules:
     - paths:
         - name: '@nestjs/cqrs'
           message: Import from our core module instead
-# TODO Enable and fix errors
-#        - name: '@nestjs/common'
-#          importNames:
-#            - BadRequestException
-#            - HttpException
-#            - UnauthorizedException
-#            - MethodNotAllowedException
-#            - NotFoundException
-#            - ForbiddenException
-#            - NotAcceptableException
-#            - RequestTimeoutException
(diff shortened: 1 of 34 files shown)


glassblowerscat:
> There's a couple log statements that pass the exception as the context instead of at the correct key. Would you mind grabbing those too while you're here?

Whoops, I totally missed this comment. I haven't done this yet

glassblowerscat:
> Whoops, I totally missed this comment. I haven't done this yet

Okay, got 'em

PR #1027, opened 2020-08-23
feat:refactor project-member readOne

izquierdoalfredo:
https://github.com/SeedCompany/cord-api-v3/issues/1023

### DIFF
--- a/src/components/project/project-member/project-member.service.ts
+++ b/src/components/project/project-member/project-member.service.ts
@@ -14,14 +14,13 @@ import {
   addAllMetaPropertiesOfChildBaseNodes,
   addAllSecureProperties,
   addBaseNodeMetaPropsWithClause,
-  addPropertyCoalesceWithClause,
-  addShapeForBaseNodeMetaProperty,
-  addShapeForChildBaseNodeMetaProperty,
   ChildBaseNodeMetaProperty,
   ConfigService,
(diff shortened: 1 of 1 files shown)


src/components/project/project-member/project-member.service.ts
```diff
@@ -279,64 +287,62 @@ export class ProjectMemberService {
...
+
+    return {
+      ...parseBaseNodeProperties(result.node),
+      ...securedProps,
       user: {
         value: await this.userService.readOne(result.userId, session),
-        canRead: !!result.canReadUser,
-        canEdit: !!result.canEditUser,
+        canRead: !!(perms as any).user?.canRead,
+        canEdit: !!(perms as any).user?.canEdit,
```
CarsonF, line 338: We should be able to handle these types for this case without using `as any`. Have you tried using `user: true` on `parseSecureProperties`? If so, I'll take a look tomorrow.

izquierdoalfredo: HI @CarsonF Please take a look at and let me know

CarsonF:
@izquierdoalfredo https://github.com/SeedCompany/cord-api-v3/pull/1034 should fix your problem here. After rebasing on master, you should be able to use `user: true` in `parseSecureProperties` and then spread it into the result.

PR #1030, opened 2020-08-23
added phone number validation for create and update

### LINKED ISSUE #684: Update User > updateUser($input: UpdateUserInput!) > User can update details with alphabetic characters in Phone field
**Describe the bug**
Update User > updateUser($input: UpdateUserInput!) > User can update details with alphabetic characters in Phone field

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
Validation message should be returned for 'phone ' field.

**Actual Result**
User details updated successfully. See video - https://www.awesomescreenshot.com/video/346451?key=f1f24df48be63a1b9353175cd9b1ac88

GraphQL Query( updateUser ):
image

**Query and Input File**
[updateUser_phone_characters.txt](https://github.com/SeedCompany/cord-api-v3/files/4884764/updateUser_phone_characters.txt)

CordAPIV3# 8b7caae

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466975432) by [Unito](https://www.unito.io)

thomas-adev:
Closes #684 

Commit uses class-validator library to add decorator to the create-person and update-user DTOs.

### DIFF
--- a/src/components/user/dto/create-person.dto.ts
+++ b/src/components/user/dto/create-person.dto.ts
@@ -1,6 +1,6 @@
 import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import { Type } from 'class-transformer';
-import { ValidateNested } from 'class-validator';
+import { IsPhoneNumber, ValidateNested } from 'class-validator';
 import { IsEmail, IsIanaTimezone, NameField } from '../../../common';
 import { Role } from '../../project/project-member/dto/role.dto';
 import { UserStatus } from './user-status.enum';
@@ -25,6 +25,7 @@ export abstract class CreatePerson {
   readonly displayLastName: string;
 
   @Field({ nullable: true })
+  @IsPhoneNumber('US')
   readonly phone?: string;
 
   @Field({ nullable: true })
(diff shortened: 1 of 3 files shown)


CarsonF:
Sorry @thomas-adev I should've noted this on the issue and will now. We can't enforce this right now. We cannot assume that phone numbers are only US numbers. And we can't confirm that will migrated numbers will be valid according to this format.
We also don't want to enforce this yet because the frontend doesn't yet support this validation. We'll do more in the future to conform our data to a standard when it becomes a priority.

thomas-adev:
@CarsonF Ok, we can drop this branch then. For the future, the library does easily support international area codes if we swap out 'US' for 'ZZ', but that would require the client to pass in the international prefix.

CarsonF:
Right and we need to handle and communicate that on the UI. All in all just a lot of complexity for not a lot of benefit. We don't yet use this number for anything programmatically so it's only for humans

CarsonF:
I appreciate the effort though. This is the right solution, just not right now.

PR #1031, opened 2020-08-23
Add product test

tjgersho:


src/components/scripture/reference.ts
```diff
@@ -124,3 +124,27 @@ export const bookCodeToName = (code: string) => {
...
 };
+// return random scriptureRefenence as an array
+export const createRandomScriptureReferences = (): ScriptureRangeInput[] => {
+  //To do
+  // const book = books[Math.floor(Math.random() * books.length)];
+  const book = books[Math.floor(Math.random() * 2)];
+  // const startChapter = book.chapters[Math.floor(Math.random() * book.chapters.length)];
+  const startChapter = book.chapters[Math.floor(Math.random() * 2)];
+  // const endChapter = book.chapters[Math.floor(Math.random() * book.chapters.length)];
+  const endChapter = book.chapters[Math.floor(Math.random() * 2)];
```
CarsonF, line 135: lodash's `random` might be easier to work with. There's also `faker.random.arrayElement`

src/components/scripture/reference.ts
```diff
@@ -128,11 +129,11 @@ export const bookCodeToName = (code: string) => {
...
   //To do
   // const book = books[Math.floor(Math.random() * books.length)];
-  const book = books[Math.floor(Math.random() * 2)];
+  const book = books[random(1)];
   // const startChapter = book.chapters[Math.floor(Math.random() * book.chapters.length)];
-  const startChapter = book.chapters[Math.floor(Math.random() * 2)];
+  const startChapter = book.chapters[random(1)];
   // const endChapter = book.chapters[Math.floor(Math.random() * book.chapters.length)];
-  const endChapter = book.chapters[Math.floor(Math.random() * 2)];
+  const endChapter = book.chapters[random(1)];
```
CarsonF, line 136: That's only 0 or 1. Why not
```suggestion
  const endChapter = book.chapters[random(book.chapters.length - 1)];
```

tjgersho: That is what I tried before on my local. Sometimes, invalid verse number error is occurring when selecting the random start and end chapters. I created an issue about it. Need to figure out the reason.

CarsonF: Ok sounds good

izquierdoalfredo:
@tjgersho How do you add tests to the product?

PR #1034, opened 2020-08-24
DbPropsOfDto - Allow props that are not native DB values as unknown

CarsonF:
Instead of omitting these props we just change their value to `unknown`.
This allows them to be used in permissions, but still requires their values to be handled explicitly some other way.

For example:
```ts
export class ProjectMember extends Resource {
  user: SecuredUser
  roles: SecuredRoles
  modifiedAt: DateTime
}

// Previously DbPropsOfDto<ProjectMember> would be this:
type ProjectMemberDbProps = {
  roles: Role[]
  modifiedAt: DateTime
}

// Now it's:
type ProjectMemberDbProps = {
  user: unknown
  roles: Role[]
  modifiedAt: DateTime
}
```

### DIFF
--- a/src/core/database/results/types.ts
+++ b/src/core/database/results/types.ts
@@ -1,6 +1,6 @@
 import type { Node } from 'cypher-query-builder';
 import type { DateTime } from 'luxon';
-import type { ConditionalPick, Except } from 'type-fest';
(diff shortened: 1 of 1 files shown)


tjgersho:
Copy that. Need to check other readOne functions.

PR #1038, opened 2020-08-24
Mutate Project.step in create update

### LINKED ISSUE #1037: Add project step field in create and update project
In the project create and update, expose step field

frankyma:
Expose step field in create and update project, in create default to `EarlyConversations`

Closes #1037


src/components/project/project.service.ts
```diff
@@ -514,9 +513,7 @@ export class ProjectService {
     const changes = {
       ...input,
       modifiedAt: DateTime.local(),
-      status: object.step.value
-        ? stepToStatus(object.step.value)
-        : object.status,
+      status: input.step ? stepToStatus(input.step) : object.status,
```
CarsonF, line 516: Let's skip the change completely by omitting the key:
```suggestion
      ...(input.step ? { status: stepToStatus(input.step) } : {}),
```

src/components/project/project.service.ts
```diff
@@ -212,8 +212,7 @@ export class ProjectService {
     const createdAt = DateTime.local();
     const createInput = {
       sensitivity: Sensitivity.High, // TODO: this needs to be calculated based on language engagement
-      step: ProjectStep.EarlyConversations,
-      status: stepToStatus(ProjectStep.EarlyConversations),
+      status: stepToStatus(input.step || ProjectStep.EarlyConversations),
```
CarsonF, line 215: Shouldn't need it both here and in the DTO

frankyma: It's giving me a typeError because input.step can be undefined since it's an optional field in the DTO

src/components/project/project.service.ts
frankyma: changed

src/components/project/project.service.ts
frankyma: actually I'll add a !

src/components/project/dto/create-project.dto.ts
```diff
@@ -27,6 +28,9 @@ export abstract class CreateProject {
 
   @DateField({ nullable: true })
   readonly estimatedSubmission?: CalendarDate;
+
+  @Field(() => ProjectStep, { nullable: true })
+  readonly step?: ProjectStep = ProjectStep.EarlyConversations;
```
CarsonF, line 33:
```suggestion
  readonly step: ProjectStep = ProjectStep.EarlyConversations;
```
This will fix the need to not-null assert below.

CarsonF (dismissed):
LGTM!

src/components/project/dto/create-project.dto.ts
frankyma: Ah ok I actually had that at one point but didn't know if that's ok

PR #1043, opened 2020-08-24
fix partnership list

tjgersho:


src/components/partnership/partnership.service.ts
```diff
@@ -600,7 +600,7 @@ export class PartnershipService {
         const resultOne = await this.readOne(item.id, session);
 
         return {
-          ...item,
+          ...resultOne,
```
CarsonF, line 603: Can't this entire block be deleted? All of this logic is already in `readOne`

PR #1052, opened 2020-08-25
update toVerseId

tjgersho:


tjgersho:
Right. It was not working for the valid input too because the (chapter input - 1) is not the chapter index in the chapter array. So I update the code to get chapter index from chapter input.

CarsonF:
Ok in other words the chapter input is 1-indexed but the chapters array is 0-indexed.

I would really this validation to go [here](https://github.com/SeedCompany/cord-api-v3/blob/1cb2c0213d2638ed61b2a82dbcea540a0a695210/src/components/scripture/dto/scripture-reference.dto.ts#L10) with a validation decorator.
This would give the appropriate exception message for the input. i.e. validation error with the correct field.
I'll make an issue.

tjgersho:
> Ok in other words the chapter input is 1-indexed but the chapters array is 0-indexed.

Not exactly, Once chapter input is given, I should see what is the chapter index for the chapter input in chapters array:
[here](https://github.com/SeedCompany/cord-api-v3/blob/e60867acb2a339c1b0a2c9b553835dd1c795b727/src/components/scripture/books.ts#L23-L73)
So for example, if chapter input is 31, I should use 0.

I should get bookIndex to check if chapter input is valid like [here](https://github.com/SeedCompany/cord-api-v3/blob/e60867acb2a339c1b0a2c9b553835dd1c795b727/src/components/scripture/reference.ts#L105-L116). Can we pass the book input value to another validator?!

CarsonF:
> Not exactly, Once chapter input is given, I should see what is the chapter index for the chapter input in chapters array:
> [here](https://github.com/SeedCompany/cord-api-v3/blob/e60867acb2a339c1b0a2c9b553835dd1c795b727/src/components/scripture/books.ts#L23-L73)
> So for example, if chapter input is 31, I should use 0.

I don't think that's right. That array is the number of chapters and each value is the number of verses in that chapter.

tjgersho:
Gotcha. I must update the logic. I think I did not get the KT about the scriptureReferences.

CarsonF:
Yeah the code is pretty obtuse, it was ported from another library.

tjgersho:
I am interested in the validator for the book, chapter and verse. The chapter's validation can be checked when the book is given.
Can we implement it? 🤔

CarsonF:
Sure you can do it. Do you want to zoom?

tjgersho:
Thanks. No problems. I just got it how to do it. The first version of scriptureReference would work once we implement the validation for chapters.

Issue #1040, opened 2020-08-24
Add test to Budget e2e


mikecoll:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466575879) by [Unito](https://www.unito.io)

izquierdoalfredo:
@CarsonF 
How I can choose tests in postman for moving to e2e suite?

reneepe:
VERIFIED with https://github.com/SeedCompany/cord-api-v3/pull/1056.

Invalid budget statuses aren't allowed when updating budget. Below are the remaining results.

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/304b96e6-7765-4518-8b73-26a0229cb2cf)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/39ef869e-c799-4d2a-91fb-fae507166865)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/c631a4b8-08ba-4718-bf59-8819d3966229)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/cbdc62d8-c5a0-4bb9-b0fd-afb0defe6c47)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/75e3524e-fee1-4242-b59a-63002004e92e)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/ce2fc458-8fec-43bd-83b4-f4175509bc39)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/780496ca-3886-4d6e-8660-c7718d397f3e)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/a7b4a679-9dc1-4803-ae64-1d3a0d1d6817)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/ef9ee0e8-e4b2-4ca4-9b68-98701c6db698)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/09d04954-822b-411c-bb32-8c2d8138b742)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/89830979-d0fe-44de-97ac-d401bf456e52)

[...]

PR #1056, opened 2020-08-26
add test case in budget

izquierdoalfredo:
https://github.com/SeedCompany/cord-api-v3/issues/1040

### DIFF
--- a/test/budget.e2e-spec.ts
+++ b/test/budget.e2e-spec.ts
@@ -1,4 +1,5 @@
 import { gql } from 'apollo-server-core';
+import { times } from 'lodash';
 import {
   CalendarDate,
   fiscalYears,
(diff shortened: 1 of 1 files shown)


izquierdoalfredo:
Yes, I relpaced toThrow method by expectNotFound

CarsonF:
This coded before the validation pipe was re-enabled on e2e tests. It should be rebased and error messaged updated to actually assert what the test says it's asserting.

CarsonF (commented):
Other than list of budgets we decided on slack that these tests aren't necessary. They are the same src code used across the entire codebase and don't need to be tested in every module

elisabeth-sorrell:
@CarsonF or @michaelmarshall Can you take another look and see if ready for merge? thanks

CarsonF (approved):

PR #1057, opened 2020-08-26
836 fix root directory error

tjgersho:


src/components/scripture/reference.ts
```diff
@@ -99,6 +101,18 @@ const bookIndexFromName = (name: string) => {
...
+const chapterIndexFromChapterNumber = (
+  bookIndex: number,
+  chapterNumber: number
+) => {
+  const chapter = books[bookIndex].chapters.find((c) => {
+    return c === chapterNumber;
+  });
+  if (chapter) {
+    return books[bookIndex].chapters.indexOf(chapter);
+  }
```
CarsonF, line 113: Is chapter index different from chapter number or is this just validating that the given index is in the array?

Why not skip the `.find()`?
```tsx
const chapterIndex = books[bookIndex].chapters.indexOf(chapter);
if (chapterIndex === -1) {
   throw new InputException('');
}
return chapterIndex;
```

test/file.e2e-spec.ts
```diff
@@ -490,6 +491,24 @@ describe('File e2e', () => {
         )
         .run();
       await expectConsistency(FileNodeType.File, false);
+      // revert the changes so consistency check will be passed for remaining file nodes.
+      await app
```
CarsonF, line 495: Since this is clean up, it should be in an `afterEach/afterAll` method so it gets ran regardless of the `expectConsistency` result.

test/file.e2e-spec.ts
```diff
@@ -118,9 +118,10 @@ describe('File e2e', () => {
   });
 
   beforeEach(async () => {
-    const db = app.get(DatabaseService);
+    // const db = app.get(DatabaseService);
     // remove old data to ensure consistency check
-    await db.query().matchNode('n', 'FileNode').detachDelete('n').run();
+    // commented out the delete code so project's root directory exists after tests.
+    // await db.query().matchNode('n', 'FileNode').detachDelete('n').run();
```
CarsonF, line 124: Don't comment out code. This is what version control is for.
```suggestion
```

src/components/scripture/reference.ts
tjgersho: Yes, chapter index is different from chapter number.
Ah, can skip the `.find()` .

Issue #789, opened 2020-07-21
query createOrganization > AvatarLetters are returned blank for Organization name contains Capital and Small letters


iPiyushPatel:
**Describe the bug**
query createOrganization > AvatarLetters are returned blank for Organization name contains Capital and Small letters.

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
AvatarLetters are returned according to the Organization name.

**Actual Result**
AvatarLetters are returned blank for Organization name contains Capital and Small letters.
See video - https://drive.google.com/file/d/1Rf0uEnmDyR-Yz0wS5ACIZwfWo4T8a5vk/view?usp=drivesdk

GraphQL Query ( createOrganization ):
![image](https://user-images.githubusercontent.com/5212585/88022105-498aaf00-cb4c-11ea-855a-dd4b513007e8.png)

**Query and Input File**
[createOrganization_avtarLettersNULL.txt](https://github.com/SeedCompany/cord-api-v3/files/4951776/createOrganization_avtarLettersNULL.txt)

CordAPIV3# 65945ca

candleshine:
Org Names that are valid do not include mix case words like `orgName`. 
Expected behavior is 
`Seed Company` `SC`
`seed company` `sc`
`Seed company` `S`
`seed Company` `C`
`Project 2020` `P2`
`project 2020` `p2`
`orgName` `` (blank)
@sethmcknight is this acceptable or do you want any of these to change?  If acceptable, please feel free to close this ticket

iPiyushPatel:
Any conclusion on this issue as I'm seeing same issue on Organization, Users, Project as well?

@candleshine @sethmcknight @CarsonF @michaelmarshall

reneepe:
Bug verified. Avatar Letters are returned blank when an organization's name has a lower case letter first (ie, "orG").

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/6e70eddc-0c6c-4703-832e-8b468ebe0652)

Avatar Letters return data when the organization's name is in one case (ie, "org," or "ORG") or it has a capital letter first (ie, "Org").  

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/fa9aa5b9-8b18-455b-8a98-5697dd436d98)

[...]

PR #1060, opened 2020-08-26
Fix avatar letters lowercase issue

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/789

### DIFF
--- a/src/common/firstLettersOfWords.ts
+++ b/src/common/firstLettersOfWords.ts
@@ -21,6 +21,9 @@ const lowercasePattern = RegExp(
   'ug'
 );
 
+const matchAndMerge = (pattern: RegExp, str: string, group = 1) =>
+  Array.from(str.matchAll(pattern), (m) => m[group]).join('');
+
 export function firstLettersOfWords(
   words: string,
   limit: number | null = 3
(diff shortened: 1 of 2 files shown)


src/common/firstLettersOfWords.ts
```diff
@@ -33,7 +35,12 @@ export function firstLettersOfWords(
   const pattern = words.match(hasUppercaseLettersPattern)
     ? uppercasePattern
     : lowercasePattern;
-  const letters = Array.from(words.matchAll(pattern), (m) => m[1]).join('');
+  let letters = Array.from(words.matchAll(pattern), (m) => m[1]).join('');
+
+  letters = isEmpty(letters)
+    ? Array.from(words.matchAll(lowercasePattern), (m) => m[1]).join('')
+    : letters;
```
CarsonF, line 42:
```suggestion
  letters = letters || Array.from(words.matchAll(lowercasePattern), (m) => m[1]).join('');
```

CarsonF: Better yet would be to abstract that function
```tsx
const matchAndMerge = (pattern: RegExp, str: string, group = 1) =>
  Array.from(str.matchAll(pattern), (m) => m[group]).join('');
```
Then it usage looks even better
```tsx
  const letters =
    matchAndMerge(pattern, words) || matchAndMerge(lowercasePattern, words);
```

PR #1067, opened 2020-08-26
Bugfix: add project member after delete

frankyma:
Couldn't add the same user as project member after deleting it because wasn't looking for the `active` tag on existing project members

### DIFF
--- a/test/project-member.e2e-spec.ts
+++ b/test/project-member.e2e-spec.ts
@@ -161,6 +161,34 @@ describe('ProjectMember e2e', () => {
     ).rejects.toThrowError();
   });
 
+  it('Can create the same projectMember after deletion', async () => {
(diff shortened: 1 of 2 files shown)


test/project-member.e2e-spec.ts
```diff
@@ -159,6 +159,13 @@ describe('ProjectMember e2e', () => {
         }
       )
     ).rejects.toThrowError();
+
+    const newProjectMember = await createProjectMember(app, {
+      userId: member.id,
+      projectId: project.id,
+    });
+
+    expect(newProjectMember.id).toBeTruthy();
```
CarsonF, line 168: Can you put this in another test? Tests should only test one thing.

frankyma: got it, done

Issue #837, opened 2020-07-28
query budgets > Sort by 'status' with 'ASC / DESC' order is not working


iPiyushPatel:
**Describe the bug**
query budgets > Sort by 'status' with 'ASC / DESC' order is not working

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap & attached notepad file and execute request.

**Expected Result**
Budget items should be sorted by 'status' in ascending / descending order.

**Actual Result**
Sort by 'status' with 'ASC / DESC' order is not working.
See video - https://drive.google.com/file/d/1L-JmVlmJxEFCbAcq5MHrZTJDDYWb-XgX/view?usp=drivesdk

GraphQL Query ( budgets - ASC ):
![image](https://user-images.githubusercontent.com/5212585/88623941-abef2c80-d0c3-11ea-97e4-2183b6ec0ec2.png)

GraphQL Query ( budgets - DESC ):
![image](https://user-images.githubusercontent.com/5212585/88624324-5f582100-d0c4-11ea-96ba-8f3ec2a92fee.png)

**Query and Input File**
[budgets_sort.txt](https://github.com/SeedCompany/cord-api-v3/files/4986243/budgets_sort.txt)

CordAPIV3# c129e39

reneepe:
@zkhin The budget list is still sorted by date instead of status.

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/b14dd274-56ae-41a4-883f-49dca86ea609)

reneepe:
@zkhin It passed. The list is sorted and ascends as expected.

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/145aa09a-f0d7-4282-8fad-d22a5ce3b311)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/2bbc8c8b-1bf5-4592-b6f4-1a33388f018e)


The list is sorted and descends as expected as well.

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/601eac83-50c9-4726-a2a4-299c5ca27f38)

![image.png](https://images.zenhubusercontent.com/5d66bac4a77bcf04870bc8bc/888042c7-5197-42be-ba7d-10ab63a2b9ff)

iPiyushPatel:
Verified and passed.

PR #1068, opened 2020-08-27
Refactor budget lists

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/837
Use new `Project.list` pattern


CarsonF (changes requested):
@zkhin could you do the budget records list too?

zkhin:
> @zkhin could you do the budget records list too?

Sure

CarsonF (approved):
🚀

PR #1069, opened 2020-08-27
add validation for scripture references

tjgersho:


src/common/validators/scripture-references.validator.ts
```diff
@@ -0,0 +1,52 @@
...
+  ValidateBy(
+    {
+      name: 'isValidChapter',
+      constraints: [book],
+      validator: {
+        validate: (value, args: ValidationArguments) => {
+          const [book] = args.constraints;
+          const bookName = (args.object as any)[book];
+          const bookIndex = bookIndexFromName(bookName);
+          return value <= books[bookIndex].chapters.length;
```
CarsonF, line 19: Could we move the actual validation logic to a `validateChapter(book, chapter)` function in the reference file?
It would be better for separation of concerns. That can handle the actual logic and this file can just convert that function to a validation decorator.

src/common/validators/scripture-references.validator.ts
```diff
@@ -0,0 +1,52 @@
...
+  validationOptions?: ValidationOptions
+) =>
+  ValidateBy(
+    {
+      name: 'isValidChapter',
+      constraints: [book],
+      validator: {
+        validate: (value, args: ValidationArguments) => {
+          const [book] = args.constraints;
+          const bookName = (args.object as any)[book];
```
CarsonF, line 17: I think it's ok to assume the shape of the object. It will only be used on `ScriptureReference` object.

src/common/validators/scripture-references.validator.ts
```diff
@@ -0,0 +1,52 @@
...
+      constraints: [book, chapter],
+      validator: {
+        validate: (value, args: ValidationArguments) => {
+          const [book, chapter] = args.constraints;
+          const bookName = (args.object as any)[book];
+          const chapterNumber = (args.object as any)[chapter];
+          const bookIndex = bookIndexFromName(bookName);
+          if (chapterNumber > books[bookIndex].chapters.length) {
+            // chapter is invalid
+            return true;
```
CarsonF, line 44: This function should return true if it is valid. Do you mean return false here?
Also do same thing here as mentioned above.

src/components/scripture/dto/scripture-reference.dto.ts
```diff
@@ -12,19 +13,20 @@ export abstract class ScriptureReferenceInput {
   @Field({
     description: 'The code of the Bible book',
   })
-  // TODO validate code against list
   book: string;
 
   @Field(() => Int, {
     description: 'The chapter number',
   })
   @IsPositive()
```
CarsonF, line 21: Not necessary anymore.
```suggestion
```

src/components/scripture/dto/scripture-reference.dto.ts
```diff
@@ -12,19 +13,20 @@ export abstract class ScriptureReferenceInput {
   @Field({
     description: 'The code of the Bible book',
   })
-  // TODO validate code against list
```
CarsonF, line 15: Book code is not being validated yet so this comment should stay.

src/components/scripture/dto/scripture-reference.dto.ts
tjgersho: I will add the logic for a positive number to the validate function. so it can be deleted.

src/common/validators/scripture-references.validator.ts
tjgersho: What I mean is checking verse number has no meaning once the chapter is invalid.

src/common/validators/scripture-references.validator.ts
tjgersho: Gotcha

src/common/validators/scripture-references.validator.ts
tjgersho: Just pushed a new commit. Can you suggest the code change to assume the shape of the object?

tjgersho:
I pushed a new commit recently again. It is ready to be checked.

CarsonF (approved):
Thanks @tjgersho! I did some more work...hope you don't mind 😁

tjgersho:
🚀  Great update.

Issue #1059, opened 2020-08-26
query project > "Could not find root directory associated to this project" without logging in


zkhin:
## Current Behavior
Try to query project without logging and get the error message: "Could not find root directory associated to this project"

## Expected Behavior

It should return `null` value

## Steps to Reproduce

Query project with id without logging
```
query project($id: ID!) {
  project(id: $id) {
    ...project
  }
}

fragment project on Project {
  id
  rootDirectory {
    id
  }
}

"message": "Could not find root directory associated to this project",

**Comment by tjgersho** [comment]
I think this issue can be fixed after merging [#1057](https://github.com/SeedCompany/cord-api-v3/pull/1057) and wiping the database.

**Comment by zkhin** [comment]
@tjgersho I think it's not related to testing cases... If the `session.userId` is `null` or `anonUser`, it will throw `Notfound` exception (https://github.com/SeedCompany/cord-api-v3/blob/master/src/components/project/project.service.ts#L834~L858)

**Comment by tjgersho** [comment]
Ahh, yes. This is another issue.

**Comment by zkhin** [comment]
@CarsonF Can you verify this bug if you have a chance?

**Comment by reneepe** [comment]
@zkhin I'm getting the following error when I query with rootDirectory. Am I supposed to receive the error message below or the error message stating that the root directory wasn't found? I'm unsure of the expected results. 
![image](https://user-images.githubusercontent.com/38309285/91690524-1b799100-eb34-11ea-83b4-8d829802acc5.png)

As expected, querying with all other fields returned data without an error regardless if the user is logged in.
![image](https://user-images.githubusercontent.com/38309285/91693772-88435a00-eb39-11ea-9919-679caa56323d.png)

**Comment by CarsonF** [comment]
@reneepe the type of `rootDirectory` changed; it's secured now.
```gql
query {
  project(id: "") {
    id
    rootDirectory {
      canRead
      value {
        id
      }
    }
  }
}
```

reneepe:
Thanks @CarsonF.

Logged in, the following is correctly returned.

![image](https://user-images.githubusercontent.com/38309285/92012316-47ef0200-ed1a-11ea-94e6-d41d1902250b.png)

Logged out, the rootDirectory id isn't returned but there's no error message thrown. Should it only be unavailable or should there be a message as well?

![image](https://user-images.githubusercontent.com/38309285/92012613-a916d580-ed1a-11ea-954e-462affa2083f.png)

PR #1070, opened 2020-08-27
Handle no permission for project root directory

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1059

### DIFF
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -38,7 +38,7 @@ import {
   EngagementService,
   SecuredEngagementList,
 } from '../engagement';
-import { Directory, FileService } from '../file';
+import { FileService, SecuredDirectory } from '../file';
 import { LocationService } from '../location';
 import {
   PartnershipListInput,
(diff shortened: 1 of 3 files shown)


src/components/project/project.service.ts
```diff
@@ -839,7 +839,11 @@ export class ProjectService {
   async getRootDirectory(
     projectId: string,
     session: ISession
-  ): Promise<Directory> {
+  ): Promise<Directory | null> {
+    if (!session.userId || session.userId === this.config.anonUser.id) {
```
CarsonF, line 843: This may not be the only reason that the user doesn't have access. Shouldn't this check be done with a database call?

zkhin: Okay, let me check

zkhin: @CarsonF I added `SecuredDirectory` type and changed the resolve field type.
And handling no permission, I changed `match` to `optionalMatch` (https://github.com/SeedCompany/cord-api-v3/pull/1070/files#diff-8d8f4e73db837875a146ee377742270bR846) for getting `Directory` associated with that project. So it will return `undefined` if there's no permission and just returned `null` value for a directory with `canRead: false, canEdit: false`

glassblowerscat:
> Would you mind updating UI to handle this change? 

Yep, I'll take a look

glassblowerscat:
@CarsonF If the user doesn't have `canRead` permissions for the project's `rootDirectory`, should we prevent them from interacting with the project's files across the board? For example, we have a `CreateProjectDirectory` component that queries the `rootDirectory` and could display an error message to the effect of "You don't have permission to add folders in this project" if `canRead` is false.

CarsonF:
Yeah I think we should block access across the board. Don't let them click on the card. Show error in project list if they navigate there directly.

PR #1073, opened 2020-08-27
Fix sorting projects by names

### LINKED ISSUE #976: query projects > Sort by Name with 'ASC / DESC' order is not working
**Describe the bug**
query projects > Sort by Name with 'ASC / DESC' order is not working

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request.

**Expected Result**
Project items should be sorted by Name in ascending / descending order.

**Actual Result**
Sort by Name with 'ASC / DESC' order is not working.
See video - https://drive.google.com/file/d/1XVMZ3ySd60vzSBdAjQaRqwCgbhShAMV2/view?usp=drivesdk

GraphQL Query ( projects - ASC ):
image

GraphQL Query ( projects - DESC):
image

**Query and Input File**
[projects_sort_ISSUE.txt](https://github.com/SeedCompany/cord-api-v3/files/5088186/projects_sort_ISSUE.txt)

CordAPIV3# 29ee3b29

elisabeth-sorrell:
Also, have the createProject function for creating projects in the test suite generate the names using a random word and random uuid combination so we have uniqueness coupled with having more of a chance that they start with a letter.

Fixes #976


src/components/project/project.service.ts
```diff
@@ -624,7 +624,7 @@ export class ProjectService {
           node('prop', 'Property', { active: true }),
         ])
         .with('*')
-        .orderBy(`prop.value ${input.order}`);
+        .orderBy(`LOWER(prop.value) ${input.order}`);
```
CarsonF, line 627: FYI this is fine for sorting by name, but it won't work for sorting by sensitivity or dates

elisabeth-sorrell: @CarsonF correct. I just tested with sensitivity, which cypher interpreted as a string (I suppose) so nothing bad happened there, but sorting by dates broke it.

elisabeth-sorrell:
@reneepe I'll need a little more time on this before testing. Need to figure out what I'm going to do with use cases other than name.

reneepe:
> @reneepe I'll need a little more time on this before testing. Need to figure out what I'm going to do with use cases other than name.

Ok no problem. Let me know when ready.

elisabeth-sorrell:
@CarsonF I have a proposed solution for handling sorting on different fields. Not sure if I handled the new database abstraction pattern correctly either, so just let me know.
Thanks

elisabeth-sorrell:
@CarsonF ughghghghgh. scratch that. I was assuming that apoc was installed in CI/Prod/etc, but apparently it isn't. I'll post when I have something else.

CarsonF:
@elisabeth-sorrell Yeah that might be an oversight. I don't think we have a problem using apoc.
However, I think for this case it would be better to determine this upfront in code rather than "reflect" the type from the DB.

Assuming the project properties are all "normal, secured" properties, the only thing that needs to change in the query is that one order by expression. A simple mapping:
```tsx
const projectSortMap: Record<typeof input.sort, string> = {
  name: 'lower(prop.value)',
  status: 'prop.value',
};
query.orderBy(projectSortMap[input.sort], input.order);
```
Taking it a step further:
```tsx
// query/sorting.ts
export enum DbSortType {
  Exact: 'prop.value',
  CaseInsensitive: 'lower(prop.value)',
  Date: '?',
};

// project.service.ts
const projectSortMap: Record<typeof input.sort, DbSortType> = {
  name: DbSortType.CaseInsensitive,
  status: DbSortType.Exact,
  ...
};
query.orderBy(projectSortMap[input.sort], input.order);
```
It may not be the best to hardcode the query expression in the enum. But maybe this helps in general for what I was thinking.
Happy to talk it out 👍

test/organization.e2e-spec.ts
```diff
@@ -209,58 +210,155 @@ describe('Organization e2e', () => {
...
+    for (const i of items) {
+      if (index < organizations.items.length - 1) {
+        const currName = i.name.value;
+        const nextName = items[index + 1].name.value;
+        const areEqual = currName.toUpperCase() < nextName.toUpperCase();
+        if (!areEqual) isAscending = false;
+        index++;
+      }
+    }
+    expect(isAscending).toBe(true);
```
CarsonF, line 265:
```suggestion
    const items = organizations.items;
    const sorted = orderBy(items, (org) => org.name.value.toLowerCase(), ['asc']);
    expect(sorted).toEqual(items);
```
https://lodash.com/docs/4.17.15#orderBy

test/utility/create-project.ts
```diff
@@ -17,7 +17,7 @@ export async function createProject(
   input: Partial<CreateProject> = {}
 ) {
   const project: CreateProject = {
-    name: faker.random.uuid(),
+    name: faker.random.word() + faker.random.uuid(),
```
CarsonF, line 20:
```suggestion
    name: faker.random.word() + ' ' + faker.random.uuid(),
```

src/components/project/project.service.ts
```diff
@@ -581,7 +581,11 @@ export class ProjectService {
         : filter.type === 'Translation'
         ? 'TranslationProject'
         : 'Project';
-
+    const projectSortMap: Partial<Record<typeof input.sort, string>> = {
+      name: 'lower(prop.value)',
+      status: 'lower(prop.value)',
```
CarsonF, line 586: Status is an enum so its case should never differ.

src/core/database/query.helpers.ts
```diff
@@ -997,6 +997,12 @@ export function listReturnBlock<T = any>(
   isSecuredSort: boolean,
   sort?: string | ((sortStr: string) => string)
 ) {
+  const securedSortMap: Partial<Record<typeof sortInput, string>> = {
+    name: `lower(node.${sortInput}.value)`,
+  };
+  const insecuredSortMap: Partial<Record<typeof sortInput, string>> = {
+    name: `lower(node.${sortInput})`,
+  };
```
CarsonF, line 1005: This is a generic function and should not be making guesses about properties.

This function is also deprecated. I didn't realize it was still in use by orgs. I'll create an issue to update it.

CarsonF (changes requested):

test/organization.e2e-spec.ts
elisabeth-sorrell: awesome! I can put this in for the project tests, too

elisabeth-sorrell:
@CarsonF okay, so nearly all of the tests started blowing up again all over me. I keep wiping out the db, or wiping out the workspace, or updating/merging master. as of right now, I had almost gotten one clean run by the time I committed.

elisabeth-sorrell:
@CarsonF mwah ha! finally fixed the rest of the tests! ready for (hopefully final ;) review
@michaelmarshall the deleteOrganizations that I had in the listAscending/Descending was failing all the tests that would run after it.

test/project.e2e-spec.ts
```diff
@@ -298,6 +298,148 @@ describe('Project e2e', () => {
...
+          sort: sortBy,
+          order: ascOrder,
+          filter: {
+            mine: true,
+          },
+        },
+      }
+    );
+    // Set a flag that's going to indicate if the projects are in order
+    const isAscending = true;
```
CarsonF, line 348:
```suggestion
```

test/project.e2e-spec.ts
```diff
@@ -298,6 +298,148 @@ describe('Project e2e', () => {
...
+    const items = projects.items;
+    const sorted = orderBy(items, (proj) => proj.name.value.toLowerCase(), [
+      'asc',
+    ]);
+    expect(sorted).toEqual(items);
+    expect(isAscending).toBe(true);
+
+    //delete all projects
+    await Promise.all(
+      projects.items.map(async (item: { id: any }) => {
```
CarsonF, line 358: Are we sure we want to delete all projects after this? What about just deleting the ones that were created above in this test case?

test/project.e2e-spec.ts
```diff
@@ -298,6 +298,148 @@ describe('Project e2e', () => {
...
+          sort: sortBy,
+          order: ascOrder,
+          filter: {
+            mine: true,
+          },
+        },
+      }
+    );
+    // Set a flag that's going to indicate if the projects are in order
+    const isDescending = true;
```
CarsonF, line 420:
```suggestion
```

src/components/organization/organization.service.ts
```diff
@@ -253,16 +257,14 @@ export class OrganizationService {
...
-              .orderBy('prop.value', order)
-          : q.with('*').orderBy(`node.${sort}`, order)
+        q
+          .match([
+            node('node'),
+            relation('out', '', sort),
+            node('prop', 'Property', { active: true }),
+          ])
+          .with('*')
+          .orderBy(sortBy, order)
```
CarsonF, line 267: This doesn't handle unsecured properties anymore, like `id` and `createdAt`

CarsonF (commented):
We're getting closer!

test/project.e2e-spec.ts
elisabeth-sorrell: it is deleting only the ones created. It's going through the projects.items and deleting only the projects with those ids (which it's getting from `projects` ) if you look above, you'll see that we created/registered a user named Tammy, and that user created four projects. when that user lists the projects, only the projects that Tammy created will be listed, and she can delete them.
.......
..... unless I totally don't understand how permissions with that user work, which is a huge possibility ;)

CarsonF: > when that user lists the projects, only the projects that Tammy created will be listed,

That's only true if the `mine` filter was implemented, which it's not right now. Actually permissions also affect it, which I don't fully understand either. So maybe it's ok.

elisabeth-sorrell:
@CarsonF back at ya!

CarsonF (approved):
Thanks @elisabeth-sorrell!

PR #1074, opened 2020-08-27
User list refactor

frankyma:
Copied from project list refactor.  Skipped the first/last name filters as Carson said we don't need them.  Was in the process of making `applyListFilters` to abstract the users but will keep the code in, maybe used later.

### DIFF
--- a/src/common/util.ts
+++ b/src/common/util.ts
@@ -26,3 +26,17 @@ export const mapFromList = <T, S = T, K extends string = string>(
     return acc;
   }, out as Record<K, S>);
 };
+
+/**
+ * Just like Object.entries except keys are strict
+ */
+export const entries: <K extends string, V>(
+  o: Record<K, V>
+) => Array<[K, V]> = Object.entries as any;
+
+/**
+ * Just like Object.keys except keys are strict
+ */
+export const keys: <K extends string>(
+  o: Record<K, unknown>
+) => K[] = Object.keys as any;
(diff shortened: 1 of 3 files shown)


src/core/array-helpers.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
CarsonF, line 1: Can you put these in `common/util`?

src/core/database/query.helpers.ts
```diff
@@ -1102,3 +1103,19 @@ export function getPropList(query: Query, ...aliases: string[]) {
     ])
     .with(['collect(prop) as propList', 'node', ...aliases]);
 }
+
+//List filtering after refactor
+export function applyListFilters<
```
CarsonF, line 1108: Let's not do this. Each filter property can have different handling so handling as a complete object will not work.

src/core/array-helpers.ts
frankyma: Moved

src/core/database/query.helpers.ts
frankyma: Removed 😢

zkhin:
@CarsonF Do we still need `displayFirstName` and `displayLastName` in the user list filter?

CarsonF:
No

CarsonF (approved):
Thanks @frankyma & @zkhin!

CarsonF (dismissed):

CarsonF:
@zkhin Can you update user tests to account for the fact that you can't filter now? Maybe the tests can just be removed? I didn't look hard enough to confirm whether they are still applying value or not.

zkhin:
> @zkhin Can you update user tests to account for the fact that you can't filter now? Maybe the tests can just be removed? I didn't look hard enough to confirm whether they are still applying value or not.

Okay, I will

CarsonF (approved):
💪

Issue #971, opened 2020-08-17
Fix project test: list view of project members by project id


michaelmarshall:
This test experiences an intermittent bug and we skipped it. Try to unskip and see if you can determine the root cause and fix it.

reneepe:
VERIFIED - The type, name, roles, creation data, project member ids, and total of project members return correctly.

![image](https://user-images.githubusercontent.com/38309285/92195726-a52ea480-ee3b-11ea-9cb0-d7f09a77c7bc.png)

Issue #972, opened 2020-08-17
fix test: list view of partnerships by project


michaelmarshall:
This test is currently skipped and needs to run and pass.

reneepe:
VERIFIED - Fixed by https://github.com/SeedCompany/cord-api-v3/pull/1076
The partnership ids, and total of partnerships return correctly.

izquierdoalfredo:
@reneepe This ticket can be closed?

PR #1076, opened 2020-08-28
fix:list view of project members by projectId test

izquierdoalfredo:
https://github.com/SeedCompany/cord-api-v3/issues/971
https://github.com/SeedCompany/cord-api-v3/issues/972

### DIFF
--- a/test/project.e2e-spec.ts
+++ b/test/project.e2e-spec.ts
@@ -602,7 +602,7 @@ describe('Project e2e', () => {
     expect(queryProject.project.team.total).toBe(numProjectMembers);
   });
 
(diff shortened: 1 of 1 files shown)


izquierdoalfredo:
I unskiped and then tested "List view of partnerships by projectId" around 10 times and it work

reneepe (commented):
VERIFIED - Fixed https://github.com/SeedCompany/cord-api-v3/issues/971 and https://github.com/SeedCompany/cord-api-v3/issues/972. Results in each ticket.

The type, name, roles, creation data, project member ids, and total of project members return correctly. The partnership ids, and total of partnerships return correctly as well.

PR #1083, opened 2020-08-28
Fix budgetRecord.organization

### LINKED ISSUE #1082: Organization always returns null when returned from project query
## Current Behavior
When querying the gql API for a Project, organization is always returning null.

## Expected Behavior

Organization should return as specified in the gql schema.

## Steps to Reproduce

1. Create an organization.
2. Create a project.
3. Create a partnership linking the organization and project you just created.
4. Query the project including the organization as a part of your return value.

## Environment
Short commit hash: c4c11b73

thomas-adev:
Added missing organization secured property to budget service method readOneRecord. 
'organizationId' was being set incorrectly which was resulting in null responses to API queries.

Updated budget record DTO and resolver to call organization id string 'organization' instead of 'organizationId'.
This matches the naming convention in the partnership DTO and resolver.

Closes #1082

### DIFF
--- a/src/components/budget/budget.service.ts
+++ b/src/components/budget/budget.service.ts
@@ -445,7 +445,15 @@ export class BudgetService {
       ])
       .with('{value: props.value, property: type(r)} as prop, permList, node')
       .with('collect(prop) as propList, permList, node')
-      .return('propList, permList, node')
+      .match([
+        node('node'),
+        relation('out', '', 'organization', { active: true }),
+        node('organization', 'Organization', { active: true }),
+      ])
+      .with('propList, permList, node, organization.id as organizationId')
+      .return(
+        'propList + [{value: organizationId, property: "organization"}] as propList, permList, node'
+      )
       .asResult<StandardReadResult<DbPropsOfDto<BudgetRecord>>>();
 
     const result = await query.first();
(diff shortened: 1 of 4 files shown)


thomas-adev:
Just a note on the renaming from 'organizationId' to 'organization'. I initially wrote a version that left 'organizationId', however this didn't match the permissions list retrieved from the database and so in order to make it work, you cannot simply rely on parsePropList and parseSecuredProperties.

Once I saw that the partnership DTO and Resolver just called it "organization", I just refactored to simplify the code.

src/components/budget/budget.service.ts
```diff
@@ -445,7 +445,15 @@ export class BudgetService {
...
-      .return('propList, permList, node')
+      .match([
+        node('node'),
+        relation('out', '', 'organization', { active: true }),
+        node('organization', 'Organization', { active: true }),
+      ])
+      .with('propList, permList, node, organization.id as organizationId')
+      .return(
+        'propList + [{value: organizationId, property: "organization"}] as propList, permList, node'
+      )
```
CarsonF, line 456: The extra with clause isn't needed.
```suggestion
      .return([
        'propList + [{value: organization.id, property: "organization"}] as propList',
        'permList',
        'node',
      ])
```

thomas-adev: Good catch! Thanks.

Issue #1090, opened 2020-08-31
query languages > Getting 'Variable `permList` not defined' error when using filter by name


iPiyushPatel:
**Describe the bug**
query languages > Getting 'Variable `permList` not defined' error when using filter by name.

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request.

**Expected Result**
Languages list should be returned.

**Actual Result**
Getting 'Variable `permList` not defined' error when using filter by name.
See video - https://drive.google.com/file/d/1rM7avG7RTLtLj2eG3eq7iHKuFIK8W9-F/view?usp=drivesdk

GraphQL Query ( languages - error ):
![image](https://user-images.githubusercontent.com/5212585/90590496-55769900-e1fe-11ea-8acf-8558c5b661b5.png)

GraphQL Query ( languages ):
![image](https://user-images.githubusercontent.com/5212585/90590616-abe3d780-e1fe-11ea-9d0c-77af97105889.png)

**Query and Input File**
[languages_permList_ERR.txt](https://github.com/SeedCompany/cord-api-v3/files/5093942/languages_permList_ERR.txt)

CordAPIV3# c4c11b73

PR #1094, opened 2020-08-31
Refactor language list

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1086
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1090

### DIFF
--- a/src/components/language/language.service.ts
+++ b/src/components/language/language.service.ts
@@ -535,6 +535,17 @@ export class LanguageService {
   ): Promise<LanguageListOutput> {
     const label = 'Language';
 
+    const secureProps = [
+      'name',
+      'displayName',
+      'isDialect',
+      'populationOverride',
+      'registryOfDialectsCode',
+      'leastOfThese',
+      'leastOfTheseReason',
+      'displayNamePronunciation',
+    ];
+
     const skip = (input.page - 1) * input.count;
 
     const query = this.db.query();
@@ -576,14 +587,18 @@ export class LanguageService {
       .raw('unwind nodes as node');
 
     if (input.sort) {
-      query
-        .match([
-          node('node'),
-          relation('out', '', input.sort),
-          node('prop', 'Property', { active: true }),
-        ])
-        .with('*')
-        .orderBy(`prop.value ${input.order}`);
+      if (secureProps.includes(input.sort)) {
+        query
+          .match([
+            node('node'),
+            relation('out', '', input.sort),
+            node('prop', 'Property', { active: true }),
+          ])
+          .with('*')
+          .orderBy(`prop.value ${input.order}`);
+      } else {
+        query.with('*').orderBy(`node.${input.sort} ${input.order}`);
+      }
     }
 
     query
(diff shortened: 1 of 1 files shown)


src/components/language/language.service.ts
```diff
@@ -537,8 +534,7 @@ export class LanguageService {
     session: ISession
   ): Promise<LanguageListOutput> {
     const label = 'Language';
-    // const baseNodeMetaProps = ['id', 'createdAt'];
-    // const unsecureProps = [''];
+
     const secureProps = [
```
CarsonF, line 538: This has already been defined in `readOne`. Just move that to a private class prop or a top-level instead of duplicating it here.

src/components/language/language.service.ts
```diff
@@ -55,6 +58,18 @@ import { EthnologueLanguageService } from './ethnologue-language';
...
+    'name',
+    'displayName',
+    'isDialect',
+    'populationOverride',
+    'registryOfDialectsCode',
+    'leastOfThese',
+    'leastOfTheseReason',
+    'displayNamePronunciation',
+    'sponsorDate',
+  ];
```
CarsonF, line 71: Sorry for the confusion. I meant move the object here https://github.com/SeedCompany/cord-api-v3/blob/e7f1230fd74a03ddf60f8885371c2ebf9cf36866/src/components/language/language.service.ts#L425-L433
to class prop. Then
```tsx
sort in this.securedProperties
```

src/components/language/language.service.ts
```diff
@@ -537,46 +546,36 @@ export class LanguageService {
...
+        sort in this.securedProperties
+          ? q
+              .match([
+                node('node'),
+                relation('out', '', sort),
+                node('prop', 'Property', { active: true }),
+              ])
+              .with('*')
+              .orderBy('prop.value', order)
+          : q.with('*').orderBy(`node.${input.sort} ${input.order}`)
```
CarsonF, line 575:
```suggestion
          : q.with('*').orderBy(`node.${sort}`, order)
```

src/components/language/language.service.ts
```diff
@@ -537,46 +546,36 @@ export class LanguageService {
...
+        ...permissionsOfNode(label),
+        ...(filter.name
+          ? [
+              relation('out', '', 'name', { active: true }),
+              node('name', 'Property', { active: true }),
+            ]
+          : []),
+      ])
+      .call((q) =>
+        filter.name ? q.where({ name: { value: contains(filter.name) } }) : q
```
CarsonF, line 563: I'm seeing now where we previously called `filterByString` which handled checking permissions.
https://github.com/SeedCompany/cord-api-v3/blob/f0174b0dfb5bfb2d79f9cb6845f480c515165d3f/src/core/database/query.helpers.ts#L802
I'm wondering if this is needed here. Unfortunately the two people who started this are on vacation lol.

CarsonF: Ok I created an issue for Michael. We can come back to it if needed. https://github.com/SeedCompany/cord-api-v3/issues/1109

PR #1096, opened 2020-08-31
refactor product readOne

tjgersho:


src/components/product/product.service.ts
```diff
@@ -368,10 +324,18 @@ export class ProductService {
...
+      produces: true,
     });
-    const baseNodeProps = parseBaseNodeProperties(result.node);
 
-    const produces = await this.db
+    const {
+      produces,
+      scriptureReferencesOverride,
+      ...rest
+    } = securedProperties;
```
CarsonF, line 336: This destructure could be merged with the assignment above.
`props` is also not being used so `result.propList` could just be passed as the first arg to `parseSecuredProperties`.

src/components/product/product.service.ts
```diff
@@ -381,18 +345,18 @@ export class ProductService {
       .return('p')
       .first();
```
CarsonF, line 346:
```tsx
.first()
.asResult<{ p: Node<BaseNode> }>();
```

src/components/product/product.service.ts
```diff
@@ -381,18 +345,18 @@ export class ProductService {
...
-    if (!produces) {
+    if (!pr) {
       const scriptureReferences = await this.listScriptureReferences(
-        baseNodeProps.id,
+        id,
         'Product',
         session
       );
+
       return {
```
CarsonF, line 355: This could be done once instead of both here and below. (Other than the scriptureReferences prop)

src/components/product/product.service.ts
```diff
@@ -441,27 +401,39 @@ export class ProductService {
...
       scriptureReferencesOverride: {
-        canRead: result.canScriptureReferencesOverrideRead,
-        canEdit: result.canScriptureReferencesOverrideEdit,
-        value: !scriptureReferencesOverride.length
+        ...(securedProperties.scriptureReferencesOverride
+          ? securedProperties.scriptureReferencesOverride
+          : {
+              canRead: false,
+              canEdit: false,
+            }),
```
CarsonF, line 433: scriptureReferencesOverride should always be defined and have default values if needed.

src/components/product/product.service.ts
```diff
@@ -441,27 +401,39 @@ export class ProductService {
...
+        value: !scriptureReferencesOverrideValue.length
           ? null
-          : scriptureReferencesOverride,
+          : scriptureReferencesOverrideValue,
```
CarsonF, line 436: An empty list is not the same thing as null here. null means there's no override. empty list means there is an override of an empty list.

src/components/product/product.service.ts
```diff
@@ -441,27 +401,39 @@ export class ProductService {
         value: securedProperties.purposes.value ?? [],
       },
       produces: {
+        ...(securedProperties.scriptureReferencesOverride
+          ? securedProperties.scriptureReferencesOverride
+          : {
+              canRead: false,
+              canEdit: false,
+            }),
```
CarsonF, line 409: Why would produces use the security props from scriptureReferencesOverride?

src/components/product/product.service.ts
```diff
@@ -441,27 +401,39 @@ export class ProductService {
...
         value: {
           ...producible,
-          id: produces.p.properties.id,
+          id: pr.p.properties.id,
           __typename: (ProducibleType as any)[typeName],
-          scriptureReferences: !scriptureReferencesOverride.length
+          scriptureReferences: !scriptureReferencesOverrideValue.length
```
CarsonF, line 414: scriptureReferences should be overridden here on the producible.

src/components/product/product.service.ts
```diff
@@ -406,30 +370,26 @@ export class ProductService {
...
 
     return {
-      ...baseNodeProps,
+      ...parseBaseNodeProperties(result.node),
       ...securedProperties,
       scriptureReferences: {
-        canRead: result.canScriptureReferencesRead,
-        canEdit: result.canScriptureReferencesEdit,
+        ...securedProperties.scriptureReferences,
         value: [],
```
CarsonF, line 393: This should be the `scriptureReferencesOverride` if it's not null or the `produces.scriptureReferences`

src/components/product/product.service.ts
tjgersho: Empty list from `listScriptureReferences` function means scriptureReferencesOverride does not exist(we are saying this is null)

src/components/product/product.service.ts
tjgersho: https://i.imgur.com/8kTFDK8.png

src/components/product/product.service.ts
tjgersho: Ahh, got it.

src/components/product/product.service.ts
tjgersho: scriptureReferencesOverride and produces have undefined type as union.

CarsonF: Ah. That's from the usage of `AnyProduct`. I would create a type alias
```tsx
type ProductDbProps = DbPropsOfDto<DirectScriptureProduct & DerivativeScriptureProduct>;
```

src/components/product/product.service.ts
CarsonF: We need to make the distinction between an empty list and null. They mean two different things. Maybe we need another db property like `isOverriding` to help.

tjgersho: It's great if we can use isOverriding to differentiate between override states instead of using null.

src/components/product/dto/create-product.dto.ts
```diff
@@ -51,6 +51,15 @@ export abstract class CreateProduct {
...
   readonly scriptureReferencesOverride?: ScriptureRangeInput[];
 
+  @Field({
+    nullable: true,
+    description: stripIndent`
+      When isOverriding is false, scriptureReferencesOverride will be ignored.
+      Note only \`DerivativeScriptureProduct\`s can use this field.
+    `,
+  })
+  readonly isOverriding?: boolean = false;
```
CarsonF, line 61: This doesn't need to be apart of the API schema. It's only for the DB to help give correct response to the API schema already in place.

CarsonF (changes requested):

src/components/product/product.service.ts
```diff
@@ -370,105 +345,102 @@ export class ProductService {
...
-      'Product',
-      session,
-      { isOverride: true }
-    );
-
-    const typeName = difference(produces.p.labels, [
-      'Producible',
-      'BaseNode',
-    ])[0];
+    const typeName = difference(pr.p.labels, ['Producible', 'BaseNode'])[0];
```
CarsonF, line 399:
```suggestion
    const typeName = difference(pr.p.labels as ProducibleType[], ['Producible', 'BaseNode'])[0];
```
This should prevent the need to typecast to any below.

src/components/product/product.service.ts
```diff
@@ -370,105 +345,102 @@ export class ProductService {
...
+        ...rest.purposes,
+        value: rest.purposes.value ?? [],
       },
       produces: {
+        ...produces,
         value: {
           ...producible,
-          id: produces.p.properties.id,
+          id: pr.p.properties.id,
           __typename: (ProducibleType as any)[typeName],
```
CarsonF, line 430:
```suggestion
          __typename: typeName,
```
With comment on `typeName` declaration.

CarsonF (changes requested):

src/components/product/product.service.ts
tjgersho: `difference` still returns string not ProducibleType value. 
https://i.imgur.com/HtRnxtM.png

tjgersho: https://i.imgur.com/zxgGObJ.png
This is working fine.

CarsonF (changes requested):
Looking good, just one more thing I think

CarsonF (approved):
Thanks @tjgersho!

PR #1099, opened 2020-08-31
re-refactor readOnes and update create and update for song

tjgersho:
re-refactor readones for Story, Song, Literacy-material and Film.
update create and update function for Song.

### DIFF
--- a/src/components/literacy-material/literacy-material.service.ts
+++ b/src/components/literacy-material/literacy-material.service.ts
@@ -8,11 +8,11 @@ import {
   ServerException,
 } from '../../common';
 import {
-  addAllMetaPropertiesOfChildBaseNodes,
-  ChildBaseNodeMetaProperty,
   ConfigService,
   createBaseNode,
   DatabaseService,
+  getPermList,
+  getPropList,
   ILogger,
   Logger,
   matchRequestingUser,
@@ -221,12 +221,9 @@ export class LiteracyMaterialService {
     }
   }
 
-  async readOne(
-    literacyMaterialId: string,
-    session: ISession
-  ): Promise<LiteracyMaterial> {
+  async readOne(id: string, session: ISession): Promise<LiteracyMaterial> {
     this.logger.debug(`Read literacyMaterial`, {
-      id: literacyMaterialId,
+      id,
       userId: session.userId,
     });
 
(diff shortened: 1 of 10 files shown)


src/components/song/song.service.ts
```diff
@@ -315,4 +353,50 @@ export class SongService {
...
+        node('song', 'Song', {
+          id: songId,
+          active: true,
+          owningOrgId: session.owningOrgId,
+        }),
+        relation('out', '', 'scriptureReferences'),
+        node('scriptureRanges', 'ScriptureRange', { active: true }),
+      ])
+      .with('collect(scriptureRanges) as items')
+      .return('items');
```
CarsonF, line 373: There's no need to collect() & first() row just to run() it. 
```suggestion
      .return('scriptureRanges')
      .asResult<{ scriptureRanges: Node<{ start: number; end: number }> }>();
```
Then
```tsx
const results = await query.run();
return results.map(item => 
  verseToScriptureRange({
    start: item.properties.start,
    end: item.properties.end,
  })
);

File: src/components/song/song.service.ts [hunk 2]
```diff
@@ -162,31 +167,39 @@ export class SongService {
...
+
+      if (input.scriptureReferences) {
+        for (const sr of input.scriptureReferences) {
+          const verseRange = scriptureToVerseRange(sr);
+          query.create([
+            node('node'),
+            relation('out', '', 'scriptureReferences', { active: true }),
+            node('sr', 'ScriptureRange', {
+              start: verseRange.start,
+              end: verseRange.end,
```
CarsonF, line 188:
```suggestion
              ...scriptureToVerseRange(sr),
```

src/components/literacy-material/literacy-material.service.ts
```diff
@@ -315,52 +268,21 @@ export class LiteracyMaterialService {
...
-              start: verseRange.start,
-              end: verseRange.end,
-              active: true,
-              createdAt: DateTime.local(),
-            }),
-          ])
-          .return('lm')
-          .first();
-      }
+    if (input.scriptureReferences) {
```
CarsonF, line 271: Can this check be moved to the service and allow nullable to be passed in?

src/components/literacy-material/literacy-material.service.ts
```diff
@@ -291,21 +239,26 @@ export class LiteracyMaterialService {
...
 
-    const secured = parseSecuredProperties(result.propList, result.permList, {
-      name: true,
-    });
-
-    const scriptureReferences = await this.listScriptureReferences(
-      literacyMaterialId,
+    const scriptureReferences = await this.scriptureRefService.list(
+      id,
+      'LiteracyMaterial',
```
CarsonF, line 244: Is this label really needed? Can't we just use the id with the 'BaseNode' label? Do we not index on `id` properties?

src/components/song/song.service.ts
```diff
@@ -158,35 +160,36 @@ export class SongService {
...
 
-      const result = await query.first();
       if (!result) {
         throw new ServerException('failed to create a song');
       }
 
-      const id = result.id;
-
-      // add root admin to new song as an admin
-      await this.db.addRootAdminToBaseNodeAsAdmin(id, 'Song');
```
CarsonF, line 186: This is the only one of the 4 that had this. Is this missing on all 4 now or is it not needed?

CarsonF (changes requested):

src/components/song/song.service.ts
tjgersho: It is duplicated in `createBaseNode` function. So not needed.

src/components/literacy-material/literacy-material.service.ts
tjgersho: Yeah. right.

tjgersho: Can use `BaseNode`.

src/components/literacy-material/literacy-material.service.ts
tjgersho: Sure

CarsonF (approved):
Thanks @tjgersho!

I just tweaked to use early returns and to pass session object

PR #1100, opened 2020-09-01
1072 add test ceremony

izquierdoalfredo:
https://github.com/SeedCompany/cord-api-v3/issues/1072

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -1203,6 +1203,23 @@ export class EngagementService {
         object,
         aclEditProp: 'canDeleteOwnUser',
       });
+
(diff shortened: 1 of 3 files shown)


izquierdoalfredo:
@CarsonF I see, So we dont need to add the separate Ceremony test.
Do you just want to add ensure ceremonies are created and deleted with engagements?

CarsonF:
> Do you just want to add ensure ceremonies are created and deleted with engagements?

Correct. No API query/mutation changes necessary.

izquierdoalfredo:
I added delete with engagement test. 
And I fixed bug in engagement to fit "ceremonies are created and deleted with engagements"

Issue #1102, opened 2020-09-01
createPartnership > Partnership created with duplicate values of 'types' field


iPiyushPatel:
**Describe the bug**
createPartnership > Partnership created with duplicate values of 'types' field

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
The duplicate value should not be allowed in the type field

**Actual Result**
createPartnership > Partnership created with duplicate values of 'types' field
See video -  https://drive.google.com/file/d/1oFvnaxAOgCK9uG_KbqanXB3A8KgUvkK6/view?usp=sharing

GraphQL Query:
https://drive.google.com/file/d/1hDMyoCZGzx6NCcjl3W5H8TlPLeh6aZcr/view?usp=sharing

GraphQL Error
```
 "data": {
        "createPartnership": {
            "partnership": {
                "id": "nSG8XxDt1l",
                "createdAt": "2020-09-01T15:42:35.349+05:30",
                "types": {
                    "value": [
                        "Technical",
                        "Technical"
                    ],
                    "canRead": true,
                    "canEdit": true
                },
                "agreementStatus": {
                    "value": "NotAttached",
                    "canRead": true,
                    "canEdit": true
                },
```

*Query and Input File*
https://docs.google.com/document/d/1rkWl4eqHVzNWY4Vimw2AM4evmQVhYlg5NnE-cr0udvA/edit?usp=sharing

CordAPIV3# 4c946db8

PR #1104, opened 2020-09-01
Create partnership with unique types

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1102


CarsonF:
How about on update too?

Issue #1103, opened 2020-09-01
mutation updateSong> Getting 'Could not find song' error when song id is blank


iPiyushPatel:
**Describe the bug**
mutation updateSong > Getting 'Could not find song' error when song `id` is blank

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request.

**Expected Result**
Error message 'Input validation failed' should be displayed when song id is blank.
Note : It is working as expected in Film module.

**Actual Result**
Getting 'Could not find song' error when song id is blank
See video - https://drive.google.com/file/d/1PH-9tMZ-23DYTEkJZkb_KAc1TvA-_cmy/view?usp=sharing

GraphQL Query ( updateSong ):
https://drive.google.com/file/d/15ofsv4jmRdNATLuwgGP-LSJkroFyoTMF/view?usp=sharing

GraphQL Error
```
{
    "errors": [
        {
            "message": "Could not find song",
            "extensions": {
                "codes": [
                    "NotFound",
                    "Input",
                    "Client"
                ],
                "status": 400,
                "field": "song.id",
                "code": "NotFound",
                "exception": {
                    "message": "Could not find song",
                    "stacktrace": [
                        "at SongService.readOne (components\\song\\song.service.ts:237:13)",
                        "at SongService.update (components\\song\\song.service.ts:259:18)",
                        "at SongResolver.updateSong (components\\song\\song.resolver.ts:58:18)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "updateSong"
            ]
        }
    ],

```

**Query and Input File**
https://docs.google.com/document/d/1q5UFXf1Ib4UmacSuo14Lo4m-VHQ5PLQ4CQL21-bixKw/edit?usp=sharing

CordAPIV3# 4c946db8

PR #1105, opened 2020-09-01
Validation error updating song with invalid id

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1103

### DIFF
--- a/src/components/song/dto/update-song.dto.ts
+++ b/src/components/song/dto/update-song.dto.ts
@@ -1,13 +1,13 @@
-import { Field, ID, InputType, ObjectType } from '@nestjs/graphql';
+import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import { Type } from 'class-transformer';
 import { ValidateNested } from 'class-validator';
-import { NameField } from '../../../common';
+import { IdField, NameField } from '../../../common';
 import { ScriptureRangeInput } from '../../scripture';
 import { Song } from './song.dto';
 
 @InputType()
 export abstract class UpdateSong {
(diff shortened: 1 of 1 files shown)


test/song.e2e-spec.ts
```diff
@@ -122,4 +122,30 @@ describe('Song e2e', () => {
...
+        {
+          input: {
+            song: {
+              id: '',
+              name: newName,
+            },
+          },
+        }
+      )
+    ).rejects.toThrowError();
```
CarsonF, line 149: A not found error was previously thrown. So this would've been green before your change. We should at least assert that it's a validation error (has `extensions.codes[0] === 'Validation'`). I'm not sure exactly the syntax of the top of my head. 

It would probably be good to have an `expectValidationError` helper.

Alternatively if you just removed this test I would look the other way haha 😅

zkhin: Oh, you're right.
Seems we disabled the validation pipe in the e2e test app, so it will not go into the validation decorator anymore... lol
https://github.com/SeedCompany/cord-api-v3/blob/master/test/utility/create-app.ts#L33

CarsonF: Ah geez. That explains a lot. I wonder if that can be re-enabled now without breaking all the tests.
I'll make a ticket for that. Let's just remove the test here for now since it isn't doing anything.

CarsonF: https://github.com/SeedCompany/cord-api-v3/issues/1107

Issue #1107, opened 2020-09-01
Re-enable validation on e2e tests


CarsonF:
- [ ] Delete this
  https://github.com/SeedCompany/cord-api-v3/blob/74d60973ddc49b785b162bf1e282db6a96ebf9c7/test/utility/create-app.ts#L30-L33
- [ ] Fix now failing tests

zkhin:
@CarsonF Seems there's an issue in the `CalendarDate`. After enabling the `ValidationPipe`, I cannot create the project with the following date.
```
...
mouStart: CalendarDate.fromISO('1991-01-01'),
mouEnd: CalendarDate.fromISO('1992-01-01'),
...
```
If I comment out the date fields, I can create the Project successfully.

![image](https://user-images.githubusercontent.com/4212008/92038748-33ad1280-edb7-11ea-9caf-526f07450b3f.png)

PR #1118, opened 2020-09-02
Fix email & date inputs in tests

CarsonF:
- Patch faker email to be a more unique, valid email
- [tests] Fix complex objects being passed in with variables

Related to #1107

### DIFF
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -1,4 +1,4 @@
-import { ISession, ServerException } from '../../../common';
+import { CalendarDate, ISession, ServerException } from '../../../common';
 import {
   DatabaseService,
   EventsHandler,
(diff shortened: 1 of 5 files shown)


src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
```diff
@@ -100,8 +100,16 @@ export class SyncBudgetRecordsToFundingPartners
...
+      let mouStartYear = partnership.mouStart.value.year;
+      let mouEndYear = partnership.mouEnd.value.year;
+
+      if (!mouStartYear || !mouEndYear) {
+        mouStartYear = CalendarDate.fromISO(
+          partnership.mouStart.value.toString()
+        ).year;
+        mouEndYear = CalendarDate.fromISO(partnership.mouEnd.value.toString())
+          .year;
+      }
```
CarsonF, line 112: Thanks for finding this @zkhin. This shouldn't be necessary though. It looks like the `mouStart/End.value` are strings but they TS says they are CalendarDates. TS types should always match what the actual types are, if it doesn't that's a problem.
Looking into why the ISO strings are not being transformed now.

zkhin: Yes, that's the weird part.

CarsonF: The transformation gets applied by the `ValidationPipe`. That's why it was working in the other branch with that pipe re-enabled in tests.

CarsonF:
Ok I'm closing and going to rebase this into your branch @zkhin

Issue #1108, opened 2020-09-01
Default budget record amount should be null by default


CarsonF:
- [ ] We want to allow the float value to be null as this indicates better that the user has not filled it out.
- [ ] The output field probably needs it's type changed to a `SecuredFloatNullable`
- [ ] The update input field needs it's type changed to a nullable float as well

glassblowerscat:
Do we also want to update the `total` returned for a budget where all records have an amount of `null`? Currently the total is also `0`

CarsonF:
> Do we also want to update the `total` returned for a budget where all records have an amount of `null`? Currently the total is also `0`

I think that's ok since it's a calculated aggregate.

PR #1123, opened 2020-09-03
Change budget record amount to be a nullable float

CarsonF:
Second half of #1108 

This updates GQL schema to reflect that amount can be a float or null.


glassblowerscat:
Derp

CarsonF:
np. I too forgot when I merged.

PR #1124, opened 2020-09-03
Enable validation on e2e

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1107


CarsonF (approved):
Awesome! Thanks @zkhin!

PR #1135, opened 2020-09-07
1128 refactor list organization ceremony

tjgersho:


CarsonF (changes requested):
Thanks @tjgersho.

Could you move the `securedProperties` to a class level property? (Like done in #1078)
It also needs to be an object. `'key' in [...]` is not the same as `[...].includes('key')`

CarsonF (approved):
🚀

PR #1136, opened 2020-09-07
add decorators to set default values for scriptures

tjgersho:


CarsonF (commented):
It's a shame the entire input object has to be defined 3 times. Could we just put a transformer on the range.start/end properties instead?

tjgersho:
Matthew 1:1-John 2:3 should be Matthew 1-John 2:3
Matthew 1:1-2:3 should be Matthew 1-2:3
Matthew 1:1-John last:last should be Matthew-John

The cases above are right?
not sure if the 7 cases in the comment are the only possible ones.

src/components/scripture/dto/scripture-reference.dto.ts
```diff
@@ -20,12 +20,14 @@ export abstract class ScriptureReferenceInput {
 
   @Field(() => Int, {
     description: 'The chapter number',
+    defaultValue: null,
```
CarsonF, line 23: Does this make the field nullable in the GQL schema?

src/components/scripture/dto/scripture-reference.transformer.ts
```diff
@@ -0,0 +1,31 @@
+import { applyDecorators } from '@nestjs/common';
+import { Transform } from '../../../common/transform.decorator';
+import { books } from '../books';
+import { bookIndexFromName } from '../reference';
+import { ScriptureReferenceInput } from './scripture-reference.dto';
+
+export const ScriptureStart = () =>
+  applyDecorators(
```
CarsonF, line 8: `applyDecorators` joins multiple decorators into a single one. Since we only have one here, this isn't needed.

src/components/scripture/dto/scripture-reference.transformer.ts
```diff
@@ -0,0 +1,31 @@
...
+    Transform((value: ScriptureReferenceInput) => {
+      value.chapter = value.chapter == null ? 1 : value.chapter;
+      value.verse = value.verse == null ? 1 : value.verse;
+      return value;
+    })
```
CarsonF, line 13: Does this come before or after validation? We don't want validation to fail because the transform hasn't happened yet.

src/components/scripture/dto/scripture-reference.dto.ts
tjgersho: Yes.

src/components/scripture/dto/scripture-reference.transformer.ts
tjgersho: Yeah, that was my concern.
I tested it and could see it comes before validation. So it works fine.

CarsonF (approved):
Thanks @tjgersho!

PR #1137, opened 2020-09-08
added handler to set initial end date

thomas-adev:
Added events for engagements being created or updated.
Updated engagement creation and update methods on engagement service to allow for updating initialEndDate.
Added event handler which sets initial end date.

### DIFF
--- a/src/components/engagement/handlers/set-initial-end-date.handler.ts
+++ b/src/components/engagement/handlers/set-initial-end-date.handler.ts
@@ -21,6 +21,48 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
   ) {}
   private readonly languageEngagementTypeName = 'LanguageEngagement';
 
+  async handle(event: SubscribedEvent) {
+    this.logger.debug('Engagement mutation, set initial end date', {
+      ...event,
+      event: event.constructor.name,
+    });
+
+    const engagement = 'engagement' in event ? event.engagement : event.updated;
+
+    const shouldUpdateInitialEndDate =
+      engagement.status === EngagementStatus.Active &&
+      engagement.initialEndDate.value == null &&
+      engagement.endDate.value != null;
+    if (!shouldUpdateInitialEndDate) {
+      return;
+    }
+
+    try {
+      const initialEndDate = engagement.endDate.value!;
+
+      const updatedEngagement = await this.updateEngagementInitialEndDate(
+        engagement,
+        initialEndDate,
+        event.session
+      );
+
+      if (event instanceof EngagementUpdatedEvent) {
+        event.updated = updatedEngagement;
+      } else {
+        event.engagement = updatedEngagement;
+      }
+    } catch (exception) {
+      this.logger.error(`Could not set initial end date on engagement`, {
+        userId: event.session.userId,
+        exception,
+      });
+      throw new ServerException(
+        'Could set initial end date on engagement',
+        exception
+      );
+    }
+  }
+
   private async updateEngagementInitialEndDate(
     engagement: Engagement,
     initialEndDate: CalendarDate,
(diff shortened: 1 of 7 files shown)


src/components/engagement/events/internship-engagement-updated.event.ts
```diff
@@ -0,0 +1,9 @@
+import { ISession } from '../../../common';
+import { InternshipEngagement } from '../dto';
+
+export class InternshipEngagementUpdatedEvent {
+  constructor(
+    readonly engagement: InternshipEngagement,
```
CarsonF, line 6: For the updated events I've been giving the `updated` & `previous` object as well as `updates` (changes).
https://github.com/SeedCompany/cord-api-v3/blob/cbb0a2dc2aa6a603a3dbc3e24ad220f1c39ac460/src/components/project/events/project-updated.event.ts#L6-L8
Let's do that in both of these updates too.

src/components/engagement/handlers/set-initial-end-date.handler.ts
```diff
@@ -0,0 +1,85 @@
...
+        };
+
+        if (
+          event instanceof LanguageEngagementCreatedEvent ||
+          event instanceof LanguageEngagementUpdatedEvent
+        ) {
+          await this.engagementService.updateLanguageEngagement(
+            updateInput,
+            event.session
+          );
```
CarsonF, line 62: By setting this here the engagement object returned from the API will be stale. Since the `initialEndDate` will have the previous value.
Maybe the event object could have a needsRefresh flag so that the engagement service can check for that and refetch it again.
Or maybe the event could have a setter for the engagement object and then the service returns that? That would eliminate an extra db call.

src/components/engagement/handlers/set-initial-end-date.handler.ts
```diff
@@ -0,0 +1,85 @@
...
+      event: event.constructor.name,
+    });
+
+    try {
+      const shouldUpdateInitialEndDate =
+        event.engagement.status === EngagementStatus.Active &&
+        event.engagement.initialEndDate.value == null &&
+        event.engagement.endDate.value != null;
+
+      if (shouldUpdateInitialEndDate) {
```
CarsonF, line 49: This could be done outside of the try/catch and negated to return early.

CarsonF (changes requested):
Thanks @thomas-adev! Looks good, just a couple things.

I'm realizing too that the `initialEndDate` needs to be updated as well when the project's `mouEnd` date changes or the engagement's `endDateOverride` changes in order to stay in sync until the engagement becomes active. You can do that in this PR or a separate one.

src/components/engagement/handlers/set-initial-end-date.handler.ts
thomas-adev: I pushed a new commit where I moved the if statement outside of the try/catch block. I believe this resolves this comment, but wanted to check if you would prefer an explicit return.

src/components/engagement/handlers/set-initial-end-date.handler.ts
thomas-adev: @CarsonF I updated so that after calling the event handler it will always fetch a fresh object. This will result in extra db reads, however the alternatives I could see all required rewriting the event bus so that the handler could set a flag to be returned from the publish function. 

The current interface returns void so I couldn't see a way to communicate out of the handler to the engagement service.
```
  async publish<T extends IEvent>(event: T): Promise<void> {
    ...
  }
```

I didn't want to update the event bus since that would effect lots of code outside of my narrow focus on this PR. Carson, let me know if you want me to focus on optimizing this more for this PR, or also let me know if I missed a trick here and there is a simpler path then the one I went down.

CarsonF: the publish returns void but the event can still have state.

```tsx
const event = new EngagementUpdated(...);
this.eventBus.publish(event);
if (event.isRefreshNeeded) {
  // do another read
}

// in handler
event.needsRefresh();
```

Or instead of requiring a blind refresh we could have the `updated` object be settable and that used for the output.
```tsx
// service
const event = ...;
publish(event);
return event.updated;

// handler
event.updated = {...};
```

CarsonF (commented):
@thomas-adev I rebased this on master, now that the other PR is merged.
You'll need to do
```
git fetch; git reset --hard origin/1062-set-intial-end-date
```
to update your local branch since I force pushed.

----

I also noticed that you are call the service update method again in the handler. This will result in an infinite loop....ok maybe not infinite, but it will publish the event again. I don't think we want that, but maybe we do.... If we had another handler listening for changes made here it couldn't unless we did this. All the listeners would have to be careful that they had an exit condition so they didn't infinitely loop. It could be more straight forward just to disallow nesting these events to keep it simple. Until something different is needed at least.

src/components/engagement/handlers/set-initial-end-date.handler.ts
thomas-adev: Ah, I understand now. I'll make that update.

thomas-adev:
@CarsonF that is a good point, if the check at the beginning of the handler is incorrect, or if the update is run incorrectly, this could end up in an infinite loop. To make the handler safer, I see a two options.

1. Duplicate the db call in the handler or elsewhere.
2. Refactor to db calls out of the engagement.service and into engagement.repository.

From our previous conversation, I believe option 2 is the direction your wanting to move. Would it make sense to push that change off to a separate PR for now and continue with event triggering itself once for now?

CarsonF:
Yeah option 2 is the direction I'm wanting to move to. I'd be ok doing that in a separate PR as long as this one doesn't cause a loop and you add a TODO saying this should be a lower-level repository call instead of a service call. At least then someone will think twice before copying it, hopefully.

CarsonF (approved):
I just fixed the double read on create lang engagement, and reordered the handler methods.

Thanks @thomas-adev!

Issue #1113, opened 2020-09-02
File Migration Logic


CarsonF:
We want to add the ability to "import" a file into our DB that already exists in S3 (for migration).

Modifications needed for `FileService.createFileVersion`:
- [ ] Along side checking for the s3 object at `temp/uploadId` we need to also check for uploadId (root).
  - It would be nice to execute both headObjects at the same time in parallel.
- [ ] If both locations do _not_ exist throw not found error
- [ ] If both locations _do_ exist throw a client error saying that upload request has already been used
- If existing S3 object (at root) exists and temp one does not exist:
  - [ ] Check DB for node and throw already uploaded error if found
  - Continue on if not found
- [ ] Skip S3 move if it's not needed

zkhin:
@CarsonF For checking uploadId(root), is it enough to only call `getFileNode`?
```
await this.getFileNode(uploadId, session);
```

CarsonF:
From the first bullet point? No, that's another call to S3 with headObject.

zkhin:
I see... so that should look like this?
```
await this.bucket.headObject(`${uploadId}`);
```

PR #1140, opened 2020-09-08
Add file migration logic to file service

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1113
Need to fix `file.e2e` failings

### DIFF
--- a/src/components/file/file.service.ts
+++ b/src/components/file/file.service.ts
@@ -210,8 +210,10 @@ export class FileService {
     session: ISession
   ): Promise<File> {
     let upload;
+    let rootUpload;
     try {
       upload = await this.bucket.headObject(`temp/${uploadId}`);
+      rootUpload = await this.bucket.headObject(`${uploadId}`);
     } catch (e) {
       if (
         (e as AWSError).code === 'NotFound' ||
@@ -220,6 +222,18 @@ export class FileService {
         throw new InputException('Could not find upload', 'uploadId');
       }
       throw new ServerException('Unable to create file version');
+    }
+
+    if (!upload && !rootUpload) {
+      throw new NotFoundException('Could not find upload', 'uploadId');
+    } else if (upload && rootUpload) {
+      throw new InputException('Upload request has already been used');
+    } else if (!upload && rootUpload) {
+      const fileNode = await this.getFileNode(uploadId, session);
+
+      if (fileNode) {
+        throw new InputException('Already uploaded');
+      }
     }
 
     const parent = await this.getParentNode(parentId, session);
--- a/src/components/file/file.service.ts
+++ b/src/components/file/file.service.ts
@@ -1,5 +1,5 @@
 import { Inject, Injectable } from '@nestjs/common';
-import { HeadObjectOutput } from 'aws-sdk/clients/s3';
+import { AWSError } from 'aws-sdk';
 import { generate } from 'shortid';
 import {
   DuplicateException,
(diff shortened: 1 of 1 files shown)


zkhin:
@CarsonF Do you have some time to look into why almost tests are failed in `file.e2e`?
Thank you

src/components/file/file.service.ts
```diff
@@ -210,8 +210,10 @@ export class FileService {
     session: ISession
   ): Promise<File> {
     let upload;
+    let rootUpload;
     try {
       upload = await this.bucket.headObject(`temp/${uploadId}`);
+      rootUpload = await this.bucket.headObject(`${uploadId}`);
```
CarsonF, line 216: Ok so this is the tricky part. Both of these calls throw exceptions but we want to catch and handle both of those together. i.e. both of the calls do _not_ have to be successful - one should fail most of the time even.

We can use `Promise.allSettled` to facilitate this. This also calls both in parallel which is a nice bonus.
```tsx
const [tempUpload, existingUpload] = await Promise.allSettled([
  this.bucket.headObject(`temp/${uploadId}`),
  this.bucket.headObject(uploadId),
]);
```
This replaces the try/catch block btw.

Both of the resulting items have a `status` of either `'fulfilled'` or `'rejected'` and either a `value` or `reason` based on the status. TS has this typed.

From there the logic you added below can be adapted to this.

zkhin: @CarsonF Thanks!!

src/components/file/file.service.ts
```diff
@@ -239,15 +252,19 @@ export class FileService {
       uploadId,
     });
 
-    const mimeType = upload.ContentType ?? 'application/octet-stream';
+    const mimeType =
+      (tempUpload as PromiseFulfilledResult<HeadObjectOutput>).value
```
CarsonF, line 256: Can `tempUpload` & `existingUpload` be resolved to `upload` in block above to not clutter this logic?

src/components/file/file.service.ts
```diff
@@ -209,17 +209,30 @@ export class FileService {
...
-        e instanceof NotFoundException
-      ) {
-        throw new InputException('Could not find upload', 'uploadId');
+    const [tempUpload, existingUpload] = await Promise.allSettled([
+      this.bucket.headObject(`temp/${uploadId}`),
+      this.bucket.headObject(uploadId),
+    ]);
+
+    if (
+      tempUpload.status === 'rejected' &&
```
CarsonF, line 218: I think we should also verify the type of the error in all of these cases. The previous code only threw that exception if it determined the error/reason was not found. Otherwise it threw a `ServerException`.

CarsonF (changes requested):
We also need to not call `moveObject` if it doesn't need to be moved (using existingUpload)

src/components/file/file.service.ts
```diff
@@ -209,17 +209,30 @@ export class FileService {
...
+    ) {
+      throw new InputException('Upload request has already been used');
+    } else if (
+      tempUpload.status === 'rejected' &&
+      existingUpload.status === 'fulfilled'
+    ) {
+      const fileNode = await this.getFileNode(uploadId, session);
+
+      if (fileNode) {
+        throw new InputException('Already uploaded');
```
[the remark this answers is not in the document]
zkhin: Yes, that would be good.

CarsonF (approved):
Awesome, looks great 🎉

Issue #1134, opened 2020-09-07
createProduct > Getting `0` value in `totalVerses` field of the `scriptureReferences`


iPiyushPatel:
**Describe the bug**
createProduct > Getting `0` value in `totalVerses` field of the `scriptureReferences`

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
The  `>0` value should be return for the `totalVerses` field of the `scriptureReferences`

**Actual Result**
Getting `0` value in `totalVerses` field of the `scriptureReferences`
See video -  https://drive.google.com/file/d/1_x4FKj0Fn-5Yda1Ztyc4QUy-44hWqK9n/view?usp=sharing

**GraphQL Query:**
![image](https://user-images.githubusercontent.com/5212585/92382370-1b782300-f12a-11ea-9167-535293040ad0.png)

**GraphQL Error**
```
"scriptureReferences": {
                    "value": [
                        {
                            "start": {
                                "book": "Genesis",
                                "chapter": 3,
                                "verse": 4,
                                "bookName": "Genesis",
                                "label": "Genesis 3:4"
                            },
                            "end": {
                                "book": "Genesis",
                                "chapter": 3,
                                "verse": 4,
                                "bookName": "Genesis",
                                "label": "Genesis 3:4"
                            },
                            "label": "",
                            "totalVerses": 0
                        }
```

**Query and Input File**
https://docs.google.com/document/d/1PTnnTCXVZJchBm44BTXAEst_gUbiHh0XKLM7jHnIBE8/edit?usp=sharing

CordAPIV3# 3a23f628

CarsonF:
https://github.com/SeedCompany/cord-api-v3/blob/master/src/components/scripture/scripture-range.resolver.ts#L30-31

PR #1149, opened 2020-09-09
Calculate total verse in scriptureReferences

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1134

### DIFF
--- a/src/components/scripture/scripture-range.resolver.ts
+++ b/src/components/scripture/scripture-range.resolver.ts
@@ -1,6 +1,7 @@
 import { Int, Parent, ResolveField, Resolver } from '@nestjs/graphql';
 import { stripIndent } from 'common-tags';
 import { ScriptureRange } from './dto';
+import { scriptureToVerseRange } from './reference';
 
 @Resolver(ScriptureRange)
 export class ScriptureRangeResolver {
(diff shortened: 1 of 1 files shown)


src/components/scripture/scripture-range.resolver.ts
```diff
@@ -26,8 +27,8 @@ export class ScriptureRangeResolver {
   @ResolveField(() => Int, {
     description: 'The total number of verses in this scripture range',
   })
-  totalVerses(@Parent() _range: ScriptureRange): number {
-    // TODO
-    return 0;
+  totalVerses(@Parent() range: ScriptureRange): number {
+    const verseRange = scriptureToVerseRange(range);
+    return Math.abs(verseRange.end - verseRange.start) + 1;
```
CarsonF, line 32: We shouldn't need to abs. End should always come after start. I'll create an issue to enforce this.

CarsonF (dismissed):
Thanks @zkhin!

src/components/scripture/scripture-range.resolver.ts
zkhin: Sure, got it

CarsonF (approved):

Issue #1150, opened 2020-09-09
Validate scripture range


CarsonF:
Add decorator to ensure `start` comes before or equal to the `end`.

PR #1152, opened 2020-09-09
Add decorator to ensure start and end order

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1150

### DIFF
--- a/src/components/scripture/dto/scripture-range.validator.ts
+++ b/src/components/scripture/dto/scripture-range.validator.ts
@@ -15,7 +15,7 @@ export const IsValidOrder = createValidationDecorator({
       return verseRange.start <= verseRange.end;
     },
     defaultMessage: () => {
-      return 'start comes before or equal to the end';
+      return 'Scripture range must end after the starting point';
     },
   },
 });
--- a/src/components/scripture/dto/scripture-range.validator.ts
+++ b/src/components/scripture/dto/scripture-range.validator.ts
@@ -1,0 +1,21 @@
+import { ValidationArguments } from 'class-validator';
+import { Merge } from 'type-fest';
+import { createValidationDecorator } from '../../../common/validators/validateBy';
+import { scriptureToVerseRange } from '../reference';
+import { ScriptureRange } from './scripture-range.dto';
+
+// We assume this is only used on the ScriptureRange object
+type ValidationArgs = Merge<ValidationArguments, { object: ScriptureRange }>;
+
+export const IsValidOrder = createValidationDecorator({
+  name: 'ScriptureRange',
+  validator: {
+    validate: (_value, { object: range }: ValidationArgs) => {
+      const verseRange = scriptureToVerseRange(range);
+      return verseRange.start <= verseRange.end;
+    },
+    defaultMessage: () => {
+      return 'start comes before or equal to the end';
+    },
+  },
+});
--- a/src/components/scripture/dto/scripture-range.validator.ts
+++ b/src/components/scripture/dto/scripture-range.validator.ts
@@ -1,7 +1,12 @@
 import { ValidationArguments } from 'class-validator';
 import { Merge } from 'type-fest';
 import { createValidationDecorator } from '../../../common/validators/validateBy';
-import { scriptureToVerseRange } from '../reference';
+import {
+  scriptureToVerseRange,
+  validateBook,
+  validateChapter,
+  validateVerse,
+} from '../reference';
 import { ScriptureRange } from './scripture-range.dto';
 
 // We assume this is only used on the ScriptureRange object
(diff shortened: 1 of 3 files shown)


src/components/scripture/dto/scripture-range.validator.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// We assume this is only used on the ScriptureRange object
+type ValidationArgs = Merge<ValidationArguments, { object: ScriptureRange }>;
+
+export const IsValidOrder = createValidationDecorator({
+  name: 'ScriptureRange',
+  validator: {
+    validate: (_value, { object: range }: ValidationArgs) => {
+      const verseRange = scriptureToVerseRange(range);
+      return verseRange.start <= verseRange.end;
```
CarsonF, line 15: Does this execute after `ScriptureReferenceInput` is validated? i.e. start & end are, at this point, valid objects?

src/components/scripture/dto/scripture-range.validator.ts
```diff
@@ -0,0 +1,21 @@
...
+    },
+    defaultMessage: () => {
+      return 'start comes before or equal to the end';
```
CarsonF, line 18: How about: "Scripture range must end after the starting point"?

src/components/scripture/reference.ts
```diff
@@ -152,10 +152,11 @@ export const validateVerse = (
 // return random scriptureRefenence as an array
 export const createRandomScriptureReferences = (): ScriptureRangeInput[] => {
   const book = books[random(books.length - 1)];
-  const startChapter = random(book.chapters.length - 1);
-  const startVerse = random(book.chapters[startChapter] - 1);
   const endChapter = random(book.chapters.length - 1);
   const endVerse = random(book.chapters[endChapter] - 1);
+  const startChapter = random(endChapter);
+  const startVerse = random(endVerse);
```
CarsonF, line 158: Nice fix

src/components/scripture/dto/scripture-range.validator.ts
zkhin: It validates following the below steps
- Validates `start`
- Validates order
- Validates `end` 

And it returns all the error messages like below
```
"errors": {
   "product.scriptureReferences.0.start.verse": {
        "ScriptureVerse": "Genesis 3 does not have verse 1000"
    },
    "product.scriptureReferences.0.end": {
         "ScriptureRange": "Scripture range must end after the starting point"
    }
},
```

zkhin: So while executing this validator, we don't know `start` and `end` are already `validated`

CarsonF: That's a problem because `scriptureToVerseRange` needs the range to be valid. Does flipping the order make a difference?
```tsx
  @Field()
  @IsValidOrder()
  @ValidateNested()
  @Type(() => ScriptureReferenceInput)
  end: ScriptureReferenceInput;
```

zkhin: It's the same. 😢

zkhin: Btw `scriptureToVerseRange ` returns `start` and `end` even if the `range` is invalid
```
"start": {
      "book": "Genesis",
      "chapter": 3,
      "verse": 1000
},

result: 
start: 1055
```

But if the chapter is invalid, it returns like this.
```
"start": {
     "book": "Genesis",
     "chapter": 1000,
     "verse": 3
},

result:
start: NaN
```

zkhin: So should I have to validate the `range` in the `IsValidOrder` first and if at least one is failed, it should return `true`?

CarsonF: Yeah that code that was copied/ported needs some love. We probably need to check for `NaN` and return true in that case for the same reason. `NaN < x === false`

zkhin: Can you let me know what is `x`? 😉 
Also, it returns a valid `start` number even if the `start` is invalid.
```
"start": {
      "book": "Genesis",
      "chapter": 3,
      "verse": 1000
},

result: 
start: 1055
```

zkhin: So I think we need to validate all the fields one by one using `validateBook`, `validateChapter` and `validateVerse`

CarsonF: Yeah I think that would be best for now

CarsonF (approved):
Gross lol. But it's the best thing right now. Thanks @zkhin

PR #1154, opened 2020-09-09
Resolve ScriptureRange.label

tjgersho:


CarsonF (changes requested):
We want to shorten the label in certain cases. Look at the examples a few lines above this.

tjgersho:
For example, 
Matthew 1-John 2 means Matthew 1:1-John 2:`end verse number`. Right?

CarsonF:
Exactly :+1:

tjgersho:
- Matthew 1:1-John 3:21 should be Matthew 1-John 3:21
- Matthew 1:1-John last:last should be Matthew-John
- Matthew 1:1-Matthew 2:4 should be Matthew 1-2:4

The cases above are right?

CarsonF:
> * Matthew 2:1-John 3:21 should be Matthew 2-John 3:21

If we are displaying verses on one side in which they aren't the first/last verse, let's display them on both. "Matthew 2:1-John 3:21"

If it was "Matthew 1:1-John 3:last" then that would become "Matthew 1-John 2"
If it was "Matthew 1:1-Matthew 3:last" then that would become "Matthew 1-3"


> * Matthew 1:1-John last:last should be Matthew-John

👍 

> * Matthew 1:1-Matthew 2:4 should be Matthew 1-2:4

Same for this as first case. Matthew 1:1-2:4

src/components/scripture/scripture-range.resolver.ts
```diff
@@ -13,15 +14,54 @@ export class ScriptureRangeResolver {
...
+        - Matthew 1:1
+        - Matthew 1:1-20
+        - Matthew 1:1-4:21
+        - Matthew-John
         - Matthew 1-John 2
-        - Matthew 1:3-John 2:4
+        - Matthew 1:1-John 2:4
     `,
   })
   label(@Parent() _range: ScriptureRange): string {
```
CarsonF, line 25:
```suggestion
  label(@Parent() { start, end }: ScriptureRange): string {
```

CarsonF (commented):
Looks good @tjgersho. Could you just make this one change?

CarsonF (approved):

Issue #1144, opened 2020-09-08
Rename Project.deptId -> departmentId (Step 2)


CarsonF:
- [x] Add field resolver to forward `deptId` to `departmentId`
  - Done https://github.com/SeedCompany/cord-api-v3/blob/a32c02431f71febad76d8f89debea17f770488a8/src/components/project/project.resolver.ts#L73-L78
- [ ] Rename `deptId` to `departmentId` in DTOs, DB calls, everywhere internally.
- [ ] Change field resolver (mentioned above) to forward `departmentId` to `deptId` and mark it as deprecated. `deprecatedReason` can be moved from field definition in DTO to here.

Later we'll remove this field resolver entirely after the UI updates.

PR #1155, opened 2020-09-09
Rename deptId to departmentId

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1144

### DIFF
--- a/src/components/project/dto/project.dto.ts
+++ b/src/components/project/dto/project.dto.ts
@@ -39,7 +39,6 @@ class Project extends Resource {
 
   @Field({
     description: 'The legacy department ID',
-    deprecationReason: 'Use `Project.departmentId` instead',
   })
   readonly departmentId: SecuredString;
 
--- a/src/components/project/dto/project.dto.ts
+++ b/src/components/project/dto/project.dto.ts
@@ -41,7 +41,7 @@ class Project extends Resource {
     description: 'The legacy department ID',
     deprecationReason: 'Use `Project.departmentId` instead',
   })
-  readonly deptId: SecuredString;
+  readonly departmentId: SecuredString;
 
   @Field()
   readonly step: SecuredProjectStep;
--- a/src/components/project/project.resolver.ts
+++ b/src/components/project/project.resolver.ts
@@ -70,16 +70,10 @@ export class ProjectResolver {
       : undefined;
   }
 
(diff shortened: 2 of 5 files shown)


src/components/project/dto/project.dto.ts
```diff
@@ -41,7 +41,7 @@ class Project extends Resource {
     description: 'The legacy department ID',
     deprecationReason: 'Use `Project.departmentId` instead',
```
CarsonF, line 42: This needs to be moved to the field resolver

src/components/project/project.resolver.ts
```diff
@@ -74,7 +74,15 @@ export class ProjectResolver {
     description: 'The legacy department ID',
   })
   departmentId(@Parent() project: Project): SecuredString {
-    return project.deptId;
+    return project.departmentId;
+  }
```
CarsonF, line 78: This one isn't needed anymore

CarsonF (changes requested):

src/components/project/dto/project.dto.ts
```diff
@@ -39,9 +39,9 @@ class Project extends Resource {
 
   @Field({
     description: 'The legacy department ID',
-    deprecationReason: 'Use `Project.departmentId` instead',
+    deprecationReason: 'Moved to the field resolver',
```
CarsonF, line 42: Sorry for the confusion, I mean't the `deprecatedReason` should be removed here (`departmentId` is not deprecated) and added to the ResolveField's options for `deptId` which, still is, deprecated.

CarsonF (changes requested):

src/components/project/project.resolver.ts
```diff
@@ -70,11 +70,12 @@ export class ProjectResolver {
       : undefined;
   }
 
+  /** @deprecated Moved from field definition in DTO to here */
   @ResolveField(() => SecuredString, {
     description: 'The legacy department ID',
   })
```
CarsonF, line 76:
```suggestion
  @ResolveField(() => SecuredString, {
    description: 'The legacy department ID',
    deprecationReason: 'Use `Project.departmentId` instead',
  })
```

zkhin: Thanks.

CarsonF (approved):
Thanks @zkhin!

Issue #1156, opened 2020-09-10
query organizations > total number is returned wrong


iPiyushPatel:
**Describe the bug**
query organizations > `total` number is returned wrong

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request.

**Expected Result**
The 'total' should be equal to the sum of all records returned on all pages.

**Actual Result**
`total` number is returned wrong.
See video - https://drive.google.com/file/d/1vl8cGoXupqbSIdjlSqTyLSDKTKGe7ALd/view?usp=drivesdk

**GraphQL Query:**
![image](https://user-images.githubusercontent.com/5212585/92709243-608f9580-f374-11ea-8bc6-8088f811f0fb.png)

**GraphQL Response**
```
"data": {
        "organizations": {
            "total": 13,
            "hasMore": false,
            "items": [
                {
                    "id": "23yKNyq25-",
                    "name": {
                        "value": "automation_Orgsort10092020111453",
                        "canRead": true,
                        "canEdit": true
                    },
                    "avatarLetters": "a",
                    "createdAt": "2020-09-10T11:14:53.565+05:30"
                },
                {
                    "id": "9TJqLA_PIH",
                    "name": {
                        "value": "automation_Orgsort10092020103209",
                        "canRead": true,
                        "canEdit": true
                    },
                    "avatarLetters": "a",
                    "createdAt": "2020-09-10T10:32:09.430+05:30"
                },
                {
                    "id": "seedcompanyid",
                    "name": {
                        "value": "Seed Company",
                        "canRead": true,
                        "canEdit": false
                    },
                    "avatarLetters": "SC",
                    "createdAt": "2020-09-10T10:20:09.809+05:30"
                },
                {
                    "id": "kRCyUEWYu6",
                    "name": {
                        "value": "OrganizationName_10092020111217",
                        "canRead": true,
                        "canEdit": true
                    },
                    "avatarLetters": "O",
                    "createdAt": "2020-09-10T11:12:17.156+05:30"
                },
                {
                    "id": "xX350wW6Vk",
                    "name": {
                        "value": "OrganizationName_10092020111207",
                        "canRead": true,
                        "canEdit": true
                    },
                    "avatarLetters": "O",
                    "createdAt": "2020-09-10T11:12:07.625+05:30"
                },
                {
                    "id": "_wj8g4KSu1",
                    "name": {
                        "value": "OrganizationName_10092020111105",
[...]

PR #1157, opened 2020-09-10
Fix: duplicates in query results

izquierdoalfredo:
Fix https://github.com/SeedCompany/cord-api-v3/issues/1156

### DIFF
--- a/src/core/database/query/lists.ts
+++ b/src/core/database/query/lists.ts
@@ -14,7 +14,9 @@ export function calculateTotalAndPaginateList(
     // .with(['node', 'total']) TODO needed?
     .call(sort, sortInput, order)
     .with([
-      `collect(distinct node.id)[${(page - 1) * count}..${page * count}] as items`,
+      `collect(distinct node.id)[${(page - 1) * count}..${
+        page * count
+      }] as items`,
       'total',
     ])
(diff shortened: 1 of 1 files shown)


src/core/database/query/lists.ts
```diff
@@ -14,7 +14,7 @@ export function calculateTotalAndPaginateList(
     // .with(['node', 'total']) TODO needed?
     .call(sort, sortInput, order)
     .with([
-      `collect(node.id)[${(page - 1) * count}..${page * count}] as items`,
+      `collect(distinct node.id)[${(page - 1) * count}..${page * count}] as items`,
```
CarsonF, line 17: Haven't the nodes already been distinct-ed on line 12?

izquierdoalfredo: @CarsonF  Yes, but results still contain duplicates.
So I added last distinct operator

izquierdoalfredo: @CarsonF 
Do you agree with my solution?

CarsonF: Ok it looks like since the `sort` callback can do more matches this is needed and does make a difference. 👍

CarsonF (approved):
Thanks @izquierdoalfredo!

PR #1169, opened 2020-09-10
Remove old update property methods with acl

frankyma:
Removed the old update property methods that searched by ACL connections.  Also refactored create unavailability because that was using the old create service with ACL connections.

Next will remove the old create service, will do that on a separate PR that includes refactoring file nodes create and file nodes read

### DIFF
--- a/src/components/user/unavailability/unavailability.service.ts
+++ b/src/components/user/unavailability/unavailability.service.ts
@@ -80,9 +80,15 @@ export class UnavailabilityService {
           [],
           session.userId === this.config.rootAdmin.id
         )
-        .return('node.id as id');
+        .return('node.id as id')
+        .asResult<{ id: string }>();
 
-      const unavailabilityId = await createUnavailability.first();
+      const createUnavailabilityResult = await createUnavailability.first();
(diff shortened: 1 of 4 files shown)


src/components/user/unavailability/unavailability.service.ts
```diff
@@ -15,59 +22,97 @@ import {
...
+          {
+            owningOrgId: session.owningOrgId,
+          },
+          [],
+          session.userId === this.config.rootAdmin.id
+        )
+        .return('node.id as id')
+        .asResult<{ id: string }>();
+
+      const createUnavailabilityResult = await createUnavailability.first();
```
CarsonF, line 86: first() can return null if nothing was matched. I would add a null check to be safe. This also prevents the need to non-null assert below

frankyma: Oh I thought the entire call would just fail and get caught by the try...catch?  But I'm not 100%, I'll add the null check to be safe

CarsonF (approved):
Thanks @frankyma!

PR #1171, opened 2020-09-10
1151 validate user roles on team

tjgersho:


src/components/project/project-member/project-member.service.ts
```diff
@@ -193,6 +194,11 @@ export class ProjectMemberService {
       );
     }
 
+    const user = await this.userService.readOne(userId, session);
+    if (difference(input.roles, user.roles.value).length) {
+      throw new InputException('No roles in the user', 'input.roles');
```
CarsonF, line 199: This error message doesn't make any sense. Something like "Project member cannot be assigned these roles" would be better. Listing the invalid roles from the `difference()` call would be helpful too.

CarsonF (changes requested):

CarsonF (approved):

Issue #1167, opened 2020-09-10
Language.signLanguageCode


CarsonF:
An optional string field that can be set on create/update.

Validate that it follows the format `XXYY` where `X` is an uppercase letter and `Y` is a number.

PR #1176, opened 2020-09-11
Add signLanguageCode to language

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1167

### DIFF
--- a/src/components/language/dto/create-language.dto.ts
+++ b/src/components/language/dto/create-language.dto.ts
@@ -6,6 +6,7 @@ import {
   IsNumberString,
   IsPositive,
   Length,
+  Matches,
   ValidateNested,
 } from 'class-validator';
 import { NameField } from '../../../common';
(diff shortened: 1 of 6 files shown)


src/components/language/dto/create-language.dto.ts
```diff
@@ -69,6 +70,10 @@ export abstract class CreateLanguage {
 
   @NameField({ nullable: true })
   readonly leastOfTheseReason?: string;
+
+  @Field({ nullable: true })
+  @IsSignLanguageCode()
```
CarsonF, line 75:
```suggestion
  @Matches(/^[A-Z]{2}\d{2}$/)
```

CarsonF (changes requested):
Since the validation is just a one liner and only used in one spot I don't think it's worth a custom validator

CarsonF:
Ah except it's used in 2 spots 🤦 . Ok why don't you keep the custom validator but just have it forward to the Matches validator

CarsonF (dismissed):

zkhin:
@CarsonF There's a conflict now, I will rebase master.

CarsonF (approved):

Issue #1166, opened 2020-09-10
Remove EthnologueLanguage.id


CarsonF:
It is no longer used.

PR #1177, opened 2020-09-11
Remove id from enthnologue language

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1166


CarsonF (approved):
Thanks @zkhin!

Waiting for UI to stop using it first.

/cc @SeedCompany/cord-ui

Issue #1161, opened 2020-09-10
Language.sponsorDate -> sponsorStartDate


CarsonF:
Rename the field

PR #1179, opened 2020-09-11
Change sponsor date to sponsor start date

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1161

### DIFF
--- a/src/components/language/language.resolver.ts
+++ b/src/components/language/language.resolver.ts
@@ -54,7 +54,7 @@ export class LanguageResolver {
     description: `The fiscal year of the sponsor date`,
   })
   beginFiscalYear(@Parent() language: Language): SecuredInt {
-    const { value, ...rest } = language.sponsorDate;
+    const { value, ...rest } = language.sponsorStartDate;
     return {
       ...rest,
       value: value ? fiscalYear(value) : undefined,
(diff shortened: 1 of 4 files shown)


src/components/language/language.resolver.ts
```diff
@@ -54,7 +54,7 @@ export class LanguageResolver {
     description: `The fiscal year of the sponsor date`,
```
CarsonF, line 54:
```suggestion
    description: `The fiscal year of the sponsor start date`,
```

CarsonF (changes requested):

src/components/language/language.resolver.ts
CarsonF: Actually can you just remove this field @zkhin? I've changed my mind that value can just be calculated by the consumer.

zkhin: So you mean I have to remove the origin `sponsorDate` field?

CarsonF: No I mean remove the `beginFiscalYear` resolver

zkhin: I see.

CarsonF (approved):

PR #1182, opened 2020-09-11
Change Language.sensitivity to be set by user instead of being calculated

### LINKED ISSUE #1160: Language.sensitivity not calculated
We thought it would be calculated but it's not going to be. It needs to be settable optionally via create & update.
It should default to `High` on create.

izquierdoalfredo:
Fixes #1160

### DIFF
--- a/src/components/language/dto/create-language.dto.ts
+++ b/src/components/language/dto/create-language.dto.ts
@@ -9,7 +9,7 @@ import {
   Matches,
   ValidateNested,
 } from 'class-validator';
-import { NameField } from '../../../common';
+import { NameField, Sensitivity } from '../../../common';
 import { Language } from './language.dto';
 
 @InputType()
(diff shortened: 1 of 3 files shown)


src/components/language/dto/create-language.dto.ts
```diff
@@ -69,6 +69,9 @@ export abstract class CreateLanguage {
 
   @NameField({ nullable: true })
   readonly leastOfTheseReason?: string;
+
+  @NameField({ nullable: true })
+  readonly sensitivity?: Sensitivity;
```
CarsonF, line 74:
```suggestion
  @Field(() => Sensitivity, { nullable: true })
  readonly sensitivity?: Sensitivity = Sensitivity.High;
```

src/components/language/dto/update-language.dto.ts
```diff
@@ -72,6 +72,9 @@ export abstract class UpdateLanguage {
 
   @NameField({ nullable: true })
   readonly leastOfTheseReason?: string;
+
+  @NameField({ nullable: true })
```
CarsonF, line 76:
```suggestion
  @Field(() => Sensitivity, { nullable: true })
```

CarsonF (changes requested):

CarsonF (dismissed):
Thanks @izquierdoalfredo!

CarsonF (approved):

PR #1183, opened 2020-09-11
1158 language.isSignLanguage

### LINKED ISSUE #1158: Language.isSignLanguage boolean field
Add `isSignLanguage` boolean field to Language. It can assigned at create & update. defaults to false.

tjgersho:
Closes #1158

### DIFF
--- a/src/components/language/dto/update-language.dto.ts
+++ b/src/components/language/dto/update-language.dto.ts
@@ -72,6 +72,9 @@ export abstract class UpdateLanguage {
   readonly leastOfTheseReason?: string;
 
   @Field({ nullable: true })
+  readonly isSignLanguage?: boolean;
+
+  @Field({ nullable: true })
   @Matches(/^[A-Z]{2}\d{2}$/)
   readonly signLanguageCode?: string;
 }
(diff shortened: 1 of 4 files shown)


src/components/language/dto/update-language.dto.ts
```diff
@@ -72,6 +72,9 @@ export abstract class UpdateLanguage {
 
   @NameField({ nullable: true })
   readonly leastOfTheseReason?: string;
+
+  @Field({ nullable: true })
+  readonly isSignLanguage?: boolean = false;
```
CarsonF, line 77:
```suggestion
  readonly isSignLanguage?: boolean;
```
We don't want it set back to false if it's omitted. Probably copy/paste error 😉

CarsonF (changes requested):
Just that one thing and the conflicts, then good to go

src/components/language/dto/update-language.dto.ts
tjgersho: 🤦

CarsonF (approved):
Thanks @tjgersho!

PR #1185, opened 2020-09-11
add test for product with scriptureRefOverride which is null

tjgersho:
Can be merged after merging https://github.com/SeedCompany/cord-api-v3/pull/1096


CarsonF (approved):
Thanks @tjgersho!

PR #1186, opened 2020-09-11
Create multiple scripture references verride bug

frankyma:
The unused `sr` node alias was causing a duplicate variable error when multiple scripture references were made


CarsonF (approved):
Nice find

PR #1187, opened 2020-09-12
fix createRandomScriptureRange

tjgersho:


src/components/scripture/reference.ts
```diff
@@ -160,7 +160,12 @@ const createRandomScriptureRange = (): ScriptureRangeInput => {
   const endChapter = random(book.chapters.length - 1);
   const endVerse = random(book.chapters[endChapter] - 1);
   const startChapter = random(endChapter);
-  const startVerse = random(endVerse);
+  let startVerse = 0;
+  if (startChapter === endChapter) {
+    startVerse = random(endVerse);
+  } else {
+    startVerse = random(book.chapters[startChapter] - 1);
+  }
```
CarsonF, line 168: I think maybe we just need random to give back a 1-N instead of 0-N. That way we wouldn't need to +1 in the output. For chapters and verses only.

```
random(1, x)
```

tjgersho: Okay.

CarsonF (approved):
Sweet! 🎉

PR #1203, opened 2020-09-14
Failing tests product scriptures

frankyma:
Product scriptures were returning unsorted so the tests failed.

Sorted the returned results and check with `expect.arrayContaining` to match array items.


CarsonF (approved):
Awesome, thanks @frankyma!

Issue #1194, opened 2020-09-14
Partner Module


CarsonF:
Partner Module with CRUD functionality.

To start we only need a relation to an organization.

PR #1204, opened 2020-09-14
Add Partner module

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1194

### DIFF
--- a/src/components/partner/dto/list-partner.dto.ts
+++ b/src/components/partner/dto/list-partner.dto.ts
@@ -1,4 +1,4 @@
-import { Field, InputType, ObjectType } from '@nestjs/graphql';
+import { InputType, ObjectType } from '@nestjs/graphql';
 import { Type } from 'class-transformer';
 import { ValidateNested } from 'class-validator';
 import {
@@ -10,7 +10,6 @@ import { Partner } from './partner';
 
 @InputType()
 export abstract class PartnerFilters {
-  @Field({ nullable: true })
   readonly organizationId?: string;
 }
 
@@ -22,7 +21,6 @@ export class PartnerListInput extends SortablePaginationInput<keyof Partner>({
 }) {
   static defaultVal = new PartnerListInput();
 
-  @Field({ nullable: true })
   @Type(() => PartnerFilters)
   @ValidateNested()
   readonly filter: PartnerFilters = defaultFilters;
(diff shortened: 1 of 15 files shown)


src/components/partner/dto/list-partner.dto.ts
```diff
@@ -0,0 +1,37 @@
...
+  PaginatedList,
+  SecuredList,
+  SortablePaginationInput,
+} from '../../../common';
+import { Partner } from './partner';
+
+@InputType()
+export abstract class PartnerFilters {
+  @Field({ nullable: true })
+  readonly organizationId?: string;
```
CarsonF, line 14: We aren't exposing ID filters through the API now. 

If you need them for internal use just remove the field decorators.
```suggestion
  readonly organizationId?: string;
```

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -0,0 +1,28 @@
...
+import { IdField } from '../../../common';
+import { Partner } from './partner';
+
+@InputType()
+export abstract class UpdatePartner {
+  @IdField()
+  readonly id: string;
+
+  @IdField({ nullable: true })
+  readonly organizationId?: string;
```
CarsonF, line 13: Mmm I don't think we want to be able to change organization on update. It's a fundament part of the relationship. Just like we don't allow changing the language or project on a language engagement.

src/components/partner/partner.resolver.ts
```diff
@@ -0,0 +1,99 @@
...
+import { PartnerService } from './partner.service';
+
+@Resolver(Partner)
+export class PartnerResolver {
+  constructor(
+    private readonly partnerService: PartnerService,
+    private readonly orgService: OrganizationService
+  ) {}
+
+  @Mutation(() => CreatePartnerOutput, {
```
CarsonF, line 29: Would you mind moving this after the query & field resolvers so that all the mutations are grouped together?

src/components/user/user.service.ts
```diff
@@ -268,6 +268,8 @@ export class UserService {
           canCreateLocation: true,
           canCreateEthnologueLanguage: true,
           canReadEthnologueLanguages: true,
+          canCreatePartner: true,
+          canReadPartners: true,
```
CarsonF, line 272: I'm not even sure if this is used anymore... @michaelmarshall?

CarsonF (changes requested):
Thanks @zkhin!

Just a couple things I see

CarsonF (approved):

src/components/user/user.service.ts
michaelmarshall: oh ya, all those props can be removed.

PR #1206, opened 2020-09-15
Update product scripture references

frankyma:
Have input checks for updating product scripture references (override) based on product type.

Have `scriptureRefService.update` take care of accepting `null` override values and setting the isOverride tag to false

### DIFF
--- a/src/components/product/product.service.ts
+++ b/src/components/product/product.service.ts
@@ -212,12 +212,17 @@ export class ProductService {
     }
 
     if (input.produces) {
-      const produce = await this.db
+      const producible = await this.db
         .query()
-        .match([node('pr', 'Producible', { id: input.produces, active: true })])
-        .return('pr')
+        .match([
+          node('producible', 'Producible', {
+            id: input.produces,
+            active: true,
+          }),
+        ])
+        .return('producible')
         .first();
-      if (!produce) {
+      if (!producible) {
         this.logger.warning(`Could not find producible node`, {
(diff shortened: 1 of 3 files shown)


src/components/product/product.service.ts
```diff
@@ -440,20 +446,51 @@ export class ProductService {
...
+        'Cannot update produces on a Direct Scripture Product'
+      );
+    }
+
+    //If current product is a Derivative Scripture Product, cannot update scriptureReferencesOverride
+    if (!isDirectScriptureProduct && scriptureReferences) {
+      throw new InputException(
+        'Cannot update Scripture References on a Derivative Scripture Product'
+      );
+    }
```
CarsonF, line 476: Can you add the field name argument to these exceptions?

test/product.e2e-spec.ts
```diff
@@ -304,6 +304,95 @@ describe('Product e2e', () => {
...
+        {
+          input: {
+            product: {
+              id: product.id,
+              scriptureReferencesOverride: ScriptureRange.randomList(),
+            },
+          },
+        }
+      )
+    ).rejects.toThrowError();
```
CarsonF, line 333: Can you specify the error message, type, or code here so that we assert exactly what we are trying to check for here?

CarsonF (changes requested):

CarsonF (approved):
Thanks @frankyma!

Issue #1213, opened 2020-09-15
Password Reset Errors


sethmcknight:
When attempting to create an account on beta.cordfield.com I am informed that the user ID is already in use.

I then attempt to reset the password but receive an error that the password reset cannot be sent because the 
>"Email address is not verified. The following identities >failed the check in region US-EAST-2: >noreply@cordfield.com, seth_mcknight@tsco.org"

<img width="1552" alt="Screen Shot 2020-09-15 at 7 50 39 AM" src="https://user-images.githubusercontent.com/7377802/93213379-634e0a00-f729-11ea-918f-46ae1e5a5e82.png">

michaelmarshall:
@zkhin please replace the last reset password query in the authentication service with this and let me know if it still works for you locally. We'll then push to beta and let Seth try.
```
          MATCH(e:EmailToken {token: $token})
          DELETE e
          WITH *
          MATCH (:EmailAddress {active: true, value: $email})<-[:email {active: true}]-(user:User {active: true})
          OPTIONAL MATCH (user)-[oldPasswordRel:password]->(oldPassword)
          SET oldPasswordRel.active = false, oldPassword.active = false
          WITH user
          LIMIT 1
          MERGE (user)-[:password {active: true}]->(password:Property {active: true})
          SET password.value = $password
          RETURN password
```

PR #1215, opened 2020-09-15
Authentication reset password test case

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1213

@CarsonF Cannot reproduce the issue on backend. As you can see in this case, it passes all the reset password workflow (https://github.com/SeedCompany/cord-api-v3/blob/363826e9c396eab800a46afb5c590f25c76e046c/test/authentication.e2e-spec.ts#L29~L89)


CarsonF (changes requested):
This is just updating a test case it doesn't fix #1213

CarsonF (approved):
Thanks @zkhin!

PR #1218, opened 2020-09-16
Set project sensitivity

frankyma:
Translation project sensitivity is calculated on read, internship allowed to be set by user

### DIFF
--- a/src/components/project/dto/create-project.dto.ts
+++ b/src/components/project/dto/create-project.dto.ts
@@ -38,8 +38,11 @@ export abstract class CreateProject {
   @Field(() => ProjectStep, { nullable: true })
   readonly step?: ProjectStep;
 
-  @Field(() => Sensitivity, { nullable: true })
-  readonly sensitivity?: Sensitivity = Sensitivity.High;
+  @Field(() => Sensitivity, {
+    description: 'Defaults to High, only available on internship projects',
+    nullable: true,
+  })
+  readonly sensitivity?: Sensitivity;
 }
 
 @InputType()
--- a/src/components/project/dto/create-project.dto.ts
+++ b/src/components/project/dto/create-project.dto.ts
@@ -1,7 +1,13 @@
 import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import { Type } from 'class-transformer';
 import { ValidateNested } from 'class-validator';
-import { CalendarDate, DateField, IdField, NameField } from '../../../common';
+import {
+  CalendarDate,
+  DateField,
+  IdField,
+  NameField,
+  Sensitivity,
+} from '../../../common';
 import { IProject, Project } from './project.dto';
 import { ProjectStep } from './step.enum';
 import { ProjectType } from './type.enum';
@@ -31,6 +37,9 @@ export abstract class CreateProject {
 
   @Field(() => ProjectStep, { nullable: true })
   readonly step?: ProjectStep;
+
+  @Field(() => Sensitivity, { nullable: true })
+  readonly sensitivity?: Sensitivity = Sensitivity.High;
 }
 
 @InputType()
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -213,12 +213,16 @@ export class ProjectService {
     }
 
     if (input.type === ProjectType.Translation && input.sensitivity) {
-      throw new InputException('Cannot set sensitivity on tranlation project');
+      throw new InputException(
+        'Cannot set sensitivity on tranlation project',
+        'project.sensitivity'
+      );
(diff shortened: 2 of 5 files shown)


src/components/project/dto/create-project.dto.ts
```diff
@@ -31,6 +37,9 @@ export abstract class CreateProject {
 
   @Field(() => ProjectStep, { nullable: true })
   readonly step?: ProjectStep;
+
+  @Field(() => Sensitivity, { nullable: true })
```
CarsonF, line 41: Can you add description here and update that says this is only for internship projects

src/components/project/project.service.ts
```diff
@@ -499,7 +515,9 @@ export class ProjectService {
     return {
       ...parseBaseNodeProperties(result.node),
       ...securedProps,
-      sensitivity: props.sensitivity,
+      sensitivity:
+        getHighestSensitivity(result.languageSensitivityList) ||
+        props.sensitivity,
```
CarsonF, line 520: Falling back to a database value doesn't make sense. Would translation projects even have this property?
Since we default to High we can just return that instead of undefined in your getHighest function.

src/components/project/project.service.ts
```diff
@@ -445,10 +449,22 @@ export class ProjectService {
...
+        relation('out', '', 'engagement', { active: true }),
+        node('', 'LanguageEngagement', { active: true }),
+        relation('out', '', 'language', { active: true }),
+        node('', 'Language', { active: true }),
+        relation('out', '', 'sensitivity', { active: true }),
+        node('sensitivity', 'Property', { active: true }),
+      ])
+      .return(
+        'propList, permList, node, country.id as countryId, collect(distinct sensitivity.value) as languageSensitivityList'
+      )
```
CarsonF, line 463:
```suggestion
      .return([
        'propList',
        'permList',
        'node',
        'country.id as countryId',
        'collect(distinct sensitivity.value) as languageSensitivityList',
      ])
```

src/components/project/project.service.ts
```diff
@@ -211,10 +212,13 @@ export class ProjectService {
       throw new UnauthenticatedException('user not logged in');
     }
 
+    if (input.type === ProjectType.Translation && input.sensitivity) {
+      throw new InputException('Cannot set sensitivity on tranlation project');
```
CarsonF, line 216:
```suggestion
      throw new InputException('Cannot set sensitivity on translation projects', 'project.sensitivity');
```

CarsonF (changes requested):

src/components/project/project.service.ts
frankyma: Yes, both projects will have this property because both create a **High** sensitivity property on create, just that internship projects will have this node changed on update.

The getHighest function is just if given a sensitivity list and return the highest, I'd like to keep this modularized.

src/components/project/dto/create-project.dto.ts
frankyma: Added to both

src/components/project/project.service.ts
frankyma: changed

src/components/project/project.service.ts
frankyma: oops missed, changed

src/components/project/project.service.ts
CarsonF: Ok can we just fallback to High here then? 
```tsx
sensitivity: getHighestSensitivity(result.languageSensitivityList) ?? Sensitivity.High,
```

I think it's confusing that it is conditionally calculated from a different stored value when the logic really is just defaulting to high.

CarsonF: Oh I see this is for both types of projects. You're just banking on the fact that an internship project will have an empty language list.

frankyma: exactly

CarsonF: Can you add a comment stating this? This isn't immediately apparent. Also that translation projects get a default of high in db on create, which is only used when there is no engagements.

frankyma: added comment, see if it makes sense

CarsonF: Beautiful, thanks Frank

CarsonF (dismissed):
Thanks @frankyma!

CarsonF:
@frankyma CI has been running twice now for over an hour. I think something is wrong here.

frankyma:
on it

frankyma:
Yep I had some 🐞 in my code.  fixed

CarsonF (approved):

Issue #1168, opened 2020-09-10
Partner.types


CarsonF:
Same as `Partnerships.types` copied to Partners. Maybe rename `PartnershipType` to `PartnerType`.

Like `ProjectMember.roles`, allowed `Partnership.types` need to be limited to a subset of `Partner.types`.

CarsonF:
Ok waiting until we have Partner module and going to put it there

zkhin:
@CarsonF I think we need to wait until this https://github.com/SeedCompany/cord-api-v3/pull/1211 PR is merged

PR #1211, opened 2020-09-15
Update partnership to have partner relation

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1196

PR #1222, opened 2020-09-16
Add types to partner module

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1168
Blocked until https://github.com/SeedCompany/cord-api-v3/pull/1211 is merged


src/components/partner/partner.service.ts
```diff
@@ -225,10 +242,27 @@ export class PartnerService {
...
+        value: secured.types.value || [],
+      },
     };
   }
 
   async update(input: UpdatePartner, session: ISession): Promise<Partner> {
+    const object = await this.readOne(input.id, session);
+    const changes = {
+      ...input,
+      types: uniq(input.types),
```
CarsonF, line 256: I wasn't going to make a big deal of it before, but I think this would be better as a transform on the DTO.
```tsx
@Transform(uniq)
```

src/components/partner/dto/partner-type.enum.ts
```diff
@@ -0,0 +1,11 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum PartnerType {
```
CarsonF, line 3: This should be merged with `PartnershipType`....probably this one kept

zkhin: So I have to move `PartnershipType` to `partner-type.enum.ts`, right?

CarsonF: What I mean is they are the same enum, so we don't need both. I think it makes sense for the type to be here since this is the "source" from the business' perspective. That means we should delete `PartnershipType` and replace usage with `PartnerType`

zkhin: Got it, thanks

CarsonF (approved):

CarsonF:
Waiting for UI to catch up

CarsonF (approved):

Issue #1217, opened 2020-09-15
Implement secret for password hashing


CarsonF:
- [ ] Add an optional string `passwordSecret` to `ConfigService`
- [ ] Use it in all argon2 calls in `AuthenticationService`. The `secret` option.

PR #1224, opened 2020-09-16
Add secret to password hashing

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1217

### DIFF
--- a/src/components/authentication/authentication.service.ts
+++ b/src/components/authentication/authentication.service.ts
@@ -106,9 +106,7 @@ export class AuthenticationService {
       throw e;
     }
 
-    const passwordHash = await argon2.hash(input.password, {
-      secret: this.config.passwordSecret,
-    });
+    const passwordHash = await argon2.hash(input.password, this.argon2Options);
     await this.db
       .query()
       .match([
@@ -163,9 +161,7 @@ export class AuthenticationService {
 
     if (
       !result1 ||
-      !(await argon2.verify(result1.pash, input.password, {
-        secret: this.config.passwordSecret,
-      }))
+      !(await argon2.verify(result1.pash, input.password, this.argon2Options))
     ) {
       throw new UnauthenticatedException('Invalid credentials');
     }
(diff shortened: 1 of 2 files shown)


src/core/config/config.service.ts
```diff
@@ -91,6 +91,13 @@ export class ConfigService {
     };
   }
 
+  @Lazy() get passwordSecret() {
+    return Buffer.from(
+      this.env.string('PASSWORD_SECRET').optional('passwordSecret'),
```
CarsonF, line 96: Let's just omit it by default as to not mess up existing databases
```suggestion
      this.env.string('PASSWORD_SECRET').optional(),
```

src/core/config/config.service.ts
```diff
@@ -91,6 +91,13 @@ export class ConfigService {
     };
   }
 
+  @Lazy() get passwordSecret() {
+    return Buffer.from(
```
CarsonF, line 95: Would you mind doing this conversion has a private getter/method in the AuthenticationService? I don't think the fact that argon2 needs it as a buffer should concern the config service.

CarsonF (changes requested):

CarsonF (dismissed):
Thanks @zkhin!

src/components/authentication/authentication.service.ts
```diff
@@ -419,9 +420,11 @@ export class AuthenticationService {
...
 
-  private get argon2Options() {
-    return this.config.passwordSecret
-      ? { secret: Buffer.from(this.config.passwordSecret, 'utf-8') }
-      : {};
+  private get argon2Options(): Except<argon2.Options, 'raw'> {
+    return {
+      secret: this.config.passwordSecret
+        ? Buffer.from(this.config.passwordSecret, 'utf-8')
+        : undefined,
```
zkhin, line 427: I don't think `undefined` works for argon2.Options

CarsonF: Really? Ok good catch.

CarsonF (approved):

Issue #643, opened 2020-07-01
Project > Get all projects > All data is returned even though I've applied filter properties


iPiyushPatel:
**Describe the bug**
Project > Get all projects > All data is returned even though I've applied filter properties

**Steps to Reproduce**
On Graphql playground or Postman - create a API request as below snap and execute request.

**Expected Result**
Projects results are returned as per requested filters.

**Actual Result**
All the projects are returned. See video - https://www.awesomescreenshot.com/video/307369?key=556a507a6580145b26d3dcf881fea8fe

I've also tested with other filter parameters. 

GraphQL Query:
![image](https://user-images.githubusercontent.com/5212585/86254498-7eea4f80-bbd3-11ea-8a5a-4651fb0940b6.png)

CordAPIV3# 1caff78

michaelmarshall:
Filter hasn't been added back to `projects` yet. This will need to use the new refactor.

punamasawa777:
@michaelmarshall,  Helper function is required to filter data if filter input is Array or other Datatype
currently filter it is working only for strings type

michaelmarshall:
@punamasawa777 yes, that is currently a limitation. feel free to update the helper function to handle more cases. thanks.

punamasawa777:
sure @michaelmarshall, I will look into it

punamasawa777:
@michaelmarshall , I have worked on List and boolean filter & got an issue Did u fixed ? if fixed please let me know as i am blocked..

candleshine:
Punam is asking @michaelmarshall to pick up the branch that he started to finish this filter

michaelmarshall:
I'm now working this alongside other filter-bug issues.

reneepe:
Additional details in https://github.com/SeedCompany/cord-api-v3/issues/768

iPiyushPatel:
Hi @michaelmarshall @CarsonF @reneepe 

As of now, 
**Working Filters** -by  type
**Not Working Filters** - by name,createdAt, modifiedAt, mine	
**Couldn't check due to static response values** - by sensitivity, status, step

I also have one question. What is 'filter by clusters'?
![image](https://user-images.githubusercontent.com/5212585/90480221-91055a80-e14d-11ea-94ba-bc0c16e3bce4.png)

Please suggest if any.

tjgersho:
When a project works on more than one language it is a 'cluster' project. So it'll have more than one engagement, naturally.

[...]

PR #1234, opened 2020-09-17
Project filters

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/643

### DIFF
--- a/src/components/project/query.helpers.ts
+++ b/src/components/project/query.helpers.ts
@@ -1,17 +1,15 @@
-import { contains, inArray, node, Query, relation } from 'cypher-query-builder';
+import { inArray, node, Query, relation } from 'cypher-query-builder';
 import { ProjectFilters } from './dto';
 
 export function projectListFilter(query: Query, filter: ProjectFilters) {
   query
-    .match([...propMatch('name')])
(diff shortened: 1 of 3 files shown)


src/components/project/query.helpers.ts
```diff
@@ -0,0 +1,46 @@
+import { inArray, node, Query, relation } from 'cypher-query-builder';
+import { ProjectFilters } from './dto';
+
+export function projectListFilter(query: Query, filter: ProjectFilters) {
+  query
+    .match([...propMatch('status')])
+    .call((q) =>
+      filter.status ? q.where({ status: { value: inArray(filter.status) } }) : q
+    )
+    .match([...propMatch('sensitivity')])
```
CarsonF, line 10: We only want to match these two if they are given

zkhin: You mean, I have to change `match` to `optionalMatch` for these two fields?

CarsonF: I mean don't call match if the value is nil

zkhin: Okay

CarsonF (approved):

Issue #1221, opened 2020-09-16
updateUser > Getting "Invalid input ';': expected whitespace, UNIQUE or a pattern" error when updating with the existing user role


iPiyushPatel:
**Describe the bug**
updateUser  > Getting `Invalid input ';': expected whitespace, UNIQUE or a pattern` error when updating with the existing user role

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
Proper validation message should be returned when updating with the existing user role

**Actual Result**
Getting `Invalid input ';': expected whitespace, UNIQUE or a pattern` error when updating with the existing user role
See video -  https://drive.google.com/file/d/1sWgm5w5NZW_eD9VZB2WJ0F04Pe8x28na/view?usp=sharing

**GraphQL Query(updateUser):**
![image](https://user-images.githubusercontent.com/5212585/93332070-e01ed800-f83e-11ea-917a-fa27f438fc8a.png)

**GraphQL response**
```

    "data": {
        "updateUser": {
            "user": {
                "id": "2nVxthkjz",
                "fullName": "FirstName1692020045801 LastName1692020045801",
                "firstName": "FirstName1692020045801",
                "avatarLetters": "FL",
                "createdAt": "2020-09-16T10:30:55.702+05:30",
                "email": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "piyush123456@mail.com"
                },
                "realFirstName": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "FirstName1692020045801"
                },
                "realLastName": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "LastName1692020045801"
                },
                "displayFirstName": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "DisplayFirstName1692020045801"
                },
                "displayLastName": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "DisplayLastName1692020045801"
                },
                "phone": {
                    "canRead": true,
                    "canEdit": true,
                    "value": "903-498-5014"
                },
                "timezone": {
                    "canRead": true,
                    "canEdit": true,
                    "value": {
                        "name": "Europe/Andorra",
                        "lat": 42.5,
                        "long": 1.5167,
                        "countries": [
                            {
                                "code": "AD",
                                "name": "Andorra",
                                "zones": [
                                    {
                                        "name": "Europe/Andorra",
[...]

PR #1238, opened 2020-09-18
Fix update user role issue

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1221


src/components/user/user.service.ts
```diff
@@ -402,15 +402,17 @@ export class UserService {
...
+        await this.db
+          .query()
+          .match([
+            node('user', ['User', 'BaseNode'], {
+              id: input.id,
+            }),
+          ])
+          .create([...this.roleProperties(newRoles)])
+          .run();
+      }
```
CarsonF, line 415: What about removing existing roles not present in the input?

zkhin: Okay, I will remove old roles.

src/components/user/user.service.ts
```diff
@@ -401,15 +400,27 @@ export class UserService {
...
           node('user', ['User', 'BaseNode'], {
             id: input.id,
           }),
+          relation('out', 'oldRoleRel', 'roles'),
+          node('oldRoles', 'Property'),
         ])
-        .create([...this.roleProperties(newRoles)])
+        .delete('oldRoleRel')
+        .delete('oldRoles')
+        .run();
```
CarsonF, line 414: @michaelmarshall you still want active=false here right?

michaelmarshall: yes `active` still exists on relationships

michaelmarshall: so don't `delete` relationships, just mark them as `active = false`

michaelmarshall (changes requested):

michaelmarshall (dismissed):

PR #1247, opened 2020-09-21
Add funding account module

zkhin:


src/components/funding-account/funding-account.service.ts
```diff
@@ -0,0 +1,216 @@
...
+    });
+  }
+
+  async delete(id: string, session: ISession): Promise<void> {
+    const fundingAccount = await this.readOne(id, session);
+    try {
+      await this.db.deleteNode({
+        session,
+        object: fundingAccount,
+        aclEditProp: 'canDeleteOwnUser',
```
michaelmarshall, line 182: we are in the middle of refactoring how we delete nodes. this function and the test can be 'non-implemented' for now.

test/funding-account.e2e-spec.ts
```diff
@@ -0,0 +1,129 @@
...
+        },
+      }
+    );
+    const updated = result.updateFundingAccount.fundingAccount;
+    expect(updated).toBeTruthy();
+    expect(updated.name.value).toBe(newName);
+  });
+
+  // Delete FundingAccount
+  it('delete funding account', async () => {
```
michaelmarshall, line 86: best to skip this until our delete refactor is done.

src/components/funding-account/funding-account.service.ts
michaelmarshall: fyi, we won't be using `acl` concepts at all.

zkhin: Okay

CarsonF (approved):

PR #1249, opened 2020-09-21
Add registry of geography module

zkhin:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3655294965) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/registry-of-geography/registry-of-geography.service.ts
+++ b/src/components/registry-of-geography/registry-of-geography.service.ts
@@ -203,20 +203,8 @@ export class RegistryOfGeographyService {
     });
   }
 
-  async delete(id: string, session: ISession): Promise<void> {
-    const RegistryOfGeography = await this.readOne(id, session);
-    try {
-      await this.db.deleteNode({
-        session,
-        object: RegistryOfGeography,
-        aclEditProp: 'canDeleteOwnUser',
-      });
-    } catch (e) {
-      this.logger.error('Failed to delete', { id, exception: e });
-      throw new ServerException('Failed to delete');
-    }
-
-    this.logger.info(`Deleted registry of geography with id`, { id });
+  async delete(_id: string, _session: ISession): Promise<void> {
+    // Not implemented
   }
 
   async list(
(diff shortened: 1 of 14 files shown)


src/components/registry-of-geography/registry-of-geography.service.ts
```diff
@@ -0,0 +1,246 @@
...
+    return await this.db.sgUpdateProperties({
+      session,
+      object: RegistryOfGeography,
+      props: ['name', 'registryId'],
+      changes: input,
+      nodevar: 'registryOfGeography',
+    });
+  }
+
+  async delete(id: string, session: ISession): Promise<void> {
```
michaelmarshall, line 206: deletes are being refacored. this can be skipped for now. we won't use ACLs in the future.

test/registry-of-geography.e2e-spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        },
+      }
+    );
+    const updated = result.updateRegistryOfGeography.registryOfGeography;
+    expect(updated).toBeTruthy();
+    expect(updated.name.value).toBe(newName);
+  });
+
+  // Delete Registry Of Geography
+  it('delete registry of geography', async () => {
```
michaelmarshall, line 89: this can be skipped for now.

CarsonF (changes requested):
"Registry" is a list, and we aren't creating a list of lists. We need to identify these objects by a noun representing their singular item.

> The function of the Registry of Geography (ROG) is to document standardized codes for geopolitical areas.
https://hisregistries.org/rog/

_GeopoliticalArea_ might be a solution.

I also can't figure out what HIS is actually doing here. It looks like they just say that GENC is the standard, ....or basically ISO 3166 codes?...

This [CountryCodeCrossReference.pdf](https://hisregistries.org/wp-content/uploads/filebase/rog/CountryCodeCrossReference.pdf) seems to define a bunch of fields.
We should probably have another discussion on what we need to support here.

Online list found here https://www.iso.org/obp/ui/#search

zkhin:
I guess we would use the ISO2 or the two letter country codes from here https://www.cia.gov/library/publications/the-world-factbook/appendix/appendix-d.html#

CarsonF:
Ok after talking with everyone we decided that Location can just store an ISO 3166-3 code instead of referencing/manually managing this legacy standard. 😁 

Thanks for the effort @zkhin 😞

PR #1250, opened 2020-09-21
Own project cannot create engagement

frankyma:
remove the old permission method in the project service that was creating disconnected SG nodes.

remove creating `writerSG` and `readerSG` nodes for all services.


CarsonF (approved):
LGTM

PR #1252, opened 2020-09-22
Cluster/Cohort projects return only with 2 or more engagements

frankyma:
Rename the project filter to `clustersOrCohort` and have it filter on both types of engagements.

**This is a breaking change**, will need to update the Client side code.

### DIFF
--- a/src/components/project/dto/list-projects.dto.ts
+++ b/src/components/project/dto/list-projects.dto.ts
@@ -64,10 +64,9 @@ export abstract class ProjectFilters {
 
   @Field({
     nullable: true,
-    description:
-      'When a project works on more than one language it is a "cluster" project. It has more than one engagement',
+    description: 'Filter for projects with two or more engagements.',
   })
-  readonly clusters?: boolean;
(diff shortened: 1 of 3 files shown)


src/components/project/dto/list-projects.dto.ts
```diff
@@ -65,9 +65,9 @@ export abstract class ProjectFilters {
   @Field({
     nullable: true,
     description:
-      'When a project works on more than one language it is a "cluster" project. It has more than one engagement',
+      'When a project has more than one engagement.  Cluster = Traslation Projects and Cohort = Internship Projects',
   })
-  readonly clusters?: boolean;
+  readonly clusterOrCohort?: boolean;
```
CarsonF, line 70: This is the only place these nouns are used in the API, maybe something like `onlyMultipleEngagements` would be better...

frankyma: agree, added

CarsonF (approved):

PR #1258, opened 2020-09-23
Language sponsor start date and projects permissions

frankyma:
Pull language sponsorStartDate field in its own resolver.

Also moved acl permissions out of language.projects list read, hardcoded to `true` until Roles services are done

### DIFF
--- a/src/components/language/language.service.ts
+++ b/src/components/language/language.service.ts
@@ -3,10 +3,12 @@ import { node, relation } from 'cypher-query-builder';
 import { first, intersection } from 'lodash';
 import { DateTime } from 'luxon';
 import {
+  CalendarDate,
   DuplicateException,
   InputException,
   ISession,
   NotFoundException,
+  SecuredDate,
   ServerException,
   simpleSwitch,
 } from '../../common';
@@ -24,6 +26,7 @@ import {
 } from '../../core';
 import {
   calculateTotalAndPaginateList,
+  collect,
   permissionsOfNode,
   requestingUser,
 } from '../../core/database/query';
@@ -35,6 +38,7 @@ import {
   runListQuery,
   StandardReadResult,
 } from '../../core/database/results';
+import { EngagementService } from '../engagement';
 import {
   Location,
   LocationListInput,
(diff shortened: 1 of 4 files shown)


src/components/language/language.service.ts
```diff
@@ -684,8 +679,56 @@ export class LanguageService {
...
+
+    if (!result) {
+      throw new ServerException('Error fetching sponsorStartDate');
+    }
+
+    const engagments = await Promise.all(
+      result.engagementIds.map((engagementId) =>
+        this.engagementService.readOne(engagementId, session)
+      )
+    );
```
CarsonF, line 711: I think readOne can throw NotFound exception. We should catch that and return canRead=false, etc response

src/components/language/language.service.ts
```diff
@@ -684,8 +679,56 @@ export class LanguageService {
...
+
+    const dates = compact(
+      engagments.map((engagement) => engagement.startDate.value)
+    );
```
CarsonF, line 715: We also need to check every `startDate.canRead` and if some cannot be read also return a redacted response

src/components/language/language.service.ts
```diff
@@ -684,8 +679,56 @@ export class LanguageService {
...
+
+    if (!dates.length) {
+      return {
+        canRead: true,
+        canEdit: true,
```
CarsonF, line 720:
```suggestion
        canEdit: false,
```

src/components/language/language.service.ts
```diff
@@ -684,8 +679,56 @@ export class LanguageService {
...
+        value: undefined,
+      };
+    }
+
+    const minDate = CalendarDate.min(...dates);
+
+    return {
+      //TODO: use the upcoming roles service to determine permissions
+      canRead: true,
+      canEdit: true,
```
CarsonF, line 730:
```suggestion
        canEdit: false,
```

CarsonF (changes requested):
Lookin good @frankyma!

src/components/language/language.service.ts
frankyma: Added

src/components/language/language.service.ts
frankyma: checked, don't need the redacted response we just need canRead = false and the UI will show redacted

src/components/language/language.service.ts
frankyma: done

src/components/language/language.service.ts
frankyma: done

CarsonF (approved):
Thanks @frankyma! Looks great

PR #1267, opened 2020-09-25
Refactor location - 4

zkhin:


michaelmarshall (dismissed):
good job Zayar!

src/components/field-zone/field-zone.service.ts
```diff
@@ -0,0 +1,231 @@
...
+
+@Injectable()
+export class FieldZoneService {
+  private readonly securedProperties = {
+    name: true,
+    director: true,
+  };
+
+  constructor(
+    @Logger('location:service') private readonly logger: ILogger,
```
CarsonF, line 51:
```suggestion
    @Logger('field-zone:service') private readonly logger: ILogger,
```

src/components/funding-account/dto/create-funding-account.dto.ts
```diff
@@ -8,6 +8,9 @@ import { FundingAccount } from './funding-account.dto';
 export abstract class CreateFundingAccount {
   @NameField()
   readonly name: string;
+
+  @Field()
+  readonly accountNumber: string;
```
CarsonF, line 13: I think we should probably change this to a number/Integer.

src/components/field-region/field-region.service.ts
```diff
@@ -0,0 +1,260 @@
...
+@Injectable()
+export class FieldRegionService {
+  private readonly securedProperties = {
+    name: true,
+    director: true,
+    fieldZone: true,
+  };
+
+  constructor(
+    @Logger('location:service') private readonly logger: ILogger,
```
CarsonF, line 52:
```suggestion
    @Logger('field-region:service') private readonly logger: ILogger,
```

src/components/location/dto/create-location.dto.ts
```diff
@@ -1,83 +1,41 @@
...
-  @NameField()
-  readonly name: string;
+  @Field({ nullable: true })
+  readonly iso31663?: string;
 
-  @IdField()
-  readonly regionId: string;
-}
+  @NameField({ nullable: true })
+  readonly geographyName?: string;
```
CarsonF, line 23: This is just the name associated with the ISO-3166-3 code. I'd rather not manage input this separately. If we need to provide it we should look it up with the code. I'm comfortable with this happening later.

src/components/location/dto/location-type.enum.ts
```diff
@@ -0,0 +1,11 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum LocationType {
+  City = 'City',
+  County = 'Country',
```
CarsonF, line 5: Do you mean 
```suggestion
  County = 'County',
```
?

src/components/project/dto/create-project.dto.ts
```diff
@@ -21,10 +21,28 @@ export abstract class CreateProject {
...
     nullable: true,
   })
-  readonly locationId?: string;
+  readonly primaryLocationId?: string;
+
+  @IdField({
+    description: 'A non primary location ID',
+    nullable: true,
+  })
+  readonly nonPrimaryLocationId?: string;
```
CarsonF, line 33: Other locations need to be a list. A list of IDs here would be ok on create, but for updating we'd need separate add/remove mutations.

src/components/project/project.resolver.ts
```diff
@@ -144,6 +155,56 @@ export class ProjectResolver {
...
+  ): Promise<SecuredLocation> {
+    const { value: id, ...rest } = project.primaryLocation;
+    const value = id
+      ? await this.locationService.readOne(id, session)
+      : undefined;
+    return { value, ...rest };
+  }
+
+  @ResolveField(() => SecuredLocationList)
+  async nonPrimaryLocations(
```
CarsonF, line 171: How about `otherLocations`? I'd rather not define a property as the opposite of another property.

src/components/project/dto/project.dto.ts
```diff
@@ -49,8 +49,13 @@ class Project extends Resource {
   @Field(() => ProjectStatus)
   readonly status: ProjectStatus;
 
-  @Field()
-  readonly location: SecuredCountry;
+  readonly primaryLocation: Secured<string>;
+
+  readonly nonPrimaryLocation: Secured<string>;
```
CarsonF, line 54: It doesn't look like this is used since the resolver calls a custom method on the service and doesn't use any properties previously resolved on the Location object from the query.

src/components/project/dto/update-project.dto.ts
```diff
@@ -20,10 +20,28 @@ export abstract class UpdateProject {
...
     nullable: true,
   })
-  readonly locationId?: string;
+  readonly primaryLocationId?: string;
+
+  @IdField({
+    description: 'A non primary location ID',
+    nullable: true,
+  })
+  readonly nonPrimaryLocationId?: string;
```
CarsonF, line 32: See comment on create.

src/components/search/dto/search-results.dto.ts
```diff
@@ -33,6 +30,7 @@ const publicSearchable = {
   Story,
   LiteracyMaterial,
   Song,
+  Location,
```
CarsonF, line 33: Location is no longer an interface so it needs be removed below in `searchableAbstracts`

src/components/project/dto/create-project.dto.ts
CarsonF: I'd also be ok completely skipping this feature for now. It's new and not needed for MVP launch.

CarsonF (changes requested):
Great work @zkhin! Just a few points...

src/components/funding-account/dto/create-funding-account.dto.ts
```diff
@@ -10,7 +10,7 @@ export abstract class CreateFundingAccount {
   readonly name: string;
 
   @Field()
-  readonly accountNumber: string;
+  readonly accountNumber: number;
```
CarsonF: Need to mark the graphql type as an `Int`

src/components/funding-account/dto/update-funding-account.dto.ts
```diff
@@ -13,7 +13,7 @@ export abstract class UpdateFundingAccount {
   readonly name?: string;
 
   @Field({ nullable: true })
-  readonly accountNumber?: string;
+  readonly accountNumber?: number;
```
CarsonF, line 16: Same here

src/components/search/dto/search-results.dto.ts
```diff
@@ -30,7 +29,6 @@ const publicSearchable = {
   Story,
   LiteracyMaterial,
   Song,
-  Location,
```
CarsonF: This is still needed here. Sorry for the confusion. The removal on `searchableAbstracts` was the only thing I was trying to point out.

src/components/project/dto/update-project.dto.ts
```diff
@@ -25,11 +25,11 @@ export abstract class UpdateProject {
...
   readonly primaryLocationId?: string;
 
-  @IdField({
-    description: 'A non primary location ID',
+  @Field(() => [ID], {
+    description: 'Other location IDs',
     nullable: true,
   })
-  readonly nonPrimaryLocationId?: string;
+  readonly otherLocationIds?: string[];
```
CarsonF, line 32: This is different than how we do all other list relationships in the system. I was ok with it for create because there's never any existing relationships, but on update there's existing to think about. i.e. "Is this whole list? and if existing items are missing in this ID list do they need to be deleted?", etc. This is why we have separate mutations to create/delete these list relationships so we know explicitly that this single item needs to be created/deleted instead of having to diff a new list with the existing db state.

The easiest thing to do is just leave this property out of the update mutation right now. We won't implement it immediately on the UI anyways.

src/components/authorization/policies/policies.ts
```diff
@@ -165,7 +165,7 @@ const policies = [
     modifiedAt: Perm.ReadAndEdit,
     engagement: Perm.ReadAndEdit,
     primaryLocation: Perm.ReadAndEdit,
-    nonPrimaryLocation: Perm.ReadAndEdit,
+    otherLocations: Perm.ReadAndEdit,
```
CarsonF, line 168: @michaelmarshall Are you ok with this name change? Does it need to be reflected in DB schema doc?

CarsonF (changes requested):

src/components/authorization/policies/policies.ts
michaelmarshall: yes, it is updated. Thanks for making sure.

CarsonF (approved):
LGTM!

michaelmarshall (approved):

PR #1270, opened 2020-09-28
1242 role based access

### LINKED ISSUE #1242: Role based access control for projects
Using our security group and permission concept, we need to grant access to properties based on a user's role on that project. 

1. A user must be approved to hold a specific role for any project.
2. A user must be granted a role on a project to have access to a property on that project.
3. Only through a role shall a user have access to properties.

michaelmarshall:
closes #1242


CarsonF (approved):
LGMT! Thanks for finishing it off @zkhin!

PR #1271, opened 2020-09-28
Partner search

frankyma:
created `readOnePartnerByOrgId` that finds the partner connected to this org id.  The the hydrator calls `readOnePartnerByPartnerIdOrOrgId`, calls both `partner.readOnePartnerByOrgId` and `partner.readOne`.  This pattern allows us to keep `partner.readOne` unchanged and consistent with the other service readOnes.

### DIFF
--- a/src/components/partner/partner.service.ts
+++ b/src/components/partner/partner.service.ts
@@ -115,17 +115,6 @@ export class PartnerService {
     return await this.readOne(result.id, session);
   }
 
-  async readOnePartnerByPartnerIdOrOrgId(
-    id: string,
-    session: ISession
-  ): Promise<Partner> {
-    try {
-      return await this.readOne(id, session);
-    } catch {
-      return await this.readOnePartnerByOrgId(id, session);
-    }
-  }
-
   async readOnePartnerByOrgId(id: string, session: ISession): Promise<Partner> {
     this.logger.debug(`Read Partner by Org Id`, {
       id: id,
--- a/src/components/partner/partner.service.ts
+++ b/src/components/partner/partner.service.ts
@@ -115,8 +115,46 @@ export class PartnerService {
     return await this.readOne(result.id, session);
   }
 
+  async readOnePartnerByPartnerIdOrOrgId(
+    id: string,
+    session: ISession
+  ): Promise<Partner> {
+    try {
+      return await this.readOne(id, session);
+    } catch {
+      return await this.readOnePartnerByOrgId(id, session);
+    }
+  }
+
+  async readOnePartnerByOrgId(id: string, session: ISession): Promise<Partner> {
+    this.logger.debug(`Read Partner by Org Id`, {
+      id: id,
+      userId: session.userId,
+    });
+    const query = this.db
+
+      .query()
+      .match([node('node', 'Organization', { id: id })])
+      .match([
+        node('node'),
+        relation('in', '', 'organization', { active: true }),
+        node('partner', 'Partner'),
+      ])
+      .return({
+        partner: [{ id: 'partnerId' }],
+      })
+      .asResult<{
+        partnerId: string;
+      }>();
+    const result = await query.first();
+    if (!result)
+      throw new NotFoundException('No Partner Exists for this Org Id');
+
+    return await this.readOne(result.partnerId, session);
+  }
+
   async readOne(id: string, session: ISession): Promise<Partner> {
-    this.logger.debug(`Read Partner`, {
+    this.logger.debug(`Read Partner by Partner Id`, {
       id: id,
       userId: session.userId,
     });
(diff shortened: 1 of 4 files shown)


src/components/search/search.service.ts
```diff
@@ -146,12 +147,16 @@ export class SearchService {
...
+      try {
+        const obj = await hydrator(...args);
+        return {
+          ...obj,
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          __typename: type,
+        };
+      } catch {
+        return null;
+      }
```
CarsonF, line 159: Swallowing all errors here is not a good idea. What are we really trying to avoid? I assume it's when a query throws a not found error due to permissions. We should check for that explicitly.

CarsonF (changes requested):

src/components/search/search.service.ts
frankyma: syntactically how do I check which type of error this is thrown?

CarsonF: With an `instanceof` check

frankyma: Thanks, error caught

src/components/partner/partner.service.ts
```diff
@@ -115,8 +115,46 @@ export class PartnerService {
...
 
+  async readOnePartnerByPartnerIdOrOrgId(
+    id: string,
+    session: ISession
+  ): Promise<Partner> {
+    try {
+      return await this.readOne(id, session);
+    } catch {
+      return await this.readOnePartnerByOrgId(id, session);
+    }
```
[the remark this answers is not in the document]
frankyma: >When searching for a partner's name, the most common use case, every partner will have to wait for the partner findByID to fail before calling findByOrgId. 
The runtime goes from n to 2n, not horrible imo.  If searching for a partner's name is the most common use case I can move `readOnePartnerByOrgId` to the top.

>Furthermore we know in the search service which ID it is, so we could be smarter about this and in turn more performant.
Do we? I'm not certain on the cypher query. Say we search the string "hi", do we know if that was a match on an org name, or a direct Partner property?  I think either way after the `flatMap` we don't know if the id is an orgId or partnerId so we check both.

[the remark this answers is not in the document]
CarsonF: > The runtime goes from n to 2n, not horrible imo. If searching for a partner's name is the most common use case I can move `readOnePartnerByOrgId` to the top.

I'm just really sensitive to this because our query performance sucks right now. I wouldn't care as much if the queries were returning in milliseconds.

> > Furthermore we know in the search service which ID it is, so we could be smarter about this and in turn more performant.
> 
> Do we? I'm not certain on the cypher query. Say we search the string "hi", do we know if that was a match on an org name, or a direct Partner property? I think either way after the `flatMap` we don't know if the id is an orgId or partnerId so we check both.

The cypher query gives us back the type either Org or Partner and the ID. It's in the flatMap logic that we confuse the ID between both types. If we removed the flatMap there, and did it in the Organization hydrator we would know exactly which type the ID belonged too.

src/components/search/search.service.ts
```diff
@@ -123,12 +147,17 @@ export class SearchService {
...
+      try {
+        const obj = await hydrator(...args);
+        return {
+          ...obj,
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          __typename: type,
+        };
+      } catch (err) {
+        if (err instanceof NotFoundException) return null;
+        else throw new ServerException(`Error searching on ${type}`);
```
CarsonF, line 159:
```suggestion
        else throw new ServerException(`Error searching on ${type}`, err);
```

frankyma: added

src/components/partner/partner.service.ts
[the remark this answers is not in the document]
frankyma: putting thoughts down for our talk: 
So say we have a search query that's asking for `type: [Partner, Organization]` with a string "hi".  We get back a result `[{id: "nodeID123", type: "Organization"}]`.  We need to return this org and the partner associated with it.  We then send this id and the partner type to the Organization hydrator?  It won't have this pattern then 
```Organization: (...args) => this.orgs.readOne(...args),```
because it'll have to call both orgs.readOne and partner.readOneByOrgId and return an array.  The reason we had the flatMap was to try to preserve that hydrator consistency.

Also my biggest hesitation is there's too many edge cases that we might miss.  The cases of asking for Organization, Partner, or both on the search query, and each result being an Organization or Partner type.  That's 6 combinations and I don't think we've thought through each one carefully.  Our flatmap method captures all of them without too much of a performance hit.

CarsonF (approved):
LGTM! Thanks @frankyma!

PR #1283, opened 2020-10-01
Ceremony readOne fix type property

frankyma:
The Role refactor exposed 2 existing issues

1. On create LanguageEngagement, the returned engagement doesn't include the ceremonyId that was created by the event handler.  This was never found because we never needed the ceremony after creating a langEngagement

2. On Ceremony readOne, the type value was returned from the parseSecuredProperties function.  But now with the new roles refactor the type property doesn't get a Permission node because according to the dto it's not a secured property, so now it's values are always undefined.  This wasn't found because we manually gave it a Permission Node in the old create process

### DIFF
--- a/src/components/ceremony/ceremony.service.ts
+++ b/src/components/ceremony/ceremony.service.ts
@@ -27,6 +27,7 @@ import {
 import {
   DbPropsOfDto,
   parseBaseNodeProperties,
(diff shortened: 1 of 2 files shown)


CarsonF (approved):
Looks good!

Issue #1275, opened 2020-09-29
Language.hasExternalFirstScripture


CarsonF:
A boolean field, defaulting to false.
description:
> Indicates that this language's first scripture was produced outside of CORD.

It should only be allowed to be set to true if the language has no engagements that have firstScripture=true.

Blocks https://github.com/SeedCompany/cord-field/issues/39 #1276

PR #1284, opened 2020-10-01
Add new hasExternalFirstScripture to language and business logic

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1275

### DIFF
--- a/src/components/language/language.service.ts
+++ b/src/components/language/language.service.ts
@@ -363,15 +363,14 @@ export class LanguageService {
     { ethnologue: newEthnologue, ...input }: UpdateLanguage,
     session: ISession
   ): Promise<Language> {
+    if (input.hasExternalFirstScripture) {
+      await this.verifyExternalFirstScripture(input.id);
+    }
+
     const { ethnologue: oldEthnologue, ...language } = await this.readOne(
       input.id,
       session
     );
-
-    if (input.hasExternalFirstScripture) {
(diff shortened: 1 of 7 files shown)


michaelmarshall (approved):

src/components/language/language.service.ts
```diff
@@ -361,6 +368,11 @@ export class LanguageService {
       session
     );
 
+    if (input.hasExternalFirstScripture) {
+      // Check if the language has no engagements that have firstScripture=true.
+      await this.checkLEFirstScripture(input.id);
```
CarsonF, line 373: - Maybe this comment could be moved to the doc of the method?
- Maybe method `verifyExternalFirstScripture`? I'm not a huge fan of the language engagement abreviation.
- By moving this above the readOne it would save a DB call if it was going to fail.

CarsonF (changes requested):
The verification is needed on create as well.
Also just some nitpicks.

src/components/language/language.service.ts
zkhin: > The verification is needed on create as well.
> Also just some nitpicks.

@CarsonF @michaelmarshall I am not sure this is still needed on creating because the new language will not have any engagements and it will always pass the verification. 🤔

CarsonF: That's an excellent point. It must be Friday lol

CarsonF (approved):
Thanks @zkhin!

PR #1287, opened 2020-10-01
Add book abbreviations

blakekrammes:
Some abbreviations found in v2 data weren't accounted for.

### DIFF
--- a/src/components/scripture/books.ts
+++ b/src/components/scripture/books.ts
@@ -261,7 +261,7 @@ const books = [
     ],
   },
   {
-    names: 'Judges Jg Jud Jdg Ju Jdgs Judg'.split(' '),
+    names: 'Judges Jg Jdg Ju Jdgs Judg'.split(' '),
     chapters: [
       36,
       23,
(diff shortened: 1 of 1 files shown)


src/components/scripture/books.ts
```diff
@@ -1206,7 +1206,7 @@ const books = [
     chapters: [14],
   },
   {
-    names: 'Jude Jude'.split(' '),
+    names: 'Jude Jud Jude'.split(' '),
```
CarsonF, line 1209: Not sure why Jude was listed twice here but the second one is not needed. `Jud` conflicts with _Judges_.

CarsonF (changes requested):

src/components/scripture/books.ts
blakekrammes: Was just thinking that. Jud should be judges. My bad.

blakekrammes (commented):
I'm thinking there should be a 3-letter code for Jude though... lemme see

blakekrammes:
I think we use jdg for judges...

CarsonF:
We can do some mapping in the migration code as well if needed

blakekrammes:
I don't think we should need to. I checked a couple examples and `jud` refers to Jude while `jdg` refers to Judges. So `jud` should be removed from the list of names for Judges.

CarsonF:
Hmmm Googling "Jud 1" gives 5 results for Jude and 2 for Judges. I'm not sure what the best answer is here.

blakekrammes:
My vote is to do it as before. :)

CarsonF (approved):
Yeah seems like the abbrivation standards are all over the place. Maybe it's a good thing that we normalize to the actual book name now.

Alright let's move forward with this until we hit a reason why it doesn't work :+1:

PR #1294, opened 2020-10-02
612 calculate department id

thomas-adev:
Added handler to calculate and set project departmentId when project step is updated to Pending Finance Confirmation.

### DIFF
--- a/src/components/project/handlers/set-department-id.handler.ts
+++ b/src/components/project/handlers/set-department-id.handler.ts
@@ -18,7 +18,16 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
     }
 
     try {
-      await this.assignDepartmentIdForProject(event.updated);
+      const departmentId = await this.assignDepartmentIdForProject(
+        event.updated
+      );
+      event.updated = {
+        ...event.updated,
+        departmentId: {
+          ...event.updated.departmentId,
+          value: event.updated.departmentId.canRead ? departmentId : undefined,
+        },
+      };
     } catch (exception) {
       throw new ServerException(
         'Could not set departmentId on project',
@@ -56,6 +65,7 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
     if (!res) {
       throw new ServerException('Unable to assign department ID');
     }
+    return res.departmentId;
   }
 
   private async getDepartmentIdPrefixForProject(project: Project) {
--- a/src/components/project/handlers/set-department-id.handler.ts
+++ b/src/components/project/handlers/set-department-id.handler.ts
@@ -42,18 +42,29 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
   }
 
   private async getDepartmentIdPrefixForProject(project: Project) {
-    // one of (1,2,3,4,5,7)
-    // TODO: fetch appropriate prefix via location.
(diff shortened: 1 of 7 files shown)


src/components/project/handlers/set-department-id.handler.ts
```diff
@@ -0,0 +1,85 @@
...
+export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
+  constructor(
+    private readonly db: DatabaseService,
+    @Logger('project:set-department-id') private readonly logger: ILogger
+  ) {}
+
+  async handle(event: SubscribedEvent) {
+    this.logger.debug('Project mutation, set department id', {
+      ...event,
+      event: event.constructor.name,
```
CarsonF, line 24: This was really only on that one handler because it was handling so many different types of events. It will just be a constant on this one though so it doesn't provide relevant info.

src/components/project/handlers/set-department-id.handler.ts
```diff
@@ -0,0 +1,85 @@
...
+      .query()
+      .raw(
+        `
+        MATCH (:Project {id: $projectId})-[:primaryLocation {active: true}]-()-[:fundingAccount {active: true}]-()-[:accountNumber {active:true}]-(accountNumberNode:Property)
+        RETURN accountNumberNode.value
+        `,
+        { projectId: project.id }
+      )
+      .run();
+    return accountNumbers[0]['accountNumberNode.value'];
```
CarsonF, line 56: We should probably check if a result wasn't found and throw an exception here.
`.first()` could also be used as a shortcut (instead of `.run()`.

src/components/project/project.service.ts
```diff
@@ -483,10 +490,13 @@ export class ProjectService {
...
-      new ProjectUpdatedEvent(result, currentProject, input, session)
+    const projectUpdatedEvent = new ProjectUpdatedEvent(
+      result,
+      currentProject,
+      input,
+      session
     );
+    await this.eventBus.publish(projectUpdatedEvent);
 
     return await this.readOne(input.id, session);
```
CarsonF, line 501: If `sgUpdateProperties` returns the updated object then why are we calling readOne again?

I know the new handler currently needs it because it update the object returned but I'd rather have it modify the event object like we are doing in other places.

If there are other reasons for this extra read call we should fix them too.
If the read call is because the `result` from `sgUpdatedProperties` doesn't have the correct info then we should just ignore it and do the readOne before publishing the event.

CarsonF (changes requested):
Just a couple things from me. I'll let @michaelmarshall review the db queries on this one.

src/components/project/project.service.ts
thomas-adev: So is your preference to have these changes reversed: https://github.com/SeedCompany/cord-api-v3/pull/1294/commits/d444fe1572ed72ae8b4a7a023018b3f2a9ec3a88

CarsonF: Well kinda. The handler is just changing the `departmentId` so there's no need to call `readOne` again. Just "modify"  (spread the object into a new one) the project object and adjust the one property which you should already have the new value for. Said another way, you already have the fresh data to update the stale parts of the object so there's no need to ask for it again from the db.

CarsonF (approved):

Issue #1291, opened 2020-10-02
Refactor tests to use Powers API


michaelmarshall:
The new Powers API changes how we authorize users to do things. To not break tests, all users where given all powers:
![image](https://user-images.githubusercontent.com/1165460/94882858-bb298800-042e-11eb-84ec-68ae9706ab49.png)
A few things need to happen now:
1. Remove the code that grants users all powers.
2. Expose the Powers API to the frontend by creating a resolver for it.
3. Use the new API in the test suite (when applicable) by logging in as the root user and granting another user the power needed. This is only needed for powers that users don't get by default, like CreateLanguage or CreateOrganization. Some tests won't need refactored, some will.

zkhin:
@michaelmarshall What is the `GrantPower` in `Powers` type? I think that's only for the root user for granting powers to other normal users.

michaelmarshall:
The root will grant the `GrantPower` power to other users so that they, in turn, can grant powers to their people. Otherwise, no other admins would be able to delegate creating a language for example.

PR #1300, opened 2020-10-05
Powers API

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1291
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1302

### DIFF
--- a/src/components/authorization/authorization.resolver.ts
+++ b/src/components/authorization/authorization.resolver.ts
@@ -1,0 +1,48 @@
+import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
+import { node } from 'cypher-query-builder';
+import { IdArg, ISession, Session } from '../../common';
+import { DatabaseService } from '../../core';
+import { Powers } from '../authorization/dto/powers';
+import { AuthorizationService } from './authorization.service';
+
+@Resolver()
+export class AuthorizationResolver {
+  constructor(
+    private readonly db: DatabaseService,
+    private readonly authorizationService: AuthorizationService
+  ) {}
+
+  @Query(() => [Powers])
+  async powers(@Session() session: ISession): Promise<Powers[]> {
+    if (!session.userId) {
+      return [];
+    }
+    const result = await this.db
+      .query()
+      .match([node('user', 'User', { id: session.userId })])
+      .return('user.powers as powers')
+      .first();
+
+    if (!result) {
+      return [];
+    }
+    return result.powers as Powers[];
+  }
+
+  @Mutation(() => Boolean)
+  async grantPower(
+    @Session() session: ISession,
+    @IdArg() id: string,
+    @Args({ name: 'power', type: () => Powers }) power: Powers
+  ): Promise<boolean> {
+    if (!session.userId) {
+      return false;
+    }
+
+    const requestingUserPowers = await this.powers(session);
+    if (requestingUserPowers.includes(Powers.GrantPower)) {
+      return this.authorizationService.grantPower(power, id);
+    }
+    return false;
+  }
+}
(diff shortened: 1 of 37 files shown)


src/components/authorization/dto/powers.ts
```diff
@@ -32,3 +34,5 @@ export enum Powers {
   CreateZone = 'CreateZone',
   GrantPower = 'GrantPower',
 }
+
+registerEnumType(Powers, { name: 'Powers' });
```
CarsonF, line 38: This should be named as a singular.

src/components/authorization/authorization.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  ): Promise<boolean> {
+    if (!session.userId) {
+      return false;
+    }
+
+    const requestingUserPowers = await this.powers(session);
+    if (requestingUserPowers.includes(Powers.GrantPower)) {
+      return this.authorizationService.grantPower(power, id);
+    }
+    return false;
```
CarsonF, line 46: We don't use boolean returns to indicate the success of a mutation. If there is a failure, then an error should be thrown.

src/components/authorization/authorization.resolver.ts
```diff
@@ -0,0 +1,48 @@
+import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
+import { node } from 'cypher-query-builder';
+import { IdArg, ISession, Session } from '../../common';
+import { DatabaseService } from '../../core';
+import { Powers } from '../authorization/dto/powers';
+import { AuthorizationService } from './authorization.service';
+
+@Resolver()
+export class AuthorizationResolver {
```
CarsonF, line 9: The logic in this file should be moved to the `AuthorizationService`.

src/components/authorization/authorization.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor(
+    private readonly db: DatabaseService,
+    private readonly authorizationService: AuthorizationService
+  ) {}
+
+  @Query(() => [Powers])
+  async powers(@Session() session: ISession): Promise<Powers[]> {
```
CarsonF, line 16: This really fulfills the TODO here
https://github.com/SeedCompany/cord-api-v3/blob/f3bd4b45e4b44975db656abede6ac1772857db50/src/components/authentication/authentication.dto.ts#L43

If we want to keep it here that's fine, but we should also add to LoginOutput and SessionOutput as it makes more sense there from a consumption perspective.

src/components/authorization/authorization.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+      .first();
+
+    if (!result) {
+      return [];
+    }
+    return result.powers as Powers[];
+  }
+
+  @Mutation(() => Boolean)
+  async grantPower(
```
CarsonF, line 33: How is power removed/ungranted?

test/partnership.e2e-spec.ts
```diff
@@ -23,11 +27,15 @@ import { createPartnership } from './utility/create-partnership';
...
   let project: Raw<Project>;
+  const password: string = faker.internet.password();
 
   beforeAll(async () => {
     app = await createTestApp();
     await createSession(app);
-    await createUser(app);
+    const user = await createUser(app, { password });
+    await grantPower(app, user.id, Powers.CreateOrganization);
+    await login(app, { email: user.email.value, password });
```
CarsonF, line 37: Since this use case is so common it should probably be abstracted and would be an opportunity to stop using the deprecated `createUser` function. It could be called `registerUserWithPower`.

CarsonF (changes requested):

michaelmarshall:
good job @zkhin !

PR #1304, opened 2020-10-05
Cannot add mentor to internship engagement

frankyma:
The old update mentor only worked a mentor already existed because it used a hard match.  Now using optional match to set the old to inactive.

Same pattern for location

### DIFF
--- a/test/engagement.e2e-spec.ts
+++ b/test/engagement.e2e-spec.ts
@@ -18,6 +18,7 @@ import { Project, ProjectType } from '../src/components/project';
 import { User } from '../src/components/user';
 import {
   createInternshipEngagement,
+  createInternshipEngagementWithMinimumValues,
   createLanguage,
   createLanguageEngagement,
   createLocation,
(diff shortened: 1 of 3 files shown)


test/engagement.e2e-spec.ts
```diff
@@ -357,14 +358,18 @@ describe('Engagement e2e', () => {
     expect(updated.status).toBe(EngagementStatus.Rejected);
   });
 
-  it('updates internship engagement', async () => {
+  it.only('updates internship engagement', async () => {
```
CarsonF, line 361:
```suggestion
  it('updates internship engagement', async () => {
```

frankyma: ok making a mental note for myself to always double check this 😬

CarsonF: We need a lint rule lol

frankyma: There's a few in the market. https://www.npmjs.com/package/eslint-plugin-no-only-tests?

michaelmarshall (approved):

Issue #1276, opened 2020-09-29
LanguageEngagement.firstScripture defaults & validation


CarsonF:
- On create or update, if `firstScripture` is true, validate that the engagement is the only engagement for the language that has `firstScripture=true` AND that the language doesn't have `hasExternalFirstScripture=true`

Blocked by #1275 
Related to https://github.com/SeedCompany/cord-field/issues/39

PR #1310, opened 2020-10-06
Verify engagement first scripture on create/update

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1276

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -1,8 +1,8 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
 import { node, Query, relation } from 'cypher-query-builder';
-import { cloneDeep } from 'lodash';
 import { DateTime } from 'luxon';
 import { generate } from 'shortid';
+import { MergeExclusive } from 'type-fest';
 import {
   DuplicateException,
   InputException,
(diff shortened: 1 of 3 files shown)


src/components/engagement/engagement.service.ts
```diff
@@ -162,6 +163,78 @@ export class EngagementService {
...
   }
 
+  /**
+   * if firstScripture is true, validate that the engagement
+   * is the only engagement for the language that has firstScripture=true
+   * that the language doesn't have hasExternalFirstScripture=true
+   */
+  protected async verifyFirstScripture(
+    engagementId?: string,
+    languageId?: string
```
CarsonF, line 173: Is one of these always required?

If so,
```tsx
verifyFirstScripture({ engagementId, languageId }: MergeExclusive<{ engagementId: string }, { languageId: string }>) {
```
Would be a better way to define it as it enforces that in TS and doesn't look awkward for one of the cases:
```tsx
await this.verifyFirstScripture({ languageId });
await this.verifyFirstScripture({ engagementId });
```

CarsonF (changes requested):

CarsonF (approved):
Sweet, thanks @zkhin!

PR #1312, opened 2020-10-06
Sort project sensitivity

frankyma:
Used to sort by the sensitivity words alphabetically.  Now use CASE to map it to a numerical value to be sorted

### DIFF
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -526,7 +526,15 @@ export class ProjectService {
     const label = `${filter.type ?? ''}Project`;
     const projectSortMap: Partial<Record<typeof input.sort, string>> = {
       name: 'lower(prop.value)',
+      sensitivity: 'sensitivityValue',
     };
+
+    const sensitivityCase = `case prop.value 
+    when 'High' then 3
+    when 'Medium' then 2
+    when 'Low' then 1
+    end as sensitivityValue`;
+
     const sortBy = projectSortMap[input.sort] ?? 'prop.value';
     const query = this.db
       .query()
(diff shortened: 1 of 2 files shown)


test/project.e2e-spec.ts
```diff
@@ -507,6 +507,65 @@ describe('Project e2e', () => {
     );
   });
 
+  it.only('List of projects sorted by Sensitivity', async () => {
```
CarsonF, line 510:
```suggestion
  it('List of projects sorted by Sensitivity', async () => {
```

src/components/project/project.service.ts
```diff
@@ -526,7 +526,15 @@ export class ProjectService {
...
     const projectSortMap: Partial<Record<typeof input.sort, string>> = {
       name: 'lower(prop.value)',
+      sensitivity: 'sensitivityValue',
     };
+
+    const sensitivityCase = `case prop.value 
+    when 'High' then 3
+    when 'Medium' then 2
+    when 'Low' then 1
+    end as sensitivityValue`;
```
CarsonF, line 536: Can a `case` expression not be used in an `orderby` statement?

CarsonF (changes requested):

src/components/project/project.service.ts
frankyma: not that i've seen, it's usually inside a RETURN or a WITH in the examples.  I played around with it in ORDERBY and couldn't get it to work.

test/project.e2e-spec.ts
frankyma: fixed

frankyma: fixed

src/components/project/project.service.ts
```diff
@@ -537,10 +545,13 @@ export class ProjectService {
         q
           .match([
             node('node'),
-            relation('out', '', sort),
+            relation('out', '', sort, { active: true }),
```
CarsonF, line 548: Is this necessary? If so it should be applied to every list

michaelmarshall: yes. When we traverse over property relationships we must always test for active true in order to return only non-deleted properties.

Issue #982, opened 2020-08-18
Fields for Partners


sethmcknight:
Add these fields to `Partner`:
- `pmcEntityCode` [string - Three capital letters (e.g. AAA)]
- `globalInnovationsClient` [boolean]
- `active` [boolean]
  > Active represents an organization we are currently working with in some capacity or have plans to work with in the future.
Inactive represents an organization that we have worked with or had plans to work with in the past but are no longer working with and have no future plans to work with.
  - Add some form of this ^ to the gql doc for the field
- `types` [enum list] [Funding, Managing, Technical, Resource, Impact] (#1168)
- `financialReportingTypes` [enum list] (if a Managing Partner) [Funded or Field Engaged]
- `address` [text field for now]
- `contact` [link to a person]

*REMOVE*
- IDPName
- Integrated Flag
- SF Owner ID
@blakekrammes @michaelmarshall note this removal for migration

sethmcknight:
blocked by #1196

sethmcknight:
blocks https://github.com/SeedCompany/cord-field/issues/372

PR #1318, opened 2020-10-07
Add new fields to partner module

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/982


CarsonF (changes requested):
`financialReportingType` needs to be `financialReportingTypes` (plural)

The Partnership.financialReportingType is a subset of its Partner's financialReportingTypes.

PR #1319, opened 2020-10-07
Use fiscalYears function for budget records

blakekrammes:
This fixes a bug where the fiscal year is calculated as being the same as the year of the date, which is not the case.

### DIFF
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -1,4 +1,4 @@
-import { ISession, ServerException } from '../../../common';
+import { fiscalYears, ISession, ServerException } from '../../../common';
 import {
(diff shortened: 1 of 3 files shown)


src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
```diff
@@ -101,13 +101,10 @@ export class SyncBudgetRecordsToFundingPartners
     } else {
       // TODO: mous starting at midnight on Jan 1st might cause issues with UTC vs local issues here.
       // double check how year should be calculated
```
CarsonF, line 103: Can you remove this TODO as well? This addresses it.

blakekrammes: TODONE

CarsonF (approved):

PR #1320, opened 2020-10-07
Refactor delete

frankyma:
Following the steps defined in the ticket description.  To use `deleteNodeNew` will need to pass in all the baseNode labels, and a dictionary of all of its unique property relations and those property node labels.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -23,7 +23,7 @@ rules:
       filter: __typename
       format: ~
     - selector: property
-      format: [camelCase, PascalCase]
+      format: [camelCase, PascalCase, snake_case]
       leadingUnderscore: forbid
     - selector: enumMember
       format: [PascalCase, UPPER_CASE]
--- a/src/components/authorization/dto/powers.ts
+++ b/src/components/authorization/dto/powers.ts
@@ -32,6 +32,7 @@ export enum Powers {
   CreateUnavailability = 'CreateUnavailability',
   CreateUser = 'CreateUser',
   CreateZone = 'CreateZone',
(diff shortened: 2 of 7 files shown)


michaelmarshall (changes requested):
Looks good so far. Now that we've gotten here, please add `DeleteX` members to the `Powers` enum. Then, in the delete function, check if a user has that power by calling the authorization service function. For now you may want to grant all users the power you are creating by going to the user service create function and adding the power to the list that all users will get. Once that is done, unskip the delete test for each object you have covered. Please add this list to your ticket too. Thanks Frank!

michaelmarshall:
By `DeleteX` I mean `DeleteLanguage`, `DeleteUser`, etc. Whatever the base node is.

frankyma:
Pushed the changes to add the check `Powers`.  I only added the `DeleteLanguage` power to keep this PR small, once this gets merged we can do the rest in a batch

michaelmarshall (approved):

.eslintrc.yml
```diff
@@ -23,7 +23,7 @@ rules:
       filter: __typename
       format: ~
     - selector: property
-      format: [camelCase, PascalCase]
+      format: [camelCase, PascalCase, snake_case]
```
CarsonF, line 26: Why was this needed?

frankyma: This showed up as a warning: https://github.com/SeedCompany/cord-api-v3/pull/1320/files#diff-3db627b34674b6a447753880e259bffb58dfe6cd8f22f35623415bce6a93b455R911

And I though if snake case will be in our database from now on might as well add the rule.

PR #1321, opened 2020-10-07
fix(project.service.ts): use new delete

frankyma:
Testing the new delete refactor on proejct


michaelmarshall (changes requested):
great, please review my comments on the language delete refactor. we'll want the tests to be brought back for each base node.

michaelmarshall (approved):

PR #1323, opened 2020-10-07
Validate iso31663 field

frankyma:


src/components/location/dto/create-location.dto.ts
```diff
@@ -16,7 +16,10 @@ export abstract class CreateLocation {
   @Field(() => Sensitivity)
   readonly sensitivity: Sensitivity;
 
-  @Field({ nullable: true })
+  @Field({ nullable: true, description: 'Must be 3 uppercase characters' })
+  @IsAlpha()
+  @IsUppercase()
+  @Length(3, 3)
```
CarsonF, line 22: Aren't they typically lowercase? It might just be better to accept either case and transform it to one or the other.

CarsonF (changes requested):
Can you apply to UpdateLocation as well?

src/components/location/dto/create-location.dto.ts
frankyma: Honestly I'm not sure.  @sethmcknight can you confirm what validation we need on this field?  When I google iso31663 the first wiki page actually says 4 letters: https://en.wikipedia.org/wiki/ISO_3166-3

CarsonF: Ok I see the confusion. We are not actually referencing the ISO 3166-3 codes here, which are the deleted codes from the ISO 3166-1 database. Instead we are referencing ISO 3166-1, specifically the alpha 3 code.

CarsonF: I've seen them written both ways (upper/lower). We should just normalize to one of those instead of requiring it to be one of the cases.

frankyma: Changed to take upper and lower but saved as upper case on both create and update.

src/components/location/location.service.ts
```diff
@@ -222,11 +222,16 @@ export class LocationService {
   async update(input: UpdateLocation, session: ISession): Promise<Location> {
     const location = await this.readOne(input.id, session);
 
+    const inputNormalized = {
+      ...input,
+      iso31663: input.iso31663?.toUpperCase(),
```
CarsonF, line 227: Would you mind putting this as an a transform decorator? Sorry to be mum on that previously.

```tsx
@Transform(lodash.toUpper)
```

frankyma: No prob, I don't know much about decorators so good to know

CarsonF (approved):
Thanks @frankyma!

PR #1330, opened 2020-10-08
fix(engagement.e2e-spec.ts): fix import

frankyma:
Fixing a missing import in the test, not sure how this slipped through the first time.

Eslint has been acting weird for me lately on the test files, anyone else?

### DIFF
--- a/test/engagement.e2e-spec.ts
+++ b/test/engagement.e2e-spec.ts
@@ -30,6 +30,7 @@ import {
   fragments,
   getUserFromSession,
   Raw,
+  registerUser,
   registerUserWithPower,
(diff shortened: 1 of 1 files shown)


CarsonF (approved):
It's because your branch that passed was based on a commit prior to #1300. CI runs on your branch (for PRs), not the result of your branch being merged into master. #1300 got merged before yours so this slipped through the checks.

GitHub has an option to "require up-to-date branches" to prevent this. But I feel like it isn't worth the effort for the edge cases that are easily caught when they rarely occur.

CarsonF:
Merging anyways because failures are unrelated.

Issue #1336, opened 2020-10-08
Rename grantPower.id to userId


CarsonF:
Rename `id` argument of `grantPower` mutation to `userId`

PR #1337, opened 2020-10-08
Rename mutation grantPower/removePower arg id to userId

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1336


CarsonF (approved):
Thanks @zkhin!

Issue #1259, opened 2020-09-23
deletePartnership > Getting "Could not synchronize budget records" error


iPiyushPatel:
**Describe the bug**
deletePartnership > Getting ` Could not synchronize budget records` error.

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
Created partnership should be deleted.

**Actual Result**
 Getting ` Could not synchronize budget records` error
See video -  https://drive.google.com/file/d/1G7IviC_Tx5EFWg5D8Q3RAeMewfzTD9RZ/view?usp=drivesdk

**GraphQL Query:(CreatePartnership)**
![image](https://user-images.githubusercontent.com/5212585/94015175-25588200-fdca-11ea-877f-3b905b0d9ad0.png)

**GraphQL Query:(DeletePartnership)**
![image](https://user-images.githubusercontent.com/5212585/94015034-f3dfb680-fdc9-11ea-9280-977d26e7bfe6.png)

**GraphQL response**

```
{
    "errors": [
        {
            "message": "Could not synchronize budget records",
            "extensions": {
                "codes": [
                    "Server"
                ],
                "status": 500,
                "code": "Server",
                "exception": {
                    "message": "Could not synchronize budget records",
                    "stacktrace": [
                        "at SyncBudgetRecordsToFundingPartners.handle (components\\budget\\handlers\\sync-budget-records-to-funding-partners.handler.ts:251:13)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "deletePartnership"
            ]
        }
```

**Query and Input File**
[DeletePartnershipError.txt](https://github.com/SeedCompany/cord-api-v3/files/5268230/DeletePartnershipError.txt)

CordAPIV3# f1e92283

PR #1339, opened 2020-10-09
Fix delete partnership bug related budget records

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1259

### DIFF
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -11,8 +11,8 @@ import { PartnershipService } from '../../partnership';
 import { Partnership } from '../../partnership/dto';
 import {
   PartnershipCreatedEvent,
-  PartnershipDeletedEvent,
   PartnershipUpdatedEvent,
+  PartnershipWillDeleteEvent,
 } from '../../partnership/events';
 import { Project } from '../../project/dto';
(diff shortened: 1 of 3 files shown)


CarsonF (commented):
This should not care what event listeners are attached to this event.
Is it not possible to update the logic in the sync budget records listener?

If not (which I would be hard pressed to believe), or if it's too hard (more likely) we can make this change, but outside of the try/catch and without the comment and we should probably rename it to something like `PartnershipWillDelete`.

zkhin:
> This should not care what event listeners are attached to this event.
> Is it not possible to update the logic in the sync budget records listener?
> 
> If not (which I would be hard pressed to believe), or if it's too hard (more likely) we can make this change, but outside of the try/catch and without the comment and we should probably rename it to something like `PartnershipWillDelete`.

This part(https://github.com/SeedCompany/cord-api-v3/blob/master/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts#L44~L55) is the source of this bug. Because partnership is already deleted and it cannot get any budget records related to that partnership. So I think we should deactivate the budget records before the partnership is deleted. So can I change the event name to `PartnershipWillDeleteEvent`?

CarsonF:
Ok that does sound like the easest solution. So we are just "un-hacking" it a bit :wink:

CarsonF (approved):
Thanks @zkhin!

PR #1341, opened 2020-10-10
DTOs canDelete

### LINKED ISSUE #1274: DTOs canDelete
All DTOs need a `canDelete` boolean field indicating that the current user/requestor has permission to delete the object.
For now the implementation can just stub this to `true` everywhere, but we'll need to go back and do it correctly eventually.

glassblowerscat:
Closes #1274 

Update all definitions and logic for secured properties to use a `canDelete` property, set for moment to always be `true`

### DIFF
--- a/src/common/secured-list.ts
+++ b/src/common/secured-list.ts
@@ -23,6 +23,11 @@ export function SecuredList<Type, ListItem = Type>(
       description: `Whether the current user can add items to this list via the appropriate mutation`,
     })
     readonly canCreate: boolean;
+
+    @Field({
+      description: `Whether the current user can remove items from this list via the appropriate mutation`,
+    })
+    readonly canDelete: boolean;
   }
 
   return SecuredListClass;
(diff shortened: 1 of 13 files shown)


CarsonF:
We've got 3 types of data. i.e.:
1. A User
2. User's unavailability object
3. Its start date

This works great for sub-relationships (No. 2).

It's meaningless for scalar properties (No. 3). I'd be fine to keep the property for consistency, but I'd also be ok removing it where it doesn't make sense.
Is there a meaning we can give canDelete for these that's different from canEdit? Maybe canDelete means it's nullable? It seems like a stretch.

What's still not covered here is No. 1. How do we communicate delete-ability on root-level resources/base nodes? These aren't wrapped in secured objects some/most of the time. I think the only option is a property (or set of properties? futureproofing?) on the object.

Thoughts?

CarsonF (changes requested):
Ok after talking on zoom, we decided to move `canDelete` to a property on `Resource` interface. @glassblowerscat is going to see if this is easy to stub on returns for now, and if not then we are just going to wait on this until the DB can support it for real.

CarsonF:
Finished off in #1421

PR #1364, opened 2020-10-15
Refactor add/remove/list locations in language

zkhin:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3655346047) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/language/language.service.ts
+++ b/src/components/language/language.service.ts
@@ -519,11 +519,6 @@ export class LanguageService {
       })
       .run();
 
-    const canCreateLocation = await this.authorizationService.checkPower(
-      Powers.CreateLocation,
-      session.userId
-    );
-
     const items = await Promise.all(
       result.map(
         async (location): Promise<Location> => {
@@ -536,8 +531,8 @@ export class LanguageService {
       items: items,
       total: items.length,
       hasMore: false,
-      canCreate: canCreateLocation,
-      canRead: canCreateLocation,
+      canCreate: true, // TODO
+      canRead: true, // TODO
     };
   }
 
--- a/src/components/authorization/roles/consultant-manager.role.ts
+++ b/src/components/authorization/roles/consultant-manager.role.ts
@@ -163,6 +163,7 @@ export const ConsultantManager = new DbRole({
         { propertyName: 'name',                       permission: { read, write, }, },
         { propertyName: 'sensitivity',                permission: { read, write, }, },
         { propertyName: 'hasExternalFirstScripture',  permission: { read, write, }, },
+        { propertyName: 'locations',                  permission: { read, write, }, },
     ]}),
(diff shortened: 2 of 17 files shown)


src/components/language/language.service.ts
```diff
@@ -555,8 +536,8 @@ export class LanguageService {
       items: items,
       total: items.length,
       hasMore: false,
-      canCreate: !!permission?.canReadLocationCreate,
-      canRead: !!permission?.canReadLocationRead,
+      canCreate: canCreateLocation,
+      canRead: canCreateLocation,
```
michaelmarshall, line 540: @CarsonF are you fine with doing it like this for now?

michaelmarshall (approved):

src/components/language/language.service.ts
CarsonF: Mmm I don't think so. I think it's important for most people to be able to read and I imagine not many people will be able to create/edit these locations

zkhin: @michaelmarshall @CarsonF so can I just set the `canRead` to `true` and remove the `CreateLocation` power while creating a user? and the `CreateLocation` power can be granted from root user like `Organization` and `Language`.

CarsonF: Hmm `canCreate` really means `canAdd` in this case. "Can the current user add a location to this specific language?"

The `CreateLocation` power is completely separate. That should only apply when you are creating a location.

michaelmarshall: @CarsonF @sethmcknight We need the business rule to determine if a user can add a location to a language. Thoughts?

CarsonF: Well maybe. "Business rule" is really vague. Typically when I've said it in the past I've meant event handlers that have "reaction" logic - like syncing budget records. In my mind this just needs to be in the authorization system some how.

CarsonF:
@zkhin I still think checking the "create location" power here is wrong, as it's a completely separate action. Maybe `canCreate` should just be hardcoded to true for now as well?

zkhin:
> @zkhin I still think checking the "create location" power here is wrong, as it's a completely separate action. Maybe `canCreate` should just be hardcoded to true for now as well?

Okay, I will

CarsonF (approved):

CarsonF:
@zkhin You're adding read/write permissions to `locations` property for a bunch of roles. Can those not be referenced/used instead of the hardcoded true?

(sorry for the back and forth)

zkhin:
> @zkhin You're adding read/write permissions to `locations` property for a bunch of roles. Can those not be referenced/used instead of the hardcoded true?
> 
> (sorry for the back and forth)

Yes, it could be used for `canRead`/`canCreate` in the future but for that I need to do a KT with Mike

PR #1369, opened 2020-10-15
Add/remove/lists other locations in project

zkhin:


michaelmarshall (approved):

src/components/project/project.resolver.ts
```diff
@@ -254,6 +255,34 @@ export class ProjectResolver {
     return true;
   }
 
+  @Mutation(() => IProject, {
+    description: 'Add a location to a project',
+  })
+  async addOtherLocationToProject(
+    @Session() session: ISession,
+    @Args('projectId', { type: () => ID }) projectId: string,
+    @Args('locationId', { type: () => ID }) locationId: string
```
CarsonF, line 264: Can you do the args thing here too?

src/components/project/project.service.ts
```diff
@@ -792,13 +793,93 @@ export class ProjectService {
...
+      })
+      .run();
+
+    const items = await Promise.all(
+      result.map(
+        async (location): Promise<Location> => {
+          return await this.locationService.readOne(location.id, session);
+        }
+      )
+    );
```
CarsonF, line 874: Why not use `calculateTotalAndPaginateList`?

CarsonF (changes requested):

src/components/project/project.service.ts
```diff
@@ -874,13 +875,89 @@ export class ProjectService {
...
-  async listOtherLocations(
-    _projectId: string,
-    _input: LocationListInput,
+  async addOtherLocation(
+    projectId: string,
+    locationId: string,
+    session: ISession
+  ): Promise<void> {
+    try {
+      await this.removeOtherLocation(projectId, locationId, session);
```
CarsonF, line 884: Still don't like this, but we can come back to it.

CarsonF (approved):

PR #1371, opened 2020-10-15
1293 speed up engagements list

thomas-adev:
Added repository with fluent api to create chainable domain specific cypher requests.

Utilized to reduce API calls made by engagements list. Utilized to make list and readOne in the engagements.service utilize common repository request code.


michaelmarshall:
holding off on this as we may do some refactoring.

michaelmarshall:
I won't be able to revisit this before the MVP and before the bigger query refactor. Closing.

PR #1373, opened 2020-10-16
Add modifiedAt to partner

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1348
@michaelmarshall can you add `modifiedAt` to partner in the schema doc?


michaelmarshall (approved):
DB schema updated.

PR #1379, opened 2020-10-19
Add location fields to User and Organization modules

zkhin:


src/components/organization/organization.resolver.ts
```diff
@@ -87,6 +103,30 @@ export class OrganizationResolver {
     return true;
   }
 
+  @Mutation(() => Organization, {
+    description: 'Add a location to a organization',
+  })
+  async addLocationToOrganization(
+    @Session() session: ISession,
+    @Args('organizationId', { type: () => ID }) organizationId: string,
+    @Args('locationId', { type: () => ID }) locationId: string
```
CarsonF, line 112: Could you do these like:
https://github.com/SeedCompany/cord-api-v3/blob/229348e2c1d577051ba770a49560e7c37c8a7997/src/components/authorization/authorization.resolver.ts#L34
https://github.com/SeedCompany/cord-api-v3/blob/229348e2c1d577051ba770a49560e7c37c8a7997/src/components/authorization/authorization.resolver.ts#L13-L16
That way IDs are validated.

CarsonF (changes requested):

michaelmarshall (dismissed):

src/components/user/user.service.ts
```diff
@@ -612,6 +619,92 @@ export class UserService {
     };
   }
 
+  async addLocation(
+    userId: string,
+    locationId: string,
+    session: ISession
+  ): Promise<void> {
+    try {
+      await this.removeLocation(userId, locationId, session);
```
CarsonF, line 628: There's gotta be a better way to do this.

CarsonF (approved):

CarsonF:
We've got list/add/remove locations on several resources. It could be better to abstract this in `LocationService` as just pass in the piece of the query define the relationship to the location.

zkhin:
> We've got list/add/remove locations on several resources. It could be better to abstract this in `LocationService` as just pass in the piece of the query define the relationship to the location.

Okay, I will create a new ticket/PR for refactoring for this

Issue #1363, opened 2020-10-15
query {module} > Data is still returned at times when the "sort" field having invalid data (i.e. any string other than actual sorting param)


iPiyushPatel:
**Describe the bug**
query {module} > Data is still returned at times when the `sort` field having invalid data (i.e. any string other than actual sorting param)

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
Error should be returned for invalid input to the `sort` field.

**Actual Result**
At times, Data is still returned when `sort` field having invalid data (i.e. any string other than actual sorting param)
See video - https://drive.google.com/file/d/17cuz-dYlq3Wy6mVAEWV7pCZXZjnNEr16/view?usp=sharing

**Note**
I can replicate the same issue in Budget, Education, Film, Organization, Language, LiteracyMaterial, Location, Organization, Partner, Partnership, Song, Story, and User.
Some request returns data and some returns error for invalid input to `sort` field.

**GraphQL Query ( partners ):**
![image](https://user-images.githubusercontent.com/5212585/96131985-7b2ad080-0f17-11eb-85fa-0c375aa588f0.png)

**GraphQL Query ( unavailablities ):**
![image](https://user-images.githubusercontent.com/5212585/96130800-04d99e80-0f16-11eb-87c6-fe0d0ff7c587.png)

**GraphQL Query ( films ):**
![image](https://user-images.githubusercontent.com/5212585/96131640-153e4900-0f17-11eb-829c-9840389ec9b7.png)

**GraphQL response( partners ):**
```
{
    "data": {
        "partners": {
            "total": 46,
            "hasMore": true,
            "items": [
                {
                    "id": "tB1pr4cUE",
                    "createdAt": "2020-10-15T18:45:03.871+05:30",
                    "types": {
                        "value": [
                            "Funding"
                        ],
                        "canRead": true,
                        "canEdit": true
                    },
                    "financialReportingTypes": {
                        "value": [],
                        "canEdit": true,
                        "canRead": true
                    },
                    "pmcEntityCode": {
                        "value": "KAK",
                        "canEdit": true,
                        "canRead": true
                    },
                    "globalInnovationsClient": {
                        "value": false,
                        "canEdit": true,
                        "canRead": true
                    },
                    "active": {
                        "value": false,
                        "canEdit": true,
                        "canRead": true
                    },
                    "address": {
                        "value": "2621 McKenzie Gardens",
[...]

PR #1407, opened 2020-10-21
Abstract sort

frankyma:


src/common/util.ts
```diff
@@ -50,3 +50,8 @@ export const iterate = <T>(
   }
   return res;
 };
+
+/** helper to put a delay in JS */
```
frankyma, line 54: unrelated to this ticket, thought I'd sneak in this helper, mostly used by FE to delay calls to check loading states

src/components/project/project.service.ts
```diff
@@ -565,18 +565,22 @@ export class ProjectService {
...
-          .with([
-            '*',
-            ...(input.sort === 'sensitivity' ? [sensitivityCase] : []),
-          ])
-          .orderBy(sortBy, order)
+      .call(
+        calculateTotalAndPaginateList,
+        input,
+        this.securedProperties,
+        (q, sort, order) =>
```
frankyma, line 572: can't use the defaultSorter on this because it doesn't take in the case statement...yet.  maybe on the next iteration

src/core/database/query/lists.ts
```diff
@@ -23,3 +33,34 @@ export function calculateTotalAndPaginateList(
...
+  return sortInput in securedProperties
+    ? q
+        .match([
+          node('node'),
+          relation('out', '', sortInput, { active: true }),
+          node('prop', 'Property'),
+        ])
+        .with('*')
+        .orderBy(sortValSecuredProp, order)
+    : q.with('*').orderBy(sortValBaseNodeProp, order);
```
frankyma, line 65: The issue with this is last orderBy statement is if we sort by something that doesn't match any property values we can all results returned unsorted. ( e.g. sort by "asdf") This exists on master on most list queries

src/common/util.ts
CarsonF: https://github.com/SeedCompany/cord-api-v3/blob/a3afcbe5eb4660d4c9d53874505e85034b9b2bce/src/common/util.ts#L9-L10

src/core/database/query/lists.ts
```diff
@@ -23,3 +33,34 @@ export function calculateTotalAndPaginateList(
...
+
+export const defaultSorter: Sorter = (
+  q,
+  sortInput,
+  order,
+  securedProperties,
+  sortValInput
+) => {
+  //The properties that are stored as strings
+  const stringProperties = ['name'];
```
CarsonF, line 45: This feels like magic. That's why I wanted to pass in the config to this abstraction so that the concrete caller defines the sorting rules based on the properties it knows about for its domain model.

An example: This doesn't work for users which have `firstName` and `lastName`.

src/core/database/query/lists.ts
CarsonF: I'd really like to define these sort keys in a stricter way. There's also this though https://github.com/SeedCompany/cord-api-v3/issues/1363#issuecomment-712307415

frankyma: Is any specific action you'd want me to take for this ticket?

src/core/database/query/lists.ts
frankyma: Not really following your first statement...but what we've been doing in the intermediary step is make the resource specific sortmap: https://github.com/SeedCompany/cord-api-v3/blob/090a4d38e6393daaf41433485b57d63baae117f4/src/components/organization/organization.service.ts#L268-L271

I tried this and user and it does work for `displayFirstName` and `displayLastName`.  `firstName` is a calculated value, which is another complexity.

frankyma: I created this `const stringProperties = ['name']` just to capture the property across resources  that we know are strings so we don't have to make that sortMap for each resource

CarsonF: The thing I don't like about this is it's coupling the sorting strategy with the specific database query. Which is why I created in my branch the separate enum. So `name` is sorted case insensitive, and separately it's a secured property or a special relation or something else. 

> `firstName` is a calculated value, which is another complexity.

You're right. We need to maybe get with @sethmcknight to determine what to do for user sorting. I think we have some work to do with display* properties anyways.

CarsonF (approved):
I think this is a good next step. I'd like to see more still on validation, flexibility with properties, and decoupling from the db query. This can come later.

CarsonF:
Thanks for DRYing this up @frankyma 💪

Issue #1385, opened 2020-10-20
Create Workflow tests


michaelmarshall:
We need tests for everything the beta testers have in their checklist.
![image](https://user-images.githubusercontent.com/1165460/96599905-014b6c00-12b6-11eb-8e87-53531441acd1.png)

There is a new suite called `project-workflow`, we'll use that for this issue.

Let's have a KT with me (Michael) before starting this. I'll explain the details.

PR #1417, opened 2020-10-22
Project workflow test

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1385


michaelmarshall:
Coming along well. I assume you don't need an approval just yet as there is still work to do.

PR #1423, opened 2020-10-22
Only give creator admin privileges on certain basenodes

frankyma:
We used to give the creator admin privileges on all basenodes they create, regardless of the user's role.  So a ProjectMember can create a partnership and has edit access on all the fields.  Now check that the baseNode is a certain kind (e.g. not a Partnership) before giving admin SGs.

### DIFF
--- a/src/components/authorization/authorization.service.ts
+++ b/src/components/authorization/authorization.service.ts
@@ -79,6 +79,17 @@ export interface ProjectChildIds {
   users: string[];
 }
 
+const projectChildNodes = [
+  DbBaseNodeLabel.Budget,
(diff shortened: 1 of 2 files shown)


michaelmarshall (approved):

CarsonF (commented):
Wouldn't be better to define a whitelist of what does get admin privileges instead of a blacklist of what doesn't?

Better yet, maybe the caller can choose whether or not to give admin powers. That way the auth module doesn't have to maintain that mapping.

michaelmarshall:
@CarsonF I agree. I'll do that in my speed refactor I'm doing.

michaelmarshall (approved):

PR #1443, opened 2020-10-27
Check is not null on sort value

frankyma:
remove null sort values when specified.  This will fix sorting by a optional value like project estimated submission date, so projects without that date will be removed from the list.  Otherwise cypher will make those projects will appear at the beginning or the end depending on the sort order.

The issue with this approach is it really breaks the Partner sort.  While it's currently not working, aka the partners are always returned in random order, this will not return any partner because a Partner doesn't have a name (it's the org's name).  We can either write a custom sort query for Partner in this case, or actually store a name field on create partner

### DIFF
--- a/src/components/budget/budget.service.ts
+++ b/src/components/budget/budget.service.ts
@@ -313,7 +313,7 @@ export class BudgetService {
 
     const records = await this.listRecords(
       {
-        sort: 'fiscalYear',
+        sort: 'createdAt',
         order: Order.ASC,
         page: 1,
         count: 25,
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -673,11 +673,12 @@ export class ProjectService {
               relation('out', '', sort, { active: true }),
               node('prop', 'Property'),
             ])
-            .where({ [sortBy]: not(isNull()) })
             .with([
               '*',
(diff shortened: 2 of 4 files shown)


src/components/budget/budget.service.ts
```diff
@@ -313,7 +313,7 @@ export class BudgetService {
 
     const records = await this.listRecords(
       {
-        sort: 'fiscalYear',
+        sort: 'createdAt',
```
frankyma, line 316: sorting on createdAt because fiscalYear is not a required field, and the new default sorter will not list records with null sort values

CarsonF:
> remove null sort values when specified. This will fix sorting by a optional value like project estimated submission date, so projects without that date will be removed from the list. 

I'm not sure it makes sense to remove items from the list just because you are sorting a certain way.

> Otherwise cypher will make those projects will appear at the beginning or the end depending on the sort order.

This is what I would expect.

src/components/budget/budget.service.ts
CarsonF: Disagree. fiscalYear is the default. If they can't see the fiscalYear then the records are just unsorted that way, but they'd never know since they can't see the actual value.

frankyma:
Closed because date sort working as expected, no need to remove null items from sort

Issue #1437, opened 2020-10-26
Project.stepChangedAt


CarsonF:
Expose a `Project.stepChangedAt` DateTime field. It should be updated (to now-ish) whenever the project step is changed.

Consult Engagement.statusChangedAt if needed.

PR #1445, opened 2020-10-27
Add stepChangedAt field/handler to project

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1437
@michaelmarshall can you update the project schema with this new field?

### DIFF
--- a/src/components/project/dto/project.dto.ts
+++ b/src/components/project/dto/project.dto.ts
@@ -7,6 +7,7 @@ import {
   Secured,
   SecuredDate,
   SecuredDateNullable,
+  SecuredDateTime,
   SecuredString,
   Sensitivity,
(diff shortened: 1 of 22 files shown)


src/components/project/dto/project.dto.ts
```diff
@@ -65,6 +65,9 @@ class Project extends Resource {
   // this should match project mouEnd, until it becomes active, then this is final.
   readonly initialMouEnd: SecuredDateNullable;
 
+  @DateTimeField()
+  readonly stepChangedAt: DateTime;
```
CarsonF, line 69: Hmm I wonder if this should be secured? `Engagement.statusModifiedAt` is...

CarsonF (approved):

Issue #1342, opened 2020-10-09
[BUG] 'My Projects' filter not working


sethmcknight:
The Projects list is displaying all projects visible to the user, whether the 'My Projects' filter is toggled on or off.

zkhin:
@CarsonF Is this backend issue? I think `mine` filter option is working fine now.

CarsonF:
I thought it was lol.

zkhin:
@iPiyushPatel Can you verify this bug? `mine` filter works fine on my local.

iPiyushPatel:
@zkhin @michaelmarshall @CarsonF 
I can still replicate this issue on the latest master 4fb39928
See video: https://drive.google.com/file/d/1MTTdMkYhB_J0M_QVEYJIkymvsHnlUDDZ/view?usp=drivesdk

Note: Projects created by other users are returned for `mine = true` and projects created by the same user also returned for `mine = false`.

sethmcknight:
This still seems to be an issue.

- We should default the filter to `mine=true`
- 'mine=true' should return any project the user is a team member of (@CarsonF this would be the same list we'll display on a users profile)
- 'mine=false' should return all projects the user has permissions to view (regardless of membership)

CarsonF:
FYI @michaelmarshall reverted the merged PR for this.

iPiyushPatel:
Hi @zkhin @michaelmarshall @CarsonF 
I've verified on the latest develop d2d4e490 branch and it's working fine.
See video: 
https://user-images.githubusercontent.com/5212585/106235840-8df4b700-6221-11eb-8365-974d2326d852.mp4

PR #1448, opened 2020-10-28
Fix mine filter in list projects

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1342

### DIFF
--- a/test/project.e2e-spec.ts
+++ b/test/project.e2e-spec.ts
@@ -1,6 +1,6 @@
 import { gql } from 'apollo-server-core';
 import * as faker from 'faker';
-import { orderBy, times } from 'lodash';
+import { compact, orderBy, times } from 'lodash';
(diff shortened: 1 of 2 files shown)


test/project.e2e-spec.ts
```diff
@@ -535,6 +536,9 @@ describe('Project e2e', () => {
           input: {
             sort: 'sensitivity',
             order,
+            filter: {
+              mine: true,
+            },
```
CarsonF, line 541: Excluding this shouldn't fail the test

zkhin: Not sure why it fails without mine filter.

zkhin: if we don't add mine filter, it returns all the projects like below.
image

michaelmarshall (approved):

PR #1456, opened 2020-10-29
#633 add logic to be able to sort on special characters.

elisabeth-sorrell:


elisabeth-sorrell:
@CarsonF could you take a quick look at this solution?... the latest of master hasn't been merged in, but I just wanted to see what you thought of the solution.

src/core/database/results/parse-base-node.ts
```diff
@@ -4,9 +4,10 @@ import { DateTime } from 'luxon';
...
   id: string;
   createdAt: DateTime;
+  sortValue?: string;
 }
 
 export const parseBaseNodeProperties = (node: Node<BaseNode>) => {
-  const { id, createdAt } = node.properties;
-  return { id, createdAt };
+  const { id, createdAt, sortValue } = node.properties;
+  return { id, createdAt, sortValue };
```
CarsonF, line 12: We shouldn't ever need to pull the sortValue out of the database. It will only be used in queries.

src/core/database/database.service.ts
```diff
@@ -156,6 +205,14 @@ export class DatabaseService {
     nodevar: string;
   }): Promise<TObject> {
     const createdAt = DateTime.local();
+    let nodePropsToUpdate: SelectedNodeProperties = { createdAt, value };
+    if (await this.hasSortValue({ session, object, key, nodevar })) {
```
CarsonF, line 209: We should be able to use the same logic as we do on create to determine the sort value. We shouldn't need to pull the current value out of the database to determine this.

CarsonF (changes requested):
Thinking about it more and after a brief discussion with Michael.

I think there's just not a problem with having a `sortValue` on every property node. For the majority of the cases it could be the same value as value, but for strings we could deburr it. 
```ts
const nodeProps = {
  createdAt,
  value,
  sortValue: determineSortValue(value),
};

const determineSortValue = (value: unknown) =>
  typeof value === 'string' ? deburr(value) : value;
```

----

Another option would be to only have `sortValue` when necessary. 

In that case we could just do:
```ts
const determineSortValue = (value: unknown) =>
  typeof value === 'string' ? deburr(value) : undefined;
```
Returning `undefined` should just omit it from the db.

We'd need the sorting logic to be smarter then, to know whether to sort on `sortValue` or `value`.

elisabeth-sorrell:
@CarsonF ready for review

michaelmarshall (approved):

elisabeth-sorrell:
@michaelmarshall rebased on master

michaelmarshall (approved):

PR #1457, opened 2020-10-29
Read user perms on currentBudget

frankyma:


src/components/project/project.service.ts
```diff
@@ -998,10 +998,40 @@ export class ProjectService {
...
+      .first();
+
+    const budgetToReturn = current || pendingBudget;
     return {
-      value: current ? current : pendingBudget,
-      canEdit: true,
-      canRead: true,
+      value: budgetToReturn,
+      canRead: (budgetToReturn && budgetPerms?.canRead) || false,
+      canEdit: budgetPerms?.canEdit || false,
```
CarsonF, line 1034: Does canRead return true even when the list is empty? That to me is a sign that there's some difference between this query and the list query that needs to be rectified.

frankyma: yes i've been getting canRead true with an empty list.  I can try to investigate?  Not sure how/if they're related

frankyma: Just talked to @michaelmarshall, here's a summary:
The issue is as a Field Operations Director, when I crate a project, which creates a budget, the budget doesn't have the right permission path back to the creator so I don't have access to this budget.

This is the budget list query that's returning null (confirmed with Michael that it's the correct path, albeit too vague.  The perms Node should have canRead: true and property:budget):
`MATCH (User { id: '33UIVXUEzKD' })<-[:member]-(:SecurityGroup)-[:permission]->(perms:Permission)-[:baseNode]->(node:Budget)`

Need to investigate how the budget is created, but that's moved to java now

michaelmarshall (approved):

Issue #1460, opened 2020-10-30
updateProject > Getting "Cannot read property 'status' of undefined" error while update any field of project


iPiyushPatel:
**Describe the bug**
updateProject > Getting `Cannot read property 'status' of undefined` error while update any field of project

**Steps to Reproduce**
1. Login as a Root admin
2. Create Project
3. Click on the created project 
4. Update the `step` of the project

**Expected Result**
Project should be updated.

**Actual Result**
Getting `Cannot read property 'status' of undefined` error while update any field of project.
See video -  https://drive.google.com/file/d/1jswSE3OTwrW_6YoSaQ1OQF1Ef_ugjPIY/view?usp=drivesdk

cord-field# 1772f10d

CarsonF:
Can you post the exception stacktrace?

PR #1462, opened 2020-10-30
Fix cannot read budget status property while updating project

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1460


CarsonF:
@zkhin Frank's on it too with #1457

zkhin:
> @zkhin Frank's on it too with #1457

Okay, I will close this PR. Thanks

Issue #1465, opened 2020-10-30
Add 'partner' read and edit permission nodes


frankyma:
Blocker for https://github.com/SeedCompany/cord-api-v3/issues/1436

michaelmarshall:
`partners` added.

PR #1466, opened 2020-10-30
1436 user partners

frankyma:
Searches the user's partners (if the user has an org assigned from `assignOrganizationToUser` and if that org has a partner)

@michaelmarshall add the new field to the db proc.

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -49,7 +49,11 @@ import {
   OrganizationService,
   SecuredOrganizationList,
 } from '../organization';
-import { PartnerService } from '../partner';
+import {
+  PartnerListInput,
+  PartnerService,
+  SecuredPartnerList,
+} from '../partner';
 import {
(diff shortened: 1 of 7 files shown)


src/components/user/user.service.ts
```diff
@@ -588,6 +595,66 @@ export class UserService {
...
+        exception,
+        userId: session.userId,
+      });
+      throw new ServerException('Could not find partner', exception);
+    }
+    if (!user) {
+      throw new NotFoundException('Could not find user', 'userId');
+    }
+
+    //TODO: add back in after the partner permission nodes are available in the DB
```
frankyma, line 634: Needs the partner permission nodes added: https://github.com/SeedCompany/cord-api-v3/issues/1465

michaelmarshall: added `partners` to procedure code.

michaelmarshall (approved):

PR #1511, opened 2020-11-04
unskip list users

frankyma:


test/user.e2e-spec.ts
```diff
@@ -206,7 +206,7 @@ describe('User e2e', () => {
   });
 
   // LIST USERS
-  it.skip('list view of users', async () => {
+  it.only('list view of users', async () => {
```
CarsonF, line 209:
```suggestion
  it('list view of users', async () => {
```

CarsonF (changes requested):

test/user.e2e-spec.ts
frankyma: sorry about that, been a distracting day.

I also added the no-only-tests linter

michaelmarshall (approved):

Issue #1507, opened 2020-11-04
User.knownLanguages


CarsonF:
[`KnownLanguages`](https://github.com/SeedCompany/cord-api-v3/blob/42ddf17a130dead6c9212b37610dc215f8477443/src/components/user/known-language.ts) needs to be connected up.

- [ ] Enum: LanguageProficiency
- [ ] Query: User.knownLanguages
- [ ] Mutation: createKnowLanguage
- [ ] Mutation: deleteKnowLanguage

The file linked above was the placeholder. It can be deleted/moved to dto/ to match the rest of the app.

PR #1515, opened 2020-11-05
Add known languages to user

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1507


src/components/user/user.resolver.ts
```diff
@@ -268,4 +280,36 @@ export class UserResolver {
     await this.userService.removeOrganizationFromUser(input.request, session);
     return true;
   }
+
+  @Mutation(() => User, {
+    description: 'Create known language to user',
+  })
+  async createKnowLanguage(
```
michaelmarshall, line 287: mispelling. should be `Known`

michaelmarshall: there are other examples of this in this PR I won't comment on.

michaelmarshall (changes requested):

PR #1524, opened 2020-11-05
Language filters

frankyma:
Branched off of: https://github.com/SeedCompany/cord-api-v3/pull/1513
Filter on sensitivity, leastOfThese, signLanguage, and Dialect

### DIFF
--- a/src/components/language/dto/list-language.dto.ts
+++ b/src/components/language/dto/list-language.dto.ts
@@ -12,7 +12,7 @@ import { Language } from './language.dto';
 @InputType()
 export abstract class LanguageFilters {
   @Field(() => [Sensitivity], {
-    description: 'Only projects with these sensitivities',
+    description: 'Only languages with these sensitivities',
     nullable: true,
   })
   readonly sensitivity?: Sensitivity[];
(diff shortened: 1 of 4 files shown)


michaelmarshall (dismissed):

frankyma:
@michaelmarshall I merged the base to master can you pls re-review?

src/components/language/dto/list-language.dto.ts
```diff
@@ -1,15 +1,38 @@
...
+  Sensitivity,
   SortablePaginationInput,
 } from '../../../common';
 import { Language } from './language.dto';
 
 @InputType()
-export abstract class LanguageFilters {}
+export abstract class LanguageFilters {
+  @Field(() => [Sensitivity], {
+    description: 'Only projects with these sensitivities',
```
CarsonF, line 15:
```suggestion
    description: 'Only languages with these sensitivities',
```

src/components/language/query.helpers.ts
```diff
@@ -0,0 +1,33 @@
...
+        ? q
+            .match(propMatch('isSignLanguage'))
+            .where({ isSignLanguage: { value: true } })
+        : q
+    )
+    .call((q) =>
+      filter.isDialect
+        ? q.match(propMatch('isDialect')).where({ isDialect: { value: true } })
+        : q
+    );
```
CarsonF, line 32: Can't these be merged into a single where clause that would be more performant?

frankyma: I'm not sure, I'm not that familiar with the where clause.  What's the query you're imagining?  I can't think of one that allows for us to match on multiple of these filters together.

CarsonF:
```ts
.where({
  a: ...,
  b: ...,
})
```
aka
```sql
WHERE a AND b
```

frankyma: Ok I think I got it to work, pushed on the latest commit

PR #1536, opened 2020-11-10
Feature/api 1526 refactor delete with permission

frankyma:


src/components/language/language.service.ts
```diff
@@ -243,6 +244,12 @@ export class LanguageService {
           isPublic: false,
           isOrgPublic: false,
         },
+        {
+          key: 'canDelete',
```
frankyma, line 248: This property is not technically needed since all we're checking is the canDelete permission, but agreed with Michael to keep this in for consistency

src/components/language/language.service.ts
```diff
@@ -373,7 +380,7 @@ export class LanguageService {
       },
       sensitivity: props.sensitivity,
       ethnologue,
-      canDelete: true, // TODO
+      canDelete: await this.db.checkDeletePermission(langId, session),
```
frankyma, line 383: Tried to move this into its own resolver to save on performance, but got an error in this return statement because the Resource dto needs a `canDelete` property.  Not sure how to get around this or if it's a good idea.

michaelmarshall (approved):

PR #1539, opened 2020-11-11
Add handler for removing powers and perms when roles are removed

### LINKED ISSUE #1516: Add handler for when roles are removed from User.roles and ProjectMember.roles
Need to delete the correct powers and permissions when a role is removed.

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1516

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3655529641) by [Unito](https://www.unito.io)


src/components/user/handlers/user-project-member-roles.handler.ts
```diff
@@ -0,0 +1,45 @@
...
+  IEventHandler,
+  ILogger,
+  Logger,
+} from '../../../core';
+import { AuthorizationService } from '../../authorization/authorization.service';
+import { ProjectMemberUpdatedEvent, UserUpdatedEvent } from '../events';
+
+type SubscribedEvent = UserUpdatedEvent | ProjectMemberUpdatedEvent;
+
+@EventsHandler(UserUpdatedEvent, ProjectMemberUpdatedEvent)
```
michaelmarshall, line 14: As written, when a project member loses a role, this will delete them from the same role grant on any project. This handler shouldn't handle the `ProjectMemberUpdatedEvent`. We need to have a specific handler for that event that only removes the user from the role(s) on the project.

So there are two scenarios this PR needs to cover:
1. A user's role is removed from their user node. This is a global removal of a role. They are no longer authorized to have this role anywhere. The apoc call currently in this PR does that.
2. A user's role is removed on their project member node on a project. In this case, we just need to remove their membership on the security group for their role on that one project.

michaelmarshall (changes requested):

michaelmarshall:
Our latest security refactor changes how we will do this. Closing.

Issue #851, opened 2020-07-28
Not Found on Invalid Session


CarsonF:
I just got a project not found error. I refreshed the page and it took me to the login screen, meaning I was either not logged in or didn't have a valid session.
An [`UnauthenticatedException`](https://github.com/SeedCompany/cord-api-v3/blob/6b294cf6f8d05a3550dfa2f41dd7fef8871c4a07/src/common/exceptions/unauthenticated.exception.ts) should be thrown/returned instead of the `NotFound` error.

michaelmarshall:
Carson was logged in as root. Its possible this was caused by another person logging in as root. We need to ensure multiple logins work as expected.

CarsonF:
Right, but that's a separate issue. If you are logged out, you should get a unauthenticated error instead of not found error.

zkhin:
I think this ticket could be closed since refactoring `readOne` methods.

CarsonF:
We still always handle non matches as not found without checking if it's a permissions thing.

elisabeth-sorrell:
@ParkeBrown changes are merged in. https://github.com/SeedCompany/cord-api-v3/pull/1543

ParkeBrown:
@iPiyushPatel would you test to confirm bug is resolved?

iPiyushPatel:
Hi @CarsonF 
I'm trying to test this ticket, can you let me know whats the session timeout we've set? 

Also I'm thinking to hit project page after timeout and see I'm getting login page or not (instead of not found page). Is that ok ?

thanks,

ParkeBrown:
Thank you, @iPiyushPatel. Feedback may need to come post launch items.

elisabeth-sorrell:
@iPiyushPatel the problem was that the 'Session Invalid' error was occurring when it was supposed to be 'Unauthenticated'.
1. Create a project
2. log out
3. query for that project's id

The error was that there was Session invalid error instead of Unauthenticated. Didn't really have anything to do with sessions in and of itself.

iPiyushPatel:
Hi, @elisabeth-sorrell @ParkeBrown 
I've verified on the latest develop a99ebf10 branch and not getting any error.
See video: https://drive.google.com/file/d/1sn8S3Pza_OkFjaK3foWUAbV3CNGl-qns/view?usp=drivesdk

elisabeth-sorrell:
Putting into Sprint Closed, then.

PR #1543, opened 2020-11-12
#851 error message on invalid/unauthenticated session

elisabeth-sorrell:


michaelmarshall (approved):

CarsonF:
This would be the first Query where we require users to be logged in

PR #1546, opened 2020-11-18
add workflow for engagement status advancement

elisabeth-sorrell:


elisabeth-sorrell:
Warning: hasn't been tested and there are compiler errors with some handlers. will need to get with a Typescript master on it because I think it should be an easy fix.

CarsonF (commented):
Looks good at first glance. The errors are because status is now a secured object. Every place that was referencing it needs to be updated.

elisabeth-sorrell:
@CarsonF ready for review. commented out one or two expect statements and am hoping to fix that in the second iteration

CarsonF (approved):

PR #1567, opened 2020-11-30
Add v2 engagement statuses

blakekrammes:


CarsonF:
Can you note the v2 ones that will only be used with legacy projects?

blakekrammes:
> Can you note the v2 ones that will only be used with legacy projects?

Are all the added ones v2 only?

CarsonF:
@sethmcknight ^

blakekrammes:
@sethmcknight are the following only from v2?

Converted = 'Converted',
Unapproved = 'Unapproved',
Transferred = 'Transferred',
NotRenewed = 'NotRenewed',
Rejected = 'Rejected',

michaelmarshall (approved):

PR #1569, opened 2020-12-01
Change paratext casing

blakekrammes:


CarsonF:
Is this field used in the UI?

blakekrammes:
The plan was to have it in the v3 UI right?

CarsonF:
Prob, but idk if it's there yet

blakekrammes:
It's not there in v2, but this is how it's cased in our database and on paratext's website.

blakekrammes:
Oh. You're wondering if this will break v3 frontend.

CarsonF:
I don't disagree with the change, just want to make sure it's handling places since it's breaking.

blakekrammes:
It's used in a couple components on the FE. I can change those.

CarsonF:
Better yet. Add a deprecated field resolver so this isn't a breaking change

CarsonF:
Example:
https://github.com/SeedCompany/cord-api-v3/blob/8a8d7cc5e1e87d5f934ee23ff85b9cae737a66ae/src/components/project/project.resolver.ts#L109-L115

blakekrammes:
In addition to changing the FE or instead of?

CarsonF:
Both. Doing it this way should be standard procedure going forward. This allows the API and consumers to be deployed independently without breaking anything.

blakekrammes:
Do you then remove the deprecation code after?

CarsonF:
Later yeah

blakekrammes:
That look right?

CarsonF (dismissed):
Yeah looks good. Technically we should support the old name on the input as well, but I'll let that slide as we aren't using it much yet.

blakekrammes:
Where does that happen? In the dto?

blakekrammes:
Like so?

CarsonF:
Well yes in the DTO, but in the create/update mutations, not the output. The field you added to the DTO is the same as adding the field resolver, so it can be removed.

CarsonF (commented):
Closer @blakekrammes. You accept the old form but then don't handle the different case in the service. They also should be nullable.

Again if we want to skip this on input I'm ok with it.

src/components/engagement/engagement.service.ts
```diff
@@ -792,7 +792,7 @@ export class EngagementService {
           'communicationsCompleteDate',
           'startDateOverride',
           'endDateOverride',
-          'paraTextRegistryId',
+          'paratextRegistryId',
```
CarsonF, line 795: Above where `changes` is defined:
```tsx
    const changes = pickBy({
      ...rest,
      paratextRegistryId: changes.paratextRegistryId || changes.paraTextRegistryId
      modifiedAt: DateTime.local(),
    }, val => val !== undefined);
```

PR #1578, opened 2020-12-11
Extract, Persist, & Expose PnP Summary Data

CarsonF:


src/components/engagement/engagement.service.ts
```diff
@@ -312,6 +315,10 @@ export class EngagementService {
       input.pnp,
```
blakekrammes, line 315: Why doesn't `createDefinedFile` check for `input.pnp`?

CarsonF: Since all defined files could be nullable we just save the if statement everywhere and do it inside the function.

blakekrammes (approved):

Issue #1583, opened 2020-12-17
Engagements error when querying Projects without language engagements


elisabeth-sorrell:
## Current Behavior
When querying a project with no language engagements, there's an error from the engagement output that no engagement id is provided to the transitions mutation

## Expected Behavior
There should be no error, just provide empty transitions instead

## Steps to Reproduce

1. Create a project
2. query for projects, engagements, and transitions
3. error happens

elisabeth-sorrell:
https://github.com/SeedCompany/cord-api-v3/pull/1584

elisabeth-sorrell:
Per @CarsonF 's comment on the above PR, I'm going to close because it's a developer mistake in graphql if we're querying an engagement's transitions when the engagement doesn't exist.

CarsonF:
There shouldn't be an error if there is just data not there though. Why is the enagagenent transitions resolver being invoked without an enagagenent?

elisabeth-sorrell:
It was being done in a bug that Piyush put in. When trying to reproduce the bug, the mutation that I had copied/pasted from Piyush had `projects {...engagements {transitions{...}}}` when there was no engagement. The transitions resolver as of now throws a ServerException if there is no engagement, hence this ticket. I was thinking just return nothing if there is not an engagement ID so that there's no error in cases like this.
I'll post the bug when I find it, as well. May take a few minutes.

elisabeth-sorrell:
@CarsonF I'm okay with just closing this if you are. I don't think it's a bug. I put this in from trying to test one of Piyush's tickets.

PR #1584, opened 2020-12-17
return empty transitions instead of error #1583

elisabeth-sorrell:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3655558330) by [Unito](https://www.unito.io)


elisabeth-sorrell:
@CarsonF do you think you could take a look at this when you get a chance? I'm not sure if it goes against business rules...

CarsonF:
The exception is there because it's a developer mistake if it's missing.

elisabeth-sorrell:
@CarsonF thanks. I wanted to verify that, but didn't know how to put it into words :)

Issue #1541, opened 2020-11-12
mutation updateLanguage > User can successfully update the language which is created by another user


iPiyushPatel:
**Roles:** ProjectManager

**Describe the bug**
mutation updateLanguage > User can successfully update the language which is created by another user.

**Steps to Reproduce**
1. Login with user1
2. Create a Language
3. Register user2
4. Login with user2
5. Update the language created by user1 (As in step 2)

**Expected Result**
Unauthorized error should be thrown for the user updating language which is not created by him.

**Actual Result**
User can successfully update the language which is created by another user.
See video - https://drive.google.com/file/d/1-VsSldtUmZud17cAa0-mk3Cr7zVIRsxH/view?usp=drivesdk

GraphQL Query ( createLanguage ):
![image](https://user-images.githubusercontent.com/5212585/98916003-7fe39580-247f-11eb-91ee-2dfd35b6330b.png)

GraphQL Query ( updateLanguage ):
![image](https://user-images.githubusercontent.com/5212585/98916162-ac97ad00-247f-11eb-9a96-11665309d285.png)

**Query and Input File**
[updateLanguage_anotherUserUpdate_ISSUE.txt](https://github.com/SeedCompany/cord-api-v3/files/5529297/updateLanguage_anotherUserUpdate_ISSUE.txt)
[createLanguage_anotherUserUpdate_ISSUE.txt](https://github.com/SeedCompany/cord-api-v3/files/5529298/createLanguage_anotherUserUpdate_ISSUE.txt)

**graphQL Response:**
```
{
    "data": {
        "updateLanguage": {
            "language": {
                "id": "jmb6aCra7Nn",
                "createdAt": "2020-11-12T14:02:43.020-08:00",
                "sensitivity": "High",
                "avatarLetters": "L",
                "name": {
                    "value": "LangName1211020241",
                    "canEdit": false,
                    "canRead": true
                },
                "displayName": {
                    "value": "LangDispName1211020241",
                    "canEdit": false,
                    "canRead": true
                },
                "displayNamePronunciation": {
                    "value": "LangDispNamePron1211020241",
                    "canEdit": false,
                    "canRead": true
                },
                "isDialect": {
                    "value": false,
                    "canEdit": false,
                    "canRead": true
                },
                "ethnologue": {
                    "code": {
                        "value": "ethcodentwmj",
                        "canEdit": false,
                        "canRead": true
                    },
                    "provisionalCode": {
[...]

PR #1588, opened 2020-12-22
Fix language authentication #1541

elisabeth-sorrell:
- check for CreateLanguage power before creating; 
- get rid of the withAcl permission check for update and instead check each property's canEdit value before update and throw an UnauthorizedException if it's false


elisabeth-sorrell:
@CarsonF ready for your review

test/language.e2e-spec.ts
```diff
@@ -263,6 +262,7 @@ describe('Language e2e', () => {
   });
 
   it('should throw error if trying to set hasExternalFirstScripture=true while language has engagements that have firstScripture=true', async () => {
+    await loginAsAdmin(app);
```
CarsonF, line 265: Is it possible to just give the current user relevant permissions instead of this? 
This makes all subsequent test cases be ran as the admin user.

michaelmarshall (approved):

Issue #1566, opened 2020-11-30
createProject > Getting "Cannot return null for non-nullable field Unavailability.canDelete" error for "canDelete" property of "unavailabilities" present in query output


iPiyushPatel:
**Roles:** ProjectManager 

**Describe the bug**
createProject > Getting `Cannot return null for non-nullable field Unavailability.canDelete` error for `canDelete` property of `unavailabilities` present in query output

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
Project should be created.

**Actual Result**
Getting `Cannot return null for non-nullable field Unavailability.canDelete` error for `canDelete` property of `unavailabilities` present in query output
See video - https://drive.google.com/file/d/1YQiN_HJXFWKHv4q-_ufl4yyfBCLts8-4/view?usp=drivesdk

**GraphQL Query ( createProject ):**
![image](https://user-images.githubusercontent.com/5212585/100597043-f0653000-3322-11eb-9bc5-2c25f5f64ca7.png)

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Cannot return null for non-nullable field Unavailability.canDelete.",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR",
                "codes": [
                    "INTERNAL_SERVER_ERROR",
                    "GraphQL",
                    "Server"
                ]
            },
            "locations": [
                {
                    "line": 452,
                    "column": 23
                }
            ],
            "path": [
                "createProject",
                "project",
                "fieldRegion",
                "value",
                "fieldZone",
                "value",
                "director",
                "value",
                "unavailabilities",
                "items",
                0,
                "canDelete"
            ]
        }
    ],
```

**Query and Input File**
[CanDeleteErrorCreateProject.txt](https://github.com/SeedCompany/cord-api-v3/files/5615334/CanDeleteErrorCreateProject.txt)

CordAPIV3# 73f6e558

elisabeth-sorrell:
@iPiyushPatel Not getting the error for me now, tried with users `root` and `field operations director`. Can you verify when you get a chance? Thanks!

iPiyushPatel:
Hi @elisabeth-sorrell 
Verified on latest develop and getting `Cannot return null for non-nullable field SecuredOrganizationList.canCreate.` error when output query has `canCreate` field.
[...]

Issue #1563, opened 2020-11-27
query projects > Getting "Could not find project member" error after creating the "Internship" Project


iPiyushPatel:
**Roles:** ProjectManager 

**Describe the bug**
query projects > Getting `Could not find project member` error after creating the `Internship` Project

**Steps to Reproduce**
1. Login as a project manager
2. Create a project with `Internship` type 
3. Hit the request of `query projects`

**Expected Result**
Projects listing should be returned.

**Actual Result**
Getting `Could not find project member` error after creating the `Internship` Project
See video - https://drive.google.com/file/d/1gOc_tWxMkzcwUxxDQf2f3vof0G2pibud/view?usp=drivesdk

**GraphQL Query (projects ):**
![image](https://user-images.githubusercontent.com/5212585/100452743-beb05700-30df-11eb-8140-b8232a0e9e97.png)

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Could not find project member",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR",
                "exception": {
                    "name": "NotFoundException",
                    "status": 400,
                    "field": "projectMember.id",
                    "stacktrace": [
                        "at ProjectMemberService.readOne (components\\project\\project-member\\project-member.service.ts:198:13)",
                        "at async Promise.all (index 0)",
                        "at Object.runListQuery (core\\database\\results\\lists.ts:24:12)",
                        "at ProjectMemberService.list (components\\project\\project-member\\project-member.service.ts:318:12)",
                        "at ProjectService.listProjectMembers (components\\project\\project.service.ts:848:20)"
                    ]
                },
                "codes": [
                    "INTERNAL_SERVER_ERROR",
                    "Server"
                ]
            },
            "locations": [
                {
                    "line": 584,
                    "column": 7
                }
            ],
            "path": [
                "projects",
                "items",
                0,
                "team"
            ]
        }
    ],
```

**Query and Input File**
[NotFoundErrorInProjects.txt](https://github.com/SeedCompany/cord-api-v3/files/5607931/NotFoundErrorInProjects.txt)

CordAPIV3# 73f6e558

elisabeth-sorrell:
@iPiyushPatel Just tried to verify this. Kept getting an error for the `canCreate` property because it was null. Deleted all of the `canCreate`s in the output and works. Not sure why I can't reproduce the error....?

iPiyushPatel:
Hi @elisabeth-sorrell 
[...]

PR #1589, opened 2020-12-31
null canCreate be false if user.edit is false #1566

elisabeth-sorrell:
also covers https://github.com/SeedCompany/cord-api-v3/issues/1563

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -497,13 +497,10 @@ export class UserService {
       },
       session
     );
-    if (user.canEdit === null) {
-      user.canEdit = false;
-    }
     return {
       ...result,
       canRead: user.canRead,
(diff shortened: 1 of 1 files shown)


src/components/user/user.service.ts
```diff
@@ -497,6 +497,9 @@ export class UserService {
       },
       session
     );
+    if (user.canEdit === null) {
+      user.canEdit = false;
+    }
```
CarsonF, line 502: Would you mind doing this as
```tsx
canCreate: user.canEdit ?? false,
```

It's shorter and it doesn't mutate the `user` object

michaelmarshall (dismissed):

CarsonF (approved):

michaelmarshall (approved):

PR #1595, opened 2021-01-12
Exclude certain engagement statuses from language sponsor start date

### LINKED ISSUE #1594: Amend language.sponsorStartDate logic
Currently we derive this field from the first engagement's start date. We just found out that this may not always yield the correct data. 

We need to check to make sure that the first engagement's status is not any of the following:

- Rejected, Unapproved, In Development, Did Not Develop

If the language's first engagement does not have a valid status we should check the next most recent engagement to see if it is valid and so forth. If there is no engagement with a valid status, this field should be `null`.

CarsonF:
Closes #1594


blakekrammes (commented):
Excellent!

blakekrammes (approved):

PR #1603, opened 2021-01-26
Security Performance readOne refactor

elisabeth-sorrell:
Just using this as a placeholder for review


elisabeth-sorrell:
@CarsonF tests are fixed, but they were due to the role they were running under not having the nodeGrant needed for the actions so that's pretty cool.

elisabeth-sorrell:
@CarsonF and @michaelmarshall all of the readOnes are static with security now.

src/components/user/user.service.ts
```diff
@@ -285,17 +284,78 @@ export class UserService {
...
+      ...securedProps,
+      roles: {
+        ...securedProps.roles,
+        value: rolesValue,
+      },
+      canDelete: await this.db.checkDeletePermission(id, session),
+    };
+  }
+
+  async readOneByUserId(id: string, userId: string): Promise<User> {
```
CarsonF, line 352: See comment on projects about merging these two methods.

CarsonF (changes requested):
Great work @elisabeth-sorrell!

It would be lovely to have another function that wraps `getPermissionsOfBaseNode` & `parseSecuredPropertiesNew`. This way we don't have call both of those methods across the entire codebase.

elisabeth-sorrell:
> It would be lovely to have another function that wraps `getPermissionsOfBaseNode` & `parseSecuredPropertiesNew`. This way we don't have call both of those methods across the entire codebase.

Done!

src/components/project/project.service.ts
```diff
@@ -429,16 +428,12 @@ export class ProjectService {
     return project as InternshipProject;
   }
 
-  async readOne(
-    id: string,
-    { userId }: Pick<Session, 'userId'>
-  ): Promise<Project> {
+  // TODO: fix these so that we have DRY code.
+  async readOne(id: string, session: Session): Promise<Project> {
```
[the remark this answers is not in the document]
elisabeth-sorrell: I'm not really sure why we have `matchRequestingUser`.... to me it looked like it was to use the `matchSecure` or whatever the method is called that matches to the security groups. I've taken out the matchRequestinguser for now in Project and user, but let me know your thoughts.

I have, though, implemented what you mentioned in the authorization.service, project, and user for other stuff

src/components/user/user.service.ts
elisabeth-sorrell: see above reply... did pretty much the same thing with user as with project

src/components/user/user.service.ts
```diff
@@ -285,17 +284,78 @@ export class UserService {
...
+    }
+
+    const rolesValue = result.propList
+      .filter((prop) => prop.property === 'roles')
+      .map((prop) => prop.value as Role);
+
+    let permsOfBaseNode = {};
+    // -- let the user explicitly see all properties only if they're reading their own ID
+    // -- TODO: need to find a more... dynamic way of doing this. Just coding like this for
+    //    now to make test pass and this isn't a huge priority at the moment.
```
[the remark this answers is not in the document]
elisabeth-sorrell: yeah, add a comment to track in code. I'll also make a ticket for it.

src/components/budget/budget.service.ts
```diff
@@ -57,8 +55,9 @@ import { DbBudgetRecord } from './model/budget-record.model.db';
 @Injectable()
 export class BudgetService {
   private readonly securedProperties = {
-    status: true,
-    universalTemplateFile: true,
+    amount: true,
+    fiscalYear: true,
+    organization: true,
```
CarsonF, line 60: These are for budget records, not budgets. Both are needed.

elisabeth-sorrell:
@CarsonF or @michaelmarshall refactored the getFileVersion too. when I went through the readOnes I naively thought that it didn't pertain to file, so went ahead and refactored that too while I was at it.

elisabeth-sorrell:
@CarsonF @michaelmarshall new commits ready for review from #1624

CarsonF (approved):
I think I'm good with this iteration 👍

michaelmarshall (approved):

PR #1604, opened 2021-01-26
Add defaultFieldRegion to Location

### LINKED ISSUE #1593: Location.defaultFieldRegion
Add `defaultFieldRegion` property to `Location` which is a relation to a `FieldRegion`. It should probably be nullable.

@kjones140's requesting this for DOMO. And since the v2 schema has this relation, so it would be good to maintain this data (albeit in a different capacity).

elisabeth-sorrell:
closes #1593

### DIFF
--- a/src/components/budget/budget.module.ts
+++ b/src/components/budget/budget.module.ts
@@ -18,7 +18,7 @@ import * as handlers from './handlers';
     FileModule,
     forwardRef(() => AuthorizationModule),
     EducationModule,
-    LocationModule,
+    forwardRef(() => LocationModule),
     forwardRef(() => PartnershipModule),
     forwardRef(() => OrganizationModule),
     forwardRef(() => ProjectModule),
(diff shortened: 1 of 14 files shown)


elisabeth-sorrell:
tested with update/create/read operations

CarsonF (approved):
Looks good to me! @michaelmarshall could double check the db queries.

CarsonF:
What about permissions? Do the role files matter right now @michaelmarshall?

What about tests? Should we add some to ensure this relationship is working?

elisabeth-sorrell:
@CarsonF I'll go ahead and add tests FYI @michaelmarshall

elisabeth-sorrell:
@CarsonF just added a test for updating defaultFieldRegion. I went as in depth as the fundingAccount test

michaelmarshall:
@elisabeth-sorrell reminder to flag property additions as schema changes.

michaelmarshall (approved):

PR #1620, opened 2021-02-04
Change somethings for migration & bug fixes found during migration runs

CarsonF:


michaelmarshall (dismissed):

src/components/partnership/partnership.service.ts
```diff
@@ -365,32 +358,26 @@ export class PartnershipService {
...
+        changes = {
+          ...changes,
+          financialReportingType: null,
+        };
       }
-      changes = {
-        ...changes,
-        financialReportingType: null,
-      };
+      throw e;
```
blakekrammes, line 380: Should you throw an error if managing type is removed and FRT omitted? In that case it's not an invalid operation, just incomplete and you're helping it out above.

CarsonF: Ah! yes this should be in an else case. Good catch.

blakekrammes: Yep. :) Everything else looks solid from what I can tell.

CarsonF: Sweet! Fixed! 👍

PR #1622, opened 2021-02-05
Remove labels from ethnologue properties

CarsonF:
They don't seem to be used and the labels are not specific


blakekrammes:
I noticed that language name and display name have their own labels as well. Is that for the search queries? Nothing like that is happening with ethnologue, right?

CarsonF:
I think those are probably for the indexes. Search doesn't use labels on properties.

blakekrammes:
Both of those are constrained for uniqueness.

CarsonF:
Which ethnologue doesn't have.

michaelmarshall (approved):

Issue #1598, opened 2021-01-19
createPartnership > Getting "Partnership for this project and partner already exists" error when no partnership for provided "projectId" and "partnerId" is created before


iPiyushPatel:
**Roles:** ProjectManager

**Describe the bug**
createPartnership > Getting `Partnership for this project and partner already exists` error when no partnership for provided `projectId` and `partnerId` is created before

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
Partnership should be created Or a better error message should be provided

**Actual Result**
Getting `Partnership for this project and partner already exists` error when no partnership for provided `projectId` and `partnerId` is created before
See video - https://drive.google.com/file/d/1SQ3pKpUYMb4z9yQD5N-n2wEDX9k9yc21/view?usp=drivesdk

**GraphQL Query ( createPartnership ):**
![image](https://user-images.githubusercontent.com/5212585/104922359-9e569780-59c0-11eb-88f8-8f8c3d55ee7c.png)

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Partnership for this project and partner already exists",
            "extensions": {
                "codes": [
                    "Duplicate",
                    "Input",
                    "Client"
                ],
                "status": 400,
                "field": "partnership.projectId",
                "code": "Duplicate",
                "exception": {
                    "message": "Partnership for this project and partner already exists",
                    "stacktrace": [
                        "at PartnershipService.create (components\\partnership\\partnership.service.ts:113:13)",
                        "at PartnershipResolver.createPartnership (components\\partnership\\partnership.resolver.ts:39:25)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createPartnership"
            ]
        }
    ],
```

**Query and Input File**
[CreatePartnershipDuplicateIssue.txt](https://github.com/SeedCompany/cord-api-v3/files/5830490/CreatePartnershipDuplicateIssue.txt)

CordAPIV3# 8ab5e110

Issue #1597, opened 2021-01-19
createPartnership > Getting "Failed to create partnership" error while the input field"`agreement uploadID" has a blank value


iPiyushPatel:
**Roles:** ProjectManager

**Describe the bug**
createPartnership > Getting `Failed to create partnership` error while the input field `agreement uploadID` has a blank value

**Steps to Reproduce**
1. Create Project
2. Create Partner
3. Upload agreement file
4. Create a partnership with blank `agreement uploadID`
5. Repeat step 1 to 3
6. Create Partnership with `agreement uploadID`

**Expected Result**
Better error/validation message should be returned.

**Actual Result**
Getting `Failed to create partnership` error while the input field `agreement uploadID` has a blank value.
See video - https://drive.google.com/file/d/1kOKhZ9_3calHA0cQDLtsrb0weVXK0-kD/view?usp=drivesdk

**Note:** I can replicate the same issue for `mou uploadID`

**GraphQL Query (createPartnership ):**
![image](https://user-images.githubusercontent.com/5212585/104920678-2b4c2180-59be-11eb-870c-64ad7d964f61.png)

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Failed to create partnership",
            "extensions": {
                "codes": [
                    "Server"
                ],
                "status": 500,
                "code": "Server",
                "exception": {
                    "message": "Failed to create partnership",
                    "stacktrace": [
                        "at PartnershipService.create (components\\partnership\\partnership.service.ts:282:13)",
                        "at PartnershipResolver.createPartnership (components\\partnership\\partnership.resolver.ts:39:25)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createPartnership"
            ]
        }
    ],
```
**Query and Input File**
[CreatePartnershipAgreementIdIssue.txt](https://github.com/SeedCompany/cord-api-v3/files/5830419/CreatePartnershipAgreementIdIssue.txt)

CordAPIV3# 8ab5e110

PR #1626, opened 2021-02-09
Throw error when uploadIds are empty #1598

elisabeth-sorrell:


CarsonF:
This is should be validated automatically but they aren't because of some missing decorators. I'll fix in another PR.

How does this correspond to #1598?

elisabeth-sorrell:
> This is should be validated automatically but they aren't because of some missing decorators. I'll fix in #1628
> 
> How does this correspond to #1598?

A little complicated, but #1597 and #1598 both correspond with the following steps:
- createPartnership with an empty value for uploadId on agreement or mou
- try to create partnership again (since there was an createFileVersion error last time)
- error from having a duplicate partnership

the problem came in when the partnership is created before the uploadId verification, so that there is a partnership created, but it throws an error. Since there's no rollback after the error, there's a duplicate partnership error when the createPartnership is attempted again.

elisabeth-sorrell:
@CarsonF oh, and would you not mind linking #1598 and #1597 to #1628?

CarsonF:
Ok so the invalid input resulted in a partially mutated db state.
Two solutions:
1. Validated complete input before starting anything, which #1628 does.
2. Apply transactions so that errors during mutations rollback the db state instead of leaving it partially applied. I have a branch for this as well. Will PR soon.

elisabeth-sorrell:
Closing since it's handled by #1628

PR #1627, opened 2021-02-09
Add pnp year and quarter

blakekrammes:


src/components/engagement/dto/pnp-data.dto.ts
```diff
@@ -16,4 +16,10 @@ export abstract class PnpData {
 
   @Field(() => Float)
   variance: number;
+
+  @Field(() => Number)
```
CarsonF, line 20: graphql has floats and ints, no numbers.
```suggestion
  @Field(() => Int)
```

src/components/engagement/pnp-extractor.service.ts
```diff
@@ -16,12 +16,47 @@ export class PnpExtractor {
     @Logger('pnp:extractor') private readonly logger: ILogger
   ) {}
 
+  private year = 0;
+  private quarter = 0;
```
CarsonF, line 20: Class properties to store temporary data is a bad practice and hard to follow. Refactor so that it's returned from this method instead. The `extract()` method could be refactored to easily join the results of this and the excel data into a single object.

src/components/engagement/pnp-extractor.service.ts
```diff
@@ -16,12 +16,47 @@ export class PnpExtractor {
...
+    const reg2021 = /2021/;
+    const reg2020 = /2020/;
+    const reg2019 = /2019/;
+    // regex for year between 1900 and 2099
+    const yearReg = /19\d{2}|20\d{2}/;
+    const fyReg = /fy19|fy20|fy21/i;
+    const quarterReg = /q[1-4]/i;
+
+    const year = reg2021.exec(fileName)
+      ? Number(reg2021.exec(fileName))
```
CarsonF, line 34: Number conversion could be done once on output instead of for every case.

src/components/engagement/pnp-extractor.service.ts
```diff
@@ -16,12 +16,47 @@ export class PnpExtractor {
     @Logger('pnp:extractor') private readonly logger: ILogger
   ) {}
 
+  private year = 0;
+  private quarter = 0;
+
+  private getYearAndQuarterFromFileName(fileName: string) {
+    // first check for 2021, 2020, 2019
+    // the reason I did this is b/c some file names had multiple sets of 4 numbers (2020 and 2003 for instance)
```
CarsonF, line 24: Maybe you could match all years and then take the latest one?

src/components/engagement/pnp-extractor.service.ts
```diff
@@ -16,12 +16,47 @@ export class PnpExtractor {
...
+    this.quarter = quarter;
+  }
+
   async extract(
     input: CreateDefinedFileVersionInput,
     session: Session
   ): Promise<PnpData | null> {
     const file = await this.files.getFileVersion(input.uploadId, session);
     const pnp = await this.downloadWorkbook(file);
+    this.getYearAndQuarterFromFileName(input.name!);
```
CarsonF, line 59: input.name is optional and defaults to the file's name. Use it instead.
```suggestion
    this.getYearAndQuarterFromFileName(file.name);
```

CarsonF (changes requested):

src/components/engagement/pnp-extractor.service.ts
blakekrammes: Yeah. That would be more solid.

blakekrammes: This gets a bit tricky though because 4 digits may not be a year. I can also check that the number isn't greater than 2021, but still not bulletproof.

CarsonF: What about?
```tsx
const possibleYears = [];

const currentYear = DateTime.local().year;
const years = possibleYears.filter(year => year <= currentYear && year > currentYear - 30);
```

If you could distill all the options we are trying to account for it would be trivial to make a test for it.

blakekrammes: Yeah. This code is going away before too long though, right?

CarsonF: Yeah don't spend too much time on it

CarsonF (approved):

PR #1629, opened 2021-02-09
query languages > Getting "Engagement ID should have been provided by Engagement resolver" 

elisabeth-sorrell:


CarsonF (changes requested):
This could be moved into `EngagementService.list` instead of making the `ProjectService` aware of the hack

src/components/engagement/engagement.resolver.ts
```diff
@@ -60,11 +60,7 @@ export class EngagementResolver {
     @AnonSession() session: Session
   ): Promise<EngagementListOutput> {
     const engagements = await this.service.list(input, session);
-    for (const engagement of engagements.items) {
-      // @ts-expect-error hack engagement id into status object
-      // so the lazy transitions field resolver can use it
-      engagement.status.engagementId = engagement.id;
-    }
+    await this.service.setEngagementIdsIntoStatusObjects(engagements);
```
CarsonF, line 63:
```suggestion
```

src/components/engagement/engagement.service.ts
```diff
@@ -122,6 +122,14 @@ export class EngagementService {
     @Logger(`engagement.service`) private readonly logger: ILogger
   ) {}
 
+  async setEngagementIdsIntoStatusObjects(engagements: EngagementListOutput) {
```
CarsonF, line 125: You can just move this into the list function or at least make this private & not async.

CarsonF (changes requested):

CarsonF (approved):
Thanks @elisabeth-sorrell!

PR #1631, opened 2021-02-10
Do falsey check instead of nullish

blakekrammes:


blakekrammes:
Derped. `fourDigitYear` will be `0` not `null` or `undefined`.

CarsonF (approved):

blakekrammes:
For the record, this was caught esLint in seed-API as I was applying it there.

PR #1632, opened 2021-02-10
Math.max returns -Infinity when you try to spread an empty array

blakekrammes:
- i.e. `Math.max(...[])` doesn't work, but `Math.max([...[]])` does (returns `0`)


blakekrammes:
Also `-Infinity` is truthy for some reason.

CarsonF (approved):

PR #1649, opened 2021-02-18
Procedure roles to java roles

elisabeth-sorrell:
Admin, consultant-manager, consultant, controller, field-operations-director completed. 
Went through admin role at least twice.

One weird discrepancy:
Java Procedure code has a `partners` property, and the API code has a `knownLanguage` property. Checked the user service, and looks like partners is handled by the partner service, and looks like knownLanguage is part of the user object. I'm assuming theses changes were made in API, but the updated not reflected in the Java code, but not sure.


src/common/resource.dto.ts
```diff
@@ -11,11 +11,6 @@ export abstract class Resource {
   @DateTimeField()
   readonly createdAt: DateTime;
 
-  @Field({
-    description: 'Whether the requesting user can delete this resource',
-  })
-  readonly canDelete: boolean;
```
CarsonF, line 17: This needs to be reverted. The UI uses this to determine if the resource can be deleted or not.

src/components/budget/dto/budget-record.dto.ts
```diff
@@ -17,4 +18,6 @@ export class BudgetRecord extends Resource {
 
   @Field()
   readonly amount: SecuredFloatNullable;
+
+  readonly canDelete: SecuredBoolean;
```
CarsonF, line 22: canDelete is not a secured boolean. The boolean represents whether the requesting user can delete the resource or not. It doesn't make sense for the user to not be able to read it and there's nothing to edit there.

I'm assuming this is just trying to sync between java. I'd suggest dropping all the canDelete changes and additions as we haven't really worked out how they will work and they are not the same as properties.

CarsonF (changes requested):

michaelmarshall:
closing. replaced by #1656

Issue #1646, opened 2021-02-17
Funding Partner Missing After User Attempted to Delete and Re-Add


ParkeBrown:
Field Project Manager: "Chakhi 3 Seed Co is not listed as partner. Neither I nor my supervisor can see it. Seed Co should be the funding partner. I admit that this is most likely my fault. When I was unable to change Seed Co from Resource to Funding partner, I deleted Seed Co as partner, then tried to add them back as Funding partner, but CORD says Seed Co already exists."

CarsonF:
I believe this should be covered by #1661 and #1647

blakekrammes:
I fully deleted the Partnership node manually in the db and added Seed Company as the funding partner on the project.

CarsonF:
Did you add via API? If not you need to run the java procedure as well

blakekrammes:
I added via UI, so yes.

ParkeBrown:
@blakekrammes does that complete this limited fix request then?

blakekrammes:
I think so, yes. I'll close.

PR #1661, opened 2021-02-23
Fix uniqueness check not accounting for deleted nodes

CarsonF:
- Fix engagement, partnership, project member uniqueness checks not accounting for deleted nodes
- Merge node existence & relationship uniqueness to same query and return better error messages to consumer

Related to #1646


michaelmarshall:
@CarsonF queries look good. good job.

PR #1682, opened 2021-02-25
Secure updates

elisabeth-sorrell:
- fix tests
- start migrating components to the new updateProperties function
- add a check for user/unavailability/education. if the user id it's associated with is the same as the session userId, then update the properties without checking the write permission. If not, then check the write permissions of the properties
- bug fix: the role name for the lead financial analyst fixed to lead instead of regular FA. (Going to fix some permission restrictions for the LFA)

### DIFF
--- a/src/components/user/unavailability/unavailability.service.ts
+++ b/src/components/user/unavailability/unavailability.service.ts
@@ -1,5 +1,5 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
-import { node } from 'cypher-query-builder';
+import { node, relation } from 'cypher-query-builder';
 import {
   generateId,
   NotFoundException,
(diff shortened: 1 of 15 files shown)


src/core/database/database.service.ts
```diff
@@ -186,6 +241,20 @@ export class DatabaseService {
...
+      ) {
+        continue;
+      }
+      if (!(await this.authorizationService.checkIfCanEdit(object, prop))) {
+        throw new UnauthorizedException(
+          'You do not have permission to update this property',
+          prop
+        );
+      }
+    }
```
CarsonF, line 257: I thought we said we wanted to keep authorization checks out of this class.

This could be an easy
```ts
authorization.verifyCanEdit(currentObject, changes);
```
Which would conditionally throw an exception.

src/core/database/database.service.ts
```diff
@@ -170,6 +175,55 @@ export class DatabaseService {
     });
   }
 
+  // Used for user, education and unavailability, when a user wants to
+  // update their own properties.
+  async updatePropertiesInsecure<TObject extends Resource>({
```
CarsonF, line 180: Can we just add an option to skip authorization checks on `updateProperties` instead of duplicating all of this?

```tsx
updateProperties({
  ...
  skipAuth: true,
});
```

test/location.e2e-spec.ts
```diff
@@ -69,6 +77,7 @@ describe('Location e2e', () => {
   it('update location', async () => {
     const st = await createLocation(app);
     const newName = faker.company.companyName();
+    await registerUser(app, { roles: [Role.Administrator] });
```
CarsonF, line 80: can just use `runAsAdmin()` instead of `register/login`
```tsx
await runAsAdmin(app, async () => {
  await mutate(...);
});

File: src/components/user/unavailability/unavailability.service.ts [hunk 4]
```diff
@@ -165,14 +165,37 @@ export class UnavailabilityService {
...
+      });
+    } else {
+      return await this.db.sgUpdateProperties({
+        session,
+        object: unavailability,
+        props: ['description', 'start', 'end'],
+        changes: input,
+        nodevar: 'unavailability',
+      });
+    }
```
CarsonF, line 198: This, and the other places like this, can be simplified.

```suggestion
    return await this.db.updateProperties({
      type: Unavailability,
      object: unavailability,
      props: ['description', 'start', 'end'],
      changes: input,
      // assume current user has permission to edit themselves since the auth doesn't account for this yet.
      skipAuth: result?.unavailabilityUserId === session.userId,
    });
```

src/core/database/database.service.ts
```diff
@@ -86,11 +87,15 @@ export interface ServerInfo {
   edition: string;
 }
 
+forwardRef(() => AuthorizationService);
+
```
CarsonF, line 91:
```suggestion
```

CarsonF:
@elisabeth-sorrell Do you want to close this in favor of #1726?

elisabeth-sorrell:
> @elisabeth-sorrell Do you want to close this in favor of #1726?

@CarsonF sounds good to me. this was just an intermediary branch in case we wanted some sort of security layer

PR #1686, opened 2021-02-26
Bugfix synchronize terminal project/engagement statuses

blakekrammes:


src/components/engagement/dto/status.enum.ts
```diff
@@ -26,6 +26,29 @@ export enum EngagementStatus {
   Rejected = 'Rejected',
 }
 
+export const TerminalEngagementStatuses: EngagementStatus[] = [
+  EngagementStatus.Terminated,
+  EngagementStatus.Completed,
+  EngagementStatus.Converted,
+  EngagementStatus.Unapproved,
```
blakekrammes, line 33: I'm actually wondering if Unapproved is terminal.

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -13,35 +14,35 @@ export class UpdateProjectStatusHandler
...
-    if (
-      previous.status !== ProjectStatus.InDevelopment ||
-      updated.status !== ProjectStatus.Active
-    ) {
-      return;
-    }
-
-    // Update engagement status to Active
-    const engagements = await this.db
+  private async getEngagements(projectId: string) {
```
CarsonF, line 17: Can these private methods be moved below the entrypoint, `handle()`?

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -53,4 +54,52 @@ export class UpdateProjectStatusHandler
...
+      // filter out any engagements that have already reached their terminus
+      const engagements = (await this.getEngagements(updates.id)).filter(
+        (e) => {
+          if (
+            !TerminalEngagementStatuses.includes(e.status as EngagementStatus)
+          ) {
+            return true;
+          }
+          return false;
+        }
```
CarsonF, line 69:
```suggestion
        (e) => !TerminalEngagementStatuses.includes(e.status)
```

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -53,4 +54,52 @@ export class UpdateProjectStatusHandler
       })
     );
   }
+
+  async handle({ previous, updated, updates, session }: ProjectUpdatedEvent) {
+    if (updated.status === ProjectStatus.Terminated) {
+      // filter out any engagements that have already reached their terminus
+      const engagements = (await this.getEngagements(updates.id)).filter(
```
CarsonF, line 61: maybe `nonTerminalEngagements` would be clearer

src/components/engagement/dto/status.enum.ts
blakekrammes: @sethmcknight is `Unapproved` a terminal engagement status?

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -53,4 +54,52 @@ export class UpdateProjectStatusHandler
       })
     );
   }
+
+  async handle({ previous, updated, updates, session }: ProjectUpdatedEvent) {
+    if (updated.status === ProjectStatus.Terminated) {
```
CarsonF, line 59: Need to check `previous` as well. We don't want to run this code every time the project is updated after it's terminated.

src/components/project/dto/step.enum.ts
```diff
@@ -61,4 +61,10 @@ export abstract class ProjectStepTransition {
 
   @Field(() => TransitionType)
   type: TransitionType;
+
+  @Field(() => Boolean, { defaultValue: false })
+  disabled?: boolean;
+
+  @Field(() => String, { defaultValue: '' })
+  disabledLabel?: string;
```
CarsonF, line 69: Maybe `disabledReason` would be clearer? This should be nullable too. Empty strings should not be used to represent a lack of a value.

src/components/project/project.rules.ts
```diff
@@ -681,6 +682,16 @@ export class ProjectRules {
           ],
         };
       case ProjectStep.FinalizingCompletion:
+        const disabled = await (async () => {
```
CarsonF, line 685: This closure is unnecessary.

src/components/project/project.rules.ts
```diff
@@ -702,6 +713,10 @@ export class ProjectRules {
               to: ProjectStep.Completed,
               type: TransitionType.Approve,
               label: 'Complete 🎉',
+              disabled,
+              disabledLabel: disabled
+                ? 'The project cannot be Completed if any engagements have a non-terminal status'
```
CarsonF, line 718: I think it would be better to replace the "if wording" with "assertion wording" since we know it's true.
```suggestion
                ? 'The project cannot be completed since some engagements have a non-terminal status'
```

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -13,35 +14,35 @@ export class UpdateProjectStatusHandler
...
         node('project'),
         relation('out', '', 'engagement', { active: true }),
         node('engagement'),
+        relation('out', '', 'status', { active: true }),
+        node('sn', 'Property'),
       ])
-      .return('engagement.id as id')
-      .asResult<{ id: string }>()
+      .return('engagement.id as id, sn.value as status')
+      .asResult<{ id: string; status: string }>()
```
CarsonF, line 29:
```suggestion
      .asResult<{ id: string; status: EngagementStatus }>()
```
Then you can drop the typecast below

CarsonF (changes requested):

src/components/engagement/handlers/update-project-status.handler.ts
blakekrammes: I just wanted to use terminus in a sentence

src/components/project/dto/step.enum.ts
blakekrammes: Hmm. Why should this be nullable, but not `disabled`? It seems like every time `disabled` is true `disabledReason` would exist.

blakekrammes: I guess then you can have consistent boolean checks for `disabled`.

src/components/engagement/dto/status.enum.ts
blakekrammes: There aren't any engagements in the db with this status, so I'm not too worried about it. Guessing it's obsolete.

CarsonF:
What's the word on this @blakekrammes? It looks really close.

src/components/engagement/dto/status.enum.ts
blakekrammes: `Unapproved` is the equivalent of `DidNotDevelop`, so yes it is terminal. This status and probably some others should most likely be removed, but not in this PR.

blakekrammes:
I think it is just about ready. Looks like the GQL schema may need to be updated, but I think that's it. Also need to merge after https://github.com/SeedCompany/cord-field/pull/742.

src/components/engagement/dto/status.enum.ts
```diff
@@ -26,6 +26,29 @@ export enum EngagementStatus {
...
+  EngagementStatus.InDevelopment,
+  EngagementStatus.Active,
+  EngagementStatus.DiscussingTermination,
+  EngagementStatus.DiscussingReactivation,
+  EngagementStatus.DiscussingChangeToPlan,
+  EngagementStatus.DiscussingSuspension,
+  EngagementStatus.FinalizingCompletion,
+  EngagementStatus.ActiveChangedPlan,
+  EngagementStatus.Suspended,
+];
```
CarsonF, line 50: If we are actually trying to partition the statuses into terminal/non-terminal, it would be better to define this as a map.
```ts
const EngagementStatusTerminalMap: Record<EngagementStatus, boolean> = {
  [EngagementStatus.InDevelopment]: false,
   ...
};
```
This will ensure that if/when statuses are added or removed that those changes are reflected here as well.

This mapping could be converted into two lists for consumption like they are now.

src/components/project/project.rules.ts
```diff
@@ -831,6 +842,22 @@ export class ProjectRules {
...
+      .match([node('project', 'Project', { id: projectId })])
+      .match([
+        node('project'),
+        relation('out', '', 'engagement', { active: true }),
+        node('engagement'),
+        relation('out', '', 'status', { active: true }),
+        node('sn', 'Property'),
+      ])
+      .return('engagement.id as id, sn.value as status')
+      .asResult<{ id: string; status: string }>()
```
CarsonF, line 857:
```suggestion
      .asResult<{ id: string; status: EngagementStatus }>()
```

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -1,47 +1,97 @@
...
-      .return('engagement.id as id')
-      .asResult<{ id: string }>()
+      .return('engagement.id as id, sn.value as status')
+      .asResult<{ id: string; status: EngagementStatus }>()
       .run();
+  }
 
+  private async updateEngagements(
+    status: EngagementStatus,
+    engagements: Array<{ id: string; status: string }>,
```
CarsonF, line 84:
```suggestion
    engagements: Array<{ id: string; status: EngagementStatus }>,
```

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -1,47 +1,97 @@
...
+      (t) => t[0] === updated.step.value && t[0] !== previous.step.value
+    ) ?? [null, null];
+    if (!engagementStatus) return;
+    // we don't allow completing a project if there are non terminal engagements unless the user can bypass workflow constraints
+    // in which case update all non terminal engagements to Completed
     if (
-      previous.status !== ProjectStatus.InDevelopment ||
-      updated.status !== ProjectStatus.Active
+      projectStep === ProjectStep.Completed &&
+      !(await this.projectRules.canBypassWorkflow(session))
```
CarsonF, line 41: I thought we said this check wasn't actually needed?

CarsonF (changes requested):

src/components/engagement/handlers/update-project-status.handler.ts
blakekrammes: If it's not accessible via the FE I guess not. If a non admin updated via playground it would have undesired consequences. Probably everyone using playground will only be making updates as an admin.

CarsonF: No we should not assume that if possible. You need check for enabled [here](https://github.com/SeedCompany/cord-api-v3/blob/1a650c80461fa19f5cfa6f4356c933c7ee557304/src/components/project/project.rules.ts#L835), which should prevent us from getting here if we aren't supposed to.

src/components/engagement/handlers/update-project-status.handler.ts
```diff
@@ -14,34 +15,73 @@ export class UpdateProjectStatusHandler
...
+    const updateStatusTuples: Array<[ProjectStatus, EngagementStatus]> = [
+      [
+        ProjectStatus.Terminated,
+        updated.step.value === ProjectStep.Rejected
+          ? EngagementStatus.Rejected
+          : EngagementStatus.Terminated,
+      ],
+      [ProjectStatus.DidNotDevelop, EngagementStatus.DidNotDevelop],
+      [ProjectStatus.Completed, EngagementStatus.Completed],
+      [ProjectStatus.Active, EngagementStatus.Active],
```
blakekrammes, line 28: I changed these to `ProjectStatus`es instead because if an override to an `Active` status with a step that isn't `Active` (like `DiscussingChangeToPlan`) goes through then the engagements wouldn't get updated. By checking statuses all cases are covered I believe. The only time we need to disambiguate is with `EngagementStatus.Rejected` and `EngagementStatus.Terminated`, which is covered on `22-24`.

CarsonF (approved):

PR #1690, opened 2021-03-01
added config for jest vscode extension

michaelmarshall:
for vscode users

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,14 @@
       "processId": "${command:PickProcess}",
       "restart": true,
       "protocol": "inspector"
+    },
+    {
+      "type": "node",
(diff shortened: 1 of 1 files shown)


blakekrammes:
This isn't perfect unfortunately with VS Code's code lens. I've been refactoring some tests using `it.each([])`, which takes an array of inputs (Carson provided me the link end of last week) and applies them to the same test. The Jest extension doesn't pick this up correctly. I've just been using `it.only.each()` as a workaround. For simple individual tests it should work fine though.

CarsonF:
I don't think there's a way to pick out individual items of an each array. It's not really intended to be used that way I don't think.

blakekrammes:
Right. I just meant the array of tests as a single entity.

Issue #1645, opened 2021-02-17
Changing/reseting password should invalidate all current sessions


CarsonF:
## Environment

| Tech      | Version / Name |
| --------- | -------------- |
| API       |                |
| FE        |                |
| OS/Device |                |
| Browser   |                |

CarsonF:
@zkhin Can you address the comment on that merged PR?

zkhin:
@CarsonF Here's [PR](https://github.com/SeedCompany/cord-api-v3/pull/1698).

PR #1693, opened 2021-03-02
Remove current session after changing password

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1645

### DIFF
--- a/src/components/authentication/authentication.service.ts
+++ b/src/components/authentication/authentication.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { EmailService } from '@seedcompany/nestjs-email';
-import { node, relation } from 'cypher-query-builder';
+import { node, not, relation } from 'cypher-query-builder';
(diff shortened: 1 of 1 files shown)


CarsonF (changes requested):
Glad to have you back @zkhin 💪 

I think the changing your current password should keep you logged in with the current session, but it should invalidate all the rest.

The same thing should happen on `resetPassword` as well.

Sorry for a description-less ticket 😬

michaelmarshall (approved):

src/components/authentication/authentication.service.ts
```diff
@@ -390,6 +403,13 @@ export class AuthenticationService {
         }
       )
       .first();
+
+    // remove all the email tokens
+    await this.db
+      .query()
+      .match([node('emailToken', 'EmailToken', { value: result.email })])
+      .delete('emailToken')
+      .run();
```
CarsonF, line 412: Old `:token` relations need to inactivated here as well.

PR #1698, opened 2021-03-02
Invalidate old tokens after reset password

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1645

### DIFF
--- a/src/components/authentication/authentication.service.ts
+++ b/src/components/authentication/authentication.service.ts
@@ -355,7 +355,10 @@ export class AuthenticationService {
     });
   }
 
-  async resetPassword({ token, password }: ResetPasswordInput): Promise<void> {
+  async resetPassword(
+    { token, password }: ResetPasswordInput,
+    session: Session
+  ): Promise<void> {
     const result = await this.db
       .query()
       .raw(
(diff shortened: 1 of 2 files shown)


src/components/authentication/authentication.service.ts
```diff
@@ -404,12 +404,24 @@ export class AuthenticationService {
...
+      .query()
+      .match([
+        node('emailAddress', 'EmailAddress', { value: result.email }),
+        relation('in', '', 'email', { active: true }),
+        node('user', 'User'),
+        relation('out', 'oldRel', 'token', { active: true }),
+        node('token', 'Token'),
+      ])
+      .setValues({ 'oldRel.active': false })
+      .run();
```
CarsonF, line 424: We still need to avoid the current session here.

CarsonF (changes requested):

src/components/authentication/authentication.service.ts
zkhin: Do we have a valid `session.token` while resetting the password?

zkhin: I don't think we have `LoggedInSession` in the `resetPassword` resolver.

CarsonF: Ah. You can add an anonymous session arg, which has a token, just `userId` is optional.
```ts
```

zkhin: Done

CarsonF (approved):
Sweet!

Issue #1694, opened 2021-03-02
Reassign all users with the role of `Development` to the role of `Fundraising`


sethmcknight:
seedcompany/management#1

blakekrammes:
@michaelmarshall If you have time to work this there are 16 users with active Development roles. There are no team members on projects that have this role thankfully, which makes this update easier. Here are the roles below.

["5c3d88a07219425288a3cc24", "5c3d88a27219425288a3cc31", "5c3d88a37219425288a3cc33", "5c3d88a77219425288a3cc57", "5c4278d69503d5cd78e82bf8", "5c4278d69503d5cd78e82bfc", "5c4278d69503d5cd78e82bff", "5c4278d69503d5cd78e82c00", "5c4278d69503d5cd78e82c08", "5c4278d69503d5cd78e82c11", "5c4741e07e96a6307636844d", "5c4a1f997e96a66acb39478a", "5c4a1fbb7e96a649f339479a", "5c4f82ec8a0ddd0b79b5c69b", "5c950f2919eefb5ae4ec52b8", "5d2787f2231ca435b5524307"]

blakekrammes:
@sethmcknight, @michaelmarshall  I think we need to make updates to the API roles as well. 

Here are the roles Seth requested to be remapped previously:
```
Writer > Marketing
Executive Development Representative & Office of the President > Fundraising (Fundraising should be updated to Advancement)
Supporting Project Manager > Project Manager
Executive Leadership > Leadership
```

This is the current list of roles we have in the API with outdated role names struck through:

Administrator 
  BibleTranslationLiaison 
  Consultant 
  ConsultantManager 
  Controller 
  ~~Development~~ 
  ~~ExecutiveDevelopmentRepresentative~~
  ~~ExecutiveLeadership~~
  FieldOperationsDirector 
  FieldPartner 
  FinancialAnalyst 
  Fundraising 
  Intern 
  LeadFinancialAnalyst 
  Leadership 
  Liaison 
  Marketing 
  Mentor 
  ~~OfficeOfThePresident~~
  ProjectManager 
  RegionalCommunicationsCoordinator 
  RegionalDirector 
  StaffMember 
  ~~SupportingProjectManager~~ 
  Translator 
  ~~Writer~~ 
  
-------------------
@sethmcknight  can you confirm the above list?

sethmcknight:
@blakekrammes @michaelmarshall I'll make the changes to these users

[...]

PR #1702, opened 2021-03-03
Remove obsolete roles

blakekrammes:
Waiting on https://github.com/SeedCompany/cord-api-v3/issues/1694 to be closed. After it's closed we should run the following queries in the db to verify no users or project members have the role of `Development`.

These are the queries I ran to verify the roles didn't exist in the db:

`match(u:User)-[r:roles{active:true}]-({value:""}) return count(u)`
```
match(u:ProjectMember)-[r:roles{active:true}]-(rn)
where any(x in rn.value where x = "")
return count(u)
```

### DIFF
--- a/src/components/authorization/dto/role.dto.ts
+++ b/src/components/authorization/dto/role.dto.ts
@@ -7,9 +7,6 @@ export enum Role {
   Consultant = 'Consultant',
   ConsultantManager = 'ConsultantManager',
   Controller = 'Controller',
-  Development = 'Development',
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
You've confirmed none of these roles exist in db?

blakekrammes:
Just left a comment ^

blakekrammes:
Seth has closed #1694 and I have verified that there are no users or project members with the role of `Development` in the db. This is ready to be merged.

CarsonF (approved):

michaelmarshall (approved):

Issue #1610, opened 2021-01-28
Make the email field an optional field when creating a person. We should be able to come back and add an email to the person later.


sethmcknight:
The email field is currently showing as required when creating a Person. If a user attempts to create a Person without an email address entered they get an error that it is required and are not allowed to create the person.

Users should be able to create a person with only a First Name and Last Name

sethmcknight:
This appears to be reverted.

The email field is currently showing as required when creating a Person. If a user attempts to create a Person without an email address entered they get an error that it is required and are not allowed to create the person.

Users should be able to create a person with only a First Name and Last Name

CarsonF:
Probably a UI change needed

mosesintech:
While investigating another open issue, [I found that creating a new user without an email address would give me a Neo4j error](https://github.com/SeedCompany/cord-api-v3/issues/1822). The GraphQL schema doesn't care if there is an email address or not, but we have two database constraints for email addresses that would cause a database error if the email is not provided: 
```cql
CREATE CONSTRAINT ON (n:EmailAddress) ASSERT EXISTS(n.value)
CREATE CONSTRAINT ON (n:EmailAddress) ASSERT n.value IS UNIQUE
```
It looks like if we remove the first constraint, Neo4j will accept multiple null values for user emails.

michaelmarshall:
I think our API is functioning as it should. Project managers can create a 'person' without creating a registered user. Let's have Carson attack this from a UI perspective first.

ParkeBrown:
Clearing assignees as is backlog. Reassign and move if picking up.

PR #1705, opened 2021-03-03
Create person email field optional

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1610

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -22,6 +22,7 @@ import {
   OnIndex,
   property,
   setPropLabelsAndValuesDeleted,
+  Transactional,
   UniquenessError,
   UniqueProperties,
 } from '../../core';
@@ -169,20 +170,6 @@ export class UserService {
       property('roles', role, 'user', `role${role}`)
     );
   };
-
-  emailProperty = (email: string | undefined, createdAt: DateTime) =>
-    email
-      ? [
-          relation('out', '', 'email', {
-            active: true,
-            createdAt,
-          }),
-          node('email', 'EmailAddress:Property', {
-            value: email,
-            createdAt,
-          }),
-        ]
-      : [];
 
   async create(input: CreatePerson, _session?: Session): Promise<string> {
     const id = await generateId();
@@ -350,6 +337,7 @@ export class UserService {
     };
   }
 
+  @Transactional()
   async update(input: UpdateUser, session: Session): Promise<User> {
     this.logger.debug('mutation update User', { input, session });
     const user = await this.readOne(input.id, session);
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -21,6 +21,7 @@ import {
   matchSession,
   OnIndex,
   property,
+  setPropLabelsAndValuesDeleted,
   UniquenessError,
   UniqueProperties,
 } from '../../core';
@@ -194,8 +195,8 @@ export class UserService {
           id,
           createdAt,
         }),
-        ...this.emailProperty(input.email, createdAt),
       ],
+      ...property('email', input.email, 'user', 'email', 'EmailAddress'),
       ...property('realFirstName', input.realFirstName, 'user'),
       ...property('realLastName', input.realLastName, 'user'),
       ...property('displayFirstName', input.displayFirstName, 'user'),
@@ -374,15 +375,14 @@ export class UserService {
     // Update email
     if (input.email) {
       // Remove old emails and relations
+      const uniqueProperties: UniqueProperties<User> = {
+        email: ['Property', 'EmailAddress'],
+      };
       await this.db
         .query()
-        .match([
-          node('user', ['User', 'BaseNode'], { id: user.id }),
-          relation('out', 'oldRel', 'email', { active: true }),
-          node('email', 'EmailAddress:Property'),
-        ])
-        .delete('oldRel')
-        .delete('email')
+        .match([node('node', ['User', 'BaseNode'], { id: user.id })])
+        .call(setPropLabelsAndValuesDeleted, uniqueProperties)
+        .return('*')
         .run();
 
       try {
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -169,6 +169,20 @@ export class UserService {
     );
   };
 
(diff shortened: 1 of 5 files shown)


src/components/user/user.service.ts
```diff
@@ -366,6 +373,49 @@ export class UserService {
...
+      // Remove old emails and relations
+      await this.db
+        .query()
+        .match([
+          node('user', ['User', 'BaseNode'], { id: user.id }),
+          relation('out', 'oldRel', 'email', { active: true }),
+          node('email', 'EmailAddress:Property'),
+        ])
+        .delete('oldRel')
+        .delete('email')
```
CarsonF, line 387: I think we just we the old relationships marked `active: false` since of deleting them. For history.

src/components/user/user.service.ts
```diff
@@ -366,6 +373,49 @@ export class UserService {
...
+            relation('out', '', 'email', {
+              active: true,
+              createdAt,
+            }),
+            node('email', 'EmailAddress:Property', {
+              value: input.email,
+              createdAt,
+            }),
+          ])
+          .run();
```
CarsonF, line 406: Actually I think this whole thing can just be merged into `sgUpdateProperties` call. You could still try/catch that to have the uniqueness check there.

src/components/user/user.service.ts
```diff
@@ -178,14 +192,7 @@ export class UserService {
...
         }),
-        relation('out', '', 'email', {
-          active: true,
-          createdAt,
-        }),
-        node('email', 'EmailAddress:Property', {
-          value: input.email,
-          createdAt,
-        }),
+        ...this.emailProperty(input.email, createdAt),
```
CarsonF, line 195: I think we always want the property defined even if it's null. This way we don't have to do optional matches later.
```suggestion
        ...property('email', input.email, 'user', 'email', 'EmailAddress'),
```

CarsonF (changes requested):

src/components/user/user.service.ts
zkhin: Hmm... if we merge into `sgUpdateProperties`, the `email` property will not have `EmailAddress` label. @michaelmarshall Can you confirm this part?

zkhin:
@CarsonF I think it will throw an error if we don't remove the old `emailAddress` following the below steps.
- Create a person without an email
- Change email to `aaa@test.com`
- Change email to `bbb@test.com`
- Change email to `aaa@test.com`: it will throw an error because of the `emailAddress` uniqueness.

CarsonF:
Can that constraint be changed to include `active: true`?

zkhin:
Do you mean the `emailAddress` node or relation?

zkhin:
@michaelmarshall @CarsonF If we don't want to remove the old emails, I think it will be a little complicated while updating the email.
- If the inputted email does not exist in the `emailAddress`, we can follow the current approach.
- If the email exists, need to check if it's already used by others (rel: { `active: true` })
   - If it's already used by others, throw an exception
   - If it's not used, create a relationship with that email with `active=true`

src/components/user/user.service.ts
michaelmarshall: Any property that has a uniqueness constraint must have a unique label. In order for email address to be unique, they must have the `EmailAddress` label.

michaelmarshall:
Ah yes, we have encountered this before with other deletes. Here is what we need to do:
- When a delete occurs on a uniqueness constrained node/property you:
1. set the relationship to it to `active: false`
If its a `:Property` node:
   2. create a property on the `:Property` node called `deletedValue` and set it to the deleted value.
   3. set the `value` property to `null`
If its a `:BaseNode` node:
   2. create a property on the `:BaseNode` called `deletedId` and set it to the id of the `:BaseNode`
   3. set the `id` property to `null`

michaelmarshall (approved):

src/components/user/user.service.ts
```diff
@@ -169,6 +170,20 @@ export class UserService {
...
+            active: true,
+            createdAt,
+          }),
+          node('email', 'EmailAddress:Property', {
+            value: email,
+            createdAt,
+          }),
+        ]
+      : [];
+
```
CarsonF, line 186:
```suggestion
```

src/components/user/user.service.ts
```diff
@@ -364,6 +372,48 @@ export class UserService {
...
+        if (e instanceof UniquenessError && e.label === 'EmailAddress') {
+          throw new DuplicateException(
+            'person.email',
+            'Email address is already in use',
+            e
+          );
+        }
+        throw new ServerException('Failed to create user', e);
+      }
+    }
```
CarsonF, line 415: Since there's the potential for the user's email to be deleted only to determine that the new email address is already in use means we are leaving the db in a broken state.

Thankfully, I think we can just add the `@Transactional()` decorator to this update() method which should rollback the entire transaction when we throw an error.

test/utility/create-person.ts
```diff
@@ -43,7 +43,7 @@ export async function createPerson(
   expect(actual).toBeTruthy();
 
   expect(isValidId(actual.id)).toBe(true);
-  expect(actual.email.value).toBe(person.email.toLowerCase());
+  expect(actual.email.value).toBe(person.email?.toLowerCase());
```
CarsonF, line 46:
```suggestion
```

There shouldn't really be any assertions here. The id is really just a sanity check that the gql call was successful. 

If we want to assert the email then that should be directly in a test case.

CarsonF (changes requested):

CarsonF:
I'd like to confirm that a user can try change email their email to one that's already in use and they get the correct error back and then logout and log back in. Just to confirm that the DB was left in a clean state.

zkhin:
> I'd like to confirm that a user can try change email their email to one that's already in use and they get the correct error back and then logout and log back in. Just to confirm that the DB was left in a clean state.

I tried to confirm this part following the below steps:
- Create A user with `aaa@test.com`
- Create B user with `bbb@test.com` 
- Update B's email with `aaa@test.com` and it throws the error `Email is already in use`
- Query B's email and it's still `bbb@test.com`

`@Transactional()` decorator works like a charm!

CarsonF (approved):
Thanks @zkhin!

PR #1706, opened 2021-03-04
Update Powers to use app logic instead of DB SecurityGroups

CarsonF:


CarsonF:
Looks like this is going to be some work to get our test suite updated for the differences in powers from our TS roles.

CarsonF:
Closing in favor of https://github.com/SeedCompany/cord-api-v3/pull/2400

PR #1713, opened 2021-03-05
Query base node labels on delete

blakekrammes:
1. I believe the reason for errors regarding the uploading of files https://github.com/SeedCompany/cord-field/issues/748 has to do with the previous deletion of files with the same name in the same directory. 

2. Along with this change we need to also run a few queries in the db to change any `FileNode` labels to `Deleted_FileNode` labels. This can be accomplished by querying for `Deleted_File`, `Deleted_Directory` and `Deleted_File_Version` nodes. 

Because the label FileNode wasn't being passed in for node labels to be marked as `Deleted`, the query here matched a deleted FileNode and then returned an error of `Not Found` to the front end: 

https://github.com/SeedCompany/cord-api-v3/blob/2ed15ae2d88f007303f5f8323593244db9ef1929/src/components/file/file.repository.ts#L61-L75

This query also matched because this function isn't deactivating the deleted file's relationships here: 

https://github.com/SeedCompany/cord-api-v3/blob/2ed15ae2d88f007303f5f8323593244db9ef1929/src/core/database/database.service.ts#L558-L588

Note the relationship on 573 is only pointing at `BaseNodes` going **into** the file node to be deleted, not **out**, which is what we need for the pertinent relationship (on 69 of `file.repository`) to be deactivated in this case: 
`(fn:FileNode)-[p:parent]->(d:Directory)`

 My first solution was to make that relation have a direction of `either`, but I didn't know what ramifications that would have since this function is used for all deletes.
 
 I wonder if we have a need for new labels with a `Delete_` prefix or if we can just deactivate all relationships and replace the `id` property with `deleted_id` on the deleted node. Michael's comment here on deletion protocol didn't mention `Delete_` prefixed node labels: https://github.com/SeedCompany/cord-api-v3/pull/1705#issuecomment-790685440. There's work to be done here still, but I think this one liner will fix the bug for users pending running the queries above in the prod db. 
 

See https://github.com/SeedCompany/cord-api-v3/issues/1717 to perhaps prevent future bugs of this sort.

### DIFF
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -557,16 +557,20 @@ export class DatabaseService {
 
   async deleteNodeNew<TObject extends Resource>({
     object,
-    baseNodeLabels,
     uniqueProperties = {},
   }: {
     object: TObject;
-    baseNodeLabels: string[];
     uniqueProperties?: UniqueProperties<TObject>;
   }) {
+    const nodeLabelsResult = await this.db
+      .query()
+      .matchNode('node', { id: object.id })
+      .return('labels(node) as nodeLabels')
+      .asResult<{ nodeLabels: string[] }>()
+      .first();
     const query = this.db
       .query()
-      .match(node('node', { id: object.id }))
+      .matchNode('node', { id: object.id })
       //Mark any parent base node relationships (pointing to the base node) as active = false.
       .optionalMatch([
         node('node'),
@@ -580,7 +584,7 @@ export class DatabaseService {
       })
       .with('distinct(node) as node')
       //Mark baseNode labels and id deleted
-      .call(setBaseNodeLabelsAndIdDeleted, baseNodeLabels)
+      .call(setBaseNodeLabelsAndIdDeleted, nodeLabelsResult!.nodeLabels)
       //Mark unique property labels and values deleted
       .call(setPropLabelsAndValuesDeleted, uniqueProperties)
       .return('*');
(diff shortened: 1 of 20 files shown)


src/core/database/database.service.ts
```diff
@@ -557,16 +557,20 @@ export class DatabaseService {
...
-    baseNodeLabels,
     uniqueProperties = {},
   }: {
     object: TObject;
-    baseNodeLabels: string[];
     uniqueProperties?: UniqueProperties<TObject>;
   }) {
+    const nodeLabelsResult = await this.db
+      .query()
+      .matchNode('node', { id: object.id })
```
CarsonF, line 567:
```suggestion
      .matchNode('node', 'BaseNode', { id: object.id })
```
Should we limit this node selection to only base nodes?

src/core/database/database.service.ts
```diff
@@ -580,7 +584,7 @@ export class DatabaseService {
       })
       .with('distinct(node) as node')
       //Mark baseNode labels and id deleted
-      .call(setBaseNodeLabelsAndIdDeleted, baseNodeLabels)
+      .call(setBaseNodeLabelsAndIdDeleted, nodeLabelsResult!.nodeLabels)
```
CarsonF, line 587: I don't think we need change the ID, only the labels. I believe this can be done in one query as well. Perhaps this whole function needs to be rewritten.

CarsonF (changes requested):

blakekrammes:
merged with https://github.com/SeedCompany/cord-api-v3/pull/1718

PR #1718, opened 2021-03-05
Query unique property node labels on delete

blakekrammes:
Depends on https://github.com/SeedCompany/cord-api-v3/pull/1713. 

This eliminates the need to enumerate  property labels when deleting nodes and reduces the chance for bugs there.


CarsonF:
Perhaps this refactor can just be merged into the other PR

blakekrammes:
close in favor of https://github.com/SeedCompany/cord-api-v3/pull/1752

Issue #1704, opened 2021-03-03
Invalid relation IDs when modifying project should be client errors


CarsonF:
## Current Behavior

When creating or updating a project any ID's referencing other resources are not validated and invalid one give a ServerException or "success".

## Expected Behavior

When creating or updating a project any ID's referencing other resources should be validated.
This can happen before the modification query or ideally confirmed after the modification query fails (this optimizes the happy path).

i.e. 
```gql
createProject({ project: { primaryLocationId: 'invalid' } })
```
results in a
```tsx
throw new NotFoundException('Primary location not found', 'project.primaryLocationId');
```

PR #1720, opened 2021-03-06
Create/Update Project - Validate other resource ids

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1704

### DIFF
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -293,7 +293,7 @@ export class ProjectService {
           otherLocationIds,
           'Location',
           'otherLocationIds',
-          'Other locations not found'
+          'One of the other locations was not found'
         );
         otherLocationIds.forEach((id) => {
           createProject.match([node(`otherLocation${id}`, 'Location', { id })]);
--- a/src/components/project/project.service.ts
+++ b/src/components/project/project.service.ts
@@ -1,11 +1,13 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
 import { node, relation } from 'cypher-query-builder';
+import { Many } from 'lodash';
 import { DateTime } from 'luxon';
 import {
   DuplicateException,
   generateId,
   getHighestSensitivity,
   InputException,
+  many,
   NotFoundException,
   Sensitivity,
   ServerException,
@@ -266,7 +268,7 @@ export class ProjectService {
 
       if (fieldRegionId) {
         await this.validateOtherResourceId(
-          [fieldRegionId],
+          fieldRegionId,
           'FieldRegion',
           'fieldRegionId',
           'Field region not found'
@@ -277,7 +279,7 @@ export class ProjectService {
       }
       if (primaryLocationId) {
         await this.validateOtherResourceId(
-          [primaryLocationId],
+          primaryLocationId,
           'Location',
           'primaryLocationId',
           'Primary location not found'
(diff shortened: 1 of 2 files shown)


src/components/project/project.service.ts
```diff
@@ -1200,4 +1243,25 @@ export class ProjectService {
       ).every((n) => n)
     );
   }
+
+  protected async validateOtherResourceId(
+    ids: string[],
```
CarsonF, line 1248: This could be
```suggestion
    ids: Many<string>,
```
allowing a single item or a list to be passed in. So we don't have to wrap them in lists if they aren't.

It could be converted to a list below like this:
```tsx
for (const id of many(ids)) {
```

src/components/project/project.service.ts
```diff
@@ -1200,4 +1243,25 @@ export class ProjectService {
...
+  ): Promise<boolean> {
+    for (const id of ids) {
+      const result = await this.db
+        .query()
+        .match([node('node', label, { id })])
+        .return('node')
+        .first();
+
+      if (!result) {
+        throw new NotFoundException(errMsg, `project.${resourceField}`);
```
CarsonF, line 1261: If the input is actually an array we should specify which index has failed. i.e. `project.otherLocationIds[3]`

```suggestion
        throw new NotFoundException(errMsg, `project.${resourceField}${Array.isArray(ids) ? `[${index}]`: ''}`);
```

src/components/project/project.service.ts
```diff
@@ -578,16 +605,26 @@ export class ProjectService {
...
+          throw new InputException(
+            'Cannot connect location without a funding account',
+            'project.primaryLocationId'
+          );
+        }
+      } catch (e) {
+        if (e instanceof NotFoundException) {
+          throw new NotFoundException(
+            'Primary location not found',
+            'project.primaryLocationId'
```
CarsonF, line 623: Let's forward the exception we are wrapping on, so we can see the true error source in the logs
```suggestion
            'project.primaryLocationId',
            e
```

src/components/project/project.service.ts
```diff
@@ -1200,4 +1243,25 @@ export class ProjectService {
       ).every((n) => n)
     );
   }
+
+  protected async validateOtherResourceId(
+    ids: string[],
+    label: string,
+    resourceField: string,
+    errMsg: string
+  ): Promise<boolean> {
```
CarsonF, line 1252: We are just throwing an exception if needed, which is the best way. The boolean return does nothing. We don't want to suggest to callers of this that they need to check that boolean to determine success.
```suggestion
  ): Promise<void> {
```

CarsonF (changes requested):
Looking good @zkhin!

src/components/project/project.service.ts
```diff
@@ -265,21 +267,45 @@ export class ProjectService {
...
         createProject.match([
           node('primaryLocation', 'Location', { id: primaryLocationId }),
         ]);
       }
       if (otherLocationIds?.length) {
+        await this.validateOtherResourceId(
+          otherLocationIds,
+          'Location',
+          'otherLocationIds',
+          'Other locations not found'
```
CarsonF, line 296:
```suggestion
          'One of the other locations was not found'
```

CarsonF (commented):
One more thing...

CarsonF (approved):
Thanks @zkhin!

Issue #1611, opened 2021-01-28
Partnership Card - Primary Partner


sethmcknight:
Add a field to define which Partnership is the 'Primary Partner' (there can only be 1 per Project)

- What is the best UX for this?

PR #1725, opened 2021-03-08
Add primary partner to project

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1611

### DIFF
--- a/src/components/partnership/partnership.service.ts
+++ b/src/components/partnership/partnership.service.ts
@@ -70,6 +70,7 @@ export class PartnershipService {
     mou: true,
     agreement: true,
     partner: true,
+    primary: true,
   };
 
   constructor(
@@ -92,6 +93,12 @@ export class PartnershipService {
     const createdAt = DateTime.local();
 
     await this.verifyRelationshipEligibility(projectId, partnerId);
+
+    const isFirstPartnership = await this.isFirstPartnership(projectId);
+    if (!isFirstPartnership && input.primary) {
+      await this.setPrimaryForOldPartnerships(projectId, false);
+    }
+    const primary = isFirstPartnership ? true : input.primary;
 
     const partner = await this.partnerService.readOne(partnerId, session);
     this.verifyFinancialReportingType(
@@ -156,6 +163,12 @@ export class PartnershipService {
       {
         key: 'canDelete',
         value: true,
+        isPublic: false,
+        isOrgPublic: false,
+      },
+      {
+        key: 'primary',
+        value: primary,
         isPublic: false,
         isOrgPublic: false,
       },
@@ -369,6 +382,11 @@ export class PartnershipService {
       }
     }
 
+    if (input.primary) {
+      const projectId = await this.getProjectIdByPartnershipId(input.id);
+      await this.setPrimaryForOldPartnerships(projectId, false);
+    }
+
     const { mou, agreement, ...rest } = changes;
     await this.db.sgUpdateProperties({
       session,
@@ -380,6 +398,7 @@ export class PartnershipService {
         'financialReportingType',
         'mouStartOverride',
         'mouEndOverride',
+        'primary',
       ],
       changes: rest,
       nodevar: 'partnership',
@@ -424,6 +443,8 @@ export class PartnershipService {
         'You do not have the permission to delete this Partnership'
       );
 
+    const projectId = await this.getProjectIdByPartnershipId(id);
+
     await this.eventBus.publish(
       new PartnershipWillDeleteEvent(object, session)
     );
@@ -438,6 +459,31 @@ export class PartnershipService {
     } catch (exception) {
       this.logger.error('Failed to delete', { id, exception });
       throw new ServerException('Failed to delete', exception);
+    }
+
+    // Assign primary to other partnership if one exists
+    if (object.primary) {
+      const result = await this.db
+        .query()
+        .optionalMatch(node('project', 'Project', { id: projectId }))
+        .optionalMatch([
+          node('project'),
+          relation('out', '', 'partnership', { active: true }),
+          node('partnership'),
+        ])
+        .return('partnership.id as partnershipId')
+        .asResult<{ partnershipId: string }>()
+        .first();
+
+      if (result) {
+        await this.update(
+          {
+            id: result.partnershipId,
+            primary: true,
+          },
+          session
+        );
+      }
     }
   }
 
@@ -602,5 +648,57 @@ export class PartnershipService {
         'Partnership for this project and partner already exists'
       );
     }
+  }
+
+  protected async isFirstPartnership(projectId: string): Promise<boolean> {
+    const result = await this.db
+      .query()
+      .optionalMatch(node('project', 'Project', { id: projectId }))
+      .optionalMatch([
+        node('project'),
+        relation('out', '', 'partnership', { active: true }),
+        node('partnership'),
+      ])
+      .return(['partnership'])
+      .asResult<{ partnership?: Node }>()
+      .first();
+
+    return result?.partnership ? false : true;
+  }
+
+  protected async setPrimaryForOldPartnerships(
+    projectId: string,
+    value: boolean
+  ): Promise<void> {
+    await this.db
+      .query()
+      .optionalMatch(node('project', 'Project', { id: projectId }))
+      .optionalMatch([
+        node('project'),
+        relation('out', '', 'partnership', { active: true }),
+        node('partnership'),
+        relation('out', '', 'primary'),
+        node('property', 'Property'),
+      ])
+      .setValues({ 'property.value': value })
+      .first();
+  }
+
+  protected async getProjectIdByPartnershipId(
+    partnershipId: string
+  ): Promise<string> {
+    const result = await this.db
+      .query()
+      .match(node('partnership', 'Partnership', { id: partnershipId }))
+      .optionalMatch([
+        node('partnership'),
+        relation('in', '', 'partnership', { active: true }),
+        node('project', 'Project'),
+      ])
+      .return('project.id as projectId')
+      .asResult<{ projectId: string }>()
+      .first();
+
+    return result!.projectId;
   }
 }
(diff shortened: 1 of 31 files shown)


CarsonF (changes requested):
I think it would be better to have a `primary` boolean on `Partnership`. The primary partnership is still a partnership and should be listed in the list of project's partnerships. I could see `primaryPartnership` existing only as a GQL helper field to pull the primary partnership from the list when it is the only one needed.

Somethings to consider:
- [x] Only one partnership can be primary.
- [x] The first partnership created should be primary by default.
- [ ] If a primary partnership is deleted it should be assigned to another existing partnership if one exists.
- [ ] Creating or updating a partnership with primary=true should remove primary from a pre-existing partnership.

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +649,56 @@ export class PartnershipService {
       );
     }
   }
+
+  protected async isFirstPartnership(projectId: string): Promise<boolean> {
+    const result = await this.db
+      .query()
+      .optionalMatch(node('project', 'Project', { id: projectId }))
+      .optionalMatch([
+        node('project'),
```
CarsonF, line 658:
```suggestion
      .match([
        node('project', 'Project', { id: projectId }),
```

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +649,56 @@ export class PartnershipService {
...
+
+  protected async setPrimaryForOldPartnerships(
+    projectId: string,
+    value: boolean
+  ): Promise<void> {
+    await this.db
+      .query()
+      .optionalMatch(node('project', 'Project', { id: projectId }))
+      .optionalMatch([
+        node('project'),
```
CarsonF, line 677:
```suggestion
      .match([
        node('project', 'Project', { id: projectId }),
```

src/components/partnership/partnership.service.ts
```diff
@@ -369,6 +382,11 @@ export class PartnershipService {
       }
     }
 
+    if (input.primary) {
+      const projectId = await this.getProjectIdByPartnershipId(input.id);
+      await this.setPrimaryForOldPartnerships(projectId, false);
+    }
```
CarsonF, line 388: What do you think about a `movePrimaryPartnership` method instead? That could work for create, update, and delete.

logically,
- match current primary partnership, its project, and all other partnerships
- set current to primary false
- set first of other partnerships to primary true

I think that could consolidate logic and reduce queries.

src/components/partnership/partnership.service.ts
```diff
@@ -93,6 +95,12 @@ export class PartnershipService {
 
     await this.verifyRelationshipEligibility(projectId, partnerId);
 
+    const isFirstPartnership = await this.isFirstPartnership(projectId);
+    if (!isFirstPartnership && input.primary) {
+      await this.movePrimaryPartnership(false, { projectId: projectId });
+    }
```
CarsonF, line 101: Could this be moved after the partnership is created in order to simplify the move query? A partnershipId could always be given then.

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +627,71 @@ export class PartnershipService {
...
+  }
+
+  /**
+   *
+   * match current primary partnership, its project, and all other partnerships
+   * set current to primary false
+   * set first of other partnerships to primary true
+   */
+  protected async movePrimaryPartnership(
+    value: boolean,
```
CarsonF, line 653: Maybe this could be changed to 
```suggestion
    direction: 'to' | 'from',
```
from that we could set the booleans on line 685 & 693 accordingly. 
I think that would be clearer as well:
```ts
// create/update
movePrimaryPartnership('to', id);

// delete
movePrimaryPartnership('from', id);
```

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +627,71 @@ export class PartnershipService {
...
+      ])
+      .setValues({ 'property.value': false })
+      .with('project, otherPartnership')
+      .limit(1)
+      .optionalMatch([
+        node('otherPartnership'),
+        relation('out', '', 'primary'),
+        node('property', 'Property'),
+      ])
+      .setValues({ 'property.value': value });
```
CarsonF, line 693: This is looking good, but we need to account for history as well. The current props need to be set to active=false and new ones created with active=true. I know this complicates the query more. See [`DatabaseService.updateProperty`](https://github.com/SeedCompany/cord-api-v3/blob/d1319a880a9ce7d5a66c93265ace25ee7ae350e5/src/core/database/database.service.ts#L306-L325) for example.

CarsonF (changes requested):
Getting closer @zkhin!

src/components/partnership/partnership.service.ts
```diff
@@ -369,6 +384,10 @@ export class PartnershipService {
       }
     }
 
+    if (input.primary) {
```
CarsonF, line 387: We should check that the previous value is not primary so we don't call move unnecessarily.

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +628,67 @@ export class PartnershipService {
...
+      sortValue: determineSortValue(direction === 'from'),
+    };
+    const query = this.db
+      .query()
+      .match([
+        node('partnership', 'Partnership', { id: partnershipId }),
+        relation('in', '', 'partnership', { active: true }),
+        node('project', 'Project'),
+        relation('out', '', 'partnership', { active: true }),
+        node('otherPartnership'),
```
CarsonF, line 670: How do we know that `otherPartnership` doesn't match the current `partnership` as well?

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +628,67 @@ export class PartnershipService {
...
+      .with('otherPartnership')
+      .limit(1)
+      .optionalMatch([
+        node('otherPartnership'),
+        relation('out', 'oldRel', 'primary', { active: true }),
+        node('property', 'Property'),
+      ])
+      .setValues({
+        'oldRel.active': false,
+      })
```
CarsonF, line 681: If our direction is `to` then shouldn't this apply for all `otherPartnership` not just `limit 1`?

src/components/partnership/partnership.service.ts
```diff
@@ -603,4 +628,67 @@ export class PartnershipService {
...
+      .with('otherPartnership')
+      .create([
+        node('otherPartnership'),
+        relation('out', '', 'primary', {
+          active: true,
+          createdAt,
+        }),
+        node('newProperty', 'Property', newPropertyNodeProps),
+      ]);
```
CarsonF, line 690: Shouldn't this create apply to the selected `partnership` if direction is `to`?

src/components/partnership/partnership.service.ts
CarsonF: Wait we need to match the other one whose property is changing, and then create a property there as well with the opposite value.

src/components/partnership/partnership.service.ts
CarsonF: I think we just need a condition here so that `otherPartnership` matches the opposite value. That way we have both partnerships that need their primary property flipped.

CarsonF:
Happy to zoom about the query if you want.

src/components/partnership/partnership.service.ts
zkhin:
```
MATCH (partnership:Partnership { id: 'id' })<-[:partnership { active: true }]-(project:Project)-[:partnership { active: true }]->(otherPartnership)
```
This always returns all partnerships attached to `project` except selected `partnership`. FYI, I cannot find a reference doc for this tricky query.

zkhin:
Added an e2e test case for primary partnership logic

src/components/partnership/partnership.service.ts
michaelmarshall: I tested how Neo4j handles this kind of query by creating the following nodes with this cypher:
`create (:child {name: "child1"})<-[:toChild]-(:mother {name: "mom"})-[:toChild]->(:child{name:"child2"})`
I then queried using this:
`match (:child {name:"child1"})<-[:toChild]-(:mother)-[:toChild]->(child) return child.name`
and got 1 match: `"child1"`
and this query:
`match (:mother)-[:toChild]->(child) return child.name`
got 2 nodes returned: `"child1", "child2"`
I believe its safe to leave off a condition, but adding it will make developer intent clear.

CarsonF (changes requested):
Ok after talking with Seth I think we can make some more changes here.

On updates, we don't want to allow the primary to be set to false. Which we do already have in place, we just need to remove it from the updateProperties call.
We should throw an error if it is requested to be false.

On deletes, if the partnership is the primary and it is not the only one we want to throw an error.
If it is the only primary one then we can remove it fine.
This means that we don't need to handle moving "from" logic anymore which should clean up that query.

I think this is better as well, because it's not surprising to the user. Removing a primary partnership and another one randomly becoming primary is surprising.

Thanks for all being flexible with all the back and forth here @zkhin.

src/components/partnership/partnership.service.ts
```diff
@@ -383,8 +383,15 @@ export class PartnershipService {
       }
     }
 
+    if (input.primary === false) {
+      throw new InputException(
+        'Partnership primary could not be false',
```
CarsonF, line 388:
```suggestion
        'To remove primary from this partnership, set another partnership as the primary',
```

src/components/partnership/partnership.service.ts
```diff
@@ -443,9 +450,18 @@ export class PartnershipService {
...
-      await this.movePrimaryPartnership('from', object.id);
+    // only primary one partnership could be removed
+    if (object.primary.value) {
+      const result = await this.otherPartnershipQuery(object.id)
+        .return('otherPartnership')
+        .first();
+
+      if (result) {
+        throw new InputException(
+          'Partnership is not the only primary one',
```
CarsonF, line 461:
```suggestion
          'Primary partnerships cannot be removed. Make another partnership primary first.',
```

CarsonF (commented):
Looks good @zkhin. Would you mind tweaking the messages?

CarsonF (approved):
🎉

PR #1726, opened 2021-03-08
Secured updates via app permissions

elisabeth-sorrell:


src/components/authorization/authorization.service.ts
```diff
@@ -104,6 +109,37 @@ export class AuthorizationService {
...
 
+  async verifyCanEditChanges<
+    TObject extends Resource,
+    Key extends keyof TObject & string
+  >(
+    baseNode: TObject,
+    props: readonly Key[],
+    changes: { [Key in keyof TObject]?: UnwrapSecured<TObject[Key]> }
+  ) {
+    for (const prop of props) {
```
CarsonF, line 120: Part of my #1651 refactor was to prevent us from having to redefine prop lists everywhere. This same concept can be applied here as well. a.k.a. Let's pass in the `User` class instead of a list of prop keys.

Here's how that could work:
```suggestion
  async verifyCanEditChanges<Resource extends ResourceShape<any>>(
    resource: Resource,
    changes: Partial<UnsecuredDto<Resource['prototype']>>
  ) {
    for (const prop of resource.SecuredProps) {
```
`Resource` here is the "static" class, so we use `Resource['prototype']` to get a shape of the "instance" of the class.
Other than that, `changes` is the same type just written another way.

I think it's fine to verify all changes here. If we don't want that then the changes passed in can just be a subset from the caller.

src/components/authorization/authorization.service.ts
```diff
@@ -104,6 +109,37 @@ export class AuthorizationService {
...
+    }
+  }
+
+  async verifyCanEdit<
+    DbNode extends Record<string, any>,
+    Key extends keyof DbNode
+  >(baseNode: DbNode, prop: Key) {
+    if (isSecured(baseNode[prop]) && !baseNode[prop].canEdit) {
+      throw new UnauthorizedException(
+        'You do not have permission to update this property',
```
CarsonF, line 137: The message could reference the object & property to be more helpful.
```suggestion
        'You do not have permission to update ${resource.name}.${prop}',
```

src/components/authorization/authorization.service.ts
```diff
@@ -104,6 +109,37 @@ export class AuthorizationService {
...
+        `${baseNode.__typename as string}.${prop as string}`
```
CarsonF, line 138: This isn't going to work. This is a path from the gql input object to the property. i.e. `user.timezone`. `user` comes from [here](https://github.com/SeedCompany/cord-api-v3/blob/5f0c861ebd5ee9ba00e0db0540e38b2a1141616e/src/components/user/dto/update-user.dto.ts#L58).

We could camelcase the `resource.name` which would work in most cases by convention. But it will still break in a few like LanguageEngagement's firstScripture which would be `engagement.firstScripture` instead of `languageEngagement.firstScripture`.

src/components/authorization/authorization.service.ts
CarsonF: Now thinking about some edge cases like defined files. Maybe we don't need to assert the type of each field value.
```tsx
changes: Partial<Record<keyof Resource['prototype'], any>>
```
This would signify that the keys need to match but the values can be anything.

src/components/authorization/authorization.service.ts
```diff
@@ -104,6 +109,37 @@ export class AuthorizationService {
...
+    Key extends keyof TObject & string
+  >(
+    baseNode: TObject,
+    props: readonly Key[],
+    changes: { [Key in keyof TObject]?: UnwrapSecured<TObject[Key]> }
+  ) {
+    for (const prop of props) {
+      if (
+        changes[prop] === undefined ||
+        unwrapSecured(baseNode[prop]) === changes[prop]
```
CarsonF, line 123: This needs to be smarter for DateTime's and string lists which will fail identity checks (`===`) even when the value the represent is the same.

CarsonF (changes requested):

src/components/authorization/authorization.service.ts
```diff
@@ -111,6 +114,41 @@ export class AuthorizationService {
     return parseSecuredProperties(props, permissions, resource.SecuredProps);
   }
 
+  async verifyCanEditChanges<TResource extends ResourceShape<any>>(
+    resource: TResource,
+    baseNode: Partial<TResource['prototype']>, // for use cases like checking defined files separately
```
CarsonF, line 119: Why partial here?

src/components/engagement/engagement.service.ts
```diff
@@ -810,6 +810,26 @@ export class EngagementService {
       session
     )) as LanguageEngagement;
 
+    const { pnp: onlyPnp, ...objectNoPnp } = object;
```
CarsonF, line 813: Not necessary

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
-      props,
       changes,
     });
   }
 
+  // expecting the changes of the "simple" properties.
+  async getActualChanges<TResource extends ResourceShape<any>>(
+    oldObject: TResource['prototype'],
+    changes: Partial<Record<keyof TResource['prototype'] & string, any>>
+  ): Promise<typeof changes> {
```
CarsonF, line 244: Does this typeof work? Maybe it would be better to define a `ChangesOf<Resource>` type alias.

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+    // we don't need id, it's not a change, it's the id of the data we want to change
+    if (props.includes('id')) {
+      delete realChanges.id;
+    }
+    for (const prop of props) {
+      if (realChanges[prop] === undefined) {
+        // TODO: maybe a better way than circumventing dynamic delete? I don't see a way around
+        //     deleting dynamically right now.
+        // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+        delete realChanges[prop];
```
CarsonF, line 259: Yes, start with an empty `realChanges` object and then assign props when they are different.

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+        continue;
+      }
+      const unwrapped = unwrapSecured(oldObject[prop]);
+      if (unwrapped instanceof Array) {
```
CarsonF, line 263:
```suggestion
      if (Array.isArray(unwrapped)) {
```
because JS

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+        if (isEqual(sortBy(unwrapped), sortBy(realChanges[prop]))) {
```
CarsonF, line 264: These if statements could be merged

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+          // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+          delete realChanges[prop];
+        }
+      }
+
+      if (unwrapped instanceof Date) {
```
CarsonF, line 270:
```suggestion
      if (unwrapped instanceof CalendarDate) {
```

We don't use native JS Date, it sucks.

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+        if (unwrapped.getTime() === (realChanges[prop] as Date).getTime()) {
```
CarsonF, line 271: Either of values could be null.

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
+          // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+          delete realChanges[prop];
+          continue;
+        }
+      }
+      if (unwrapped instanceof DateTime) {
+        if (unwrapped.equals(realChanges[prop] as DateTime)) {
```
CarsonF, line 278: Either of these could be null. We also don't want to use .equals() because that checks if both sides have the same timezone which we don't care about. You can use `a.valueOf() === b.valueOf()`, after checking both are DateTimes.

src/core/database/database.service.ts
```diff
@@ -218,48 +222,86 @@ export class DatabaseService {
...
-      if (
-        changes[prop] === undefined ||
-        unwrapSecured(object[prop]) === changes[prop]
-      ) {
-        continue;
-      }
-
+    const propsToUpdate: Array<keyof TObject & string> = keys(changes) as Array<
+      keyof TObject & string
+    >;
```
CarsonF, line 303:
```suggestion
    const propsToUpdate = keys(changes) as Array<keyof TObject & string>;
```
You don't need to typecast and have an explicit type.

CarsonF (changes requested):

src/components/engagement/engagement.service.ts
```diff
@@ -810,6 +810,26 @@ export class EngagementService {
...
+    const { pnp: changesOnlyPnp, ...changesNoPnp } = changes;
```
CarsonF, line 814: `pnp` has already been pulled out on line 799

src/components/engagement/engagement.service.ts
```diff
@@ -822,25 +842,10 @@ export class EngagementService {
...
-          'paratextRegistryId',
-          'historicGoal',
-          'modifiedAt',
-          'status',
-          'initialEndDate',
-        ],
-        changes,
-        nodevar: 'LanguageEngagement',
+      await this.db.updateProperties({
+        type: 'LanguageEngagement',
```
CarsonF, line 846:
```suggestion
        type: LanguageEngagement,
```

src/components/engagement/engagement.service.ts
```diff
@@ -888,6 +893,46 @@ export class EngagementService {
...
+      objectNoExtra,
+      realChanges
+    );
+    if (mentorId) {
+      await this.authorizationService.verifyCanEdit({
+        resource: InternshipEngagement,
+        baseNode: object,
+        prop: 'mentor',
+        propName: 'mentorId',
+        resourceName: 'Engagement',
```
CarsonF, line 916: This is defines the first part of the path for the field. It should be lowercase so the field in the exception becomes `engagement.mentorId`

src/components/authorization/authorization.service.ts
elisabeth-sorrell: I guess at the time I wanted to be very clear that these are for the "simple" properties for the object and changes. I'm fine with taking out partial here, though.

src/core/database/database.service.ts
elisabeth-sorrell: It has been working, but I'm fine with making an alias. Honestly though, aliases kind of take away readability for me personally.

src/core/database/database.service.ts
```diff
@@ -218,48 +226,79 @@ export class DatabaseService {
...
       changes,
     });
   }
 
+  // expecting the changes of the "simple" properties.
+  async getActualChanges<TResource extends Resource>(
+    oldObject: TResource,
+    changes: ChangesOf<TResource>,
+    props: ReadonlyArray<keyof TResource & string>
+  ): Promise<ChangesOf<TResource>> {
```
CarsonF, line 249: I'd rather pass in the static resource class here instead of its prop list. This is more consistent with our other functions which accept the same thing. It gives us flexibility in the future to grab more info from the static resource.
It's also terser in use across the app.

```suggestion
  async getActualChanges<TResourceStatic extends ResourceShape>(
    resource: TResourceStatic,
    existingObject: TResourceStatic['prototype'],
    changes: ChangesOf<TResourceStatic['prototype']>,
  ): Promise<ChangesOf<TResourceStatic['prototype']>> {
```

src/components/budget/budget.service.ts
```diff
@@ -419,19 +419,33 @@ export class BudgetService {
...
+      Budget,
+      budget,
+      realChanges
+    );
+    if (universalTemplateFile) {
+      await this.authorizationService.verifyCanEdit({
+        resource: Budget,
+        baseNode: budget,
+        prop: 'universalTemplateFile',
+      });
```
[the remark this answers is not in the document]
CarsonF: Or maybe since defined files are a common thing the `getActualChanges` can account for that and we don't spread them out until auth verification.

I know there's the mismatch in name with relation modifications. Those would still be necessary like you have them.

src/components/language/ethnologue-language/ethnologue-language.service.ts
```diff
@@ -149,32 +150,19 @@ export class EthnologueLanguageService {
...
-        .first();
-      if (!q) {
-        throw new UnauthorizedException(
-          `You do not have permission to edit his object', '${key}`
-        );
-      }
-    }
+    // even though we already have a cleaned value set, we still need to get them in a readable way for the verifyCanEditChanges function
+    const realChanges = await this.db.getActualChanges(
+      { ...ethnologueLanguage, createdAt: CalendarDate.now() }, // hacking in a createAt field, because getActualChanges is expecting a Resource shape.
```
CarsonF, line 156: I think this would not be needed with the change I suggested to getActualChanges

src/components/project/project.service.ts
```diff
@@ -613,6 +613,43 @@ export class ProjectService {
...
 
+    const { primaryLocationId, fieldRegionId, ...changesSimpleProps } = changes;
+
+    // get the 'real' changes... nothing undefined or the same as the old object
+    let props;
+    if (currentProject.type === ProjectType.Internship) {
+      props = InternshipProject.Props;
+    } else {
+      props = TranslationProject.Props;
+    }
```
CarsonF, line 624:
```suggestion
    const props = currentProject.type === ProjectType.Internship ? InternshipProject.Props : TranslationProject.Props;
```

src/components/project/project.service.ts
```diff
@@ -721,34 +736,19 @@ export class ProjectService {
...
-
     const event = new ProjectUpdatedEvent(
       result,
       currentProject,
       input,
       session
     );
     await this.eventBus.publish(event);
-    return event.updated;
+    return await this.readOne(input.id, session);
```
CarsonF, line 751: Why? We are trying to save this additional readOne call with the former code

src/core/database/database.module.ts
```diff
@@ -1,13 +1,14 @@
...
+import { AuthorizationModule } from '../../components/authorization/authorization.module';
 import { ConfigModule } from '../config/config.module';
 import { CypherFactory } from './cypher.factory';
 import { DatabaseService } from './database.service';
 import { IndexerModule } from './indexer/indexer.module';
 import { ParameterTransformer } from './parameter-transformer.service';
 
 @Module({
-  imports: [IndexerModule, ConfigModule],
+  imports: [IndexerModule, ConfigModule, forwardRef(() => AuthorizationModule)],
```
CarsonF, line 11: This isn't needed anymore

src/core/database/database.service.ts
```diff
@@ -92,6 +98,8 @@ export class DatabaseService {
   constructor(
     private readonly db: Connection,
     private readonly config: ConfigService,
+    @Inject(forwardRef(() => AuthorizationService))
+    private readonly authorizationService: AuthorizationService,
```
CarsonF, line 102: This isn't needed anymore.

```suggestion
```

CarsonF (changes requested):

src/core/database/database.service.ts
elisabeth-sorrell: @CarsonF I've spent several hours trying to do just that without success, but I'll try again and see.

CarsonF (approved):
Thanks for your work here @elisabeth-sorrell!

CarsonF (approved):

Issue #1721, opened 2021-03-07
PROJECT: Pinned Items


michaelmarshall:
As a user, I want to pin certain items in a list so that I can view those items in a separate list of pinned items.

For the v1 of this feature, we will only let users pin `Projects`, but we will write the code abstract enough for future versions to easily allow anything to be pinned to anything else.

Requirements:
- When viewing a list of projects, the user will be able to determine whether a project has been pinned or not.
- A user will be able to pin and unpin a project when viewing it in a list or on its detail page.
- A user will have the option of viewing all of their pinned items.

![image](https://user-images.githubusercontent.com/1165460/110253295-28e25e80-7f4f-11eb-95f5-c40e4d11b939.png)

^^ the pin on the project card and the pinned tab on the tab bar are features this project will add.

The database schema below is how we will store `pinned` relationships in the database. User -> Projects is shown in blue.
![image](https://user-images.githubusercontent.com/1165460/110253821-f38b4000-7f51-11eb-93de-d91a1ba26d5e.png)

A branch in the API and FE repo is being scaffolded with classes and functions. There will be a new `pinned` service that will be responsible for CRUD operations on the `pinned` relationship between any two `:BaseNode`s. In addition to the new service, the project service will need to be modified to return `pinned` data on projects relevant to the requesting user. Which also means model classes will need to be updated. The pinned service will be exposed via GraphQL resolvers so that the FE can before CRUD operations on it.

At a minimum, the following functions or concepts will need to be created and integrated with the rest of the codebase:
- [x] Function to create a pinned relationship between any two base nodes
- [x] Function to delete a pinned relationship 
- [x] Modification to the project model to include a boolean for if the requesting user has pinned it 
- [x] Modification to the project list view and project detail view to return the pinned boolean of the requesting user
- [x] UI component to show whether a project is pinned, and functionality to pinned/unpin it, both on list and detail views
- [x] UI component to show a list of pinned projects

Milestones:
- [x] API for pinning items and listing pinned projects.
- [x] Abstract UI components to implement the API.
- [x] Designs completed by the design team.
- [x] Implement the FE designs.

CarsonF:
API branch started at `feature/pin`. I left TODOs for work to be done.

sethmcknight:
Designs are completed and available on SeedCompany/cord-field#718

sethmcknight:
Blocked by seedcompany/cord-field#749

PR #1734, opened 2021-03-10
Feature pinning

zkhin:
https://github.com/SeedCompany/cord-api-v3/issues/1721

### DIFF
--- a/src/components/pin/pin.repository.ts
+++ b/src/components/pin/pin.repository.ts
@@ -1,0 +1,22 @@
+import { Injectable } from '@nestjs/common';
+import { NotImplementedException, Session } from '../../common';
+import { DatabaseService } from '../../core';
+
+/* eslint-disable @seedcompany/no-unused-vars -- Remove this when implementing */
+
+@Injectable()
+export class PinRepository {
+  constructor(private readonly db: DatabaseService) {}
+
+  async isPinned(id: string, session: Session): Promise<boolean> {
+    throw new NotImplementedException();
+  }
+
+  async togglePinned(
+    id: string,
+    pinned: boolean,
+    session: Session
+  ): Promise<void> {
+    throw new NotImplementedException();
+  }
+}
--- a/src/components/pin/pin.repository.ts
+++ b/src/components/pin/pin.repository.ts
@@ -28,11 +28,14 @@ export class PinRepository {
       .query()
       .call(matchRequestingUser, session)
       .match([node('node', 'BaseNode', { id })])
-      .create([
+      .merge([
         node('requestingUser'),
-        relation('out', '', 'pinned', { createdAt }),
+        relation('out', 'rel', 'pinned'),
         node('node'),
       ])
+      .onCreate.setValues({
+        'rel.createdAt': createdAt,
+      })
       .run();
   }
 
--- a/src/components/pin/pin.repository.ts
+++ b/src/components/pin/pin.repository.ts
@@ -23,10 +23,6 @@ export class PinRepository {
   }
 
   async add(id: string, session: Session): Promise<void> {
-    const isPinned = await this.isPinned(id, session);
-    if (isPinned) {
-      return;
-    }
     const createdAt = DateTime.local();
     await this.db
       .query()
@@ -41,10 +37,6 @@ export class PinRepository {
   }
 
   async remove(id: string, session: Session): Promise<void> {
-    const isPinned = await this.isPinned(id, session);
-    if (!isPinned) {
-      return;
-    }
     await this.db
       .query()
       .call(matchRequestingUser, session)
--- a/src/components/pin/pin.repository.ts
+++ b/src/components/pin/pin.repository.ts
@@ -1,15 +1,25 @@
 import { Injectable } from '@nestjs/common';
-import { NotImplementedException, Session } from '../../common';
-import { DatabaseService } from '../../core';
-
-/* eslint-disable @seedcompany/no-unused-vars -- Remove this when implementing */
+import { node, relation } from 'cypher-query-builder';
+import { DateTime } from 'luxon';
+import { Session } from '../../common';
+import { DatabaseService, matchRequestingUser } from '../../core';
 
 @Injectable()
 export class PinRepository {
   constructor(private readonly db: DatabaseService) {}
 
   async isPinned(id: string, session: Session): Promise<boolean> {
-    throw new NotImplementedException();
+    const result = await this.db
+      .query()
+      .call(matchRequestingUser, session)
+      .match([
+        node('requestingUser'),
+        relation('out', '', 'pinned', { active: true }),
+        node('node', 'BaseNode', { id }),
+      ])
+      .return('node')
+      .first();
+    return result ? true : false;
   }
 
   async togglePinned(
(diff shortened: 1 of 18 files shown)


src/components/pin/pin.resolver.ts
```diff
@@ -0,0 +1,64 @@
...
+    pinned?: boolean
+  ): Promise<boolean> {
+    return await this.pins.togglePinned(id, session, pinned);
+  }
+
+  // @Query(() => PinnedListOutput, {
+  //   name: 'pins',
+  //   description: "A list of the requesting user's pinned items",
+  // })
+  async list(
```
zkhin, line 53: @CarsonF Do we need to complete this query in this PR?

src/components/project/project.service.ts
```diff
@@ -582,6 +584,7 @@ export class ProjectService {
         value: result.owningOrganizationId,
       },
       canDelete: await this.db.checkDeletePermission(id, sessionOrUserId),
+      pinned: await this.pinSerivce.isPinned(id, sessionOrUserId as Session), // TODO: More safe way to pass session
```
zkhin, line 587: @CarsonF What is the best way to pass the session here?

src/components/pin/pin.resolver.ts
CarsonF: Nope!

src/components/project/project.service.ts
CarsonF: Can we match the relation in the same query instead of doing another lookup?

CarsonF (changes requested):

src/components/project/query.helpers.ts
```diff
@@ -44,6 +44,14 @@ export function projectListFilter(query: Query, filter: ProjectFilters) {
       node('node'),
     ]);
   }
+
+  if (filter.pinned) {
```
CarsonF, line 48: I think it makes sense to have `pinned == false` not match any pinned projects. That's why I had `pinned != null`, so both true and false could be handled.
What do you think?

src/components/project/project.service.ts
```diff
@@ -526,6 +532,7 @@ export class ProjectService {
       ])
       .asResult<
         StandardReadResult<DbPropsOfDto<Project>> & {
+          pinnedRel: Relation;
```
CarsonF, line 535: It's an optional match so the relation returned is nullable
```suggestion
          pinnedRel?: Relation;
```

src/components/pin/pin.repository.ts
```diff
@@ -22,28 +22,29 @@ export class PinRepository {
...
+      ])
+      .run();
   }
 
-  protected async removeOldPinned(id: string, session: Session): Promise<void> {
+  async remove(id: string, session: Session): Promise<void> {
+    const isPinned = await this.isPinned(id, session);
+    if (!isPinned) {
+      return;
+    }
```
CarsonF, line 47: Repositories should just contain low level db queries. They should not be smart, which is why they are private to the module. All business-logic / smarts should go in the service.

It also looks like it's ok to call the query below even if the user hasn't pinned this right?

src/components/pin/pin.repository.ts
```diff
@@ -22,28 +22,29 @@ export class PinRepository {
...
+    const isPinned = await this.isPinned(id, session);
+    if (isPinned) {
+      return;
     }
+    const createdAt = DateTime.local();
+    await this.db
+      .query()
+      .call(matchRequestingUser, session)
+      .match([node('node', 'BaseNode', { id })])
+      .create([
```
CarsonF, line 35: Maybe `createUnique()` would work here to prevent duplicates while not having to check is pinned?

src/components/pin/pin.repository.ts
zkhin: yeah, but there will be duplicated relations.

CarsonF: How? It's just deleting if found right?

src/components/pin/pin.repository.ts
zkhin: `createUnique` is deprecated. I think it will be better to check if it's already pinned or unpinned in the service before the action.

CarsonF: Oh right. I think we want `merge` here.

src/components/project/query.helpers.ts
zkhin: This [commit](https://github.com/SeedCompany/cord-api-v3/pull/1734/commits/f41c61cc1edc0b17919ed844442087ddff9118b5) will handle this function.

zkhin:
@michaelmarshall @CarsonF Added pinning e2e test cases

michaelmarshall (dismissed):

src/components/project/project.module.ts
```diff
@@ -30,6 +31,7 @@ import { ProjectService } from './project.service';
     forwardRef(() => AuthorizationModule),
     PartnerModule,
     forwardRef(() => OrganizationModule),
+    PinModule,
```
CarsonF, line 34: This import is no longer used
```suggestion
```

src/components/project/project.service.ts
```diff
@@ -127,6 +128,7 @@ export class ProjectService {
     @Inject(forwardRef(() => AuthorizationService))
     private readonly authorizationService: AuthorizationService,
     private readonly projectRules: ProjectRules,
+    private readonly pinSerivce: PinService,
```
CarsonF, line 131: This isn't used anymore
```suggestion
```

src/components/project/query.helpers.ts
```diff
@@ -44,6 +45,27 @@ export function projectListFilter(query: Query, filter: ProjectFilters) {
...
+      query
+        .optionalMatch([
+          node('node'),
+          relation('in', 'rel', 'pinned', { active: true }),
+          node('requestingUser'),
+        ])
+        .with('node, requestingUser')
+        .where({
+          rel: isNull(),
+        });
```
CarsonF, line 66: I believe this could simply be
```suggestion
      query.raw('where not (requestingUser)-[:pinned]->(node)`);
```

CarsonF (changes requested):
I don't think we need the `active` property on the relation here since we aren't keeping history.

src/components/pin/pin.repository.ts
```diff
@@ -0,0 +1,51 @@
...
+    return result ? true : false;
+  }
+
+  async add(id: string, session: Session): Promise<void> {
+    const createdAt = DateTime.local();
+    await this.db
+      .query()
+      .call(matchRequestingUser, session)
+      .match([node('node', 'BaseNode', { id })])
+      .create([
```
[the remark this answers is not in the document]
zkhin: I don't think we can use merge here like below
```
.merge([
        node('requestingUser'),
        relation('out', '', 'pinned', { createdAt }),
        node('node', 'BaseNode', { id }),
])
```
Actually, there is no node with only the `BaseNode` label with this `id`... so it will create a node with only `BaseNode` label with this id and create a relation with the requestingUser instead of the project node. So the project node will miss the `:pinned` relation with the user.

CarsonF (approved):
Looks great! Thanks @zkhin!

Issue #1729, opened 2021-03-09
Random authorization procedure failure preventing projects from being fully created


ParkeBrown:
~Carson Full Today at 2:39 PM
@Blake Krammes we are having a problem with root directories not being created/connected to new projects. Can you look to this? (assuming it’s a priority but Seth can confirm)~

ParkeBrown:
Related conversation
Ken Jones:spiral_calendar_pad:  1 hour ago
Hi @Carson Full FYI, for some reason the Projects, LanguageEngagements and Products API data is late for the last push.  The other 6 were all received by 11:39 AM ET.






Carson Full  40 minutes ago
It looks like a project was renamed during the sync, causing it to show up twice, which stopped the sync.

Carson Full  40 minutes ago
I’m going to change the sort order to not be based on the name.

Parke Brown  33 minutes ago
Still exploring and may be unrelated but we’re seeing some reported project creation issues.
https://github.com/SeedCompany/management/issues/86
https://github.com/SeedCompany/management/issues/85 (edited) 

Carson Full  33 minutes ago
This is the project whose name changed https://cordfield.com/projects/5da491e738339de3954f82d7

CarsonF:
The authorization procedure is failing after the project is created. This prevents the event handlers from running (i.e. to create the root directory, etc.) This also explains why search is failing, because the search query still uses db security groups. Since those failed to create there's no match.

CarsonF:
Instead of waiting for that refactor, I went ahead and looked into what it would take to make the procedure work inside of the current transaction. It turns out it mostly just involved removing code. https://github.com/SeedCompany/cord-procedures/pull/13

CarsonF:
This is actually because of a deadlock on the user node. The procedure wasn't failing it was just waiting for the lock to be released.

CarsonF:
I believe this was fixed with #1735

PR #1735, opened 2021-03-10
Fix user role update deadlock

CarsonF:
We were getting a deadlock in production trying to update user roles, because the code to modify the user runs in a transaction and its uncommitted when we ask the db to update security groups for that user. The transaction in progress locks the user node and then the security group query, which uses apoc.periodic.iterate (which uses separate transactions), waits for that lock to be released.

Unfortunately I was unable to reproduce this deadlock locally.

This is the cause of #1729

### DIFF
--- a/src/components/authorization/authorization.service.ts
+++ b/src/components/authorization/authorization.service.ts
@@ -50,26 +50,33 @@ export class AuthorizationService {
     baseNodeId: string,
     creatorUserId: string
   ) {
-    const label = baseNodeObj.__className.substring(2);
+    await this.afterTransaction(async () => {
+      await this.db
+        .query()
+        .raw(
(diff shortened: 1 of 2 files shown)


src/components/authorization/authorization.service.ts
```diff
@@ -187,6 +194,10 @@ export class AuthorizationService {
   }
 
   async roleAddedToUser(id: string, roles: Role[]) {
+    await this.afterTransaction(() => this.doRoleAddedToUser(id, roles));
```
blakekrammes, line 197: So is the thing that's changing is that adding roles to users is being retried?

CarsonF: Yeah but mostly to change to run after the current transaction finishes.

PR #1748, opened 2021-03-15
Ensure duplicate budget records cannot be created

### LINKED ISSUE #1743: BudgetsInProgress API has not successfully pushed to DOMO since March 10th at 3pm

CarsonF:
Helps fix #1743

### DIFF
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -1,6 +1,7 @@
 import { node, relation } from 'cypher-query-builder';
 import { difference } from 'lodash';
 import {
+  DuplicateException,
   fiscalYears,
   NotFoundException,
   Secured,
@@ -154,14 +155,23 @@ export class SyncBudgetRecordsToFundingPartners
   ) {
     await Promise.all(
       additions.map((fiscalYear) =>
-        this.budgets.createRecord(
-          {
-            budgetId: budget.id,
-            fiscalYear,
-            organizationId,
-          },
-          session
-        )
+        this.budgets
+          .createRecord(
+            {
+              budgetId: budget.id,
+              fiscalYear,
+              organizationId,
+            },
+            session
+          )
+          .catch((e) => {
+            if (e instanceof DuplicateException) {
+              // If this record already exists, this user probably just doesn't
+              // have permission to see it yet. Ignore and move on.
+              return;
+            }
+            throw e;
+          })
       )
     );
   }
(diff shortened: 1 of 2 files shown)


src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
```diff
@@ -154,14 +155,23 @@ export class SyncBudgetRecordsToFundingPartners
...
+            {
+              budgetId: budget.id,
+              fiscalYear,
+              organizationId,
+            },
+            session
+          )
+          .catch((e) => {
+            if (e instanceof DuplicateException) {
+              // If this record already exists, this user probably just doesn't
```
blakekrammes, line 169: So the user can't see the budget, but they're able to change the mou dates? Trying to understand the flow.

CarsonF: They can see the budget but this is for an edge case where don't have the permissions on the newly created record from a previous call. Because that query had finished yet, it's waiting and it times out. And now they try again to update, so this code tries again to create the record, which already exists it's just not in the list due to the missing SecurityGroups at the present moment.

blakekrammes: Right. Thanks for the reminder. If a user can see a budget, why do we have to check permissions for the records?

CarsonF: It's just because we haven't updated any of the lists yet to use the new app code. Also maybe they could see the total but not the individual break down. But yes we don't even [secure the record lists](https://github.com/SeedCompany/cord-api-v3/blob/248d9ec1ffca67612505612a6c76ea3be8dfb3d2/src/components/budget/dto/budget.dto.ts#L19), it's actually an edge case.

So lots of coming changes will prevent this problem, but this sanity check is good too.

blakekrammes: And if we return here, the rest of the records create successfully. What if we throw? Should there be some cleanup of the other records?

blakekrammes: Just read again that `Promise.all` will reject if any of it's promises reject. So nvm. :)

blakekrammes (approved):

src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
CarsonF: Well if the fail is due to duplicate, it's not a failure. Other than that we are just going to wait for transactions which have retry and rollback built in.

blakekrammes: Gotcha. Those transactions aren't in place yet, but will be right?

Issue #1723, opened 2021-03-07
PROJECT: Project Reports


michaelmarshall:
As a project member, I want to be able to upload specific reports for my project to meet my reporting obligations, so that my supervisors can see which report is for each specific requirement.

As a project manager, I want to ensure a report is for a specific requirement and has not be uploaded for an incorrect requirement so that I know I'm getting all the reports I have required of the project. I should also be able to preview a report without downloading it, as well as simply download the report.

This feature adds a section the to project details page and provides a mechanism for the project member to upload a file and tag it (or equivalent) in such a way to identify it as fulfillment of a specific requirement. The project member should be able to see which reports they've submitted, and which requirements they have yet to submit a report for. 

This feature should also include the ability to preview a report without downloading it by anyone with access to it.

This feature is similar to the 'defined files' concept and as such, much of the work may already be accomplished with defined files. Yet to be added is the fact that every project as a few reports periodically (each may be a different period) that need their own identification so everyone knows if a report has been uploaded or not.

We are still awaiting designs on this feature. We can begin API development and make abstract UI components until we have the final FE designs.

For communication purposes, a simple UI would provide for a file upload dialog with a dropdown for which specific report is being uploaded. Another component would show the list of reports that have been uploaded and which are missing.

The database schema will attach `:ProjectReport` nodes off of the `:Project` node.
![image](https://user-images.githubusercontent.com/1165460/110255825-503f2880-7f5b-11eb-9462-505312eba4e2.png)

GraphQL is used to expose data to the FE. After the model is created, resolvers will need to be written that call the new project service functions for report file CRUDL.

Milestones:
- [x] API for project report CRUDL. 
- [x] Abstract UI components to implement the new API.
- [x] Design completes the final FE designs.
- [ ] Implement FE designs.

michaelmarshall:
@sethmcknight can we list all report types, their periods, and any other meta data needed?
e.g.:
1. financial, quarterly
2. progress, monthly

sethmcknight:
> @sethmcknight can we list all report types, their periods, and any other meta data needed?
> e.g.:
> 
> 1. financial, quarterly
> 2. progress, monthly

Financial Report
- Monthly or Quarterly
- Month/Quarter the Report Covers
- Date/Time Submitted
- User Submitting

Progress
- Quarterly
- Quarter the Report Covers
- Date/Time Submitted
- User Submitting

Narrative
- Quarterly
- Quarter the Report Covers
[...]

PR #1751, opened 2021-03-16
Add project report component

tmpacifitech:
#1723


michaelmarshall:
queries look fine to me.

CarsonF:
Ok after talking with Seth, I think we need some significant changes. Sorry I didn't a chance to talk through it with him before you started. 

# Report Shape

```gql
interface PeriodicReport {
  id: ID!;
  start: Date!;
  end: Date!;
  reportFile: SecuredFile!;
}
type FinancialReport implements PeriodicReport { ... }
type NarrativeReport implements PeriodicReport { ... }
type ProgressReport implements PeriodicReport { ... }
```
This will allow each report type to define additional fields in the future, such as form replacements for these files.
Typically how we do this is two labels in the DB: one for the interface and another for the concrete type.

# Queries

The main way we want to access these is via the project/engagement they are associated with
```gql
interface Project {
  ...
  financialReports(input: PeriodicReportListInput): SecuredPeriodicReportList!
  narrativeReports(input: PeriodicReportListInput): SecuredPeriodicReportList!
}
interface Engagement {
  ...
  progressReports(input: PeriodicReportListInput): SecuredPeriodicReportList!
}
```

# Mutations

## Financial Report Period

The period length of narrative and progress reports are always quarterly. Financial reports can be quarterly or monthly _based on the project_.
So we need a new project property to define this.
```gql
enum FinancialReportPeriod {
  Quarterly
  Monthly
}
interface Project {
  ...
  financialReportPeriod: SecuredFinancialReportPeriod!
}


input UpdateProject {
  ...
  financialReportPeriod: FinancialReportPeriod
}
```
See enum & secured example [here](https://github.com/SeedCompany/cord-api-v3/blob/248d9ec1ffca67612505612a6c76ea3be8dfb3d2/src/components/user/education/dto/education.dto.ts#L11-L25).
This can be should be defaulted to `Monthly` for new projects. We'll need to retroactively add this to existing projects.

## Creating / Deleting Reports

Since the all reports are known once we have the project's date range, let's go ahead a create them automatically then. We can manage the list for users since we don't need any input from them.
This can work similarly to [SyncBudgetRecordsToFundingPartners](https://github.com/SeedCompany/cord-api-v3/blob/d4a56a9495772c133148912eb19cbcfa293de2cc/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts).
This means that we can have create/delete methods in the service file, but we don't need API mutations for them.

Progress reports are different because they are on engagements instead of projects. This can use be done similarly in a separate event handler listening for engagement create/update events.

When a projects/engagements date range changes periodic reports should be created & removed to match the new range. Removals should only be done if the report doesn't have any files uploaded.

## Adding Report Files

```gql
type Mutation {
  uploadPeriodicReport(reportId: ID!, file: CreateDefinedFileVersionInput!): File!
}
```

----

I know I've glossed over some details here. Let me know if/when you have questions or need help converting to TS / app code.

src/components/periodic-report/dto/type.enum.ts
```diff
@@ -0,0 +1,21 @@
...
+  Quarterly = 'Quarterly',
+}
+
+registerEnumType(ReportType, { name: 'ReportType' });
+registerEnumType(ReportPeriod, { name: 'PeriodType' });
+
+@ObjectType({
+  description: SecuredProperty.descriptionFor('report period'),
+})
+export abstract class SecuredReportPeriod extends SecuredEnum(ReportPeriod) {}
```
CarsonF, line 21: Can you move ReportPeriod to a different enum file?

src/components/periodic-report/dto/update-periodic-report.dto.ts
```diff
@@ -0,0 +1,44 @@
...
+  readonly id: string;
+
+  @Field(() => ReportType)
+  readonly type: ReportType;
+
+  @Field()
+  readonly start?: CalendarDate;
+
+  @Field()
+  readonly end?: CalendarDate;
```
CarsonF, line 21: These shouldn't ever change, they should be required on create and readonly from then on.

Actually I see `UploadPeriodicReport` below so this file shouldn't even be needed

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+      ...event,
+      event: event.constructor.name,
+    });
+
+    const project = event.updated;
+    const previous = event.previous;
+
+    const isDateRangeChanged =
+      project.mouStart !== previous.mouStart ||
+      project.mouEnd !== previous.mouEnd;
```
CarsonF, line 39: These are date objects and should not be compared with identity. 
https://moment.github.io/luxon/docs/manual/math.html#comparing-datetimes
```tsx
d1.toMillis() === d2.toMillis() // are d1 and d2 the same instant in time?
+d1 === +d2 // same test, using object coercion
```

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (isDateRangeChanged) {
+      await this.removeOldReports(project, event.session);
+      await this.syncRecords(project, event.session);
+    }
+  }
+
+  private async removeOldReports(project: Project, session: Session) {
```
CarsonF, line 47: We only want to remove reports based on the date range that's being removed. This could probably be a where clause.

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+      .run();
+
+    await Promise.all(
+      reports.map((report) =>
+        this.periodicReports.delete(report.reportId, session)
+      )
+    );
+  }
+
+  private async syncRecords(project: Project, session: Session) {
```
CarsonF, line 73:
```suggestion
  private async addReports(project: Project, session: Session) {
```

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+    await this.addReport(
+      project.id,
+      {
+        start: project.mouStart.value!,
+        end: project.mouEnd.value!,
+        type: ReportType.Financial,
+      },
+      session
+    );
```
CarsonF, line 82: This will create a single report for the entire range of the project. We actually want multiple reports at a monthly/quarterly interval. Narrative & Progress reports can be hardcoded to quarterly, but financial needs to be looked up from the project property.

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+      .match(node('periodicReport', 'PeriodicReport', { id: report.id }))
+      .create([
+        node('project'),
+        relation('out', '', 'report', {
+          active: true,
+          createdAt: DateTime.local(),
+        }),
+        node('periodicReport'),
+      ])
+      .run();
```
CarsonF, line 113: A report should always be attached to a project. This logic should be encapsulated in the service. I understand this might need type changes, feel free to make those as well for creation.

src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -0,0 +1,124 @@
...
+import { Engagement } from '../../engagement';
+import { EngagementUpdatedEvent } from '../../engagement/events';
+import { CreatePeriodicReport } from '../dto';
+import { ReportType } from '../dto/type.enum';
+import { PeriodicReportService } from '../periodic-report.service';
+
+type SubscribedEvent = EngagementUpdatedEvent;
+
+@EventsHandler(EngagementUpdatedEvent)
+export class SyncProgressReportToProject
```
CarsonF, line 20:
```suggestion
export class SyncProgressReportsToEngagementDateRange
```

src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -0,0 +1,124 @@
...
+  implements IEventHandler<SubscribedEvent> {
+  constructor(
+    private readonly db: DatabaseService,
+    private readonly periodicReports: PeriodicReportService,
+    @Logger('periodicReport:sync-project') private readonly logger: ILogger
```
CarsonF, line 25: Convention for logger names is `kebab-case`.
```suggestion
    @Logger('periodic-report:engagement-sync') private readonly logger: ILogger
```

src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -0,0 +1,124 @@
...
+      previousEndDate,
+    ] = this.determineEngagementDateRange(previous);
+
+    const isDateRangeChanged =
+      startDate !== previousStartDate || endDate !== previousEndDate;
+
+    if (isDateRangeChanged) {
+      await this.removeOldReports(engagement, event.session);
+      await this.syncRecords(engagement, event.session);
+    }
```
CarsonF, line 49: Same thing here as project handler

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -0,0 +1,306 @@
...
+  async uploadFile(
+    reportId: string,
+    file: CreateDefinedFileVersionInput,
+    session: Session
+  ) {
+    const reportFileId = await generateId();
+
+    await this.files.createDefinedFile(
+      reportFileId,
+      file.name ?? 'Report File',
```
CarsonF, line 119: This file name should probably reference the end date.

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -0,0 +1,306 @@
...
+      start: props.start,
+      end: props.end,
+      reportFile: await this.files.resolveDefinedFile(
+        {
+          value: result.reportFileId,
+          canEdit: true,
+          canRead: true,
+        },
+        session
+      ),
```
CarsonF, line 208: If we cannot grab the relation object in the same db query then we should defer that to another resolver. It allows this db query to be skipped entirely if `reportFile` is not asked for. 
See these examples:
https://github.com/SeedCompany/cord-api-v3/blob/bcfef17a3c44066d8f5f247825f9e5035f18621e/src/components/engagement/dto/engagement.dto.ts#L45-L45
https://github.com/SeedCompany/cord-api-v3/blob/bcfef17a3c44066d8f5f247825f9e5035f18621e/src/components/engagement/engagement.resolver.ts#L65-L76

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -0,0 +1,306 @@
...
+      canDelete: await this.db.checkDeletePermission(id, session),
+    };
+  }
+
+  async update(
```
CarsonF, line 213: Shouldn't need an update for these reports.

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -0,0 +1,306 @@
...
+    reportType: ReportType,
+    { filter, ...input }: PeriodicReportListInput,
+    session: Session
+  ): Promise<SecuredPeriodicReportList> {
+    const query = this.db
+      .query()
+      .match([
+        node('project', 'Project', { id: projectId }),
+        relation('out', '', 'report', { active: true }),
+        node('node', `PeriodicReport:${reportType}Report`),
```
CarsonF, line 264: Prefer just using an array.
```suggestion
        node('node', ['PeriodicReport', `${reportType}Report`]),
```

CarsonF (changes requested):
Overall looks good other than changes noted!

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
tmpacifitech: @CarsonF This makes sense to me but if the project's date range is 2020-02-13 ~ 2020-06-29, how many progress/narrative reports should be created with what start/end date fields?

CarsonF: I believe the date range should expand to the end of the quarter/month. If the project has been worked one week into the month, we still need the last report for that partial month.

Luxon has `endOf('month')`, etc. btw. I think quarter refers to our fiscal quarters though which are offset. I'll confirm with @sethmcknight.

tmpacifitech:
@CarsonF Resolved all comments and ready for review again.

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -22,4 +31,25 @@ export class PeriodicReportResolver {
     );
     return reportFile;
   }
+
+  @ResolveField(() => SecuredFile, {
+    nullable: true,
+  })
```
CarsonF, line 37: Secured* objects are not nullable. canRead/Edit are always given.
```suggestion
  @ResolveField(() => SecuredFile)
```

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -22,4 +31,25 @@ export class PeriodicReportResolver {
...
+      ? await this.files.resolveDefinedFile(
+          {
+            value: id,
+            ...rest,
+          },
+          session
+        )
+      : null;
+
+    return value;
```
CarsonF, line 53: `resolveDefinedFile` already handles all that.
```suggestion
  ): Promise<SecuredFile> {
  return await this.files.resolveDefinedFile(report.reportFile);
```

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -198,33 +241,15 @@ export class PeriodicReportService {
...
-          canEdit: true,
-          canRead: true,
-        },
-        session
-      ),
+      reportFile: {
+        value: result.reportFileId,
+        canEdit: true,
+        canRead: true,
+      },
```
CarsonF, line 248: Pass `reportFileId` as `reportFile` to `secureProperties()` so that authorization can be properly applied

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -101,6 +103,45 @@ export class PeriodicReportService {
...
       }
 
+      if (
+        input.type === ReportType.Financial ||
+        input.type === ReportType.Narrative
+      ) {
+        await this.db
+          .query()
+          .match(
+            node('project', 'Project', { id: input.projectOrEngagementId })
```
CarsonF, line 113: These two queries could be combined with a single condition on the label.
```suggestion
            node('node', input.type === ReportType.Progress ? 'Engagement' : 'Project', { id: input.projectOrEngagementId })
```

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -303,4 +328,97 @@ export class PeriodicReportService {
...
+  }
+
+  async removeFinancialReports(project: Project, session: Session) {
+    let reports;
+    if (project.financialReportPeriod.value === ReportPeriod.Monthly) {
+      reports = await this.getProjectReportsQuery(project)
+        .where({
+          rel: isNull(),
+          'date(start.value)': greaterEqualTo(
+            project.mouStart.value?.startOf('month')
```
CarsonF, line 370: These queries could be combined by putting the duration unit on a variable
```ts
const period: DurationUnit = project.financialReportPeriod.value === ReportPeriod.Monthly ? 'month' : 'quarter';
...
.startOf(period)
.endOf(period)
```

src/components/periodic-report/handlers/sync-periodic-report-to-project.handler.ts
```diff
@@ -34,82 +31,67 @@ export class SyncPeriodicReportToProject
...
+      (project.mouStart.value.toMillis() !==
+        previous.mouStart.value?.toMillis() ||
+        project.mouEnd.value.toMillis() !== previous.mouEnd.value?.toMillis());
 
     if (isDateRangeChanged) {
-      await this.removeOldReports(project, event.session);
-      await this.syncRecords(project, event.session);
+      await this.periodicReports.removeFinancialReports(project, event.session);
+      await this.periodicReports.removeNarrativeReports(project, event.session);
+      await this.addRecords(project, event.session);
```
CarsonF, line 45: I don't think this is exactly right. I'd like to diff to determine the additions / removals. That way we aren't deleting and recreating a bunch of reports for the same time range.
i.e.
01-01-2020 - 01-01-2024
to
01-01-2020 - 01-01-2025
should only add 4 reports. Instead of deleting 16 & creating 20.

CarsonF (changes requested):
Getting closer @tmpacifitech!

tmpacifitech:
@CarsonF Ready for review agian. 😇

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    await this.files.createDefinedFile(
+      reportFileId,
+      report.end.toISODate(),
+      session,
+      reportId,
+      'reportFile',
+      file,
+      'periodicReport.reportFile'
+    );
```
CarsonF, line 143: I know this is a little confusing, but we actually want to do this on report create. All of our _files_ are actually just holders for _file versions_. By creating this "holder" when the object is created we can always reference the file/holder ID to connect the current upload/file version to.

Does that make sense?
See `growthPlan` in EngagementService for a concrete example you could follow.

CarsonF (changes requested):
I just rebased on develop.

I also refactored the financial & narrative syncs. Could you follow suit with progress reports sync?

src/components/periodic-report/periodic-report.service.ts
CarsonF: Nevermind did it myself 37529aaf319c0aad67293f695f1619b2907783ca

CarsonF (changes requested):
I'm done making changes @tmpacifitech. Thanks for letting me just hop in like this.

I think we are almost done here 🎉 

Just a few things:
- [ ] I did notice that when a project's date range changes it could affect the engagement's date range. The progress report sync handler needs to listen for project update events and fetch the engagement and update the reports if the range has changed.
- [ ] Could you add E2E tests to verify expected functionality?
- There's a potential code path where a report gets created, file uploaded, range change, report retained, range changed back, duplicate report gets created. This is a very small edge case though so we probably don't need to worry about it right now.

PR #1752, opened 2021-03-16
Dynamically set Deleted_ labels and remove original labels on delete

blakekrammes:
- Includes base node labels and property labels. 
- I've changed the deletion of properties to just specify base node property relationships, which I believe are the same as 
   resource keys. That way you don't have to specify the property node labels.

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -14,7 +14,6 @@ import {
   UnauthorizedException,
 } from '../../common';
 import {
-  BaseNodePropertyRelationships,
   ConfigService,
   DatabaseService,
   deleteProperties,
@@ -362,15 +361,11 @@ export class UserService {
 
     // Update email
     if (input.email) {
-      const relationshipsOfPropertiesToDelete: BaseNodePropertyRelationships<User> = [
-        'email',
-      ];
-
       // Remove old emails and relations
       await this.db
         .query()
         .match([node('node', ['User', 'BaseNode'], { id: user.id })])
-        .call(deleteProperties, relationshipsOfPropertiesToDelete)
+        .call(deleteProperties(User, 'email'))
         .return('*')
         .run();
 
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -14,17 +14,17 @@ import {
   UnauthorizedException,
 } from '../../common';
 import {
+  BaseNodePropertyRelationships,
   ConfigService,
   DatabaseService,
+  deleteProperties,
   ILogger,
   Logger,
   matchSession,
   OnIndex,
   property,
-  setPropLabelsAndValuesDeleted,
   Transactional,
   UniquenessError,
-  UniqueProperties,
 } from '../../core';
 import {
   calculateTotalAndPaginateList,
@@ -362,14 +362,15 @@ export class UserService {
 
     // Update email
     if (input.email) {
+      const relationshipsOfPropertiesToDelete: BaseNodePropertyRelationships<User> = [
+        'email',
+      ];
+
       // Remove old emails and relations
-      const uniqueProperties: UniqueProperties<User> = {
-        email: ['Property', 'EmailAddress'],
-      };
       await this.db
         .query()
         .match([node('node', ['User', 'BaseNode'], { id: user.id })])
-        .call(setPropLabelsAndValuesDeleted, uniqueProperties)
+        .call(deleteProperties, relationshipsOfPropertiesToDelete)
         .return('*')
         .run();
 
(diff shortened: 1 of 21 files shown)


src/core/database/query.helpers.ts
```diff
@@ -103,77 +103,27 @@ export function matchRequestingUser(
...
-        },
-      })
-      .removeProperties({
-        [currentPropertyNodeAlias]: 'value',
-      })
-      .with('*');
-  });
+  return query.raw(
+    `
+    match(node)-[propertyRel:${relationships.join(
```
blakekrammes, line 118: I used a JS var here because I don't believe it's possible to use a `.raw()` variable as a string literal in this way. https://jamesfer.me/cypher-query-builder/classes/query.html#raw

src/components/user/user.service.ts
```diff
@@ -362,14 +362,15 @@ export class UserService {
...
+
       // Remove old emails and relations
-      const uniqueProperties: UniqueProperties<User> = {
-        email: ['Property', 'EmailAddress'],
-      };
       await this.db
         .query()
         .match([node('node', ['User', 'BaseNode'], { id: user.id })])
-        .call(setPropLabelsAndValuesDeleted, uniqueProperties)
+        .call(deleteProperties, relationshipsOfPropertiesToDelete)
```
blakekrammes, line 373: I believe this is the only place we're deleting an individual property in the API so far.

src/core/database/query.helpers.ts
blakekrammes: Is this just not supposed to be done?

src/core/database/database.service.ts
```diff
@@ -557,32 +552,45 @@ export class DatabaseService {
...
-      //Mark baseNode labels and id deleted
-      .call(setBaseNodeLabelsAndIdDeleted, baseNodeLabels)
-      //Mark unique property labels and values deleted
-      .call(setPropLabelsAndValuesDeleted, uniqueProperties)
+      // after setting propRel.active false, we need to distinct propertyNode (not sure why)
+      .with('baseNode, collect(distinct propertyNode) as propertyNodes')
+      // combine baseNode with propertyNodes to set Deleted_ labels on all of them
+      .with(
+        'reduce(nodes = [baseNode], pn in propertyNodes | nodes + pn) as nodeList'
+      )
```
CarsonF, line 582: I believe you can just concat the lists together with `+`
```suggestion
      // combine baseNode with propertyNodes to set Deleted_ labels on all of them
      .with('[baseNode] + collect(distinct propertyNode) as nodeList')
```

blakekrammes: I think that will yield `[baseNode, [[propertyNode, propertyNode, ...]]]`

blakekrammes: It needs to be flattened

blakekrammes: Or maybe not. Hmm. Let me see

CarsonF: You sure? I don't think so.

CarsonF: https://neo4j.com/docs/cypher-manual/current/syntax/operators/#syntax-concatenating-two-lists

blakekrammes: Ok. You were right.

src/components/user/user.service.ts
```diff
@@ -363,13 +362,10 @@ export class UserService {
...
     if (input.email) {
       // Remove old emails and relations
-      const uniqueProperties: UniqueProperties<User> = {
-        email: ['Property', 'EmailAddress'],
-      };
       await this.db
         .query()
         .match([node('node', ['User', 'BaseNode'], { id: user.id })])
-        .call(setPropLabelsAndValuesDeleted, uniqueProperties)
+        .call(deleteProperties(User, 'email'))
```
blakekrammes, line 368: How does this work since 'email' is not an array?

src/core/database/query.helpers.ts
```diff
@@ -103,77 +103,27 @@ export function matchRequestingUser(
...
-    })
-    .removeLabels({
-      [nodeAlas]: label,
-    });
+/**
+ * This will set all relationships given to active false
+ * and add deleted prefix to its labels.
+ */
+export const deleteProperties = <Resource extends ResourceShape<any>>(
+  _resource: Resource,
```
blakekrammes, line 111: This looks great. Just tested it out. I'd say this is ready to go. Not sure why the checks aren't going through.

src/components/user/user.service.ts
CarsonF: It's defined as variable args `(...args: string[])`. You pass them in one by one and JS gives them to the function as an array.

CarsonF (approved):
Sweet, thanks @blakekrammes

PR #1755, opened 2021-03-22
On delete changes

blakekrammes:


src/core/database/database.service.ts
```diff
@@ -555,34 +555,46 @@ export class DatabaseService {
...
       })
-      // after setting propRel.active false, we need to distinct propertyNode (not sure why)
-      // combine baseNode with propertyNodes to set Deleted_ labels on all of them
-      .with('[baseNode] + collect(distinct propertyNode) as nodeList')
-      // yielding a node from this procedure is necessary I believe, but not used in the rest of the query
-      // they're aliased to avoid colliding with the unwound node var
+      // if we set anything on property nodes or property relationships in the query above
+      // (as was done before, but now now)
+      // we need to distinct propertyNode since each resulting row has both propertyNode and baseNode I believe
+      .with('[baseNode] + collect(propertyNode) as nodeList')
```
CarsonF, line 583: You say we need `distinct` but it's been removed from the query...?

CarsonF (approved):
Looks good other than that one confusing comment on distinct

src/core/database/database.service.ts
blakekrammes: It's only needed if we do anything to property nodes/relationships (as the query was previously doing). As the query is written now distinct does nothing. Maybe the comment is unnecessary, but I wanted to prevent bugs if we modify it down the road.

CarsonF: Can you update the comment to reflect the current thought?

blakekrammes: Yep. Should be clearer now.

CarsonF: > we need to distinct propertyNode to avoid collecting and labeling each propertyNode more than once

Then why aren't we using `distinct`?

blakekrammes: It needs to be understood with the first line of the comment.

CarsonF: Ah I see.

CarsonF (approved):

PR #1774, opened 2021-03-31
First rev of discussion (Post) API

richardgirges:
I called the discussion node `Post` - and I made it so that `Post`s are recursive (e.g. a Post can belong to another Post), although I'm not sure we truly need recursive posts. I figured it's harmless to implement regardless.

Figured there will probably be some feedback so opening PR now to initiate discussions. This should be enough for me to move forward on the frontend.

### DIFF
--- a/src/components/post/dto/create-post.dto.ts
+++ b/src/components/post/dto/create-post.dto.ts
@@ -1,0 +1,37 @@
+import { Field, InputType, ObjectType } from '@nestjs/graphql';
+import { Type } from 'class-transformer';
+import { ValidateNested } from 'class-validator';
+import { IdField } from '../../../common';
+import { Post } from './post.dto';
+import { PostType } from './type.enum';
+
+@InputType()
+export class CreatePost {
+  @IdField({ nullable: true })
+  readonly parentId?: string;
+
+  @Field(() => PostType)
+  readonly type: PostType;
+
+  @Field(() => Boolean)
+  readonly shareable: boolean;
+
+  @Field({
+    description: 'the post body',
+  })
+  readonly body: string;
+}
+
+@InputType()
+export abstract class CreatePostInput {
+  @Field()
+  @Type(() => CreatePost)
+  @ValidateNested()
+  readonly post: CreatePost;
+}
+
+@ObjectType()
+export abstract class CreatePostOutput {
+  @Field()
+  readonly post: Post;
+}
--- a/src/components/post/post.service.ts
+++ b/src/components/post/post.service.ts
@@ -1,0 +1,269 @@
+import { forwardRef, Inject, Injectable } from '@nestjs/common';
+import { ConfigService } from 'aws-sdk';
+import { node, relation } from 'cypher-query-builder';
+import { DateTime } from 'luxon';
+import {
+  generateId,
+  ID,
+  InputException,
+  NotFoundException,
+  ServerException,
+  Session,
+} from '../../common';
+import {
+  createBaseNode,
+  DatabaseService,
+  ILogger,
+  Logger,
+  matchRequestingUser,
+} from '../../core';
+import {
+  calculateTotalAndPaginateList,
+  matchPropList,
+} from '../../core/database/query';
+import {
+  DbPropsOfDto,
+  parseBaseNodeProperties,
+  parsePropList,
+  runListQuery,
+  StandardReadResult,
+} from '../../core/database/results';
+import { AuthorizationService } from '../authorization/authorization.service';
+import { UserService } from '../user';
+import { CreatePost, Post, UpdatePost } from './dto';
+import { PostListInput, SecuredPostList } from './dto/list-posts.dto';
+
+@Injectable()
+export class PostService {
+  private readonly securedProperties = {
+    body: true,
+    user: true,
+    modifiedAt: true,
+  };
+
+  constructor(
+    private readonly db: DatabaseService,
+    private readonly config: ConfigService,
+    @Inject(forwardRef(() => UserService))
+    private readonly userService: UserService,
+    @Inject(forwardRef(() => AuthorizationService))
+    private readonly authorizationService: AuthorizationService,
+    @Logger('post:service') private readonly logger: ILogger
+  ) {}
+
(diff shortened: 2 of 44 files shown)


src/components/post/dto/create-post.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+@InputType()
+export class CreatePost {
+  @IdField({ nullable: true })
+  readonly projectId?: string;
+
+  @IdField({ nullable: true })
+  readonly engagementId?: string;
+
+  @IdField({ nullable: true })
+  readonly parentPostId?: string;
```
michaelmarshall, line 16: One way I've implemented discussion in the past is to have the parentId take any kind of entity, that way, a starting post can be attached to anything and you don't need to hard code exactly what type the parent is, since we'll be putting discussion on an ever increasing number of things. Something to think about, but not sure I want to slow anything down right now.

src/components/post/post.service.ts
```diff
@@ -0,0 +1,313 @@
...
+        .create([
+          node('post'),
+          relation('out', '', 'user', {
+            active: true,
+            createdAt: DateTime.local(),
+          }),
+          node('user'),
+        ])
+        .return('post.id as id')
+        .first();
```
michaelmarshall, line 106: We've gone back and said its better to store the user.id as a prop than to link back to the user because it avoids the user node having many, many relationships coming into it.

michaelmarshall (changes requested):

src/components/post/post.service.ts
```diff
@@ -0,0 +1,313 @@
...
+      //   session.userId
+      // );
+
+      return await this.readOne(postId, session);
+    } catch (exception) {
+      this.logger.warning('Failed to create post', {
+        exception,
+      });
+
+      throw new ServerException('Failed to create post', exception);
```
CarsonF, line 179: Need to confirm that `parentId` is valid. If it's not, we need to throw `InputException` instead.

src/components/post/post.service.ts
```diff
@@ -0,0 +1,313 @@
...
+      ...securedProps,
+      body: {
+        value: props.body,
+        canRead: true,
+        canEdit: true,
+      },
+      user: {
+        ...securedProps.user,
+        value: await this.userService.readOne(result.userId, session),
+      },
```
CarsonF, line 221: If we cannot grab the user in the same db query then we should defer that to another resolver. It allows this query to be skipped entirely if `user` is not asked for. 
See these examples:
https://github.com/SeedCompany/cord-api-v3/blob/bcfef17a3c44066d8f5f247825f9e5035f18621e/src/components/engagement/dto/engagement.dto.ts#L45-L45
https://github.com/SeedCompany/cord-api-v3/blob/bcfef17a3c44066d8f5f247825f9e5035f18621e/src/components/engagement/engagement.resolver.ts#L65-L76

CarsonF (changes requested):
Yeah mostly looks good. The biggest thing is pointing to a single parent object instead of N relations.

src/components/post/post.service.ts
richardgirges: Done! will push up update shortly.

richardgirges:
All requested changes made

.gitignore
```diff
@@ -45,7 +45,7 @@ lerna-debug.log*
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
```
CarsonF, line 48: Should settings not be committed? Can you not override them if this file is committed?

richardgirges: Ah yeah, I should've checked with you on that. You can't override settings if this file is committed, which is why I made this update. Unless there's a different way of adding project-level settings that I'm not aware of

CarsonF: Mmmm I like this idea, but I hate having to fight shared settings. It's hard for me to know what's best here because I don't use vscode. Which of your settings didn't match, if I may ask?

CarsonF (changes requested):
We need a way to connect posts to projects. I think this means that `parentId` should always be required and the UI can give the project id as the parent. Then we need a `Project.posts` field returning a `SecuredPostList`.

.gitignore
richardgirges: The only setting I updated was this one:
```json
  "editor.formatOnSave": true,
```

I like to auto-format on every save.

CarsonF: Ah. we have that on the frontend. Ah you were the one committing it too lol. Would you mind just committing it here too?

I'm not sure about the `defaultFormatter`. If we could invoke eslint instead I think that would be best. It has other logic like removing unused imports, etc.
Maybe it's too heavy though.

richardgirges:
> We need a way to connect posts to projects. I think this means that `parentId` should always be required and the UI can give the project id as the parent. Then we need a `Project.posts` field returning a `SecuredPostList`.

I was building the frontend in a way where the "PostList" component could just be loaded in modularly, and you simply pass in a `parentId` as a filter. Thus allowing us to have a comment section under virtually any Node type. This would mean that something like `Project.posts` isn't necessary with my current approach.

I could change gears, and go with the sub-resolver for `Project.posts` approach - and simply rely on separate components for piecing together a comments section (e.g. PostListItem, PostReplyBox, etc).

Let me know how to proceed

CarsonF:
Hmm that sounds really flexible, I like that. I'm concerned however about security. I'm sure we'll get request to not allow people to "read posts from a high sensitivity project"

.gitignore
richardgirges: Word, I updated it! I actually tried to enable eslint formatting at first, but I couldn't get it to work for some reason and kept the default.

richardgirges:
Updated the backend with the following:
1. `Postable` abstract type with a `posts` sub-resolver that returns a `SecuredList` 
2. Extending `Postable` from `Project` type
3. Eliminated root level `posts` query that was previously returning the unsecured `PostList` as it didn't seem necessary anymore

richardgirges:
@CarsonF updated this PR with the latest list + DB-related refactors

Issue #1744, opened 2021-03-15
Add information to the email subject line for Workflow Notifications.


sethmcknight:
The email subject currently reads "Project Status Change - CORD Field", which is unhelpful for users who receive multiple status change emails at a time.

We need to update the subject line so that users can see what happened without having to open the email.

One option would be "`Project Name` is `Step Name` in CORD Field"

PR #1776, opened 2021-03-31
Add information to the email subject line for Workflow Notifications.

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1744


CarsonF (approved):
Thanks @zkhin!

Issue #1756, opened 2021-03-22
updateUser mutation doesn't work with empty role array.


michaelmarshall:
If I want to remove all roles from a user by passing an empty array to `updateUser` the server gives a 500. Its a cypher error, but its probably because the api code isn't checking for an empty array. This might need a special query to mark `active:false` for all the roles.
![image](https://user-images.githubusercontent.com/1165460/112044954-4cdd9c80-8b18-11eb-99ff-42bc2aa2c728.png)

PR #1777, opened 2021-03-31
updateUser mutation doesn't work with empty role array.

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1756

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -417,15 +417,17 @@ export class UserService {
         })
         .run();
 
-      await this.db
-        .query()
-        .match([
-          node('user', ['User', 'BaseNode'], {
-            id: input.id,
-          }),
-        ])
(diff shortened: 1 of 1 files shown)


CarsonF (changes requested):
We do need a way to say that this user now has no roles.

CarsonF (commented):
Nevermind. I missed the removal above.

I don't like though that I could have `[A, B, C]` and update to `[A, B]` and that would remove & re-add A & B property nodes.

It would better if we could diff to add/remove only necessary. 
Here's some similar logic 
https://github.com/SeedCompany/cord-api-v3/blob/bcfef17a3c44066d8f5f247825f9e5035f18621e/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts#L143-L144

CarsonF (approved):
Awesome, thanks @zkhin!

PR #1813, opened 2021-04-08
Change Engagement.products to get permissions from app

### LINKED ISSUE #1772: Project Manager Can't View Products
There appears to be a permissions configuration issue with the Project Manager role.

A Project Manager has reported they are unable to view Products for Language Engagements of Projects they are not a member of. They should be able to read Products globally, but only write if they are a Member.

Helpdesk Ticket here: SeedCompany/Management#120

CarsonF:
Fixes #1772 which has messed up security groups.

Currently this gives more access than expected, because our project managers, globally can read & create products.
https://github.com/SeedCompany/cord-api-v3/blob/d8d17718e15201d7f1d79524d81d04b978d19ffd/src/components/authorization/roles/project-manager-global.role.ts#L230-L230

### DIFF
--- a/src/components/engagement/engagement.repository.ts
+++ b/src/components/engagement/engagement.repository.ts
@@ -1,7 +1,8 @@
 import { Injectable } from '@nestjs/common';
 import { inArray, node, relation } from 'cypher-query-builder';
-import { ID } from '../../common';
+import { ID, Session } from '../../common';
 import { DatabaseService } from '../../core';
+import { Role, rolesForScope } from '../authorization';
 import { OngoingEngagementStatuses } from './dto';
 
 @Injectable()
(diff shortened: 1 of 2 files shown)


michaelmarshall (approved):

src/components/engagement/engagement.repository.ts
```diff
@@ -28,4 +29,30 @@ export class EngagementRepository {
...
+        node('projectMember', 'ProjectMember'),
+        relation('out', '', 'user', { active: true }),
+        node('user', 'User', { id: session.userId }),
+      ])
+      .match([
+        node('projectMember'),
+        relation('out', 'r', 'roles', { active: true }),
+        node('roles', 'Property'),
+      ])
+      .return('apoc.coll.flatten(collect(roles.value)) as memberRoles')
```
blakekrammes, line 50: What if a project member has multiple roles? Would `collect(roles.value)` yield an array of arrays? I don't remember if project members have a list for the value of roles if they have more than one role.

CarsonF: Hence the `apoc.coll.flatten`call

blakekrammes: Does it need to be recursive if `collect(roles.value) = [[],[],[]]`?

https://neo4j.com/labs/apoc/4.0/overview/apoc.coll/apoc.coll.flatten/

CarsonF: No because we only need to flatten one level.

PR #1820, opened 2021-04-09
Add property labels on update

### LINKED ISSUE #1754: Uniqueness constraints not enforced on updates
## Current Behavior

Not sure how widespread this is but at least `ProjectName` is only being applied at creation. Any changes to the project name do not create new `Property` nodes with the `ProjectName` label. Thus the uniqueness constraint is not enforced

## Expected Behavior

Labels should be applied on update as well as create.

CarsonF:
Fixes #1754

### DIFF
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -8,7 +8,7 @@ import {
   relation,
 } from 'cypher-query-builder';
 import type { Pattern } from 'cypher-query-builder/dist/typings/clauses/pattern';
-import { cloneDeep, Many, upperFirst } from 'lodash';
+import { cloneDeep, last, Many, startCase, upperFirst } from 'lodash';
 import { DateTime } from 'luxon';
 import { Driver, Session as Neo4jSession } from 'neo4j-driver';
 import { assert } from 'ts-essentials';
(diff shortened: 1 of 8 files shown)


src/core/database/database.service.ts
```diff
@@ -302,14 +306,22 @@ export class DatabaseService {
           active: true,
           createdAt,
         }),
-        node('newPropNode', 'Property', newPropertyNodeProps),
+        node('newPropNode', propLabels, newPropertyNodeProps),
```
blakekrammes, line 309: Should I use this instead of grabbing node labels dynamically in the query to mark as `Deleted_`?

CarsonF: Up to you. This certainly seems easier... Do we need to work about an additional Deleted_ prefix as we have in other places?

blakekrammes: You're asking if we should worry about prefixing properties on update?

blakekrammes: Sems like deletion and updating are the same because we never truly delete things. I suppose we could label with `Updated_` property and an `updatedAt` property on the old prop.

CarsonF: Nah I wouldn't introduce another prefix. Stick with `Deleted_` label prefix and deletedAt property

PR #1821, opened 2021-04-09
Prefix updated property node labels with Deleted_

blakekrammes:
Query has been tested with updating ProjectName.

### DIFF
--- a/src/components/engagement/engagement.rules.ts
+++ b/src/components/engagement/engagement.rules.ts
@@ -501,7 +501,7 @@ export class EngagementRules {
       .match([
         node('node', 'Engagement', { id }),
         relation('out', '', 'status', { active: false }),
-        node('prop', 'Property'),
+        node('prop', 'Deleted_Property'),
       ])
       .with('prop')
       .orderBy('prop.createdAt', 'DESC')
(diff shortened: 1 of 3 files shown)


blakekrammes:
Tests failed because getPrviousProjectStep is querying for `Property` nodes. Fixing it.

CarsonF (approved):

blakekrammes:
@CarsonF, did you get a chance to do any work on handling other cases of updating properties that don't use `database.service.updateProperties`? I just want to remember it for later. https://api.codestream.com/c/YG_W8EbROVBzgXfC/2ItcFNbYTK2vzV-kNVvg2Q


*[cord-api-v3] src/components/user/user.service.ts*
```
    // Update roles
    if (roles) {
      const removals = difference(user.roles.value, roles);
      const additions = difference(roles, user.roles.value);
      if (removals.length > 0) {
        await this.db
          .query()
          .match([
            node('user', ['User', 'BaseNode'], {
              id: input.id,
            }),
            relation('out', 'oldRoleRel', 'roles', { active: true }),
            node('oldRoles', 'Property'),
          ])
          .where({
            oldRoles: {
              value: inArray(removals),
            },
          })
          .set({
            values: {
              'oldRoleRel.active': false,
            },
          })
          .run();
      }


      if (additions.length > 0) {
        await this.db
          .query()
          .match([
            node('user', ['User', 'BaseNode'], {
              id: input.id,
            }),
          ])
          .create([...this.roleProperties(additions)])
          .run();
      }


      await this.authorizationService.roleAddedToUser(input.id, roles);
    }


    return await this.readOne(input.id, session);
  }
```

PR #1875, opened 2021-05-12
Limit concurrency for report sync

blakekrammes:


blakekrammes:
@tmpacifitech if you want to make changes and own this, feel free too. This is just what I wrote up to test out more limited concurrency and it looks like it's working.

tmpacifitech (approved):
This looks good. Thank you @blakekrammes !

blakekrammes:
I believe the logic is correct now, but we should do another check once we're ready to migrate.

blakekrammes:
Closing in favor of one migration branch that has this code and the code for moving existing P&P files to periodic reports.

PR #1882, opened 2021-05-17
1861 repository files

michaelmarshall:
- Pull DB code out of service files into new repository files. 
- The database service should no longer be called in any service file. The service file only calls the repo functions.
- The repo file shouldn't call any services.
- We will be reducing the number of abstractions after this. The repo file concept shouldn't create any more abstractions.


src/components/user/user.service.ts
```diff
@@ -159,132 +149,26 @@ export class UserService {
...
-    }
+    await this.userRepo.create(id, input);
     input.roles &&
       (await this.authorizationService.roleAddedToUser(id, input.roles));
-
     const dbUser = new DbUser();
     await this.authorizationService.processNewBaseNode(dbUser, id, id);
-
-    return result.id;
+    return id;
```
blakekrammes, line 157: This is returning the generated id instead of the id from the result of the db query. Do we need to return it from the db query if it's been created successfully?

src/components/user/user.service.ts
```diff
@@ -860,28 +523,10 @@ export class UserService {
...
-      .asResult<{
-        languageProficiency: LanguageProficiency;
-        languageId: ID;
-      }>()
-      .run();
+    const results = await this.userRepo.listKnownLanguages(userId, session);
 
     const knownLanguages = await Promise.all(
-      results.map(async (item) => {
+      results.map(async (item: any) => {
```
blakekrammes, line 529: Should probably clean up these `any`s at some point too. Probably need to define return types for the repo methods.

Issue #1897, opened 2021-05-25
Periodic Reports | Change Timing of Report that Displays on Card


sethmcknight:
We may need to change the way we select which report displays on each report card.

Most likely we will show the Previous reporting periods report.
- Quarterly we will show Q2 during Q3
- Monthly we will show Month 6 during Month 7

sethmcknight:
Update on the timeframes that we show.

Quarterly
- The Reporting Period displayed one the card should be one quarter behind

Monthly
- The Reporting Period displayed on the card should be

CarsonF:
In talking with @sethmcknight:
The reports will never be uploaded until after the period ends. Like submitting a time card for 2 weeks, you can’t do it until after the 2 weeks are done. So the current report due is always the previous period. So I think both of those words are important. So how about `current(Narrative|Progress|Financial)ReportDue` and `nextXReportDue` for the names.
Now the actual due date is a month after the report period ends. This IMO is still business logic so I’d like it in the API. It could easily be a field resolver called dueDate that just adds a month to the endDate.

Seth says we want to show the current & next report due for the entire reporting period. So these are always just a report behind, completely separate from +1 month logic.

sethmcknight:
Fixed in SeedCompany/cord-field#853

Issue #1896, opened 2021-05-25
Periodic Reports | Fix Monthly List Order


sethmcknight:
When viewing all Periodic Reports in the list view they appear mis-ordered.

For example M10, M11, M12, M7, M8, M9, M4, M5, M6...

Users expect these to show up in order (M4, M5, M6, M7, M8...)

We should consider showing the Month name rather than a number.

sethmcknight:
Fixed in SeedCompany/cord-field#853

PR #1902, opened 2021-05-26
Update periodic reports apis

tmpacifitech:
https://github.com/SeedCompany/cord-field/issues/860
https://github.com/SeedCompany/cord-api-v3/issues/1896
https://github.com/SeedCompany/cord-api-v3/issues/1897


CarsonF:
Why are we removing pagination?

tmpacifitech:
Added following resolvers
Project: `dueNarrativeReport` & `dueFinancialReport`
Engagement: `dueProgressReport`

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -111,17 +111,72 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
       );
   }
 
+  async currentProjectReport(
```
CarsonF, line 114:
```suggestion
  async reportForDate(
```

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -111,17 +111,72 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
...
+    projectId: string,
+    reportType: ReportType,
+    startDate: DateTime
```
CarsonF, line 117:
```suggestion
    startDate: CalendarDate
```

Use CalendarDates instead of DateTimes. This is the property type being matched on report (start date) and you won't need to cast to a date in cypher this way (`date($date)`).

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -111,17 +111,72 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
...
+  ) {
+    return await this.db
+      .query()
+      .match([
+        node('project', 'Project', { id: projectId }),
```
CarsonF, line 122:
```suggestion
        node('baseNode', 'BaseNode', { id: projectId }),
```
Now it works for engagements as well.

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -152,13 +157,68 @@ export class PeriodicReportService {
...
+  ): Promise<DueReport> {
+    let interval: DurationUnit = 'quarter';
+    if (reportType === ReportType.Financial) {
+      const project = await this.project.readOne(projectId, session);
+      if (project.financialReportPeriod.value === ReportPeriod.Monthly) {
+        interval = 'month';
+      }
+    }
+
+    const currentReportStartDate = DateTime.now()
```
CarsonF, line 192:
```suggestion
    const currentReportStartDate = CalendarDate.local()
```

I think DateTime.now() is new, if you want to use CalendarDate.now() you'll need to add it to our class.

src/components/periodic-report/dto/due-report.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+import { SecuredProperty } from '../../../common';
+import { IPeriodicReport, PeriodicReport } from './periodic-report.dto';
+
+@ObjectType()
+export class DueReport {
```
CarsonF, line 6: Does it make more sense for this DTO & field resolvers to be plural since they are referring to multiple reports?


Also just FYI since this doesn't have an ID field we'll want it to be cached on the parent object in the UI
```ts
// api/typePolicies.base.ts
DueReport: { keyFields: false },
```

src/components/periodic-report/dto/due-report.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+  @Field(() => IPeriodicReport, { nullable: true })
+  readonly current: PeriodicReport | null;
```
CarsonF, line 8: I think each of these should be secured.

src/components/periodic-report/dto/due-report.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  @Field(() => IPeriodicReport, { nullable: true })
+  readonly next: PeriodicReport | null;
+}
+
+@ObjectType({
+  description: SecuredProperty.descriptionFor('due report'),
+})
+export class SecuredDueReport extends SecuredProperty(DueReport) {}
```
CarsonF, line 17: Probably doesn't make sense since this is just a holder/grouping of individually secured props.

CarsonF (changes requested):

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -111,17 +111,46 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
       );
   }
 
+  async reportForDate(
+    parentId: string,
```
CarsonF, line 115:
```suggestion
    parentId: ID,
```

src/components/project/project.resolver.ts
```diff
@@ -312,6 +317,30 @@ export class ProjectResolver {
...
+  async dueNarrativeReport(
+    @AnonSession() session: Session,
+    @Parent() project: Project
+  ): Promise<PeriodicReport | null> {
+    return await this.periodicReportService.getDueReport(
+      project.id,
+      ReportType.Narrative,
+      session
+    );
+  }
```
tmpacifitech, line 342: These resolvers response should be SecuredPeriodicReport.. right?

CarsonF: Correct

tmpacifitech: kk

src/components/project/project.resolver.ts
```diff
@@ -312,6 +312,40 @@ export class ProjectResolver {
     );
   }
 
+  @ResolveField(() => SecuredPeriodicReport, { nullable: true })
```
CarsonF, line 315:
```suggestion
  @ResolveField(() => SecuredPeriodicReport)
```
Not nullable, we'll always get an object back for canRead.

Can you add a description here too?

src/components/project/project.resolver.ts
```diff
@@ -312,6 +312,40 @@ export class ProjectResolver {
...
+    @Parent() project: Project
+  ): Promise<SecuredPeriodicReport> {
+    const value = await this.periodicReportService.getCurrentReport(
+      project.id,
+      ReportType.Financial,
+      session
+    );
+    return {
+      canRead: true,
+      canEdit: true,
```
CarsonF, line 327:
```suggestion
      canEdit: false,
```
This prop cannot be modified since it's calculated.

src/components/engagement/engagement.resolver.ts
```diff
@@ -158,9 +159,25 @@ export class EngagementResolver {
...
     return this.periodicReports.listEngagementReports(
       engagement.id,
-      ReportType.Progress,
       input,
       session
     );
   }
+
+  @ResolveField(() => SecuredPeriodicReport)
+  async dueProgressReport(
```
CarsonF, line 168:
```suggestion
  async currentProgressReport(
```

CarsonF (changes requested):

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -152,13 +157,42 @@ export class PeriodicReportService {
...
+    if (reportType === ReportType.Financial) {
+      const project = await this.project.readOne(parentId, session);
+      if (project.financialReportPeriod.value === ReportPeriod.Monthly) {
+        interval = 'month';
+      }
+    }
+
+    const currentReportStartDate = CalendarDate.local()
+      .minus({ [interval]: 1 })
+      .startOf(interval);
```
CarsonF, line 175: Instead of needing to figure this out can't we just add a condition to the db query? `start < current < end`

src/common/promise.ts
```diff
@@ -0,0 +1,10 @@
+import * as plimit from 'p-limit';
+
+export const runPromises = async <T>(
+  promises: Array<Promise<T>>,
+  concurrency = 10
```
CarsonF, line 5: This doesn't do what you think it does. Since all the promises have been created they are all already in progress and the concurrency limit is meaningless.

tmpacifitech: Removed p-limit usages

CarsonF (approved):
Looks great, thanks @tmpacifitech!

tmpacifitech:
> Looks great, thanks @tmpacifitech!

Woot! thankyou sir!

PR #1904, opened 2021-05-26
1877/1883 periodic report migration

blakekrammes:
This includes code to create periodic reports on all existing projects and engagements. It also includes code to move previously uploaded P&Ps that match a reporting period to the correct periodic report.

Replaces https://github.com/SeedCompany/cord-api-v3/pull/1880


src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -26,10 +32,14 @@ export class SyncProgressReportToEngagementDateRange
   constructor(
     private readonly periodicReports: PeriodicReportService,
     private readonly engagements: EngagementService,
+    private readonly config: ConfigService,
     @Logger('progress-report:engagement-sync') private readonly logger: ILogger
   ) {}
 
   async handle(event: SubscribedEvent) {
+    if (this.config.periodicReportMigration) {
```
blakekrammes, line 40: During migration, we delete engagement overrides if they are incorrect since we have bad data from V2. This is to prevent unwanted effects during the migration when we update the engagement. This will also prevent periodic reports from being generated when the project dates are added as well, so we need to remove it when the periodic report migration is done. Perhaps there's a better way to accomplish this.

src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -27,10 +33,15 @@ export class SyncProgressReportToEngagementDateRange
...
     private readonly engagements: EngagementService,
+    private readonly config: ConfigService, // Remove after periodic report migration
     @Logger('progress-report:engagement-sync') private readonly logger: ILogger
   ) {}
 
   async handle(event: SubscribedEvent) {
+    // Remove after periodic report migration
+    if (this.config.periodicReportMigration) {
+      return;
+    }
```
blakekrammes, line 44: During migration, we delete engagement overrides if they are incorrect since we have bad data from V2 (discussed this with Seth). 

This check for this new config var is to prevent unwanted effects during the migration when we update the engagement. However, this will also prevent periodic reports from being generated when the project dates are added so we need to remove it when the periodic report migration is done. 

Perhaps there's a better way to prohibit this handler from running during the migration that doesn't also prohibit generating reports when a project has dates assigned to it. **Although, if we're freezing Cord during migration and remove this right after, we should have no problems.** I've added comments at all the places we need to remove code when the migration is finished.

PR #1908, opened 2021-06-01
Add sensitivity access restriction to budget and budget record

elisabeth-sorrell:


CarsonF:
What about `Project.currentBudget`? https://github.com/SeedCompany/cord-api-v3/blob/8ad6c6329732934b1c703bf87ebd27b430e6d08f/src/components/project/project.service.ts#L800-L808

src/components/authorization/authorization.service.ts
```diff
@@ -254,6 +265,23 @@ export class AuthorizationService {
...
 
+  isSensitivityAllowed(
+    grant: Partial<
+      Record<Action, boolean> & Record<'sensitivityAccess', Sensitivity>
+    >,
+    sensitivity?: Sensitivity
+  ): boolean {
+    const sensitivityRank = { High: 3, Medium: 2, Low: 1 };
+    if (
+      sensitivity &&
```
CarsonF, line 276: Should we throw an error if sensitivity check is required but none was given?

elisabeth-sorrell: done

src/components/authorization/authorization.service.ts
```diff
@@ -265,13 +271,22 @@ export class AuthorizationService {
...
     grant: Partial<
       Record<Action, boolean> & Record<'sensitivityAccess', Sensitivity>
     >,
+    resource: TResource,
+    prop: string,
     sensitivity?: Sensitivity
   ): boolean {
     const sensitivityRank = { High: 3, Medium: 2, Low: 1 };
+    if (grant.sensitivityAccess && !sensitivity) {
+      throw new InputException(
```
CarsonF, line 284:
```suggestion
      throw new ServerException(
```
Input is for user input. This would be a developer mistake.

CarsonF (changes requested):

PR #1912, opened 2021-06-02
Add ProductStep component

tmpacifitech:


CarsonF:
Thanks @tmpacifitech for working on this. Sadly, I had actually designed the schema updates for this feature already in the [product-progress](https://github.com/SeedCompany/cord-api-v3/tree/product-progress) branch before my time off. It seems like that was lost in the shuffle. I just rebased it.
I'm looking through the differences between what I initially wrote and what you have. Maybe you could see what I have and bring up any concerns if find?

PR #1922, opened 2021-06-05
Lower concurrency for report migration

blakekrammes:


CarsonF (approved):

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -61,10 +61,15 @@ export class PeriodicReportResolver {
     description: 'Create periodic reports for projects and engagements',
   })
   async syncAllReports(@LoggedInSession() session: Session) {
```
blakekrammes, line 63: Does it matter if this is async?

blakekrammes: Looks like it doesn't matter. https://stackoverflow.com/questions/45594596/async-function-without-await-in-javascript

CarsonF: Doesn't matter

CarsonF (approved):

PR #1934, opened 2021-06-09
Fix latest report submitted query

CarsonF:
CQB doesn't like empty string for missing label, but rather undefined.
Better to match BaseNode label anyways.

### DIFF
--- a/src/components/periodic-report/periodic-report.repository.ts
+++ b/src/components/periodic-report/periodic-report.repository.ts
@@ -159,7 +159,7 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
     return await this.db
       .query()
       .match([
-        node('', '', { id: parentId }),
+        node('', 'BaseNode', { id: parentId }),
         relation('out', '', 'report', { active: true }),
         node('pr', `${type}Report`),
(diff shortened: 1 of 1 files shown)


blakekrammes:
Ah. Ok.

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -159,7 +159,7 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
     return await this.db
       .query()
       .match([
-        node('', '', { id: parentId }),
+        node('', 'BaseNode', { id: parentId }),
```
blakekrammes, line 162: It's weird that this fails when `undefined` wouldn't.

blakekrammes (approved):

src/components/periodic-report/periodic-report.repository.ts
CarsonF: I guess `[]` would work too. 
https://github.com/jamesfer/cypher-query-builder/blob/3986fea0a4d97eadff829de0e2fe66665529ab4f/src/utils.ts#L90

PR #1958, opened 2021-06-19
Query Result Types

CarsonF:
# Query Result Types
I took the plunge and patched cypher-query-builder package to fix some of the type issues we have with it.
- `Query` now has a generic `Result` parameter.
- The forced `Dictionary` output was removed. The generic is just passed straight out now.
- `Query.return()` method now has a generic to change the query result generic.
- Our `Query.asResult()` is now deprecated in favor of `return()`.

### In practice it looks like this

```ts
query.return<{ id: ID }>('node.id as id')
```

## No more `asResult()`
```diff
- .return('node.id as id')
- .asResult<{ id: ID }>()
+ .return<{ id: ID }>('node.id as id')
```

This change also removes the limitation that asResult() had where it had to be the last method called before first/run().
See here how the type shape is defined on the return method and then orderBy & limit are called before executing the query.
https://github.com/SeedCompany/cord-api-v3/blob/9121810c524ba2e6095bc26cf4406cb7aae0bfda/src/components/file/file.repository.ts#L189-L192

## Type safety is default now (instead of opt-in)
The default result type is now `unknown` instead of `any/Dictionary`.
If you want to pull data out you need to define the shape on the `return()` method.
This makes our queries much safer as type safety is no longer opt-in.

I've updated all of our unsafe DB queries to specify return shape.
In doing so bugs where revealed which fixed in [this commit](https://github.com/SeedCompany/cord-api-v3/pull/1958/commits/204bf87a77b4da5d16071b6794f98e29f11509c4).

# Query.map

I added a new map method to help pull out single columns as this what most of our queries do.
```ts
const id = 
  this.db.query()...
  .return<{ id: ID }>('node.id as id')
  .map(row => row.id)
  .first();
```
There's also a string shortcut
```ts
const id = 
  this.db.query()...
  .return<{ id: ID }>('node.id as id')
  .map('id')
  .first();
```

# Misc

- Updated all of our unsafe DB queries to specify return shape
- Removed `listProjectsWithDateRange` query used in migration
- Consolidated DB queries for project list relation permissions
  ```diff
  - getEngagementPermission
  - getTeamMemberPermission
  - getPartnershipPermission
  + permissionsForListProp
  ```
- Deprecated `DatabaseService.list` it's only used in unavailability module and needs to be removed. Noted as such.
- Split our query augmentation methods into their own directory so they are more managable.
- Fixed relative path conversion for GraphQL stack traces for anonymous frames
  ```diff
  - at /Users/carsonfull/Repos/seedco/cord-api-v3/src/components/user/education/education.repository.ts:60:13
  + at components/user/education/education.repository.ts:60:13
    at EducationRepository.readOne (components/user/education/education.repository.ts:59:12)
  ```


blakekrammes:
I like the map. :)

PR #1965, opened 2021-06-21
New createNode & createRelationships fragments

CarsonF:
# `createNode`
`createBaseNode` has gone through a few iterations. Most of its intended functionality has been scratched. It's property list is unnecessarily verbose. `isPublic` & `isOrgPublic` are required but are unused.

## Current/Previous
```ts
const secureProps: Property[] = [
  {
    key: 'status',
    value: BudgetStatus.Pending,
    isPublic: false,
    isOrgPublic: false,
    label: 'BudgetStatus',
  },
  {
    key: 'universalTemplateFile',
    value: await generateId(),
    isPublic: false,
    isOrgPublic: false,
  },
  {
    key: 'canDelete',
    value: true,
    isPublic: false,
    isOrgPublic: false,
  },
];
const result = await this.db
  .query()
  .apply(createBaseNode(await generateId(), 'Budget', secureProps))
  .return<{ id: ID }>('node.id as id')
  .first();
```

## New
Here's an example of the new function:
```ts
const initialProps = {
  status: BudgetStatus.Pending,
  universalTemplateFile: await generateId(),
  canDelete: true,
};
const result = await this.db
  .query()
  .apply(await createNode(Budget, { initialProps }))
  .return<{ id: ID }>('node.id as id')
  .first();
```
- Initial property values are a simple object.
- We give the function the resource class. With this class it can lookup the DB labels needed for both the BaseNode and for properties. See [here](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/budget/dto/budget.dto.ts#L25) that status has a label.
- The function is async so it can generate the ID for us
- With the resource class and the options object this new function should be able to sustain us over a few iterations

# `createRelationships`

Relationship creation between _BaseNodes_ has never been attempted to be abstracted, yet it's pretty much the same everywhere.
1. If ID is given, match that ID. 
2. Create relationship between node and matched relationship node

With that in mind, I've created a `createRelationships` function. It works like this:
```ts
createRelationships(Location, 'out', {
  FundingAccount: { // base node label to match
    fundedBy: // label of the relationship to create 
      input.fundingAccountId, // id of the base node to match 
  }, 
  User: { createdBy: input.creatorId } // simple one liner
})
```
This creates two relationships labeled `fundedBy` & `createdBy` from the previously matched `(node: Location)` to the `FundingAccount` and `User` by their IDs.

- Zero or more IDs can be given to under the relationship label.
  - Zero (`[]/null/undefined`) omits the match & create clauses.
  - 1+ matches all and creates all
- The BaseNode resources being connected to a strictly keyed from the `ResourceMap` object
- Passing the resource is future proofing. Not having this in other query fragments, like `createBaseNode`, renders them needing to be replaced.
  - The near future use could be strictly typing the relationship names from the `Relationships` static prop on the DTO. Currently any string is allowed there.

While this does look very different from the cypher produced, I'd argue that it is not any less readable.

The biggest caveat here is that the IDs given here must exist in DB. If not then the query result will return an empty set, which would be bad for `createNode` as we'd say the resource failed to create but the base node part created fine. Using a transaction would also fix this.

# Real App Example
https://github.com/SeedCompany/cord-api-v3/blob/ba056d3bf3ccbf40dde73ddcbbe25deb144b998f/src/components/engagement/engagement.repository.ts#L217-L228
```cql
CALL {
  CREATE (node:InternshipEngagement:Engagement:BaseNode { createdAt: datetime(), id: '8i5sKkr4m5g' }),
      (node)-[:position { active: true, createdAt: datetime() }]->(:InternPosition:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:methodologies { active: true, createdAt: datetime() }]->(:ProductMethodology:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:growthPlan { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: 'AX5OOZB60No', sortValue: 'AX5OOZB60No' }),
      (node)-[:completeDate { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:disbursementCompleteDate { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:communicationsCompleteDate { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:startDateOverride { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:endDateOverride { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:status { active: true, createdAt: datetime() }]->(:EngagementStatus:Property { createdAt: datetime(), value: 'InDevelopment', sortValue: 'InDevelopment' }),
      (node)-[:initialEndDate { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:statusModifiedAt { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:lastSuspendedAt { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:lastReactivatedAt { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null, sortValue: null }),
      (node)-[:modifiedAt { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: datetime(), sortValue: datetime() }),
      (node)-[:canDelete { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: true, sortValue: true })
  RETURN node
}
CALL {
  WITH node
  MATCH (engagement:Project { id: 'y9IBRDZnUIN' }),
      (intern:User { id: 'T8DJ28wpSrW' })
  CREATE (node)<-[:engagement { active: true, createdAt: datetime() }]-(engagement),
      (node)-[:intern { active: true, createdAt: datetime() }]->(intern)
  RETURN engagement, intern
}
RETURN node.id as id
```
(BTW this query output is straight from log, I didn't tweak the formatting at all when pasting here ☺️)
Notice the base node labels are determined automatically, and some properties have special labels.
country of origin & mentor ID's were omitted from API mutation & thus not in the DB query.

# End Result
Here we can see how much of a line saving it is. That reduced number of lines helps with cognitive overhead.
```
 src/components/engagement/engagement.repository.ts | 151 +++++++++++++++++++++++++-----------------------------------------------------------------------
 src/components/project/project.repository.ts       |  77 ++++++++++++++++++++++++++++++++++++++++++++++---
 src/components/project/project.service.ts          | 295 +++++++++++++++++++++++++-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 3 files changed, 150 insertions(+), 373 deletions(-)
```

### DIFF
--- a/src/common/db-label.decorator.ts
+++ b/src/common/db-label.decorator.ts
@@ -1,5 +1,6 @@
 import { uniq, without } from 'lodash';
 import { getParentTypes } from './parent-types';
+import { isResourceClass } from './resource.dto';
 import { AbstractClassType } from './types';
 
 const DbLabelSymbol = Symbol('DbLabelSymbol');
@@ -27,7 +28,9 @@ export const getDbClassLabels = (
   const labels =
     decorated?.flatMap((l) => l.split(':')) ??
     without(
-      getParentTypes(type).map((t) => t.name),
+      getParentTypes(type)
+        .filter(isResourceClass)
+        .map((t) => t.name),
       'Resource'
     );
   return uniq([...labels, 'BaseNode']);
--- a/src/common/db-label.decorator.ts
+++ b/src/common/db-label.decorator.ts
@@ -1,4 +1,5 @@
-import { uniq } from 'lodash';
+import { uniq, without } from 'lodash';
+import { getParentTypes } from './parent-types';
 import { AbstractClassType } from './types';
 
 const DbLabelSymbol = Symbol('DbLabelSymbol');
@@ -12,6 +13,22 @@ export const getDbPropertyLabels = (
 ) => {
   // @ts-expect-error property decoration is on instance of object
   const obj = new type();
-  const labels = Reflect.getMetadata(DbLabelSymbol, obj, property);
-  return uniq([...(labels ?? []), 'Property']);
+  const labels = Reflect.getMetadata(DbLabelSymbol, obj, property) as
+    | string[]
+    | undefined;
+  const normalized = labels?.flatMap((l) => l.split(':'));
+  return uniq([...(normalized ?? []), 'Property']);
+};
+
+export const getDbClassLabels = (
+  type: AbstractClassType<unknown>
+): readonly string[] => {
+  const decorated: string[] | null = Reflect.getMetadata(DbLabelSymbol, type);
+  const labels =
+    decorated?.flatMap((l) => l.split(':')) ??
+    without(
+      getParentTypes(type).map((t) => t.name),
+      'Resource'
+    );
+  return uniq([...labels, 'BaseNode']);
 };
--- a/src/common/resource.dto.ts
+++ b/src/common/resource.dto.ts
@@ -5,6 +5,7 @@ import { ID, IdField } from './id-field';
 import { DateTimeField } from './luxon.graphql';
 import { SecuredProps, UnsecuredDto } from './secured-property';
 import { AbstractClassType } from './types';
+import { has } from './util';
 
 @InterfaceType()
 export abstract class Resource {
@@ -34,6 +35,11 @@ export type ResourceShape<T> = AbstractClassType<T> & {
   BaseNodeProps?: string[];
   Relations?: Record<string, any>;
 };
+
+export const isResourceClass = <T>(
+  cls: AbstractClassType<T>
+): cls is ResourceShape<T> =>
+  has('Props', cls) && Array.isArray(cls.Props) && cls.Props.length > 0;
 
 export type MaybeUnsecuredInstance<TResourceStatic extends ResourceShape<any>> =
   TResourceStatic['prototype'] | UnsecuredDto<TResourceStatic['prototype']>;
--- a/src/components/authorization/authorization.service.ts
+++ b/src/components/authorization/authorization.service.ts
@@ -11,9 +11,9 @@ import {
 } from 'lodash';
 import {
   getParentTypes,
-  has,
   ID,
   isIdLike,
+  isResourceClass,
   isSecured,
   keys,
   mapFromList,
@@ -206,9 +206,7 @@ export class AuthorizationService {
     // convert resource to a list of resource names to check
     const resources = getParentTypes(resource)
       // if parent defines Props include it in mapping
-      .filter(
-        (r) => has('Props', r) && Array.isArray(r.Props) && r.Props.length > 0
-      )
+      .filter(isResourceClass)
       .map((r) => r.name);
 
     const normalizeGrants = (role: DbRole) =>
--- a/src/components/engagement/engagement.repository.ts
+++ b/src/components/engagement/engagement.repository.ts
@@ -176,7 +176,7 @@ export class EngagementRepository extends CommonRepository {
       .query()
       .apply(await createNode(LanguageEngagement, { initialProps }))
       .apply(
-        createRelationships({
+        createRelationships(LanguageEngagement, {
           in: { Project: { engagement: projectId } },
           out: { Language: { language: languageId } },
         })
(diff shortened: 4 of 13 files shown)


blakekrammes:
This is great! I'm glad this is getting cleaned up and abstracted. I think cognitively there's a bit of a bump here though.

```typescript
.query() 
.apply(await createNode(InternshipEngagement, { initialProps })) 
.apply( 
  createRelationships(InternshipEngagement, { 
    in: { Project: { engagement: projectId } }, 
    out: { 
      User: { intern: internId, mentor: mentorId }, 
      Location: { countryOfOrigin: countryOfOriginId }, 
    }, 
  }) 
) 
.return<{ id: ID }>('node.id as id');
 ```
 
It's not immediately clear what the different parts of the relationship create query are. I think the main confusion is the disassociation of the target node label and the target node id and the association between the node id and the relationship label. At first glance it looks like you're creating an engagement property on the Project node. It'd be great if there was a clean a way to communicate the parts of the query explicitly:
 
 ```typescript
.query() 
.apply(await createNode(InternshipEngagement, { initialProps })) 
.apply( 
  createRelationships(InternshipEngagement, { 
    in: [{ targetNode: { label: Project, id: projectId }, relationship: engagement }], 
    out: [{...}, {...}], 
  }) 
) 
.return<{ id: ID }>('node.id as id');
 ```
 
 I could get used to it of course, but thought it was worth mentioning.

**Comment by CarsonF** [comment]
Yeah I hear you. It doesn't exactly read left to right.

However, the IDs and base node labels are strictly typed, so they can't be messed up, it's only the relationship name that is dynamic.

I also didn't want to have to re-type it for multiple relationships too. For example:
https://github.com/SeedCompany/cord-api-v3/blob/ba056d3bf3ccbf40dde73ddcbbe25deb144b998f/src/components/project/project.repository.ts#L193-L201
vs what you have proposed:
```ts
createRelationships(IProject, 'out', [
  { targetNode: { label: 'FieldRegion', id: fieldRegionId }, relationship: 'fieldRegion' },
  { targetNode: { label: 'Location', id: primaryLocationId }, relationship: 'primaryLocation' },
  { targetNode: { label: 'Location', id: otherLocationIds }, relationship: 'otherLocations' },
  { targetNode: { label: 'Location', id: marketingLocationId }, relationship: 'marketingLocation' },
  { targetNode: { label: 'Organization', id: this.config.defaultOrg.id }, relationship: 'owningOrganization' },
])
```
A lot more syntax I have to ignore while trying to understand the business logic. What do you think? Does it add more clarity?

CarsonF:
Really the only reason for that structure is to allow multiple relations to a base node at once.
If we wanted to forgo that then something like this would work:
```ts
createRelationships(IProject, 'out', {
  fieldRegion: ['FieldRegion', fieldRegionId],
  primaryLocation: ['Location', primaryLocationId],
  otherLocations: ['Location', otherLocationIds],
  marketingLocation: ['Location', marketingLocationId],
  owningOrganization: ['Organization', this.config.defaultOrg.id],
});

blakekrammes:
I like that a lot. It's also clearer because the relationship name comes first, so it's much easier to sort out.

blakekrammes:
IMO the above reads more like cypher and is more easily processed despite the bit of redundancy.

zkhin:
That really looks good to me and it will make the code more readable. Great job. 🎉

PR #1971, opened 2021-06-23
Make financialReportPeriod nullable

blakekrammes:
What needed to happen in the UI again? I'm ready with the queries.


src/components/project/dto/project.dto.ts
```diff
@@ -128,7 +128,7 @@ class Project extends PinnablePostableResource {
   @Field()
   readonly financialReportReceivedAt: SecuredDateTime;
 
-  @Field()
+  @Field({ nullable: true })
```
CarsonF, line 131: Revert. Secured fields are not nullable. Their values are by default.

CarsonF (changes requested):

PR #1972, opened 2021-06-23
Report received date

blakekrammes:


CarsonF (changes requested):
Can you revert the breaking changes here? Just adding the receivedDate field to UploadPeriodicReport should be sufficient.

CarsonF (approved):
Made a couple changes, looks good!

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -22,50 +24,36 @@ import {
...
-      .create([
-        node('parent'),
-        relation('out', '', 'report', {
-          active: true,
-          createdAt,
-        }),
-        node('newPeriodicReport'),
-      ])
-      .return('newPeriodicReport.id as id');
+      .apply(await createNode(Report, { initialProps }))
```
blakekrammes, line 45: Does createNode add `PeriodicReport` and the specific report type label (i.e. `ProgressReport`)?

blakekrammes: `BaseNode labels are pulled from the resource's class hierarchy or manually`. That sounds like it might do the trick.

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -72,25 +74,36 @@ export class PeriodicReportService {
...
+    await this.authorizationService.verifyCanEditChanges(
+      resolveReportType(current),
+      current,
+      changes
     );
 
-    return report;
+    const { reportFile, ...simpleChanges } = changes;
+
+    const updated = await this.repo.updateProperties(current, simpleChanges);
```
blakekrammes, line 88: What if there are no `simpleChanges`?

CarsonF: Then this will do nothing.

src/components/periodic-report/periodic-report.repository.ts
CarsonF: Yes and yes

blakekrammes:
Ok. Let's merge it!

PR #1974, opened 2021-06-25
Product Progress

tmpacifitech:


src/components/product-progress/product-progress.repository.ts
```diff
@@ -1,7 +1,189 @@
...
+    const id = await generateId();
+    const createdAt = DateTime.local();
+    const query = this.db
+      .query()
+      .match([node('product', 'Product', { id: input.productId })])
+      .match([node('report', 'PeriodicReport', { id: input.reportId })])
+      .create([
+        [
+          node('node', ['ProductProgress', 'BaseNode'], {
+            id,
```
CarsonF, line 44: If we are actually creating an ID we should add it the DTO. But I left it off because `ProductProgress` isn't really created. It's just an artificial grouping of each product's step's progress for the reporting period.

src/components/product-progress/product-progress.service.ts
```diff
@@ -16,34 +11,44 @@ export class ProductProgressService {
...
+      reportId,
+      steps,
+    };
+  }
 
-    throw new NotImplementedException().with(input, session, steps);
+  async update(input: ProductProgressInput) {
+    await this.repo.removeOldProgress(input.productId, input.reportId);
+    await this.repo.create(input);
   }
```
CarsonF: Ah yeah we don't want to delete & recreate the `ProductProgress` node on update. I'm not sure if I stated this but I was thinking that the steps input array could be partial - only changing / setting values for the steps given. Or maybe you coded the UI to send all the step values at once?

Maybe we do a merge and on create set the ID and createdAt?

src/components/product-progress/product-progress.repository.ts
```diff
@@ -1,7 +1,189 @@
...
+      ])
+      .match([
+        node('node'),
+        relation('out', '', 'step', { active: true }),
+        node('stepProgress', 'StepProgress'),
+      ])
+      .with('collect(stepProgress.id) as ids')
+      .return('ids')
+      .asResult<{ ids: [ID] }>()
+      .first();
```
CarsonF, line 114: This call `this.hydrate()` here instead of returning only the IDs. Takes this query from O(n+1) to O(1)

CarsonF (changes requested):

PR #1978, opened 2021-06-28
Delete in Changeset

zkhin:
Need to implement in `Partnership.delete` as well.

### DIFF
--- a/src/components/ceremony/handlers/delete-engagement-default-ceremony.handler.ts
+++ b/src/components/ceremony/handlers/delete-engagement-default-ceremony.handler.ts
@@ -17,6 +17,7 @@ export class DetachEngagementRootDirectoryHandler
       return;
     }
 
+    // TODO Ceremony should be changeset aware
     if (engagement.changeset) {
       return;
     }
--- a/src/components/ceremony/handlers/delete-engagement-default-ceremony.handler.ts
+++ b/src/components/ceremony/handlers/delete-engagement-default-ceremony.handler.ts
@@ -17,6 +17,10 @@ export class DetachEngagementRootDirectoryHandler
       return;
     }
 
+    if (engagement.changeset) {
+      return;
+    }
+
     await this.ceremonies.delete(ceremonyId, session);
   }
 }
--- a/src/core/database/common.repository.ts
+++ b/src/core/database/common.repository.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@nestjs/common';
-import { ID, Session } from '../../common';
+import { node, relation } from 'cypher-query-builder';
+import { ID, isIdLike, ServerException, Session } from '../../common';
 import { DatabaseService } from './database.service';
 
 /**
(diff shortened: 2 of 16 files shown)


src/core/database/common.repository.ts
```diff
@@ -13,7 +14,31 @@ export class CommonRepository {
...
+            node('changeset'),
+            relation('out', 'oldRel', 'changeset', { deleting: true }),
+            node('node'),
+          ])
+          .delete('oldRel')
+          .create([
+            node('changeset'),
+            relation('out', '', 'changeset', { deleting: true }),
+            node('node'),
+          ])
```
CarsonF, line 37: It looks like we are trying to delete and then re create. What if we just did a merge on the relationship and then set the deleting prop after in a setValues?

Also are we doing active here? Does that matter?

src/core/database/common.repository.ts
```diff
@@ -13,7 +14,31 @@ export class CommonRepository {
     return await this.db.checkDeletePermission(id, session);
   }
 
-  async deleteNode(objectOrId: { id: ID } | ID) {
-    await this.db.deleteNode(objectOrId);
+  async deleteNode(objectOrId: { id: ID } | ID, changeset?: ID) {
+    if (!changeset) {
+      await this.db.deleteNode(objectOrId);
+    } else {
```
CarsonF, line 20: Would you mind returning above and removing the else statement? It's a pet peeve of mine 😬

CarsonF (changes requested):

src/components/ceremony/handlers/delete-engagement-default-ceremony.handler.ts
```diff
@@ -17,6 +17,10 @@ export class DetachEngagementRootDirectoryHandler
       return;
     }
 
+    if (engagement.changeset) {
```
CarsonF, line 20: Could the changeset just be passed though here to the ceremony delete? I know we still need to get ceremonies completely changeset aware, so if not can you leave a todo stating this is temporary until then?

src/core/database/common.repository.ts
```diff
@@ -13,7 +14,26 @@ export class CommonRepository {
...
+      await this.db
+        .query()
+        .match([node('node', 'BaseNode', { id })])
+        .match([node('changeset', 'Changeset', { id: changeset })])
+        .merge([
+          node('changeset'),
+          relation('out', 'rel', 'changeset'),
+          node('node'),
+        ])
+        .onCreate.setValues({ 'rel.active': true, 'rel.deleting': true })
```
CarsonF, line 33: I don't think we need the `.onCreate`. If the relationship previously existed we still want to mark it as deleting.

src/core/database/query/lists.ts
```diff
@@ -66,3 +66,10 @@ export const defaultSorter =
           .orderBy(sortValSecuredProp, order)
       : q.with('*').orderBy(sortValBaseNodeProp, order);
   };
+
+export const whereNotDeletedInChangeset = (changeset?: ID) => (query: Query) =>
+  changeset
+    ? query.raw(
+        `WHERE NOT (node)<-[:changeset { active: true, deleting: true }]-(:Changeset { id: '${changeset}' })`
+      )
```
CarsonF, line 74: Please bind values. I'm pretty sure it's not just safer, it helps with performance as well (since a hard coded value would make each query it would need to be parsed every time).
```suggestion
        'WHERE NOT (node)<-[:changeset { active: true, deleting: true }]-(:Changeset { id: $changesetId })',
        { changesetId: changeset }
      )
```

src/core/database/common.repository.ts
zkhin: Then we need both `.onMatch` and `.onCreate`?

CarsonF: I think they can just be two completely separate clauses

CarsonF (approved):

Issue #1830, opened 2021-04-13
FA Unable to See Full Partner List


ParkeBrown:
**URL:** https://cordfield.com/partners
**Log Rocket:**
https://app.logrocket.com/x6d22y/cord-prod/s/4-d289cdc3-2ee6-455e-a27a-194a78d56996/10?filters=%255B%257B%2522id%2522%253A%2522HJtXu_QId%2522%252C%2522type%2522%253A%2522userID%2522%252C%2522operator%2522%253A%257B%2522name%2522%253A%2522is%2522%252C%2522type%2522%253A%2522IS%2522%252C%2522hasStrings%2522%253Atrue%252C%2522autocompleteEnabled%2522%253Atrue%257D%252C%2522strings%2522%253A%255B%25225f7ce4ea4d64ae7b06572bd7%2522%255D%257D%255D&fromTab=sessions

**User Report:**
"Though I am still learning how Cord works, I first presumed that when I click on the 'Partners' line of the menu on the left side of the main Cord screen, that I should be able to view all of the partners related to Projects that I am assigned as a Financial Analyst. However, as you can see from the screen shot below, only two are listed. If this should be a list of my projects, there should be more. Is this the purpose of the Partners list, to show me my partners? Or what is this Partners line for? Thanks."

ParkeBrown:
Any FA should be able to see all partners regardless so this can be tested with any FA role, assigned or not.

ParkeBrown:
FA assigned to the team experiencing 

On Log Rocket: https://app.logrocket.com/x6d22y/cord-prod/s/4-7864ec3e-f782-49d5-bcdf-62be85eea1de/16?filters=%255B%257B%2522id%2522%253A%2522H1oxj66Lu%2522%252C%2522type%2522%253A%2522name%2522%252C%2522operator%2522%253A%257B%2522name%2522%253A%2522is%2522%252C%2522type%2522%253A%2522IS%2522%252C%2522hasStrings%2522%253Atrue%252C%2522autocompleteEnabled%2522%253Atrue%257D%252C%2522strings%2522%253A%255B%2522Bety%2520Garcia%2522%255D%257D%255D&fromTab=latest&t=1618953956128.4065

At time stamp 1:36:666 she attempts to open Amerek of which she is an assigned FA and LFA and gets the same error as above.

ParkeBrown:
Two other user test:

[...]

PR #1985, opened 2021-06-30
Partner secure list refactor

elisabeth-sorrell:
@CarsonF the partners per project function is not generic yet, but there's going to be another PR for that for the other bugs I'm working this sprint.

Also, I think that this fixes https://github.com/SeedCompany/cord-api-v3/issues/1830, even though I've never been able to reproduce it locally. I think just moving away from security groups for partner list is going to fix it, but I won't know that for sure unless somebody can check prod ....... after it merges, I guess.


src/components/partner/partner.service.ts
```diff
@@ -189,9 +189,48 @@ export class PartnerService {
...
+      const result = await this.repo
+        .listPartnersOfAllUserProjects(session, { filter, ...input })
+        .run();
 
-    return await runListQuery(query, input, (id) => this.readOne(id, session));
+      // I have to put an await, because the next block will execute before this one is finished....
+      // eslint-disable-next-line @typescript-eslint/await-thenable, @typescript-eslint/no-misused-promises
+      await void result.forEach(async (project) => {
+        // eslint-disable-next-line @typescript-eslint/await-thenable, @typescript-eslint/no-misused-promises, @typescript-eslint/no-confusing-void-expression
+        await project.resources.forEach(async (resource) => {
```
CarsonF, line 207: The linter is yelling at you because this doesn't work. You'll get passed line 220 before line 215 ever executes. 

Simplist option is to switch to a for-of-await syntax.
```ts
for await (const project of result) {
  for await (const resource of project.resources) {
    await canList()
  }
}
```

src/components/partner/partner.service.ts
```diff
@@ -189,9 +189,48 @@ export class PartnerService {
...
+    if (await this.authorizationService.canList(Partner, session)) {
+      const query = this.repo.list({ filter, ...input }, session);
+      return await runListQuery(query, input, (id) =>
+        this.readOne(id, session)
+      );
+    } else {
+      const listPartnerIds = [] as ID[];
+      const result = await this.repo
+        .listPartnersOfAllUserProjects(session, { filter, ...input })
+        .run();
```
CarsonF, line 201: I don't think this direction is a good idea. It's an enormous amount of work to put onto the API every time a partner list is requested. Most of the reason to paginate is to limit the load on the API. In summary we are pulling the entire list only to filter it in app code and go back to the DB sort & paginate, and worse the do another lookup per item. That's O(n+2).

I think the logic here is if you are a member of this project with X sensitivity and Y roles then you can pass. I believe the auth service could compile that into a simple structure, which the DB query could then convert into a WHERE clause. This would remove the back and forth from the DB and still leaves us in a spot where we can merge the readOnes into the query as well bringing the number of queries to O(1).

CarsonF (changes requested):

elisabeth-sorrell:
@CarsonF Can you take a look again when you can? The mapping function and query is written. I'm going to follow this same pattern for Project Member, Partnerships and Budget, and upgrade the list query function to be probably be generic and fancy and stuff and put it in another PR

src/components/partner/partner.repository.ts
```diff
@@ -270,4 +270,70 @@ export class PartnerRepository extends DtoRepository(Partner) {
...
+              scopeSensitivityMap
+            )}') as sensMap`,
+            `(${rankSens('sensitivity')}) as sens`,
+          ])
+      )
+      .raw(
+        `MATCH (node) WHERE any(x in scopedRoles where x IN keys(sensMap) and sens <= ${rankSens(
+          'apoc.map.get(sensMap, x)'
+        )})`
+      )
```
CarsonF, line 313:
```suggestion
      .raw(
        `
          MATCH (node) WHERE any(x in scopedRoles where x IN keys($sensMap)
              and sens <= ${rankSens('apoc.map.get($sensMap, x)')}
          )
        `, {
          sensMap: scopeSensitivityMap
        }
      )
```

Does this work instead of 
```tsx
`apoc.convert.fromJsonMap('${JSON.stringify(scopeSensitivityMap)}') as sensMap`
```

src/core/database/query/match-project-based-props.ts
```diff
@@ -5,13 +5,17 @@ import { merge } from './cypher-functions';
 import { matchProps, MatchPropsOptions } from './matching';
 
 export const matchPropsAndProjectSensAndScopedRoles =
```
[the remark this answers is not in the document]
elisabeth-sorrell: ... yeah, maybe a few iterations from now

src/core/database/transformer.ts
```diff
@@ -48,7 +48,7 @@ export class MyTransformer extends PatchedTransformer {
       : timeZoneId;
     return DateTime.fromObject({
       ...rest,
-      millisecond: nanosecond / 1e6,
+      //millisecond: nanosecond / 1e7,
```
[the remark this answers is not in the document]
elisabeth-sorrell: Yeah, that's my bad. Fixed now

src/components/project-change-request/project-change-request.repository.ts
```diff
@@ -60,18 +64,20 @@ export class ProjectChangeRequestRepository extends DtoRepository(
...
+          'scopedRoles as scope',
+        ]
       );
     const result = await query.first();
     if (!result) {
       throw new NotFoundException('Could not find project change request');
     }
 
-    return result.dto;
+    return { ...result.dto, scope: result.scope };
```
CarsonF, line 81: This change doesn't do anything...

src/core/database/query/match-project-based-props.ts
```diff
@@ -98,6 +102,13 @@ export const matchProjectSens = (projectVar: string) => (query: Query) =>
       // Prevent single row with project from expanding to more here via multiple engagement matches
       .raw('LIMIT 1')
       .return(coalesce('langSens.value', '"High"').as('sensitivity'))
+      // If the cardinality of this subquery is zero, no rows will be returned at all.
+      // So, if no projects are matched (optional matching), we still need to have a cardinality > 0 in order to continue
+      // https://neo4j.com/developer/kb/conditional-cypher-execution/#_the_subquery_must_return_a_row_for_the_outer_query_to_continue
+      .union()
+      .with(projectVar)
+      .with(projectVar)
+      .return('"High" as sensitivity')
```
CarsonF: Why do we need sensitivity without project?

src/components/budget/budget-record.repository.ts
```diff
@@ -136,12 +136,13 @@ export class BudgetRecordRepository extends DtoRepository(BudgetRecord) {
...
-          .return<{ dto: UnsecuredDto<BudgetRecord> }>(
+          .return<{ dto: UnsecuredDto<BudgetRecord>; scope: ScopedRole[] }>([
             merge('props', 'changedProps', {
               organization: 'organization.id',
               changeset: 'changeset.id',
-            }).as(outputVar)
-          )
+            }).as(outputVar),
+            'scopedRoles as scope',
+          ])
```
CarsonF: I think you are probably behind develop. BudgetRecord already has a scope array. https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/budget/budget-record.repository.ts#L140-L145

src/components/project-change-request/project-change-request.repository.ts
elisabeth-sorrell: yeah, fixed that already and going to push after testing

src/core/database/query/match-project-based-props.ts
elisabeth-sorrell: I believe you have to import the same stuff into each union-ed query, but I'll double check that.

src/components/budget/budget-record.repository.ts
elisabeth-sorrell: Yeah, I did a bunch of crazy stuff like that then realized we want to cache the scope in the dto. the next push should fix most of that.

src/core/database/query/match-project-based-props.ts
CarsonF: I think we just need to add a where clause in the 3rd union so that it doesn’t return any rows if there is a project. But looks good otherwise!

src/components/project/project.service.ts
```diff
@@ -233,7 +233,7 @@ export class ProjectService {
   }
 
   async secure(
-    project: UnsecuredDto<Project>,
+    project: UnsecuredDto<Project> & { scope: ScopedRole[] },
```
elisabeth-sorrell, line 236: clean up

src/core/database/query/match-project-based-props.ts
```diff
@@ -14,24 +15,41 @@ import {
 import { matchProps, MatchPropsOptions } from './matching';
 
 export const matchPropsAndProjectSensAndScopedRoles =
-  (session?: Session | ID, propsOptions?: MatchPropsOptions) =>
+  (
+    skipProps = false,
+    skipSensitivity = false,
```
elisabeth-sorrell, line 20: don't need skipSensitivity

src/components/partner/partner.repository.ts
```diff
@@ -80,10 +84,17 @@ export class PartnerRepository extends DtoRepository(Partner) {
...
     return (query: Query) =>
       query
-        .apply(matchProps())
+        .optionalMatch([
+          node('project', 'Project'),
+          relation('out', '', 'partnership'),
+          node('', 'Partnership'),
+          relation('out', '', 'partner'),
+          node('node'),
+        ])
```
CarsonF, line 96: I'm not sure the changes from 7ee2fd2 make sense for partners & organizations. There can be multiple projects matched for these `nodes` and therefore the `scope` would just be the from the first project found. I think it's going to take some extra work on these two objects.

I think the correct solution is to merge the scope from all projects right? As being X on any project grants you rights to org/partner as that role.

CarsonF (commented):
@elisabeth-sorrell I rebased on develop and cleaned up/tweaked some things.
Everything looks good except for the one spot I commented on.

Issue #1984, opened 2021-06-29
Cannot edit objects created in changeset


CarsonF:
Objects created inside of changesets cannot be found right now. I believe if this is the case we should just apply property changes normally.

Specifically the problem is they are being identified as _changed_ in the changeset diff query. Then when the original object is looked up it cannot be found. I think this is probably just a symptom though.

PR #1989, opened 2021-07-01
Fix edit objects created in changeset

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1984

Instead of creating a new relationship prop, I would like to add a new `check` before update property in the changeset.
```
if (changeset) {
      const result = await this.db
        .query()
        .match([
          node('changeset', 'Changeset', { id: changeset }),
          relation('out', '', 'changeset', { active: true }),
          node('node', label, { id }),
        ])
        .return('node.id')
        .first();

      if (result) {
        changeset = undefined;
      }
}
```
We could put this logic in the parent method(`updateProperties`) so that it could be called only once.

### DIFF
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -289,6 +289,23 @@ export class DatabaseService {
         // They'll get them when they are applied for real.
         ['Property'];
 
+    // check if the node is created in changeset, update property normally
+    if (changeset) {
+      const result = await this.db
+        .query()
+        .match([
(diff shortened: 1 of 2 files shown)


CarsonF (approved):
Yeah I was thinking that too. But what would be even better is to combine this into the DB query. It would also be great to combine all the property updates to happen at once. This is an important bug fix though so this is fine for now 👍

PR #1990, opened 2021-07-01
Sub query import vars shortcut

CarsonF:
This adds an optional parameter to `subQuery()` to import variables.

```diff
- query.subQuery((sub) => sub
-   .with('node')
+ query.subQuery('node', (sub) => sub
```

Importing vars is a common use case and this is shorter.

Downside is it deviates more from the Neo4j doc examples. Weirdly though this `with` clauses used as the import into the sub query is different from normal `with` clauses. It can't do anything other than plainly reference an existing variable; no `as` aliasing or function calls allowed.

Thoughts?

### DIFF
--- a/src/core/database/database.service.ts
+++ b/src/core/database/database.service.ts
@@ -312,9 +312,8 @@ export class DatabaseService {
       .match(node('node', label, { id }))
       // Deactivate existing prop(s) & adjust labels as needed
       // This is an optional step
-      .subQuery((sub) =>
+      .subQuery('node', (sub) =>
         sub
-          .with('node')
           .match([
             node('node'),
             relation('out', 'oldToProp', key, { active: !changeset }),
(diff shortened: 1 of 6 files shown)


blakekrammes:
You can have multiple variables in the `with`? What about something like this?

```cql
query.subQuery(['node', 'node2', ...]).create([node('node'), relation('out', '', 'start', { active: true } ), node('n', 'Property' { value: "x" } )]).return('n')
```

The issue with the above might be that it obscures the nested nature of a subquery. I was just trying to get around having to use an inner arrow function. I think what you have is an improvement though.

CarsonF:
Without the inner function you'd need to explicitly tell the sub query when you are done with it so it can return the main query context. This way you don't have to do that, also formatting automatically makes sense as the inner function is indented more instead of being the same indentation as the main query.

PR #1993, opened 2021-07-01
Change mutations from boolean returns to have output types

CarsonF:
This allows each mutation to return more information in the future if/when needed.
Doing it like this to handle breaking change all at once.

⚠️ This needs a synchronized deployment with the UI ⚠️


CarsonF:
Merging this now as we have another breaking change with `availableMethodologySteps` that is about to go out

Issue #1983, opened 2021-06-29
Partnership Changeset Bugs


CarsonF:
These methods need to account for changesets
https://github.com/SeedCompany/cord-api-v3/blob/57a5f53a731be1f7ed1eb7512fd88a64b6baa83a/src/components/partnership/partnership.service.ts#L54-L54
https://github.com/SeedCompany/cord-api-v3/blob/57a5f53a731be1f7ed1eb7512fd88a64b6baa83a/src/components/partnership/partnership.service.ts#L56-L56
https://github.com/SeedCompany/cord-api-v3/blob/57a5f53a731be1f7ed1eb7512fd88a64b6baa83a/src/components/partnership/partnership.service.ts#L214-L214

PR #1997, opened 2021-07-05
Partnership changeset bugs

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/1983

### DIFF
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -27,7 +27,7 @@ import {
 import { ProjectUpdatedEvent } from '../../project/events';
 import { BudgetRepository } from '../budget.repository';
 import { BudgetService } from '../budget.service';
-import { Budget, BudgetRecord, BudgetStatus } from '../dto';
+import { Budget, BudgetRecord } from '../dto';
 
 type PartialBudget = UnsecuredDto<Pick<Budget, 'id' | 'status'>> & {
   records: ReadonlyArray<UnsecuredDto<BudgetRecord>>;
@@ -87,10 +87,11 @@ export class SyncBudgetRecordsToFundingPartners
       event.session,
       changeset
     );
-    if (budget.status !== BudgetStatus.Pending) {
-      this.logger.debug('Budget is not pending, skipping sync', budget);
-      return;
-    }
+    // Disable budget status check due to apply changeset
+    // if (budget.status !== BudgetStatus.Pending) {
+    //   this.logger.debug('Budget is not pending, skipping sync', budget);
+    //   return;
+    // }
 
     const partnerships = await this.determinePartnerships(event, changeset);
 
--- a/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
+++ b/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
@@ -87,11 +87,6 @@ export class SyncBudgetRecordsToFundingPartners
       event.session,
       changeset
     );
-    // Disable budget status check due to apply changeset
-    // if (budget.status !== BudgetStatus.Pending) {
-    //   this.logger.debug('Budget is not pending, skipping sync', budget);
-    //   return;
-    // }
 
     const partnerships = await this.determinePartnerships(event, changeset);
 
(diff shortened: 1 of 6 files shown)


src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts
```diff
@@ -87,10 +87,11 @@ export class SyncBudgetRecordsToFundingPartners
...
     );
-    if (budget.status !== BudgetStatus.Pending) {
-      this.logger.debug('Budget is not pending, skipping sync', budget);
-      return;
-    }
+    // Disable budget status check due to apply changeset
+    // if (budget.status !== BudgetStatus.Pending) {
+    //   this.logger.debug('Budget is not pending, skipping sync', budget);
+    //   return;
+    // }
```
CarsonF, line 94:
```suggestion
```

src/components/partnership/partnership.repository.ts
```diff
@@ -259,33 +259,80 @@ export class PartnershipRepository extends DtoRepository(Partnership) {
...
+                    ])
+                    .optionalMatch([
+                      node('partnership'),
+                      relation('out', '', 'partner', { active: true }),
+                      node('partner'),
+                    ])
+                    .return(['partner', 'project', 'partnership'])
+                : q
+            )
+        )
```
CarsonF, line 302: The partner and project matched on the left hand side of the union are not considered on the right hand side of the union. Therefore, the current logic matches _any_ project & partner.

Maybe move partner & project out of the subquery & import them on both sides of the union?

CarsonF (changes requested):

src/components/partnership/partnership.repository.ts
```diff
@@ -265,33 +265,81 @@ export class PartnershipRepository extends DtoRepository(Partnership) {
...
+              relation('out', '', 'partnership', { active: true }),
+              node('partnership'),
+              relation('out', '', 'partner', { active: true }),
+              node('partner'),
+            ])
+            .return(['partnership'])
+            .apply((q) =>
+              changeset
+                ? q
+                    .union()
```
CarsonF, line 292: Is a with clause not needed after the union to import the `project, partner` vars again?

zkhin: Honestly, I haven't tested it yet but I don't think it will be needed in the `apply` clause. If the `changeset` is defined, the subquery will be
```
.with('project, partner')
.optionalMatch([
       node('project'),
       relation('out', '', 'partnership', { active: true }),
       node('partnership'),
       relation('out', '', 'partner', { active: true }),
       node('partner'),
])
.return(['partnership'])
.union()
.match([node('changeset', 'Changeset', { id: changeset })])
.optionalMatch([
        node('project'),
        relation('out', '', 'partnership', { active: false }),
        node('partnership'),
        relation('in', '', 'changeset', { active: true }),
        node('changeset'),
 ])
.optionalMatch([
        node('partnership'),
        relation('out', '', 'partner', { active: true }),
        node('partner'),
])
.return(['partnership'])
```

CarsonF: [This](https://neo4j.com/developer/kb/post-union-processing/#_post_union_processing_enhancements_in_4_1) has it after the union clause within the subquery. Almost like two separate subqueries being ran independently with their results merged.

I think with the way you have it, those vars don't exist in the unioned part and so they match anything.

I wish there was a way with cypher to be explicit about expecting this var to be previously defined vs creating a new reference.

src/components/partnership/partnership.repository.ts
```diff
@@ -265,33 +265,82 @@ export class PartnershipRepository extends DtoRepository(Partnership) {
...
+          .match([
+            node('project', 'Project', { id: projectId }),
+            relation('out', '', 'partnership', { active: true }),
+            node('partnership'),
+          ])
+          .return(['partnership'])
+          .apply((q) =>
+            changeset
+              ? q
+                  .union()
```
CarsonF, line 331: import needed here too

CarsonF (changes requested):

src/components/partnership/partnership.repository.ts
zkhin: hmm... what should we import?

CarsonF: Oh I guess nothing. I wonder if it would be better to match project first before the subquery...

CarsonF (approved):

PR #2006, opened 2021-07-08
Make User.email optional

tmpacifitech:
https://github.com/SeedCompany/cord-api-v3/issues/1610

### DIFF
--- a/src/components/user/user.repository.ts
+++ b/src/components/user/user.repository.ts
@@ -24,7 +24,8 @@ import {
 import {
   collect,
   createNode,
-  deleteProperties,
+  createProperty,
+  deactivateProperty,
   matchProps,
   merge,
   paginate,
@@ -205,32 +206,20 @@ export class UserRepository extends DtoRepository(User) {
 
   async updateEmail(
     user: User,
-    email: any,
-    createdAt: DateTime
+    email: string | null | undefined
   ): Promise<void> {
-    //Remove old emails and relations
     await this.db
       .query()
-      .match([node('node', ['User', 'BaseNode'], { id: user.id })])
-      .apply(deleteProperties(User, 'email'))
+      .matchNode('node', 'User', { id: user.id })
+      .apply(deactivateProperty({ resource: User, key: 'email' }))
+      .apply((q) =>
+        email
+          ? q.apply(
+              createProperty({ resource: User, key: 'email', value: email })
+            )
+          : q
+      )
       .return('*')
-      .run();
-
-    //Update email
-    await this.db
-      .query()
-      .match([node('user', ['User', 'BaseNode'], { id: user.id })])
-      .create([
-        node('user'),
-        relation('out', '', 'email', {
-          active: true,
-          createdAt,
-        }),
-        node('email', 'EmailAddress:Property', {
-          value: email,
-          createdAt,
-        }),
-      ])
       .run();
   }
 
--- a/src/components/user/user.repository.ts
+++ b/src/components/user/user.repository.ts
@@ -81,7 +81,7 @@ export class UserRepository extends DtoRepository(User) {
 
   async create(input: CreatePerson) {
     const initialProps = {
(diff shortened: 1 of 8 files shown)


CarsonF (changes requested):
- [x] No changes are needed for the validation decorator. Empty value is already allowed there.
In fact what you added _enforces_ that email is empty. So the current diff can just be reverted.

- [ ] We need to remove the constraint for EmailAddresses having a value.
https://github.com/SeedCompany/cord-api-v3/blob/817c13edf15cf1938867ff71e3c3f2e6cc4b0e77/src/components/user/user.repository.ts#L68
That might require us enforcing that email addresses are unique ourselves instead of letting neo4j do it and catching that uniqueness error.

- [x] We also need to now check that the user has an email address before sending them notifications. Here is one of a few places that needs a change.
https://github.com/SeedCompany/cord-api-v3/blob/817c13edf15cf1938867ff71e3c3f2e6cc4b0e77/src/components/project/project.rules.ts#L985

CarsonF (changes requested):
Hey @tmpacifitech I'm throwing this back again. I'm not sure what I was thinking but the email constraint creates an index that is important. I don't think we can get rid of it.

I think part of my previous thought process was informed by historical constraints/perspectives that no longer apply. Historically we made a decision that every property would be created even if it was null. This made it easier to do lookups for sorting and updates. With Neo4j v4's introduction of subqueries, our queries have become a lot smarter. Many of these constraints have been removed out of necessity for other features.
For example, `updateProperty` will deactivate an active prop if found, but continues on if not.

This email case is also a bit unique too. It is often a starting point for matching instead of "just a prop on X object". It's often looked up directly, independent of a User node.
And there's no real need to sort on it, though the new `sorting()` function could manage it.

So all in all sorry for the bait and switch. That's totally on me.

src/components/user/user.repository.ts
```diff
@@ -74,37 +75,54 @@ export class UserRepository extends DtoRepository(User) {
...
+                'email',
+                input.email,
+                'node',
+                'email',
+                'EmailAddress'
+              ),
+            ])
+          : q
+      )
+      .apply((q) =>
```
tmpacifitech, line 120: I don't think `createNode` can handle current `input.roles` behavior.

src/components/user/user.repository.ts
```diff
@@ -74,37 +75,54 @@ export class UserRepository extends DtoRepository(User) {
...
-        ...property('canDelete', true, 'user'),
-      ])
-      .return('user.id as id')
+      .apply(
+        await createNode(User, {
+          initialProps,
+          baseNodeProps,
+        })
+      )
+      .apply((q) =>
```
tmpacifitech, line 107: Cannot add `email` to `initialProps` because `getDbPropertyLabels` does not get `EmailAddress` label with empty db which means `devops@tsco.org` is not added to `EmailAddress` label when the server starts. I think we need to refactor `createNode` function?

CarsonF: Just need to add
```ts
@DbLabel('EmailAddress')
readonly email: string
```
[here](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/user/dto/user.dto.ts#L39)

src/components/user/user.repository.ts
```diff
@@ -74,37 +75,54 @@ export class UserRepository extends DtoRepository(User) {
...
+      displayFirstName: input.displayFirstName,
+      displayLastName: input.displayLastName,
+      phone: input.phone,
+      timezone: input.timezone,
+      about: input.about,
+      status: input.status,
+      title: input.title,
+      canDelete: true,
+    };
+    const baseNodeProps = { id };
```
CarsonF, line 97: ID is generated inside of createNode and the DB query already returns it. So use that instead of generating it here.

CarsonF (approved):
🎉

PR #2017, opened 2021-07-11
Set previous relation on changeset approval

zkhin:


CarsonF:
Hey @zkhin I think I miss communicated the need here. I'm sorry about that.

# Problem

The existing GQL API was sufficient, it just has a problem. Here's a scenario:
1. Changeset A is created.
2. Project's name is changed from "Foo" to "Bar" in changeset A
3. Changeset A is approved. 
4. View diff of changeset A. 
5. It appears that project's A name changed from "Bar" to "Bar".

See how the project's previous state was lost in the `previous` object?

# Possible solution

I don't think a simple query will solve this. I think we need to add more state during the approval process.
This also means that the diff query will need to have two sets of logic for the determining the previous objects. 
Is this changeset pending?
Yes -> "previous" is the currently active object.
No -> "previous" is a saved view of the object.

Maybe during the approval process we do a special update query. 
Maybe instead of creating new property nodes that match the values of properties attached to the changeset, we just change the relationships.
The property is already attached to the project just inactive. So we could make it active and the currently active property could get attached to the changeset under a "previous" label.

```js
// Normal
(:Project)-[:name { active: true }]->(:Property { value: "Foo" })
// Created in pending changeset
(:Project)-[:name { active: false }]->(:Property { value: "Bar" })<-[:changeset { active: true }]-(:Changeset)
// Applied from approved changeset
(:Project)-[:name { active: true }]->(:Property { value: "Bar" })<-[:changeset { active: true }]-(:Changeset)
// Saved as previous from active state before changeset was approved
(:Project)-[:name { active: false }]->(:Deleted_Property { value: "Foo" })<-[:previous]-(:Changeset)
// Normal deleted property for reference
(:Project)-[:name { active: false }]->(:Deleted_Property { value: "Foo" })
```

Right now the approve handlers:
1. query for the changed props
2. call service.update to apply them. 

I think we need bypass the service.update call and do our own thing instead. As of last week [updateProperty](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/core/database/query/update-property.ts) got moved to a query fragment. It should make it easier to compose a single query to query to do this.
Something like:
1. match changed props
2. for each prop
3. match respective currently active property
3. [deactivateProperty](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/core/database/query/deactivate-property.ts)(activeProp)
4. Apply previous label to prop
4. Set changed prop to active

# Reading back inactive state

The question then becomes how to read the project back for this previous state. This would also be the first time we are querying for a `Deleted_Property`, not a bad thing just worth noting.

It's starting to seem like we have a handful of different views of an object. 
1. active view
2. view from a changeset
3. view before a changeset was applied (previous)
4. Looking to history, even something like as of X date. Which would filter deleted properties by their deleted at date.

```ts
// Previous signature
readOne(id: ID, session: Session, changeset?: ID)
// New signature
type ObjectView =
  | { active: true } // default
  | { changeset: ID }
  | { beforeChangeset: ID }
  | { asOf: DateTime }
readOne(id: ID, session: Session, view?: ObjectView)
```

---

How does this sound? We don't have to change reads for launch as long as we capture the data.

CarsonF (changes requested):
Marking this as needing work.

src/components/project/handlers/apply-approved-changeset-to-project.handler.ts
```diff
@@ -35,17 +44,61 @@ export class ApplyApprovedChangesetToProject
...
-        undefined,
-        false
+
+      await Promise.all(
+        keys(actualChanges).map(async (key) => {
+          const query = this.db
+            .query()
+            .match(node('node', 'Project', { id }))
+            // Apply previous label to active prop
+            .apply(addPreviousLabel(key, changesetId))
```
zkhin, line 54: Added `previous` label before deactivating active prop because it will be tricky to get deleted prop after applying to deactivate query.

src/components/project/handlers/apply-approved-changeset-to-project.handler.ts
```diff
@@ -35,17 +44,61 @@ export class ApplyApprovedChangesetToProject
...
+      // const query = this.db
+      //   .query()
+      //   .match([
+      //     node('node', 'Project'),
+      //     relation('out', '', 'changeset', { active: true }),
+      //     node('changeset', 'Changeset', { id: changesetId }),
+      //   ])
+      //   .apply(matchProps({ changeset: changesetId, optional: true }))
+      //   .with('props')
+      //   .forEach('prop', 'props', (prop) =>
```
zkhin, line 86: @CarsonF Tried to use your `forEach` query but not sure how to get `key`.

src/components/project/handlers/apply-approved-changeset-to-project.handler.ts
```diff
@@ -35,17 +44,61 @@ export class ApplyApprovedChangesetToProject
...
+      //     prop
+      //       // Deactivate active property
+      //       .apply(
+      //         deactivateProperty({
+      //           key: variable('type(prop)').name as DbChanges<
```
zkhin, line 91: @CarsonF How can we get `key(type of prop)` using `variable` here?

src/components/project/handlers/apply-approved-changeset-to-project.handler.ts
CarsonF: `matchProps` composes the a prop list to an object. Which for most cases is what we want.
Here however we want a list. I would just do [the match](https://github.com/SeedCompany/cord-api-v3/blob/02087ef8f091f5253f2404da3a7aee5570564e07/src/core/database/query/matching.ts#L81-L96) yourself here.

That would make each row be an individual property, which means you wouldn't need the foreach.

src/components/project/handlers/apply-approved-changeset-to-project.handler.ts
CarsonF: Using what I said above you could do 
```ts
key: variable('type(relationToProp)')
```
As long as `relationToProp` is what you used in the match above.

zkhin:
@CarsonF Fixed a minor query issue.

CarsonF:
Thanks @zkhin, didn't want to spend anymore time on it until tomorrow :)

I'm realizing we need to do half of this on changeset denied as well. Set the previous label but leave the currently active prop alone.

Let's do that in another PR

zkhin:
Sure

CarsonF (approved):

PR #2028, opened 2021-07-15
Allow administrator to assign roles

tmpacifitech:
https://github.com/SeedCompany/cord-field/issues/722

### DIFF
--- a/src/components/project/project-member/project-member.service.ts
+++ b/src/components/project/project-member/project-member.service.ts
@@ -98,7 +98,7 @@ export class ProjectMemberService {
 
     await this.verifyRelationshipEligibility(projectId, userId);
 
-    await this.assertValidRoles(input.roles, () =>
+    await this.assertValidRoles(input.roles, session, () =>
       this.userService.readOne(userId, session)
     );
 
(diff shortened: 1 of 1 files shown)


src/components/project/project-member/project-member.service.ts
```diff
@@ -207,7 +208,10 @@ export class ProjectMemberService {
     const user = await forUser();
     const availableRoles = user.roles.value ?? [];
     const forbiddenRoles = difference(roles, availableRoles);
-    if (forbiddenRoles.length) {
+    if (
+      forbiddenRoles.length &&
+      !session.roles.includes('global:Administrator')
```
CarsonF, line 213: Move this check to the if statement above so we don't call User.readOne when we don't need to.

CarsonF (changes requested):

src/components/project/project-member/project-member.service.ts
CarsonF: Nevermind. This isn't needed. See comment on UI PR.

There may need to be some changes to the role files to limit who can change User.roles.

tmpacifitech:
@CarsonF Seems like `DbUser.roles` permission is configured properly in roles files and we can close this PR.

PR #2031, opened 2021-07-16
Replaced usages of calculateTotalAndPaginateList(deprecated) with sorting and paginate

### LINKED ISSUE #2030: calculateTotalAndPaginateList -> sorting, paginate
Replace usages of ~[`calculateTotalAndPaginateList`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/lists.ts#L75-L75)~ with [`sorting`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/lists.ts#L105-L105) and [`paginate`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/lists.ts#L26-L26).

Both the new and the old are very similar in what they output. They've been split for separation of concerns and because both are complicated enough on their own. The biggest difference is the old method needed to be ran via a `runListQuery()` function so it could provide defaults on no results. The new paginate is better and doesn't need that.

Language list is a good example.
https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/language/language.repository.ts#L164-L173

So for each usage we need to:

- calculateTotalAndPaginateList -> sorting, paginate
- make the function async
- await the first result and return it (non-null assert it with comment - as language list does)
- Replace `runListQuery` with `mapListResults`. Just like this:
https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/language/language.service.ts#L245-L246

daniel-leod:
Fixes #2030

### DIFF
--- a/src/components/organization/organization.service.ts
+++ b/src/components/organization/organization.service.ts
@@ -15,8 +15,8 @@ import {
   OnIndex,
 } from '../../core';
 import {
+  mapListResults,
   parseBaseNodeProperties,
-  runListQuery,
 } from '../../core/database/results';
 import { AuthorizationService } from '../authorization/authorization.service';
 import { Powers } from '../authorization/dto/powers';
@@ -173,9 +173,8 @@ export class OrganizationService {
     { filter, ...input }: OrganizationListInput,
     session: Session
   ): Promise<OrganizationListOutput> {
-    const query = this.repo.list({ filter, ...input }, session);
-
-    return await runListQuery(query, input, (id) => this.readOne(id, session));
+    const results = await this.repo.list({ filter, ...input }, session);
+    return await mapListResults(results, (id) => this.readOne(id, session));
   }
 
   async addLocation(
--- a/src/components/organization/organization.service.ts
+++ b/src/components/organization/organization.service.ts
@@ -170,10 +170,11 @@ export class OrganizationService {
   }
 
   async list(
-    { filter, ...input }: OrganizationListInput,
+    input: OrganizationListInput,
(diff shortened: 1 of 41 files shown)


src/components/organization/organization.service.ts
```diff
@@ -173,9 +173,8 @@ export class OrganizationService {
     { filter, ...input }: OrganizationListInput,
     session: Session
   ): Promise<OrganizationListOutput> {
-    const query = this.repo.list({ filter, ...input }, session);
-
-    return await runListQuery(query, input, (id) => this.readOne(id, session));
+    const results = await this.repo.list({ filter, ...input }, session);
```
CarsonF, line 176: There's no need to destructure filter above, since we are just merging it back in here. I know this wasn't you, but would you mind cleaning this up?
```suggestion
    const results = await this.repo.list(input, session);
```

CarsonF (commented):
Looks great @daniel-leod, thanks!

One more small ask below

src/components/organization/organization.service.ts
daniel-leod: Sure, I'm on it.

CarsonF (approved):
Thanks @daniel-leod!

Issue #2024, opened 2021-07-14
Expose Deleted Items in Approved Changeset Diffs


CarsonF:
Currently deleted items are no longer shown in the `changeset.difference.removed` list.

PR #2033, opened 2021-07-19
Expose Deleted Items in Approved Changeset Diffs

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/2024


zkhin:
@CarsonF @michaelmarshall When the `deleting` items are approved, they will be deleted like origin, and the `Deleted_` prefix will be added to all labels. That means deleted items will not exist in `BaseNode` anymore but in `Deleted_BaseNode`.
So I refactored getting `removed` items query in the `difference` method and [resource-resolver](https://github.com/SeedCompany/cord-api-v3/pull/2033/files#diff-db4a7220b437d01e3a7e22e9781d3cbc0aa9bc20c023c27de2f26efda257a11bR121).
But still, need to figure out reading deleted items back with `changeset` in all `readOne` methods.

zkhin:
This [assert](https://github.com/SeedCompany/cord-api-v3/pull/2033/files#diff-8e42da0acb3d27f445ade436c879835b54244c8a726aaa03cd00ab30fa9b1d8cR392) will be failed due to `Cannot read engagement` with deleted item id.

CarsonF:
What do you think about a [new readOne signature I proposed here](https://github.com/SeedCompany/cord-api-v3/pull/2017#issuecomment-878284539)?

zkhin:
> What do you think about a [new readOne signature I proposed here](https://github.com/SeedCompany/cord-api-v3/pull/2017#issuecomment-878284539)?

Yes, I believe we can figure it out in this new refactor. Maybe in the third (view before a changeset was applied (previous))?

CarsonF:
Maybe we add a `deleted: true` option to `ObjectView`. That seems like the most straight forward place to start, and would allow deleted items to be read.

`ObjectView` could be passed into `ResourceResolver` which in turn could pass it to the readOne's. 

`beforeChangeset` seems like a hard query as does `asOf` date. But maybe not. Maybe it's just a prop lookup without the `active: true` filter and with a prop.createdAt filter. `beforeChangeset` could do the samething we just need to lookup the changeset finalized time. Maybe finalized time is a new prop we need to set on the Changeset?

src/components/changeset/changeset.repository.ts
```diff
@@ -27,10 +27,25 @@ export class ChangesetRepository extends DtoRepository(Changeset) {
...
+          .optionalMatch([
+            node('changeset'),
+            relation('out', '', [], { active: true, deleting: true }),
+            node('deletingNode', 'BaseNode'),
+          ])
+          .optionalMatch([
+            node('changeset', 'Changeset', { id }),
+            relation('out', '', [], { deleting: true }),
+            node('deletedNode', 'Deleted_BaseNode'),
+          ])
```
CarsonF, line 42: Maybe the BaseNode label constraint is just dropped here and we merge into a single match? Is the `active: true` constraint significant?

zkhin: Do you mean...
```
.optionalMatch([
            node('changeset'),
            relation('out', '', [], { deleting: true }),
            node('node', 'BaseNode'), // How can we merge `BaseNode` and `Deleted_BaseNode` here?
])
```

CarsonF:
```cql
(changeset)-[{ deleting: true }]->(node)
```

zkhin: I don't think `active: true` is necessary here. Let me test with this one.
```
.optionalMatch([
            node('changeset'),
            relation('out', '', [], { deleting: true }),
            node('node'),
])
```

CarsonF: Cool 👍 . I don't think optional is needed either.

zkhin: Correct

zkhin: It works as you expected. 👍

CarsonF:
Ok that work is in progress in #2049. This PR will depend on it and needs to be rebased once merged.

PR #2035, opened 2021-07-19
[RFC] Add User.isSystem flag

CarsonF:
We have many objects in production that normally have users attached to them but have been migrated in where we didn't know the user.
I didn't and still don't want to make that relationship optional, esp for an edge case.

However, it would still be nice to be able to opt-in to hiding "root root" in the UI. Or somehow convey that this user is "unknown" or the action was taken by the system, etc.

So I added an `isSystem` boolean to the `User` object. Currently it's just a check to see if the user is the root admin, but it could be implemented differently in the future.

Thoughts? Is this a good idea or bad? Is there a better way to schema this?

### DIFF
--- a/src/components/user/user.resolver.ts
+++ b/src/components/user/user.resolver.ts
@@ -7,6 +7,7 @@ import {
   ResolveField,
   Resolver,
 } from '@nestjs/graphql';
+import { stripIndent } from 'common-tags';
 import {
   AnonSession,
   firstLettersOfWords,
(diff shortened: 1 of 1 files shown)


blakekrammes:
Yes. I like system over root root. I can't think of a better schema off the top of my head.

CarsonF:
Meh I haven't merged because I'm not thrilled about it. In the mean time the root user was changed to "Root Admin" instead of "root root" so at least it looks a little better. Closing for now.

PR #2036, opened 2021-07-19
User cannot edit their own roles unless GrantRole power

CarsonF:
Correctly convey that a user can only edit their roles if they have the GrantRole power.

Related to https://github.com/SeedCompany/cord-field/issues/722

### DIFF
--- a/src/components/authorization/authorization.service.ts
+++ b/src/components/authorization/authorization.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { Connection } from 'cypher-query-builder';
 import {
   compact,
+  difference,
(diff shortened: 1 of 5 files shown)


CarsonF:
@tmpacifitech It looks like actually confirming the user has access to grant roles is breaking some tests. Would you mind taking a look and fixing?

It might just be replacing `registerUser` calls with `registerUserWithPower`.

tmpacifitech:
> @tmpacifitech It looks like actually confirming the user has access to grant roles is breaking some tests. Would you mind taking a look and fixing?
> 
> It might just be replacing `registerUser` calls with `registerUserWithPower`.

Sure

Issue #2012, opened 2021-07-08
matchPropList -> matchProps


CarsonF:
Replace usage of ~[`matchPropList`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/matching.ts#L25-L25)~ with new [`matchProps`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/matching.ts#L70-L70).

There are examples of the new function all over the codebase. [Posts](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/post/post.repository.ts#L69-L78) is a good example. It calls it, returns the `props` in the query and then the method returns the props.

Really the bigger goal here is to have the `readOne()` return an `UnsecuredDto<X>`. Sometimes that means the cypher query needs to shape the object. Like with [`User`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/user/user.repository.ts#L194-L205), we merge the list role nodes into a list of strings.

The end, end goal is to pull this out into a `hydrate()` method given a `node` it returns a shaped `dto`. This will let us use it in multiple places, so X dto can be returned from different starting queries based on different use cases.

- [x] [Education](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/user/education/education.repository.ts#L55)
- [x] [FieldRegion](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/field-region/field-region.repository.ts#L84)
- [x] [FieldZone](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/field-zone/field-zone.repository.ts#L76)
- [x] [File](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/file/file.repository.ts#L197)
- [x] [Film](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/film/film.repository.ts#L55)
[...]

PR #2038, opened 2021-07-20
Refactor readOnes to output dto via hydrate method with new matchProps

daniel-leod:
Fixes [#2012 ](https://github.com/SeedCompany/cord-api-v3/issues/2012)


src/components/field-region/field-region.repository.ts
```diff
@@ -77,33 +83,40 @@ export class FieldRegionRepository extends DtoRepository(FieldRegion) {
     return await query.first();
   }
 
-  async readOne(id: ID, session: Session) {
+  async readOne(id: ID, session: Session): Promise<UnsecuredDto<FieldRegion>> {
```
CarsonF, line 86: If the method's return type wasn't previously declared let's leave it as inferred. It usually doesn't add much since we are essentially declaring the DB return type in the method (which could not reflect reality). So it just adds more typing and another place to maintain.

src/components/field-zone/field-zone.repository.ts
```diff
@@ -69,25 +75,34 @@ export class FieldZoneRepository extends DtoRepository(FieldZone) {
...
-          directorId: ID;
-        }
-      >();
+      .apply(this.hydrate());
 
-    return await query.first();
+    const result = await query.first();
+    if (!result) {
+      throw new NotFoundException('Could not find field zone');
+    }
```
CarsonF, line 88: I like converting the null to an exception here. For any place that you've added it here, can you remove it from the service? https://github.com/SeedCompany/cord-api-v3/blob/e1635597e3d5086ebd397b7a231625a6c8c925c2/src/components/field-zone/field-zone.service.ts#L88-L90

CarsonF (commented):
Looking good @daniel-leod

CarsonF:
Since the return shape of the DB query is being changed we need to change the services as well. The services were previously doing what you've moved to the DB query. [Here's](https://github.com/SeedCompany/cord-api-v3/blob/e1635597e3d5086ebd397b7a231625a6c8c925c2/src/components/periodic-report/periodic-report.service.ts#L108-L128) an example of the new way.

src/components/user/unavailability/unavailability.service.ts
```diff
@@ -93,18 +93,22 @@ export class UnavailabilityService {
...
     const result = await this.repo.readOne(id, session);
-    if (!result) {
-      throw new NotFoundException('Could not find user', 'user.id');
-    }
+    return await this.secure(result, id, session);
+  }
 
+  private async secure(
+    dto: UnsecuredDto<Unavailability>,
+    id: ID,
```
CarsonF, line 101: Separate `id` param is not needed because `dto.id`

CarsonF (changes requested):

src/components/user/unavailability/unavailability.service.ts
daniel-leod: Fixed

src/components/field-region/field-region.service.ts
```diff
@@ -91,33 +89,32 @@ export class FieldRegionService {
...
-        value: result.directorId,
+        ...securedProps.director,
+        value: dto.director,
       },
       fieldZone: {
-        ...secured.fieldZone,
-        value: result.fieldZoneId,
+        ...securedProps.fieldZone,
+        value: dto.fieldZone,
       },
```
CarsonF, line 115: These type of explicit call outs are no longer required. Previously there was a mismatch in prop names: the DB returned `fieldZoneId` but our DTO/security code was looking for `fieldZone` (without the `Id` suffix). Now the DB returns with matching keys so these don't need to be "fixed" anymore.

So all of these type of explicit props can be removed. Should just need `dto`, `secured` and `canDelete`

src/components/partner/partner.service.ts
```diff
@@ -92,35 +89,37 @@ export class PartnerService {
...
-        ...secured.pointOfContact,
-        value: result.pointOfContactId,
+        ...securedProps.pointOfContact,
+        value: dto.pointOfContact,
       },
       types: {
-        ...secured.types,
-        value: secured.types.value || [],
+        ...securedProps.types,
+        value: securedProps.types.value || [],
```
CarsonF, line 116: These types, I think, are still required since they default the value to an empty array. This is a rough edge from our security layer that has yet to be handled.

So they need to be left as is.

CarsonF (changes requested):

CarsonF (approved):
Very excited about this one! 🎉

Issue #2013, opened 2021-07-08
createBaseNode -> createNode


CarsonF:
Replace ~[`createBaseNode`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query.helpers.ts#L24-L24)~ usage with new [`createNode`](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/core/database/query/create-node.ts#L42-L42).

Here's an example to follow where I updated `BudgetRecord` to use it. https://github.com/SeedCompany/cord-api-v3/commit/00af5fb94943ad4d3b7311c83db50670f1483263

You can see in that commit that I'm also using the `createRelationships` function as well. Bonus points for that, but it's not as straight forward since it requires a bit more knowledge about cypher & our schema.

Note also that the ID is now created inside of the createNode call so to get access to it you'll need to return it from the query.
This is what most of our create queries do anyways.

Also if any properties have special labels, like [this](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/organization/organization.repository.ts#L54-L54), they need to be moved to a Label decorator, like [this](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/budget/dto/budget.dto.ts#L29-L29).

I'm counting 21 places it's used:
- [x] [Budget](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/budget/budget.repository.ts#L58)
- [x] [Ceremony](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/ceremony/ceremony.repository.ts#L26)
- [x] [Education](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/user/education/education.repository.ts#L39)
- [x] [EthnologueLanguage](https://github.com/SeedCompany/cord-api-v3/blob/7e94dddb241c7f1ff5f96623370284248ae9cfee/src/components/language/ethnologue-language/ethnologue-language.repository.ts#L23)
[...]

PR #2042, opened 2021-07-22
Replace createBaseNode usage with new createNode

daniel-leod:
Fixes [#2013](https://github.com/SeedCompany/cord-api-v3/issues/2013)


CarsonF:
Duplicate of #2044

PR #2047, opened 2021-07-27
Create or update final reports on project/engagement sync

tmpacifitech:


tmpacifitech:
@CarsonF The final report is created with project/engagement's end date ~ end date. On frontend, this report  is also visible.. let me know if we are going to hide the final report on frontend reports list view.
image

CarsonF:
> @CarsonF The final report is created with project/engagement's end date ~ end date. On frontend, this report is also visible.. let me know if we are going to hide the final report on frontend reports list view.

We do need to show it. Maybe we could check if it's the only a day long and show "Final Report" in that case.

Does Q1 FY2023 start after Nov 2022? I was hoping we could get it last just based on the date range.

CarsonF (approved):
Looks good!

Issue #2025, opened 2021-07-14
Changeset reading back inactive state


zkhin:
The question then becomes how to read the project back for this previous state. This would also be the first time we are querying for a `Deleted_Property`, not a bad thing just worth noting.

It's starting to seem like we have a handful of different views of an object. 
1. active view
2. view from a changeset
3. view before a changeset was applied (previous)
4. Looking to history, even something like as of X date. Which would filter deleted properties by their deleted at date.

```ts
// Previous signature
readOne(id: ID, session: Session, changeset?: ID)
// New signature
type ObjectView =
  | { active: true } // default
  | { deleted: true }
  | { changeset: ID }
  | { beforeChangeset: ID }
  | { asOf: DateTime }
readOne(id: ID, session: Session, view?: ObjectView)
```

PR #2049, opened 2021-07-28
ObjectView for reading specific states of an object

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/2025

### DIFF
--- a/src/common/object-view.ts
+++ b/src/common/object-view.ts
@@ -33,3 +33,6 @@ export const typenameForView = (
     ''
   )
 `;
+
+export const viewOfChangeset = (changeset?: ID): ObjectView =>
+  changeset ? { changeset } : { active: true };
--- a/src/common/object-view.ts
+++ b/src/common/object-view.ts
@@ -1,0 +1,16 @@
+import { DateTime } from 'luxon';
+import { ID } from './id-field';
+
+interface ObjectViewTypeMap {
+  active: true;
+  deleted: true;
+  changeset: ID | undefined;
+  beforeChangeset: ID | undefined;
+  asOf: DateTime;
+}
+
+type ObjToTypeValueUnion<T extends ObjectViewTypeMap> = {
+  [K in keyof T]-?: { type: K; value: T[K] };
+}[keyof T];
+
+export type ObjectView = ObjToTypeValueUnion<ObjectViewTypeMap>;
--- a/src/common/object-view.ts
+++ b/src/common/object-view.ts
@@ -1,4 +1,6 @@
+import { stripIndent } from 'common-tags';
 import { DateTime } from 'luxon';
+import { RequireExactlyOne } from 'type-fest';
 import { ID } from './id-field';
 
 interface ObjectViewTypeMap {
@@ -9,8 +11,25 @@ interface ObjectViewTypeMap {
   asOf: DateTime;
 }
 
-type ObjToTypeValueUnion<T extends ObjectViewTypeMap> = {
-  [K in keyof T]-?: { type: K; value: T[K] };
-}[keyof T];
+export type ObjectView = RequireExactlyOne<ObjectViewTypeMap>;
 
-export type ObjectView = ObjToTypeValueUnion<ObjectViewTypeMap>;
+export const labelForView = (label: string, view?: ObjectView) =>
+  view?.deleted ? `Deleted_${label}` : label;
+
+const generateLabels = (labels: string[], view: ObjectView) =>
+  [...labels, ...(view.deleted ? labels.map((l) => 'Deleted_' + l) : [])]
+    .map((label) => `'${label}'`)
+    .join(',');
+
+export const typenameForView = (
+  labels: string[],
+  view: ObjectView = { active: true },
+  nodeVar = 'node'
+) => stripIndent`
+  replace(
+    [l in labels(${nodeVar})
+      where l in [${generateLabels(labels, view)}]][0],
+    'Deleted_',
+    ''
+  )
+`;
--- a/src/common/object-view.ts
+++ b/src/common/object-view.ts
@@ -6,8 +6,8 @@ import { ID } from './id-field';
 interface ObjectViewTypeMap {
   active: true;
   deleted: true;
-  changeset: ID | undefined;
-  beforeChangeset: ID | undefined;
+  changeset: ID;
+  beforeChangeset: ID;
   asOf: DateTime;
 }
 
--- a/src/components/engagement/engagement.repository.ts
+++ b/src/components/engagement/engagement.repository.ts
@@ -7,6 +7,7 @@ import {
   ID,
   mapFromList,
   NotFoundException,
+  ObjectView,
   ResourceShape,
   ServerException,
   Session,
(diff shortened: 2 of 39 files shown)


src/components/engagement/engagement.resolver.ts
```diff
@@ -44,7 +44,10 @@ export class EngagementResolver {
     @Args() { id, changeset }: ChangesetIds,
     @AnonSession() session: Session
   ): Promise<Engagement> {
-    const engagement = await this.service.readOne(id, session, changeset);
+    const engagement = await this.service.readOne(id, session, {
+      type: 'changeset',
+      value: changeset,
+    });
```
CarsonF, line 50: This needs to use active if changeset is not given. Maybe there's someway we could convert `ChangesetIds` to an `ObjectView`

src/components/engagement/engagement.repository.ts
```diff
@@ -76,15 +83,19 @@ export class EngagementRepository extends CommonRepository {
...
                   ])
                   .return('project, node')
               : q
           )
       )
-      .apply(matchPropsAndProjectSensAndScopedRoles(session))
+      .apply(
+        matchPropsAndProjectSensAndScopedRoles(session, {
+          deleted: view?.type === 'deleted',
+        })
```
CarsonF, line 97: Maybe it would be better to give this the view object.

src/components/engagement/engagement.repository.ts
```diff
@@ -123,7 +134,13 @@ export class EngagementRepository extends CommonRepository {
...
       .return<{ dto: UnsecuredDto<LanguageEngagement & InternshipEngagement> }>(
         merge('props', 'changedProps', {
-          __typename: `[l in labels(node) where l in ['LanguageEngagement', 'InternshipEngagement']][0]`,
+          __typename: `
+            replace(
+              [l in labels(node)
+                where l in ['LanguageEngagement', 'InternshipEngagement', 'Deleted_LanguageEngagement', 'Deleted_InternshipEngagement']][0],
+              'Deleted_',
+              ''
+            )`,
```
CarsonF, line 143:
```suggestion
          __typename: stripIndent`
            replace(
              [l in labels(node)
                where l in ['LanguageEngagement', 'InternshipEngagement', 'Deleted_LanguageEngagement', 'Deleted_InternshipEngagement']][0],
              'Deleted_',
              ''
            )
          `,
```

If we keep doing this too we should move it to a function.

src/components/engagement/engagement.service.ts
```diff
@@ -236,14 +235,15 @@ export class EngagementService {
   async readOne(
     id: ID,
     session: Session,
-    changeset?: ID
+    view?: ObjectView
+    // changeset?: ID
```
CarsonF, line 239:
```suggestion
```

src/core/database/query/matching.ts
```diff
@@ -54,7 +57,7 @@ export const matchProps = (options: MatchPropsOptions = {}) => {
           [
             node(nodeName),
             relation('out', 'r', { active: !changeset }),
-            node('prop', 'Property'),
+            node('prop', deleted ? 'Deleted_Property' : 'Property'),
```
CarsonF, line 60: I don't think this is necessary. When we delete nodes we leave the properties alone (active). It's only when properties change that we do this.

CarsonF (changes requested):

src/core/database/query/matching.ts
zkhin: If the node is deleted, its properties are moved to `Deleted_Property` and cannot find with the `Property` label. Can you check on your side?

src/components/engagement/engagement.repository.ts
```diff
@@ -71,32 +71,33 @@ export class EngagementRepository extends CommonRepository {
...
               : q
           )
       )
       .apply(
-        matchPropsAndProjectSensAndScopedRoles(session, {
-          deleted: view?.type === 'deleted',
-        })
+        matchPropsAndProjectSensAndScopedRoles(
+          session,
+          view?.deleted ? { view } : {}
```
zkhin, line 97: We should pass the `view` only for `deleted`... I am not sure what is the best option here.

src/components/engagement/engagement.resolver.ts
```diff
@@ -44,7 +44,7 @@ export class EngagementResolver {
     @Args() { id, changeset }: ChangesetIds,
     @AnonSession() session: Session
   ): Promise<Engagement> {
-    const engagement = await this.service.readOne(id, session, changeset);
+    const engagement = await this.service.readOne(id, session, { changeset });
```
zkhin, line 47: How can we convert `ChangesetIds` to `ObjectView`?

src/core/database/query/matching.ts
CarsonF: Ah you are right. Labels change. The relation stays active true though, that's what was thinking of.

src/components/engagement/engagement.repository.ts
CarsonF: I was suggesting always pass it. It could still be nullable though.

src/components/engagement/engagement.resolver.ts
CarsonF: Hmm maybe a pipe in the decorator or a getter on the class. I could play with it quickly, shortly to see what feels best.

zkhin: I am thinking this should work with the `Transform` decorator... but it does not... 🧐
```
@ArgsType()
export class ChangesetIds {
  @IdField()
  id: ID;

  @IdField({ nullable: true, name: 'changeset' })
  @Transform(({ value }) => {
    const view: ObjectView = value ? { changeset: value } : { active: true };
    return view;
  })
  view: ObjectView;
}
```
Can you take a look at what I am missing?

CarsonF: Mmm `@Transform` may not apply for `@ArgsType` since individual arguments are determined by graphql instead of an `@InputType` object sent straight through.

zkhin: Yeah... that makes sense... Do we have other options instead of `@Transform`?

CarsonF: Use a Pipe transform. Example [here](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/common/id.arg.ts#L13-L25), though it does something different.

zkhin: Hmm... can you write Pipe transform shortly? Not sure where I have to start... 😅

CarsonF: Sure pushed a commit.

src/common/object-view.ts
```diff
@@ -0,0 +1,35 @@
+import { stripIndent } from 'common-tags';
+import { DateTime } from 'luxon';
+import { RequireExactlyOne } from 'type-fest';
+import { ID } from './id-field';
+
+interface ObjectViewTypeMap {
+  active: true;
+  deleted: true;
+  changeset: ID | undefined;
+  beforeChangeset: ID | undefined;
```
CarsonF, line 10: I think these should be required
```suggestion
  changeset: ID;
  beforeChangeset: ID;
```

src/core/database/query/matching.ts
```diff
@@ -44,21 +44,21 @@ export const matchProps = (options: MatchPropsOptions = {}) => {
...
     query.comment`matchProps(${nodeName})`.subQuery(nodeName, (sub) =>
       sub
         .match(
           [
             node(nodeName),
-            relation('out', 'r', { active: !changeset }),
-            node('prop', 'Property'),
-            ...(changeset
+            relation('out', 'r', { active: !view.changeset }),
+            node('prop', view.deleted ? 'Deleted_Property' : 'Property'),
```
CarsonF, line 57:
```suggestion
            node('prop', labelForView('Property')),
```

CarsonF (commented):
Because of the signature change in `ResourceResolver` we also need to update every place we have `@HandleIdLookup()` to use a view arg instead of changeset.

CarsonF (approved):
Let's do it!

PR #2051, opened 2021-07-28
Hydrate inside of List queries

### LINKED ISSUE #2048: Hydrate inside of List queries
Currently all of our list queries are `O(N+1)` meaning there's a query to get a list of IDs then a query per ID to get the full item back. We did this for performance when we were querying `SecurityGroups`. Now that we aren't I believe there's a performance gain to be made from doing this all in one query `O(1)`.

Essentially it's taking this
```ts
// repository.ts
const results = this.db.query()
...
.paginate(input)
.first();

// service.ts
return await mapListResults(results, (id) => this.readOne(id, session));
```

And changing it to
```ts
// repository.ts
const results = this.db.query()
...
.paginate(input, this.hydrate())
.first();

// service.ts
return await mapListResults(results, (id) => this.secure(id, session));
```

In the first snippet, the `results` is a paginated list of `IDs` that are then passed through `readOne` to "hydrate".
In the second snippet, the `results` is a paginated list of unsecured `DTOs`, so all that's left to do is secure them (which doesn't need the DB).

The big pieces of this are:
1. [x] Define hydrate() query fragments. This is already done in #2038 
2. [x] Split `readOne` methods to `readOneUnsecured` and `secure`. It looks like this was mostly/completely done in #2038 as well.
3. [x] Just connect things up like I showed above.

daniel-leod:
Fixes #2048

### DIFF
--- a/src/common/secured-property.ts
+++ b/src/common/secured-property.ts
@@ -51,7 +51,28 @@ export type UnsecuredDto<Dto> = {
   [K in keyof Dto as Exclude<K, 'canDelete'>]: UnwrapSecured<Dto[K]>;
 };
 
-export type UnwrapSecured<T> = T extends Secured<infer P> ? P : T;
+/**
+ * Specify this on a property to override the value type for UnsecuredDto on
+ * the object.
+ *
+ * @example
+ * class Foo {
+ *   color: string & SetUnsecuredType<number>
+ * }
+ * const unsecured: UnsecuredDto<Foo> = {
+ *   hexColor: 0xFFFFFF,
+ * };
+ */
+export interface SetUnsecuredType<Value> {
+  // eslint-disable-next-line @typescript-eslint/naming-convention -- shush we are hiding this. It's only for TS types.
+  __unsecured_type__?: Value;
+}
+
+export type UnwrapSecured<T> = T extends SetUnsecuredType<infer Override>
+  ? Override
+  : T extends Secured<infer P>
+  ? P
+  : T;
 
 export const isSecured = <T>(value: T | Secured<T>): value is Secured<T> =>
   isObject(value) && 'canRead' in value && 'canEdit' in value;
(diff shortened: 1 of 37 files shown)


src/components/budget/budget-record.repository.ts
```diff
@@ -102,12 +102,12 @@ export class BudgetRecordRepository extends DtoRepository(BudgetRecord) {
...
       .apply(this.recordsOfBudget({ changeset }))
       .apply(sorting(BudgetRecord, input))
-      .apply(paginate(input))
+      .apply(paginate(input, this.hydrate()))
       .first();
     return result!; // result from paginate() will always have 1 row.
   }
 
-  hydrate({
+  hydrateRecord({
```
CarsonF, line 110: Revert this, it is suppose to override the inherited one.

src/components/budget/budget-record.repository.ts
```diff
@@ -102,12 +102,12 @@ export class BudgetRecordRepository extends DtoRepository(BudgetRecord) {
       .matchNode('budget', 'Budget', { id: budgetId })
       .apply(this.recordsOfBudget({ changeset }))
       .apply(sorting(BudgetRecord, input))
-      .apply(paginate(input))
+      .apply(paginate(input, this.hydrate()))
```
CarsonF, line 105: Pass the required args here

src/components/budget/budget-record.repository.ts
```diff
@@ -71,7 +71,7 @@ export class BudgetRecordRepository extends DtoRepository(BudgetRecord) {
     return !!result;
   }
 
-  async readOne(id: ID, session: Session, changeset?: ID) {
+  async readOneUnsecured(id: ID, session: Session, changeset?: ID) {
```
CarsonF, line 74: I meant this change in the service files where there could be both a readOne & readOneUnsecured, both of which could be used depending on the need. readOne would become just shortcut for readOneUnsecured() + secure().

Repositories have no concept of securing objects so this suffix doesn't make sense here.

CarsonF:
I wanna wait for #2108 to get merged. Then I'll rebase this if needed and figure out why the test is failing.

CarsonF (approved):

PR #2063, opened 2021-08-04
Add directory to periodic report and rename reportFile to pnp for progress report

tmpacifitech:


src/components/periodic-report/dto/upload-periodic-report.dto.ts
```diff
@@ -1,18 +0,0 @@
...
-  @IdField()
-  readonly reportId: ID;
-
-  @Field({
-    description: 'New version of the report file',
-  })
-  @Type(() => CreateDefinedFileVersionInput)
-  @ValidateNested()
-  readonly file: CreateDefinedFileVersionInput;
-}
```
CarsonF, line 18: Could you try to keep this as deprecated?

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -30,15 +30,15 @@ import {
...
+    const reportDirectoryId = await generateId();
 
     const Report = resolveReportType(input);
     const initialProps = {
       type: input.type,
       start: input.start,
       end: input.end,
       receivedDate: null,
-      reportFile: reportFileId,
+      reportDirectory: reportDirectoryId,
```
CarsonF, line 41: I know this is the way that we do defined files, but it doesn't really make much sense. We end up having both:
```cql
(:BaseNode)-[:reportFile]->(:Property { id: 'asdf' })
(:BaseNode)-[:reportFileNode]->(:File { id: 'asdf' })
```
I think there was a period of time in dev where the file wasn't created until the first version was uploaded.

Since we aren't bound by the _defined file_ abstractions here, let's just make this a normal relationship.
The directory can be created first and then attached with `createRelationships`.

This does mean adding an explicit match in the hydrate query.

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -53,11 +53,18 @@ export class PeriodicReportResolver {
...
-    return await this.files.resolveDefinedFile(report.reportFile, session);
+  ): Promise<SecuredDirectory> {
+    return {
+      canEdit: false,
+      canRead: true,
+      value: await this.files.getDirectory(
+        report.reportDirectory.value!,
+        session
+      ),
+    };
```
CarsonF, line 68: We should forward on the auth properties just like before.
https://github.com/SeedCompany/cord-api-v3/blob/master/src/components/file/file.service.ts#L481-L490

It's a bit ambiguous but `canEdit` could be interrupted as can files be uploaded.

src/components/periodic-report/periodic-report.service.ts
```diff
@@ -84,25 +82,7 @@ export class PeriodicReportService {
...
-        input.reportFile,
-        session
-      );
-      const newVersion = await this.files.getFileVersion(
-        reportFile.uploadId,
-        session
-      );
-      await this.eventBus.publish(
-        new PeriodicReportUploadedEvent(updated, newVersion, session)
-      );
```
CarsonF, line 104: I forgot about this. This is important. This is how we know to extract the pnp data. I think the best course of action is to add a `pnp: DefinedFile` to `ProgressReports` specifically. `reportFile` resolver could forward to this for progress reports.

The event we use internally can change. Maybe it becomes `PnpProgressUploadedEvent`.

CarsonF (changes requested):

src/components/periodic-report/events/periodic-report-uploaded.event.ts
```diff
@@ -1,13 +1,13 @@
 import { Session } from '../../../common';
 import { FileVersion } from '../../file';
-import { PeriodicReport } from '../dto';
+import { ProgressReport } from '../dto';
 
 /**
  * Dispatched when a new file is uploaded for a periodic report
```
CarsonF, line 6: update comment and rename file to match new event name

src/components/progress-summary/handlers/extract-pnp-file-on-upload.handler.ts
```diff
@@ -1,17 +1,17 @@
...
   constructor(
     private readonly repo: ProgressSummaryRepository,
     private readonly extractor: ProgressExtractor
   ) {}
 
-  async handle(event: PeriodicReportUploadedEvent) {
+  async handle(event: PnpProgressUploadedEvent) {
     if (event.report.type !== ReportType.Progress) {
       return;
     }
```
CarsonF, line 17: This if statement isn't needed with the new event.

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -53,11 +69,50 @@ export class PeriodicReportResolver {
     return await this.service.update(input, session);
   }
 
-  @ResolveField(() => SecuredFile)
+  @ResolveField(() => SecuredFile, {
+    description:
+      'Returns Pnp file for progress report. Otherwise returns first report file of directory',
+  })
```
CarsonF, line 75: This can be deprecated. `directory` or `pnp` should be recommended instead. Speaking of which we need to expose `pnp` in a new `ProgressReportResolver`.

src/components/periodic-report/periodic-report.resolver.ts
```diff
@@ -53,11 +69,50 @@ export class PeriodicReportResolver {
...
+      return await this.files.resolveDefinedFile(report.pnp, session);
+    } else {
+      const children = await this.files.listChildren(
+        directoryId,
+        undefined,
+        session
+      );
+
+      const item = children.items[0];
+      const file = await this.files.getFile(item.id, session); // TODO: is there a better way to pass `children.items[0]` to the value directly with typechecking passed?
```
CarsonF, line 95: If you use the `type` filter in the `listChildren` input you can guarantee the list will be of a certain type and then it's safer to typecast it.

Currently this could break if the first node is directory instead of a file.

src/components/periodic-report/dto/index.ts
```diff
@@ -2,7 +2,6 @@ export * from '../../authorization';
 export * from './periodic-report.dto';
 export * from './create-periodic-report.dto';
 export * from './list-periodic-reports.dto';
-export * from './upload-periodic-report.dto';
```
CarsonF, line 5: This file still exists

src/components/periodic-report/dto/periodic-report.dto.ts
```diff
@@ -50,7 +51,7 @@ class PeriodicReport extends Resource {
   @Field()
   readonly receivedDate: SecuredDateNullable;
 
-  readonly reportFile: DefinedFile;
+  readonly directory: DefinedFile;
```
CarsonF, line 54: A DB migration class needs to be written to:
- Move `ProgressReport.reportFile` to `pnp`
- Create directories
- Move other report files into the directory

CarsonF (changes requested):
Getting closer!

src/components/periodic-report/periodic-report.resolver.ts
tmpacifitech: @CarsonF hmm.. I wasn't able to make this happens.

elisabeth-sorrell:
Covered in this PR: https://github.com/SeedCompany/cord-api-v3/pull/2419

PR #2068, opened 2021-08-05
Final periodic reports & 87six comments migrations

blakekrammes:
- 87Six comments-->Cord V3 posts
- Generating final periodic reports

### DIFF
--- a/src/components/periodic-report/migrations/AddFinalReport.migration.ts
+++ b/src/components/periodic-report/migrations/AddFinalReport.migration.ts
@@ -1,8 +1,74 @@
+import { node, relation } from 'cypher-query-builder';
 import { BaseMigration, Migration } from '../../../core';
+import { ReportType } from '../dto';
 
 @Migration('2021-08-03T15:30:05')
 export class AddFinalReportMigration extends BaseMigration {
   async up() {
-    // TODO implement
+    await this.migrate(ReportType.Financial);
+    await this.migrate(ReportType.Narrative);
+    await this.migrate(ReportType.Progress);
+  }
+
+  // the finalReportMigration property is present on all the migrated :PeriodicReport nodes if there is a need to revert
+  private async migrate(type: ReportType) {
+    const before = await this.getBaseNodeCount(type, false);
+    this.logger.info(
+      `${before ?? 0} ${
+        type !== ReportType.Progress ? 'projects' : 'engagements'
+      } with ${type} reports before migration`
+    );
+    await this.db
+      .query()
+      .raw(
+        `
+          match(u:User)-[:email { active: true }]->(:Property { value: "devops@tsco.org" })
+          match(b:BaseNode)-[:report]->(:${type}Report)
+          with u, collect(distinct b) as baseNodes
+          unwind baseNodes as baseNode
+          call {
+            with baseNode
+            match(baseNode)-[:report]->(:${type}Report)-[:end]->(en:Property)
+            with en
+            order by en.value desc limit 1
+            return en.value as lastEnd
+          }
+          create
+            (baseNode)-[:report { active: true, createdAt: datetime() }]->(pr:${type}Report:PeriodicReport:BaseNode{ finalReportMigration: true, createdAt: datetime(), id: apoc.create.uuid() }),
+              (pr)-[:type { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: "${type}" }),
+              (pr)-[:start { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: date(lastEnd) }),
+              (pr)-[:end { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: date(lastEnd) }),
+              (pr)-[:receivedDate { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: null }),
+              (pr)-[:reportFile { active: true, createdAt: datetime() }]->(reportFile:Property { createdAt: datetime(),  value: apoc.create.uuid() }),
+              (pr)-[:reportFileNode { active: true } ]->(rfn:BaseFile:BaseNode:File:FileNode{ createdAt: datetime(), id: reportFile.value }),
+                (rfn)-[:createdBy { active: true, createdAt: datetime() }]->(u),
+                (rfn)-[:name { active: true, createdAt: datetime() }]->(:Property { value: toString(lastEnd) }),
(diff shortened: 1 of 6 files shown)


src/components/periodic-report/migrations/Post.migration.ts
```diff
@@ -0,0 +1,72 @@
...
+              (po)-[:body { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: post.body, sortValue: post.body }),
+              (po)-[:modifiedAt { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: p.createdAt, sortValue: p.createdAt }),
+              (po)-[:creator { active: true, createdAt: datetime() }]->(:Property { createdAt: datetime(), value: rootId, sortValue: rootId })
+        `
+      )
+      .run();
+    const after = await this.getPostCount(false);
+    this.logger.info(`${after ?? 0} Cord V3 posts after migration`);
+
+    // this can be omitted if we want to keep data here in case of reversion
```
blakekrammes, line 38: This should be uncommented to remove 87Six comments, but I wasn't sure if we wanted to do that at the same time.

blakekrammes:
I've tested both queries out FYI.

src/components/periodic-report/migrations/AddFinalReport.migration.ts
```diff
@@ -1,8 +1,74 @@
...
+          match(b:BaseNode)-[:report]->(:${type}Report)
+          with u, collect(distinct b) as baseNodes
+          unwind baseNodes as baseNode
+          call {
+            with baseNode
+            match(baseNode)-[:report]->(:${type}Report)-[:end]->(en:Property)
+            with en
+            order by en.value desc limit 1
+            return en.value as lastEnd
+          }
```
CarsonF, line 35: I know I said previously that we don't need to filter on reports that already have a final report. But now I'm thinking it might be better safe than sorry...

src/components/periodic-report/migrations/AddFinalReport.migration.ts
```diff
@@ -1,8 +1,74 @@
...
+    this.logger.info(
+      `${before ?? 0} ${
+        type !== ReportType.Progress ? 'projects' : 'engagements'
+      } with ${type} reports before migration`
+    );
+    await this.db
+      .query()
+      .raw(
+        `
+          match(u:User)-[:email { active: true }]->(:Property { value: "devops@tsco.org" })
```
CarsonF, line 25:
```suggestion
          match(u:RootUser)
```

src/components/periodic-report/migrations/Post.migration.ts
```diff
@@ -0,0 +1,72 @@
+import { BaseMigration, Migration } from '../../../core';
+
+@Migration('2021-08-03T15:30:05')
+export class PostMigration extends BaseMigration {
```
CarsonF, line 4: Move to posts module & rename to 87SixCommentsMigration

src/components/periodic-report/migrations/AddFinalReport.migration.ts
```diff
@@ -1,8 +1,74 @@
...
+        relation('out', '', 'report'),
+        node(
+          'rn',
+          `${type}Report`,
+          after ? { finalReportMigration: true } : {}
+        ),
+      ])
+      .return<{ count: number }>('count(distinct b) as count')
+      .first();
+    return res?.count;
```
CarsonF, line 72:
```suggestion
    return res?.count ?? 0;
```

src/components/periodic-report/migrations/AddFinalReport.migration.ts
```diff
@@ -1,8 +1,74 @@
...
+  private async getBaseNodeCount(type: ReportType, after: boolean) {
+    const res = await this.db
+      .query()
+      .match([
+        node('b', 'BaseNode'),
+        relation('out', '', 'report'),
+        node(
+          'rn',
+          `${type}Report`,
+          after ? { finalReportMigration: true } : {}
```
CarsonF, line 67: base node props get merged into read responses FYI

src/components/periodic-report/migrations/Post.migration.ts
CarsonF: Yes

CarsonF (changes requested):

src/components/periodic-report/migrations/AddFinalReport.migration.ts
blakekrammes: ?

CarsonF:
```ts
reports.readOne('id') // { ..., finalReportMigration: true }
```

src/components/periodic-report/migrations/Post.migration.ts
blakekrammes: Yes to un-commenting it?

CarsonF: I mean we should remove the old data

src/components/periodic-report/migrations/AddFinalReport.migration.ts
blakekrammes: So that prop will always exist on the reads. Will that create a problem?

src/components/periodic-report/migrations/AddFinalReport.migration.ts
```diff
@@ -1,8 +1,74 @@
...
+          with u, collect(distinct b) as baseNodes
+          unwind baseNodes as baseNode
+          call {
+            with baseNode
+            match(baseNode)-[:report]->(:${type}Report)-[:end]->(en:Property)
+            with en
+            order by en.value desc limit 1
+            return en.value as lastEnd
+          }
+          create
```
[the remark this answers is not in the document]
blakekrammes: It's not an easy refactor, at least for me. I'm not able to access the variable returned from the previous subquery. I dunno if the unwind is adding complications too.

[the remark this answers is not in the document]
blakekrammes: Screen Shot 2021-08-11 at 4 58 32 PM

Using `variable()`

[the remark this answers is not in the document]
CarsonF: Ah right variables for initial props needed. Ok let's skip it.

CarsonF (approved):

CarsonF:
Thanks @blakekrammes!

Issue #2064, opened 2021-08-04
Goals & Progress | Parse Overall Progress from PnP


sethmcknight:
When a new Progress Report PnP is uploaded to an Engagement's Progress Report for a specific Period, parse the PnP Spreadsheet to pull the Overall Progress for the Translation Goals.

sethmcknight:
Here is an example of the PnP Spreadsheet to Parse as an .xlsx file: ExamplePnP.xlsx

The actual PnP is an .xlsm file, if an example in that format is needed please reach out to me. .xlsm files cannot be attached to GitHub issues.

Parsing for Overall Progress should be based on the most current version of each Progress Report. We want to parse and save this data for each reporting period.

In the PnP Workbook you will look at the 'Progress' tab.

There are 3 sections of data we want to capture for each Fiscal Year:
- Progress each Quarter
- Progress each Year
- Cumulative Progress

Each row on the Progress section represents a Fiscal Year. Start on Row 20 and go down until the Fiscal Years are outside of the range of the Projects MOU Dates, or a merged cell is encountered.

Look in Column AG for the Fiscal Year.

Progress each Quarter by Fiscal Year:
Each quarter in a Fiscal Year has the % of the translation work that has been completed during that quarter.
Column AH: Q1 by FY
Column AI: Q2 by FY
Column AJ: Q3 by FY
Column AK: Q4 by FY

Progress by Fiscal Year:
Each Fiscal Year has a sum of the progress planned and progress recorded as a %.
Column AL: Planned Progress by FY
Column AM: Actual Progress by FY

Cumulative Progress by Fiscal Year:
Each Fiscal Year has a sum of the cumulative planned and recorded progress to date, as well as the variance between the two recorded as a %.
Column AN: Cumulative Planned Progress by FY
Column AO: Cumulative Actual Progress by FY
Column AP: Cumulative Variance by FY

Total Cumulative Progress:
The total cumulative progress can be found on the Harvest Tab, and is currently parsed from the PnP.
This included the Planned Progress, Actual Progress, and Variance to-date.

sethmcknight:
@CarsonF updated with example PnP and info on the cells to parse. This should be ready to work.

CarsonF:
[...]

PR #2073, opened 2021-08-10
Goals & Progress | Parse Overall Progress from PnP

daniel-leod:
#2064

### DIFF
--- a/src/components/progress-summary/progress-extractor.service.ts
+++ b/src/components/progress-summary/progress-extractor.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@nestjs/common';
-import { read, WorkBook, WorkSheet } from 'xlsx';
+import { CellObject, read, WorkBook, WorkSheet } from 'xlsx';
+import { CalendarDate, fiscalQuarter, fiscalYear } from '../../common';
 import { ILogger, Logger } from '../../core';
 import { FileService, FileVersion } from '../file';
 import { ProgressSummary as Progress } from './dto';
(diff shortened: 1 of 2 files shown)


CarsonF (changes requested):
I was just made aware today that this extracts formatted numbers instead of the actual number values. Several decimal places are lost which are important. 

We can access the actual value much easier than what is done here. 
```ts
sheet.AL20
```
will give us cell `AL20`. There's a `v` property that is the raw number which has all of the precision and doesn't require us to parse it with "parsePercent".

Additionally all the data we require is on a single dynamic row based on fiscal year.
Once we find that row we can grab the data based on the static columns.

Something like this could work.
```ts
const findFiscalYearRow = (sheet, fiscalYear: number) => {
  let i = 20;
  while (true) {
    const cell = sheet[`AG${i}`];
    if (typeof cell.v !== 'number') {
      return null;
    }
    if (cell.v === fiscalYear) {
      return i;
    }
  }
});
```

After that it's simply
```ts
sheet[`AL${fiscalYearRow}`].v
```
For all the columns we need.

This doesn't require `sheet_to_json` either.
We could still have null checks on `v` for sanity though.

---

Given how close these 3 data points actually are I'm totally fine if you want to squash them into one method. If not, then definitely extracting the common pieces into separate methods.

src/components/progress-summary/progress-extractor.service.ts
```diff
@@ -65,15 +58,103 @@ export class ProgressExtractor {
...
+          } else if (parseInt(row.AI) === fiscalQuarter) {
+            return this.parseRawData(row.AI, row.AI);
+          } else if (parseInt(row.AJ) === fiscalQuarter) {
+            return this.parseRawData(row.AJ, row.AJ);
+          } else if (parseInt(row.AK) === fiscalQuarter) {
+            return this.parseRawData(row.AK, row.AK);
+          }
+        }
+      }
+    } catch (e) {
```
CarsonF, line 95: @blakekrammes what is this catch for?

src/components/progress-summary/progress-extractor.service.ts
```diff
@@ -12,7 +12,11 @@ export class ProgressExtractor {
...
   ) {}
 
-  extractCumulative(pnp: WorkBook, file: FileVersion): Progress | null {
+  extractCumulative(
+    pnp: WorkBook,
+    file: FileVersion,
+    fiscalYear: number
+  ): Progress | null {
     // new standard 2020 version has new sheet "Harvest" which isolates relevant progress data
     const sheet = pnp.Sheets.Harvest ?? pnp.Sheets.Progress;
```
CarsonF, line 21: We shouldn't need to worry about the Harvest tab either. It doesn't have the data we are looking for.

src/components/progress-summary/progress-extractor.service.ts
blakekrammes: `parsePercent`, which is called by `parseRawData`, can throw an error.

src/components/progress-summary/progress-extractor.service.ts
blakekrammes: It doesn't? Isn't `cumulative` the same as `summary` was before?

blakekrammes: Oh right, because you're wanting to grab the longer decimal values now.

CarsonF (approved):

Issue #1737, opened 2021-03-11
PROJECT: Product Progress


michaelmarshall:
- #2009 
- #1950 
- #1949
- #2064 
- #2055 
- #1951 
- #1760 
- #1811
- #1741
- #2194 
- #2190 
- #2074 
- #2069 
- #1974 

- seedcompany/cord-field#961
- seedcompany/cord-field#957
- seedcompany/cord-field#924
   seedcompany/cord-field#945
- seedcompany/cord-field#925
   seedcompany/cord-field#961
- seedcompany/cord-field#910
- seedcompany/cord-field#909
- seedcompany/cord-field#908
- seedcompany/cord-field#902
- seedcompany/cord-field#898
- seedcompany/cord-field#920
- seedcompany/cord-field#957
- seedcompany/cord-field#956

***DRAFT***

As a project manager, I want to be able to assign goals to my project so that I can give goal-specific updates. 

This feature gives the PM the ability to add goals to a project, and then give updates on each goal as translation work completes. Goals vary in type, and so will the updates. One goal may be to give a % complete on an item, which means the update will be an integer (0-100), while another goal may be on whether a certain item is completed (boolean). 

Although goals are based on the products a project's engagements have, there is a finite list of known goals that PMs can choose from. Therefore, the list of possible goals will be kept in code, not the database, and there will be no way for a user to add new goals to the list for re-use across projects. We will still need a custom goal field that will be simple strings for projects that do things a little different (storytelling). Updates to those goals will be custom as well, just simple strings.

Viewing and editing goals will be another permission in our role definitions in the authorization service.

sethmcknight:
@michaelmarshall we have a request to look at the possibility of pulling some of this data from the PnP Spreadsheet instead of/in addition to users inputting via FE

michaelmarshall:
@sethmcknight copy that. when the feature is done/stable we can have Blake give us a level of effort on adding PnP data.

sethmcknight:
[Data Points for Translation Goals and Progress.pdf](https://github.com/SeedCompany/cord-api-v3/files/6231564/Data.Points.for.Translation.Goals.and.Progress.pdf)

sethmcknight:
**Designs Complete**

[Prototype](https://www.figma.com/proto/sWiPVOrIojuU8ityKQAnBP/Draft-Cord-Field?node-id=396%3A25994&scaling=min-zoom&page-id=149%3A0)

[...]

PR #2074, opened 2021-08-10
Dynamic number & boolean product progress measurements

CarsonF:
Part of #1737

- Add `Product.progressStepMeasurement` to declare that the product's step progress is measured by a _number_, _boolean_, or _percent_.
- Add `Product.progressTarget` to allow a dynamic target for the progress to end at. This is hardcoded to 100/1 for percent/boolean measurements.
- Rename `StepProgress.percentDone` to `completed` as the number may not be a "percent" now.
- Validate that `StepProgress.completed` is <= `Product.progressTarget`

For "migration":
- Kept `percentDone` as a deprecated field on input/output so this isn't a GQL breaking change.
- Existing products will default to their measurement to percents.
- This is a breaking change for all existing product progress. I didn't bother writing a DB migration since this is only in dev so far.

@sethmcknight These changes sound like what you are expecting / what's needed for new tracking?


sethmcknight:
@CarsonF yes, that should cover it

PR #2085, opened 2021-08-15
add mutation for flag

Rhuan288souza:


CarsonF:
#2086 is on the right track

Issue #2070, opened 2021-08-10
Preset Inventory Flag


sethmcknight:
Preset Inventory Flag

Field will be selecting Preset Inventory for Advancement to connect Investors to as part of our Language Offering.

Project Managers will make this selection at the Project level, but it will be applied to the Language Engagements. If a Language Engagement has an Investor connected we may need to automatically flag new Language Engagements with that Language as Approved Inventory.

This flag will need to be:
- Selected, stored, and displayed on the Project Details page
- Displayed on the Project Card
- Displayed on the Language Engagement Card
- Displayed on the Language Engagement Details
- Displayed on the Language Card
- Displayed on the Language Details

We'd also like to add Preset Inventory as a filter to the Projects list and Languages list

If a language has a project marked as Preset Inventory, that language should show (read-only) as Preset Inventory.

sethmcknight:
Design examples:

<img width="685" alt="Screen Shot 2021-08-10 at 2 26 37 PM" src="https://user-images.githubusercontent.com/7377802/128930351-43f6dc9a-5e3a-4d96-b760-f9cacfd61c96.png">
<img width="684" alt="Screen Shot 2021-08-10 at 2 27 15 PM" src="https://user-images.githubusercontent.com/7377802/128930355-c01f1800-92d3-4c1e-9615-920d615fcb2e.png">
<img width="684" alt="Screen Shot 2021-08-10 at 2 27 06 PM" src="https://user-images.githubusercontent.com/7377802/128930356-5c81e709-167a-49ad-be58-1cc0ebdb7d92.png">
<img width="679" alt="Screen Shot 2021-08-10 at 2 26 44 PM" src="https://user-images.githubusercontent.com/7377802/128930357-1d71e03d-54f4-4051-a98e-d577a5831af4.png">

sethmcknight:
Use the blue pills, don't add red pills

CarsonF:
I think the flag is a poor icon to describe this. "Flag" in general is just a boolean, yes/no, for a certain data point. Every place I I've seen it used requires a user action. i.e. "flag for abuse" = review needed; or in email, where you flag a message to revisit later.

Something describing the actual use of this data point would be more relevant, aka "**approved/available for investors**".

Some suggestions from a quick search:
![Screen Shot 2021-08-11 at 4 04 45 PM](https://user-images.githubusercontent.com/932566/129104453-84b181ae-ae43-4006-acc9-fc0816f9823c.png)
[...]

PR #2086, opened 2021-08-15
Preset inventory flag

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/2070

### DIFF
--- a/src/components/language/language.repository.ts
+++ b/src/components/language/language.repository.ts
@@ -86,7 +86,7 @@ export class LanguageRepository extends DtoRepository(Language) {
     const result = await this.db
       .query()
       .match([requestingUser(session), ...permissionsOfNode('Language')])
-      .apply(languageListFilter(input.filter))
+      .apply(languageListFilter(input.filter, this))
       .apply(sorting(Language, input))
       .apply(paginate(input))
       .first();
--- a/src/components/language/language.repository.ts
+++ b/src/components/language/language.repository.ts
@@ -1,19 +1,24 @@
 import { Injectable } from '@nestjs/common';
-import { node, relation } from 'cypher-query-builder';
+import { inArray, node, Query, relation } from 'cypher-query-builder';
 import { ID, Session } from '../../common';
 import { DtoRepository, matchRequestingUser } from '../../core';
 import {
   ACTIVE,
+  any,
   collect,
   createNode,
   createRelationships,
+  exp,
   matchProps,
+  merge,
   paginate,
   permissionsOfNode,
   requestingUser,
   sorting,
+  variable,
 } from '../../core/database/query';
 import { DbPropsOfDto } from '../../core/database/results';
+import { ProjectStatus } from '../project';
 import { CreateLanguage, Language, LanguageListInput } from './dto';
 import { languageListFilter } from './query.helpers';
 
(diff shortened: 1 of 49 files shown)


src/components/engagement/dto/create-engagement.dto.ts
```diff
@@ -55,6 +55,9 @@ export abstract class CreateLanguageEngagement extends CreateEngagement {
 
   @Field({ nullable: true })
   readonly historicGoal?: string;
+
+  @Field({ nullable: true })
+  readonly approvedInventory?: boolean;
```
zkhin, line 60: Do we need to expose this field or create/update only in the handler?

src/components/engagement/dto/engagement.dto.ts
```diff
@@ -152,6 +152,9 @@ export class LanguageEngagement extends Engagement {
 
   @Field()
   readonly historicGoal: SecuredString;
+
+  @Field()
+  readonly approvedInventory: SecuredBoolean;
```
zkhin, line 157: Is this `boolean` or `enum?

src/components/engagement/dto/create-engagement.dto.ts
CarsonF: It can be on create and update

src/components/engagement/dto/engagement.dto.ts
CarsonF: Boolean is good

CarsonF:
This should be on project instead of engagement

CarsonF (changes requested):
Move to project

zkhin:
Adding calculated/read only property to language for approved inventory also

src/components/language/language.repository.ts
```diff
@@ -111,6 +111,27 @@ export class LanguageRepository extends DtoRepository(Language) {
...
+        node('project', 'Project'),
+        relation('out', '', 'approvedInventory', { active: true }),
+        node('prop', 'Property'),
+      ])
+      .where({
+        prop: {
+          value: true,
+        },
+      })
+      .return('project')
```
CarsonF, line 131: This could be done in the Language.hydrate method and we could save the separate lookup. I think it would be faster.
```ts
approvedInventory: `exists((node)
  <-[:language { active: true }]-(:LanguageEngagement)
  <-[:engagement { active: true }]-(:Project)
  -[:approvedInventory { active: true }]->(:Property { active: true })
)`
```

CarsonF (commented):
I think they changed their minds again and actually want to call it "preset inventory". I'm still on the fence too about how meaningless that phrase is.

sethmcknight:
Here are the icons. @CarsonF removed all of Adobe Illustrator's junk (99% of the file size)

```html
<svg viewBox="0 0 24 24">
  <g>
    <path d="M7.1 15h4.4v2H7.1zM7.1 11h7v2h-7zM7.1 7h7v2h-7z"/>
    <path d="M19 3H5c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2zm0 11v5H5V5h14v9z"/>
  </g>
</svg>
```

```html
<svg viewBox="0 0 24 24">
  <g>
    <path d="M20 3H4c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h16c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2zM10 17H5v-2h5v2zm0-4H5v-2h5v2zm0-4H5V7h5v2zm4.8 6L12 12.2l1.4-1.4 1.4 1.4L18 9l1.4 1.4-4.6 4.6z" />
  </g>
</svg>
```

They look like this:
Screen Shot 2021-09-02 at 10 07 26 AM

CarsonF (approved):

Issue #2052, opened 2021-07-29
createProduct > Returned all value of "steps" field  even particular "methodologies" is given "(i.e. Written ...Paratext )" in the query input


iPiyushPatel:
**Roles:** All

**Describe the bug**
createProduct > Returned all value of `steps` field  even particular `methodologies` is given `(i.e. Written ...Paratext )` in the query input

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
 Given `methodologies` according  `Steps` should be returend.

**Actual Result**
 Returned all value of `steps` field  even particular `methodologies` is given `(i.e. Written ...Paratext )` in the query input

See video -  

https://user-images.githubusercontent.com/5212585/127483879-920dc36e-5bb8-4fa5-9f25-9e9e99dc4514.mp4

**GraphQL Query:(createProduct )**
![Create Product](https://user-images.githubusercontent.com/5212585/127478987-1dae44b5-9a39-4013-9a76-50e1406ce405.png)

**GraphQL response**
```
{
    "data": {
        "createProduct": {
            "product": {
                "id": "hRyeSLxPdbR",
                "createdAt": "2021-07-29T11:22:39.295Z",
                "canDelete": true,
                "approach": "Written",
                "legacyType": "JesusFilm",
                "methodology": {
                    "value": "Paratext",
                    "canEdit": true,
                    "canRead": true
                },
                "steps": {
                    "value": [
                        "BackTranslation",
                        "Check",
                        "CommunityTesting",
                        "Completed",
                        "ConsistencyCheckAndFinalEdits",
                        "ConsultantCheck",
                        "Craft",
                        "ExegesisAndFirstDraft",
                        "InternalizationAndDrafting",
                        "PeerRevision",
                        "Record",
                        "TeamCheck",
                        "Test"
                    ],
                    "canRead": true,
                    "canEdit": true
                }
            }
        }
    },
```

**Query and Input File**
[CreateProduct.txt](https://github.com/SeedCompany/cord-api-v3/files/6900284/CreateProduct.txt)

CordAPIV3# 2a79e04

CarsonF:
On save we just need to verify given steps are available based on https://github.com/SeedCompany/cord-api-v3/blob/08f111e3953adf33088cb1e96e59cc5b65b8b21d/src/components/product/dto/available-methodology-steps.ts#L24-L24

PR #2100, opened 2021-08-24
Validate product steps are available with methodology

daniel-leod:
#2052


src/components/product/product.repository.ts
```diff
@@ -150,11 +152,14 @@ export class ProductRepository extends CommonRepository {
     const Product = input.produces
       ? DerivativeScriptureProduct
       : DirectScriptureProduct;
+
+    const availableSteps = MethodologyAvailableSteps[input.methodology!];
```
CarsonF, line 156: `methodology` is nullable, so we need to account for that.

CarsonF (changes requested):

Issue #2004, opened 2021-07-07
projectMembers > Filter by "roles" not working


iPiyushPatel:
**Roles:** All 

**Describe the bug**
projectMembers > Filter by `roles` not working

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
ProjectMembers list with requested `Roles` should be returned.

**Actual Result**
Filter by `roles` not working

See video -  

https://user-images.githubusercontent.com/5212585/124769322-012ff100-df57-11eb-84b0-677df57329b7.mp4

**GraphQL Query:(projectMembers)**
![image](https://user-images.githubusercontent.com/5212585/124769633-40f6d880-df57-11eb-8624-1e4811713818.png)

**GraphQL response**
```
{
    "data": {
        "projectMembers": {
            "total": 67,
            "hasMore": true,
            "items": [
                {
                    "id": "lorJp4sAXHM",
                    "roles": {
                        "value": [
                            "Administrator"
                        ],
                        "canRead": true,
                        "canEdit": true
                    }
                },
                {
                    "id": "Mx1dfCAdVqo",
                    "roles": {
                        "value": [
                            "Administrator"
                        ],
                        "canRead": true,
                        "canEdit": true
                    }
                },
                {
                    "id": "ovGMQ9ZhF6M",
                    "roles": {
                        "value": [
                            "Administrator"
                        ],
                        "canRead": true,
                        "canEdit": true
                    }
                },
                {
                    "id": "ijzRW9muHA8",
                    "roles": {
                        "value": [
                            "Administrator"
                        ],
                        "canRead": true,
                        "canEdit": true
                    }
                },
```

**Query and Input File**
[BugProjectmembersRoleFilter.txt](https://github.com/SeedCompany/cord-api-v3/files/6777573/BugProjectmembersRoleFilter.txt)

CordAPIV3# 136c06a1

PR #2105, opened 2021-08-31
filter project members by role

daniel-leod:
#2004


elisabeth-sorrell:
@CarsonF rebased on develop. Looks good to merge to me

elisabeth-sorrell (approved):
@CarsonF Rebased on develop. Looks good to me.

src/components/project/project-member/project-member.repository.ts
```diff
@@ -146,6 +146,12 @@ export class ProjectMemberRepository extends DtoRepository(ProjectMember) {
             ]
           : []),
         node('node', 'ProjectMember'),
+        ...(filter.roles
+          ? [
+              relation('out', '', 'roles', ACTIVE),
+              node('role', 'Property', { value: filter.roles }),
```
CarsonF, line 152: `role.value` is an array, as is the `filter.roles`. They need to be intersected to see if any of member roles are in the filter. What's currently here will only exact match, and probably not even that.

Probably something like this would work.
```gql
WHERE size(apoc.coll.intersection(role.value, $filteredRoles)) > 0
```

CarsonF (changes requested):

CarsonF:
@daniel-leod Did you try this? does it work as expected?

daniel-leod:
> @daniel-leod Did you try this? does it work as expected?
Yes, it works as expected

CarsonF (approved):
Thanks @daniel-leod!

PR #2108, opened 2021-08-31
Listing secured by sensitivity of associated project and project-scoped roles

elisabeth-sorrell:
Components refactored:
- project
- partnership
- budget/budgetRecords
- project members

### DIFF
--- a/src/components/partner/partner.repository.ts
+++ b/src/components/partner/partner.repository.ts
@@ -3,7 +3,6 @@ import { node, Query, relation } from 'cypher-query-builder';
 import { DateTime } from 'luxon';
 import {
   ID,
-  isIdLike,
   NotFoundException,
   ServerException,
   Session,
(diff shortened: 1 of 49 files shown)


src/components/partner/partner.repository.ts
```diff
@@ -106,8 +106,8 @@ export class PartnerRepository extends DtoRepository(Partner) {
           sub
             .with('projList')
             .raw('UNWIND projList as project')
-            // TODO: have this match from line 92 so we don't have to match this twice
-            .match([
+            // TODO: I don't think this is needed.... more thought on this later.
+            .optionalMatch([
```
CarsonF, line 110: Shouldn't `matchProjectScopedRoles` & L103 be moved into this subquery after deduping projects matched?

elisabeth-sorrell: I don't think it should. if `matchProjecScopedRoles` andL103 were moved into this subquery, we would have to double-collect the scoped roles in order to collect all scopedRoles of all projects that member is connected to. 

.... fixed the match with partnership. Fixing org now.

elisabeth-sorrell: org fixed

CarsonF (approved):
Looks great @elisabeth-sorrell, thanks! 🎉

PR #2113, opened 2021-09-02
List ceremony based on global and scoped roles and sensitivity of project

elisabeth-sorrell:
also fix sort of listCeremonies to have a custom matcher that returns the project node's name to sort on


CarsonF:
FYI ceremony list isn't currently used anywhere, so this is going down low on my priority list.

Issue #2102, opened 2021-08-26
Set previous relation on changeset rejected


CarsonF:
Similar to #2017, `previous` relations need to be set on reject as well. Essentially anytime the changeset is _finalized_.

PR #2124, opened 2021-09-07
Add previous relation for rejected changeset

zkhin:
Fixes: https://github.com/SeedCompany/cord-api-v3/issues/2102


src/components/engagement/handlers/apply-rejected-changeset-to-engagement.handler.ts
```diff
@@ -0,0 +1,137 @@
...
+            ])
+            .delete('changesetRel')
+            .create([
+              node('node'),
+              relation('in', '', 'previous', {
+                active: true,
+                createdAt: DateTime.local(),
+              }),
+              node('changeset'),
+            ])
```
CarsonF, line 71: I'm don't think this is right. The previous relation should be applied to properties, not a changeset. This change isn't being removed from the changeset.

CarsonF (changes requested):
These changeset rejected handlers look to be 90% the same as the approved handlers.

I think it would be better to consolidate to a "finalized" event. After all the matching is done, there could be a simple conditional based on the new status of the change request to determine what action to take on the matched data.

Take engagements for example, on approval this action would be taken.
https://github.com/SeedCompany/cord-api-v3/blob/3ae9ccbdbca40ce915d80f0d65cb14d7df82bed7/src/components/engagement/handlers/apply-approved-changeset-to-engagement.handler.ts#L64-L66
On rejection, that ^ action would be swapped out.

src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts
```diff
@@ -61,21 +62,28 @@ export class ApplyApprovedChangesetToEngagement
...
             ])
-            .setValues({
-              'engagementRel.active': true,
-            })
+            .apply((q) =>
+              status === ProjectChangeRequestStatus.Approved
+                ? q.setValues({
+                    'engagementRel.active': true,
+                  })
+                : q.apply(rejectChangesetProps())
```
CarsonF, line 70: Is it weird that we _reject_ here but _commit_ below? it feels like there should be some symmetry with this logic.

zkhin: I think it will be more clear to reject [here](https://github.com/SeedCompany/cord-api-v3/blob/e6113e787bb30618ab4cc5ce30595db9f9cb045c/src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts#L92) instead of here.

CarsonF: Yeah I think so. Is there a reason that's a separate query? Shouldn't it be able to be dropped in right here?

zkhin: I added a separate [query](https://github.com/SeedCompany/cord-api-v3/blob/e6113e787bb30618ab4cc5ce30595db9f9cb045c/src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts#L80~L96) for not looping nodes that are created in changeset but I think we can just do it here.

CarsonF: I think so too

src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts
```diff
@@ -64,37 +64,19 @@ export class ApplyFinalizedChangesetToEngagement
...
               status === ProjectChangeRequestStatus.Approved
-                ? q.setValues({
-                    'engagementRel.active': true,
-                  })
+                ? q
+                    .setValues({
+                      'engagementRel.active': true,
+                    })
+                    .with('node, changeset')
+                    .apply(commitChangesetProps())
```
zkhin, line 72: Not sure why it does not apply to the pending changeset values.
```
MATCH (project:Project)-[:changeset { active: true }]->(changeset:Changeset { id: 'RquE9ZVBog7' })
    CALL {
      WITH project
      MATCH (project)-[engagementRel:engagement { active: true }]->(node:Engagement)
      RETURN node
      UNION
      WITH project, changeset
      MATCH (project)-[engagementRel:engagement { active: false }]->(node:Engagement)<-[changesetRel:changeset { active: true }]-(changeset)
      SET engagementRel.active = true
      WITH node, changeset
      CALL {
        WITH node, changeset
        // For all changed properties of node & changeset
        MATCH (node)-[changedRel { active: false }]->(changedProp:Property)<-[:changeset { active: true }]-(changeset)
        // Apply previous label to active prop
        CALL {
          WITH changeset, node, changedRel
          MATCH (node)-[previouslyActiveRel { active: true }]->(previouslyActiveProp:Property)
          WHERE type(previouslyActiveRel) = type(changedRel)
          CREATE (previouslyActiveProp)<-[:previous { active: true, createdAt: datetime() }]-(changeset)
          RETURN previouslyActiveProp, previouslyActiveRel
        }
        WITH changedRel, previouslyActiveProp, previouslyActiveRel
        // Set changed prop to active
        SET changedRel.active = true
        // Deactivate currently active prop
        SET previouslyActiveRel.active = false
        WITH previouslyActiveProp
        // prefixNodeLabelsWithDeleted(previouslyActiveProp)
        CALL {
          WITH previouslyActiveProp
          WITH previouslyActiveProp, reduce(
            deletedLabels = [], label in labels(previouslyActiveProp) |
              case
                when label starts with "Deleted_" then deletedLabels + label
                else deletedLabels + ("Deleted_" + label)
              end
          ) as deletedLabels
          call apoc.create.removeLabels(previouslyActiveProp, labels(previouslyActiveProp)) yield node as nodeRemoved
          WITH previouslyActiveProp, deletedLabels
          call apoc.create.addLabels(previouslyActiveProp, deletedLabels) yield node as nodeAdded
          RETURN 1
        }
        RETURN 1
      }
      RETURN node
    }
    RETURN node.id as id
```
Can you let me know what I am wrong with the query?

CarsonF: BaseNodes created in changesets don't need their props committed as they are [just active](https://github.com/SeedCompany/cord-api-v3/blob/narrative-report-form/src/core/database/database.service.ts#L292-L308).


Is that the problem?

zkhin: Hmm, I don't think so because the [test](https://github.com/SeedCompany/cord-api-v3/blob/develop/test/engagement-changeset-aware.e2e-spec.ts#L222~L270) is failed in updating properties in which BaseNodes are created in normal situations.

zkhin: Nvm, figure it out.

src/components/partnership/handlers/apply-finalized-changeset-to-partnership.handler.ts
```diff
@@ -7,88 +7,74 @@ import {
...
               node('node', 'Partnership'),
             ])
-            .return('node')
-            .union()
-            .with('project, changesetNode')
+            .apply((q) =>
+              status === ProjectChangeRequestStatus.Approved
+                ? q.apply(commitChangesetProps())
+                : q.apply(rejectChangesetProps())
+            )
```
CarsonF, line 55:
```suggestion
            .apply((q) =>
              status === ProjectChangeRequestStatus.Approved
                ? commitChangesetProps()
                : rejectChangesetProps()
            )
```

zkhin: Hmm... does it work?
`previous: TypeError: sub.with(...).match(...).apply(...).return is not a function`

CarsonF: Oh nope. I mean this 😁 
```suggestion
            .apply(
              status === ProjectChangeRequestStatus.Approved
                ? commitChangesetProps
                : rejectChangesetProps
            )
```

CarsonF (approved):
LGTM thanks @zkhin!

PR #2126, opened 2021-09-08
Secure Project Root Directory via TS roles & Fetch via DataLoader

elisabeth-sorrell:
To be merged after https://github.com/SeedCompany/cord-api-v3/pull/2108 and  https://github.com/SeedCompany/cord-api-v3/pull/2113


src/components/project/project.service.ts
```diff
@@ -654,14 +658,21 @@ export class ProjectService {
 
   async getRootDirectory(
     projectId: ID,
+    sensitivity: Sensitivity,
     session: Session
   ): Promise<SecuredDirectory> {
     const rootRef = await this.repo.getRootDirectory(projectId, session);
+    const membershipRoles = await this.getMembershipRoles(projectId, session);
```
CarsonF, line 665: How come we just don't do this in project hydrate?

src/components/project/project.resolver.ts
```diff
@@ -205,9 +205,9 @@ export class ProjectResolver {
   })
   async rootDirectory(
     @AnonSession() session: Session,
-    @Parent() { id }: Project
+    @Parent() { id, sensitivity }: Project
   ): Promise<SecuredDirectory> {
-    return await this.projectService.getRootDirectory(id, session);
+    return await this.projectService.getRootDirectory(id, sensitivity, session);
```
CarsonF, line 210: Let's just pass the entire project object here instead

src/components/project/project.service.ts
```diff
@@ -654,14 +658,21 @@ export class ProjectService {
...
+      otherRoles: membershipRoles,
+      sensitivity,
+    });
 
-    if (!rootRef) {
+    if (!permsOfProject.rootDirectory.canRead) {
       return {
-        canEdit: false,
-        canRead: false,
+        ...permsOfProject.rootDirectory,
```
CarsonF, line 675: I think we should just revert this change and the one below. `canEdit` is always false for this property. And for both of these we already know the `canRead` value.

src/components/project/project.repository.ts
```diff
@@ -344,7 +344,7 @@ export class ProjectRepository extends CommonRepository {
   async getRootDirectory(projectId: ID, session: Session) {
     return await this.db
       .query()
-      .match(matchSession(session, { withAclRead: 'canReadProjects' }))
+      .match(matchSession(session))
```
CarsonF, line 347: I don't see a reason to look this up here. Let's just drop this completely.

src/components/project/project.repository.ts
```diff
@@ -323,7 +323,7 @@ export class ProjectRepository extends CommonRepository {
     const query = this.db
       .query()
       .match([
-        node('node', 'Project', { projectId }),
+        node('node', 'Project', { id: projectId }),
```
CarsonF, line 326: Big OOF 😞

src/components/project/project.service.ts
```diff
@@ -657,22 +657,20 @@ export class ProjectService {
   }
 
   async getRootDirectory(
-    projectId: ID,
-    sensitivity: Sensitivity,
+    project: Project,
     session: Session
   ): Promise<SecuredDirectory> {
-    const rootRef = await this.repo.getRootDirectory(projectId, session);
-    const membershipRoles = await this.getMembershipRoles(projectId, session);
```
CarsonF: This roles are already included in `project.scope` from the hydrate query?

CarsonF: Yes exactly.

CarsonF (approved):
🎉

PR #2130, opened 2021-09-09
Expose more directory aggregate data

CarsonF:
Added these fields to `Directory` which will benefit CORD UI & DOMO.

```gql
type Directory {
  """
  The `DateTime` a file was most recently modified in this directory or any subdirectories
  """
  modifiedAt: DateTime!

  """
  The user who uploaded the most recent file in this directory or any subdirectories
  """
  modifiedBy: User!

  """
  The total size in bytes of all files under this directory and all its subdirectories
  """
  size: Int!

  """
  The total number of files under this directory and all its subdirectories
  """
  totalFiles: Int!

  """
  The first file created in this directory or any subdirectories.
  
  This can be useful to determine the "first modified date" of a directory.
  """
  firstFileCreated: File
}
```

## `modified*` Caveats

`modifiedAt/By` ignores actual directory nodes under the current directory. i.e. If I create a directory under another it won't affect the `modifiedAt/By` of the parent directory. This could be considered a feature or a bug...

Also `modifiedAt/By` isn't persisted if a file/directory is deleted. i.e. If I create a file in a directory, the directory will say I modified it just now. But if I then delete that file, the directory will revert to whatever it said previously.

### DIFF
--- a/src/components/file/directory.resolver.ts
+++ b/src/components/file/directory.resolver.ts
@@ -6,14 +6,18 @@ import {
   ResolveField,
   Resolver,
 } from '@nestjs/graphql';
+import { stripIndent } from 'common-tags';
 import { AnonSession, ID, IdArg, LoggedInSession, Session } from '../../common';
 import { Loader, LoaderOf } from '../../core';
+import { User, UserLoader } from '../user';
 import {
   asDirectory,
   CreateDirectoryInput,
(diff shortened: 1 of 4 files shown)


CarsonF (commented):
Current query projects the single dir node row into multiple rows

PR #2138, opened 2021-09-10
Secure organization props for sensitivity access across relevant roles

elisabeth-sorrell:


elisabeth-sorrell:
Note that some of the security tests failed for rootDirectory, but once this is rebased on `develop` once https://github.com/SeedCompany/cord-api-v3/pull/2136 is merged it they will pass.

CarsonF (approved):

Issue #2134, opened 2020-09-18
Edit Permissions for Partnerships


sethmcknight:
Restrict access to edit the MOU Signed & Financial Reporting Type fields on a Partnership to users who have one of the following roles:
- Lead Financial Analyst (Global R/W)
- Financial Analyst (Global R/Member W)
- Controller (Global R/W)
- Admin (Global R/W)

sethmcknight:
@CarsonF Updated roles/permissions above 👆🏼

PR #2142, opened 2021-09-13
Update role permissions for mou and financialReportingType

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2134


CarsonF:
@sethmcknight I'd like to verify this with you

CarsonF (changes requested):
Looks good @tmpacifitech. I verified with Seth, and he made a couple tweaks to the ask. It looks like there are a couple more roles we need to account for with this change.
Can you take a look?

tmpacifitech:
@CarsonF can you take a look again?

CarsonF (approved):
Thanks @tmpacifitech!

PR #2145, opened 2021-09-14
Hydrate latest file version with first file id

tmpacifitech:
Fixes https://github.com/SeedCompany/cord-field/issues/786

### DIFF
--- a/src/components/file/file.repository.ts
+++ b/src/components/file/file.repository.ts
@@ -175,7 +175,8 @@ export class FileRepository {
           node('modifiedBy'),
         ])
         .return<{ dto: File }>(
-          merge('versionProps', 'props', {
+          merge('versionProps', {
+            id: 'props.id',
             type: `"${FileNodeType.File}"`,
             latestVersionId: 'version.id',
             modifiedById: 'modifiedBy.id',


CarsonF (changes requested):
This isn't a meaningful change. I believe the issue mentioned is actually just a problem with stale cache in the UI.

tmpacifitech:
> This isn't a meaningful change. I believe the issue mentioned is actually just a problem with stale cache in the UI.

@CarsonF The issue wasn't cache. `createFileVersion` mutation and `children` file resolver returned the first version's name, size, metatype, etc. So I had to fix `hydrate` query to return first version's id with latest version's props (name, size, metatype, ...)

CarsonF:
Well this was an issue before the hydrate refactor #2118, so that tells me the problem isn't here - at least not the same problem.

`props` in the query _should_ only contain `id`, `createdAt`, & `name`. `size` & `metadata` are only stored on the versions.

CarsonF:
Closing this as we've uncovered what the real ask is in https://github.com/SeedCompany/cord-field/issues/786. @tmpacifitech I think it would be best to start fresh with that.

PR #2159, opened 2021-09-17
Implement DataLoaders for most objects

### LINKED ISSUE #2156: Implement DataLoaders for most objects
Read #2125  #2155 

Goals:
- [x] Create `readMany` methods in services & repositories.
  - If a repository doesn't have a hydrate abstraction, or it's difficult to create a `readMany` query based on the `readOne` query, this can be skipped.
- [x] Create `DataLoader` for every object.
  - If the object is `ChangesetAware` the loader needs to _extend_ `ObjectViewAwareLoader`.
  - If `readMany` wasn't able to be created then _extend_ `SingleItemLoader`.
  - Otherwise _extend_ `OrderedNestDataLoader`.
- [x] Use loader `Query` & `ResolveField` methods.
  - Favor `LoaderOf<XLoader>` for defining type instead of `DataLoader<X>`
- [x] Use `primeAll` for `Query`s that return lists
- [x] Remove unused service imports in Resolver constructors

Please use separate commits for each object.

daniel-leod:
Closes #2156

### DIFF
--- a/src/components/budget/budget-record.resolver.ts
+++ b/src/components/budget/budget-record.resolver.ts
@@ -5,8 +5,9 @@ import {
   ResolveField,
   Resolver,
 } from '@nestjs/graphql';
-import { AnonSession, LoggedInSession, Session } from '../../common';
-import { OrganizationService } from '../organization';
+import { LoggedInSession, mapSecuredValue, Session } from '../../common';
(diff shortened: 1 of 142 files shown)


CarsonF (approved):
Thanks @daniel-leod!

PR #2164, opened 2021-09-21
Fix product resolvers for OtherProduct

tmpacifitech:


CarsonF (approved):
Thanks @tmpacifitech!

PR #2170, opened 2021-09-27
Periodic Reports | Skipped Boolean & Reason

### LINKED ISSUE #1998: Periodic Reports | Skipped Boolean & Reason

daniel-leod:
API part of #1998

### DIFF
--- a/src/components/periodic-report/dto/periodic-report.dto.ts
+++ b/src/components/periodic-report/dto/periodic-report.dto.ts
@@ -7,6 +7,7 @@ import {
   SecuredDateNullable,
   SecuredProperty,
   SecuredProps,
+  SecuredStringNullable,
   Sensitivity,
   SensitivityField,
   ServerException,
@@ -54,10 +55,7 @@ class PeriodicReport extends Resource {
   readonly receivedDate: SecuredDateNullable;
 
   @Field()
-  readonly skipped: boolean;
-
-  @Field()
-  readonly skippedReason?: string;
+  readonly skippedReason: SecuredStringNullable;
 
   readonly reportFile: DefinedFile;
 
--- a/src/components/periodic-report/dto/periodic-report.dto.ts
+++ b/src/components/periodic-report/dto/periodic-report.dto.ts
@@ -53,6 +53,12 @@ class PeriodicReport extends Resource {
   @Field()
   readonly receivedDate: SecuredDateNullable;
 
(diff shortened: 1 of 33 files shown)


src/components/periodic-report/dto/periodic-report.dto.ts
```diff
@@ -53,6 +53,12 @@ class PeriodicReport extends Resource {
   @Field()
   readonly receivedDate: SecuredDateNullable;
 
+  @Field()
+  readonly skipped: boolean;
+
+  @Field()
+  readonly skippedReason?: string;
```
CarsonF, line 60: These should be secured, and unfortunately that means right now that all the role files need to be updated to allow these two new props.

CarsonF (changes requested):
- [ ] Please add a migration to add these two new props to all existing reports. See [AddPresetInventory](https://github.com/SeedCompany/cord-api-v3/blob/narrative-report-form/src/components/project/migrations/AddPresetInventory.migration.ts) as an example.
- [ ] Secure them

@sethmcknight Should skipped reason be required when skipped is being set to true?

sethmcknight:
> * [ ]  Please add a migration to add these two new props to all existing reports. See [AddPresetInventory](https://github.com/SeedCompany/cord-api-v3/blob/narrative-report-form/src/components/project/migrations/AddPresetInventory.migration.ts) as an example.
> * [ ]  Secure them
> 
> @sethmcknight Should skipped reason be required when skipped is being set to true?

@CarsonF @daniel-leod yes, we should required a reason.

CarsonF:
Since it's required, let's just drop the `skipped` boolean. It can be easily calculated from `skippedReason != null`

src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -39,6 +39,8 @@ export class PeriodicReportRepository extends DtoRepository(IPeriodicReport) {
       type: input.type,
       start: input.start,
       end: input.end,
+      skipped: input.skippedReason != null,
```
CarsonF, line 42: Err sorry I mean that in general. We don't have to write this to the DB as we can check it on read.

CarsonF (changes requested):
Needs migration still too

CarsonF (approved):
Thanks @daniel-leod!

Issue #2171, opened 2021-09-27
Props of Partner secured by sensitivity across relevant roles


elisabeth-sorrell:
As a {role}, I want {feature} so that {reason}.

PR #2172, opened 2021-09-27
Secure Partner based on Sensitivity written in static role definitions

elisabeth-sorrell:
Closes https://github.com/SeedCompany/cord-api-v3/issues/2171


CarsonF (approved):
Thanks @elisabeth-sorrell!

PR #2174, opened 2021-09-28
Rename file on new verision upload

tmpacifitech:
Part of https://github.com/SeedCompany/cord-field/issues/786

### DIFF
--- a/src/components/file/file.service.ts
+++ b/src/components/file/file.service.ts
@@ -265,9 +265,6 @@ export class FileService {
     }
 
     // Change the file's name to match the latest version name
-    // Since defined files are explicitly referenced by a named property
-    // a consistent name is not required and automatically updating it is more
-    // convenient for consumption.
     await this.rename({ id: fileId, name }, session);
 
     return await this.getFile(fileId, session);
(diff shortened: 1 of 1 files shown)


CarsonF (changes requested):
This is no longer needed
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/file/file.service.ts#L401-L405

tmpacifitech:
> This is no longer needed https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/file/file.service.ts#L401-L405

Fixed.

src/components/file/file.service.ts
```diff
@@ -264,6 +264,12 @@ export class FileService {
       await this.bucket.moveObject(`temp/${uploadId}`, uploadId);
     }
 
+    // Change the file's name to match the latest version name
+    // Since defined files are explicitly referenced by a named property
+    // a consistent name is not required and automatically updating it is more
+    // convenient for consumption.
```
CarsonF, line 270: This comment doesn't make sense here
```suggestion
```

CarsonF (changes requested):

CarsonF (approved):
🤘

PR #2182, opened 2021-10-01
Narrative Report Form API

### LINKED ISSUE #2062: Quarterly Report Form
As part of providing better reporting and communications with Investors, we need to capture the answers to a set of Narrative questions for each Language Engagement in a Project. These should be stored in a manner that allows them to be shared to other systems via an API. One location these may be shared is a set of Google Spreadsheets.

Questions:
Team Update

* Biographies of team members (1st report)
* Personal Updates of team members (all other reports)
* General update on team (Testimony, Biography, Personal reflection of a challenge or success, etc.)

Impact of Work

* Narrative, Stories, & Stats
* Story about partner interaction, achievement, challenge, gathering, etc.

Prayer Update

* Answers
* New Requests
* Update on past requests

Translation Example

* Reference
* Translation

Annual Highlight (Q3 only)

* What went well
* What didn’t go well
* What was learned
* Accomplishments
* Goals for next year

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466890361) by [Unito](https://www.unito.io)

CarsonF:
Part of #2062 

* [x] Moved Narrative Reports to be on Engagements
  * Existing reports on projects will stay there, but will not be synced with any date changes
  * [ ] Create migration to add status property to these. I thought about creating a new type two differentiate between these new "form" reports & legacy "file" reports. I couldn't come to terms with something workable there. I'm thinking we'll just move forward as a single type and those on Language Engagements will have questions & status and those elsewhere will not.
* [x] Add `NarrativeReport.status` to track draft/review/finalized workflow on questionnaire
* [x] Add `QuestionAnswer` module
  * Designed to allow question/answer pairs to be attached to any object.
  * Authorization logic is defined by events so any object in the system can respond to mutation requests. This keeps the module decoupled from usage. This is a new pattern and could/should be applied to Posts as well.
  * Queries are omitted from this module, and instead are be defined by the connector object. Such as `NarrativeReport.questions`.
  * Module defines a `QuestionBank` object type, but defers to other objects/modules to define them.
  * Question/Category types are `string` instead of `enums`. They will definitely change over time and this works for different banks.
* [x] Add `NarrativeReport.questionBank` to expose the questions available for a specific report.
  * These are hardcoded currently based on what's defined in #2062. These can be updated without consequences in future.
  * At some point we could expose an API to allow users to define question bank. The hard part there is conditional questions, such as those for specific reports.
  * [ ] Verify that questions being added to narrative reports are currently available in question bank.
         Even though I imagine the UI will only allow those available in question bank we should verify in API to prevent bad acting.
* [x] Defined authorization workflow based on narrative report status.
  * [ ] Define role for "answerer" role
* [ ] Notifications
  * [ ] Send notification to defined reviewers when a report changes to in review state. This should be simple.
  * [ ] Send notification at beginning of quarter for all narrative reports that are now due. This will be harder and probably needs scheduling & queuing logic to be implemented by me.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466906851) by [Unito](https://www.unito.io)


ParkeBrown:
Available for other developers to work on now

CarsonF:
We went with a different direction with this. Mainly merging NarrativeReports into ProgressReports. Bulk implementation was #2554

Issue #2169, opened 2021-09-24
Update project rules to be changeset aware


daniel-leod:
Project rules need to be changeset aware in order to update steps properly when a change request is created.

Issue #2037, opened 2021-07-20
Project > ChangeRequest > Getting "The project AeclOsV1kKx has never been in any of these previous steps: Active, ActiveChangedPlan" error when user change project status


iPiyushPatel:
**Roles:** Admin

**Describe the bug**
Project > ChangeRequest > Getting `The project AeclOsV1kKx has never been in any of these previous steps: Active, ActiveChangedPlan` error when user change project status

**Steps to Reproduce**
1. Login as Admin user
2. Create a project and Open created project
3. Update project status `Earlyconversation` to `Active` and  Add change request
4. Select `Time` type and Click on the `View` button
5. Update project status `Active`  to `Finalize Completion` and Refresh the page

**Expected Result**
Project status should be changed.
 
**Actual Result**
 Getting `The project AeclOsV1kKx has never been in any of these previous steps: Active, ActiveChangedPlan` error when user change project status

**Note:**
`Discuss Change to Plan` and `Discuss Termination` status also give similar types of errors.

**Stacktraces:**
```
{
  "errors": [
    {
      "message": "The project JtSP5RoQND8 has never been in any of these previous steps: Active, ActiveChangedPlan",
      "extensions": {
        "codes": [
          "Server"
        ],
        "status": 500,
        "code": "Server",
        "exception": {
          "message": "The project JtSP5RoQND8 has never been in any of these previous steps: Active, ActiveChangedPlan",
          "stacktrace": [
            "at ProjectRules.getMostRecentPreviousStep (components/project/project.rules.ts:1005:13)",
            "at ProjectRules.getStepRule (components/project/project.rules.ts:751:19)",
            "at ProjectRules.getNotifications (components/project/project.rules.ts:925:48)",
            "at ProjectStepChangedNotificationHandler.handle (components/project/handlers/step-changed-notification.handler.ts:33:24)",
            "at SyncEventBus.publish (core/events/event-bus.service.ts:24:7)",
            "at ProjectService.update (components/project/project.service.ts:378:5)",
            "at ProjectResolver.updateProject (components/project/project.resolver.ts:286:21)"
          ]
        }
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "updateProject"
      ]
    }
  ],
```

See video - 

https://user-images.githubusercontent.com/5212585/126274597-70a79c62-2dd3-4be6-a0e3-e7f75c0808cc.mp4

[...]

PR #2186, opened 2021-10-04
Project rule changeset aware

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2169


src/components/project/handlers/step-changed-notification.handler.ts
```diff
@@ -24,6 +24,7 @@ export class ProjectStepChangedNotificationHandler
   async handle(event: ProjectUpdatedEvent) {
     if (
       event.updated.step === event.previous.step ||
+      event.updated.changeset ||
```
zkhin, line 27: If the project step is updated in changesets, I don't think we need to send notifications and it will fix https://github.com/SeedCompany/cord-api-v3/issues/2037.

src/components/project/project.rules.ts
```diff
@@ -1010,14 +1032,28 @@ export class ProjectRules {
...
+            ]
+          : []),
         node('node', 'Project', { id }),
-        relation('out', '', 'step', INACTIVE),
+        relation('out', '', 'step', changeset ? undefined : INACTIVE),
         node('prop'),
       ])
+      .apply((q) =>
+        changeset
+          ? q.raw('WHERE NOT (changeset)-[:changeset {active:true}]->(prop)')
```
zkhin, line 1054: `getPreviousSteps` is updated being changeset aware. In changeset mode, it gets all(`ACTIVE` & `INACTIVE`) step props except the current active changeset step.

src/components/project/project.rules.ts
```diff
@@ -458,10 +462,11 @@ export class ProjectRules {
...
             {
-              to: await this.getMostRecentPreviousStep(id, [
-                ProjectStep.Active,
-                ProjectStep.ActiveChangedPlan,
-              ]),
+              to: await this.getMostRecentPreviousStep(
+                id,
+                [ProjectStep.Active, ProjectStep.ActiveChangedPlan],
+                changeset
+              ),
```
CarsonF, line 469: Could we do a wrapping function within this scope?
```ts
const mostRecentPreviousStep = (...steps: ProjectStep[]) =>
  this.getMostRecentPreviousSteep(id, steps, changeset);
```

zkhin: Do you mean like below? 🤔 
```
to: this.mostRecentPreviousStep(
     id,
    [ProjectStep.Active, ProjectStep.ActiveChangedPlan],
    changeset
),
```

src/components/project/handlers/step-changed-notification.handler.ts
zkhin: Yes, I agree with `ProjectChangesetFinalized` + `ProjectChangesetAfterFinalized`

PR #2208, opened 2021-10-12
Allow field region and primary location to be null when updated

### LINKED ISSUE #2206: Project > "Location" field value is not updated if the user clear texts from the "PRIMARY LOCATION" field
Roles All

**Describe the bug**
Project > `Location` field value is not updated if the user clear texts from the `PRIMARY LOCATION` field

**Steps to Reproduce**
1. Login as Admin user
2. Create a project and Open created project
3. Add `Location` and `FieldRegion`
4. Click on the  `Location` and `FieldRegion` button and clear the `Location` and `FieldRegion` textbox
5. Click on the `Save` button

**Expected Result**
`Location` values should be updated  as per `Start-End Dates` when the user clear texts from the `PRIMARY LOCATION` field

**Actual Result**
`Location` field value is not updated if the user clear texts from the `PRIMARY LOCATION` field

**Note:** same issue is for the `FieldRegion` field

See video -

https://user-images.githubusercontent.com/5212585/135454845-f5f06c07-b2ec-4570-ba18-f72305cf4556.mp4

**FYI**@ParkeBrown @michaelmarshall

cord-field# 684cfb8
CordAPIV3# 32bf73a

daniel-leod:
Fixes #2206

### DIFF
--- a/src/components/project/project.repository.ts
+++ b/src/components/project/project.repository.ts
@@ -219,58 +219,56 @@ export class ProjectRepository extends CommonRepository {
     });
   }
 
-  async updateLocation(input: UpdateProject, createdAt: DateTime) {
-    const query = this.db
+  async updateLocation(id: ID, primaryLocationId: ID | null) {
+    await this.db
       .query()
-      .match(node('project', 'Project', { id: input.id }))
-      .match(node('location', 'Location', { id: input.primaryLocationId }))
-      .with('project, location')
-      .limit(1)
+      .match(node('project', 'Project', { id: id }))
+      .match(
+        primaryLocationId
+          ? [node('location', 'Location', { id: primaryLocationId })]
+          : []
+      )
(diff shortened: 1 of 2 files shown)


elisabeth-sorrell (changes requested):
I tried running your branch and updating with an empty region and location, but nothing seemed to happen.....?

elisabeth-sorrell:
> @elisabeth-sorrell I'm going to check right now. Are you updating those fields through the UI? If so please use run this [branch ](https://github.com/SeedCompany/cord-field/pull/959)too

Yes, I had tried with the UI, but not with this branch. Trying that now

elisabeth-sorrell (approved):

CarsonF (approved):
I flattened the double if statements, merged the db query methods, and re-apply the check that `limit 1` clause was doing previously.

PR #2209, opened 2021-10-12
Add Language.firstScripture

CarsonF:
This is allows this to be more easily accessed without having to go through all of the language's projects and all of the project's engagements.

Can be accessed like
```gql 
fragment Example on Language {
  firstScripture {
    canRead
    value {
      hasFirst
      ...on InternalFirstScripture {
        engagement {
          id
        }
      }
    }
  }
}
```

### DIFF
--- a/src/components/language/dto/language.dto.ts
+++ b/src/components/language/dto/language.dto.ts
@@ -155,6 +155,9 @@ export class Language extends Resource {
   @Field()
   readonly hasExternalFirstScripture: SecuredBoolean;
 
+  // Internal First Scripture == true via this engagement
+  readonly firstScriptureEngagement?: ID;
+
   @Field()
   readonly tags: SecuredTags;
 
--- a/src/components/language/dto/first-scripture.dto.ts
+++ b/src/components/language/dto/first-scripture.dto.ts
@@ -1,0 +1,44 @@
+import { Field, InterfaceType, ObjectType } from '@nestjs/graphql';
+import { has, ID, SecuredProperty } from '../../../common';
+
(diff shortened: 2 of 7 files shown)


src/components/language/dto/language.dto.ts
```diff
@@ -155,6 +155,9 @@ export class Language extends Resource {
   @Field()
   readonly hasExternalFirstScripture: SecuredBoolean;
```
elisabeth-sorrell, line 156: is this still necessary? would the `ExternalFirstScripture` in the first-scripture.dto cover for this?

src/components/language/dto/language.dto.ts
```diff
@@ -155,6 +155,9 @@ export class Language extends Resource {
...
 
+  // Internal First Scripture == true via this engagement
+  readonly firstScriptureEngagement?: ID;
```
elisabeth-sorrell, line 159: does this need to be a `Secured<ID>`? I was noticing that `securedProps` in the languageService doesn't pick this up as being secured.

elisabeth-sorrell (changes requested):
One other thing, if hasExternalFirstScripture is not included in the role defs with `canRead` and `canEdit`, no roles can access this prop...... just thought I'd mention it since I've been buried in security for so long ;)

src/components/language/dto/language.dto.ts
CarsonF: I was just basing it on the `hasExternalFirstScripture`. Since this is based on that and one other thing, it felt like it was sufficient.

PR #2212, opened 2021-10-13
Get rid of security groups for listing fieldZone, film, fundingAccount, and language

elisabeth-sorrell:
Also, refactor the language hydrate() function to check all associated projects for lowest sensitivity and return an `effectiveSensitivity` that is used to determine security for the language component. Needed `effectiveSensitivity` because we also need to return the Language's sensitivity prop.


CarsonF (approved):
🎉

Issue #2205, opened 2021-10-11
Changesets | Language


CarsonF:
Update language module to be changeset aware

PR #2222, opened 2021-10-16
Language changeset aware

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2205

### DIFF
--- a/test/language-changeset-aware.e2e-spec.ts
+++ b/test/language-changeset-aware.e2e-spec.ts
@@ -1,0 +1,135 @@
+import { gql } from 'apollo-server-core';
+import { Connection } from 'cypher-query-builder';
+import * as faker from 'faker';
+import { Powers, Role } from '../src/components/authorization';
+import { ProjectStep } from '../src/components/project';
+import {
+  approveProjectChangeRequest,
+  createFundingAccount,
+  createLanguage,
+  createLanguageEngagement,
+  createLocation,
+  createProject,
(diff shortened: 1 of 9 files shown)


test/language-changeset-aware.e2e-spec.ts
```diff
@@ -0,0 +1,135 @@
...
+          changeset: changeset.id,
+        },
+      }
+    );
+
+    // read language without changeset
+    let result = await readLanguage(app, language.id);
+    expect(result.language.name.value === language.name.value);
+    // read language with changeset
+    result = await readLanguage(app, language.id, changeset.id);
```
zkhin, line 129: I am not sure why I am getting `Could not find language` exception in [ordered-data-loader](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/core/data-loader/ordered-data-loader.ts#L84).
FYI, I can get correct `Language` props in language readOne [resolver](https://github.com/SeedCompany/cord-api-v3/pull/2222/files#diff-72d934c30019fd8ce38c2f5e3ee139f1f621d344f3999934775c5fb5fb1dc9deR62) without any errors.

CarsonF (approved):
Thanks @zkhin!

> @CarsonF I think `Language` should be changeset aware only for reading & updating when it's connected to `LanguageEngagement`.

I think there's potential for languages to need their own "change requests", but that can come later. I agree this is good enough for now.

PR #2230, opened 2021-10-21
Deprecate Song and LiteracyMaterial in favor of EthnoArts

### LINKED ISSUE #2224: G&P | Deprecate Song + Literacy Material for EthnoArts
- [x] Update all Songs & Literacy Materials resolvers to point to EthnoArts service/loader.
- [x] Mark all Songs & Literacy Materials resolvers as deprecated in favor of EthnoArts
- [x] Add a DB migration in EthnoArts the relabels all Songs & Literacy Materials nodes to have EthnoArt label instead.

daniel-leod:
Closes #2224


src/components/ethno-art/migrations/RenameSongAndLiteracyMaterialToEthnoArts.migration.ts
```diff
@@ -0,0 +1,39 @@
+import { node, relation } from 'cypher-query-builder';
+import { BaseMigration, Migration } from '../../../core';
+import { ACTIVE } from '../../../core/database/query';
+
+@Migration('2021-10-20T16:28:16')
+export class RenameSongAndLiteracyMaterialToEthnoArts extends BaseMigration {
```
CarsonF, line 6: This looks good! I think you are missing the export in the migrations/index.ts file which connects it to the module

src/components/literacy-material/literacy-material.resolver.ts
```diff
@@ -2,79 +2,73 @@ import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
...
+/** @deprecated Use EthnoArtResolver instead */
 @Resolver(LiteracyMaterial)
 export class LiteracyMaterialResolver {
-  constructor(
-    private readonly literacyMaterialService: LiteracyMaterialService
-  ) {}
+  constructor(private readonly ethnoArtService: EthnoArtService) {}
 
-  @Query(() => LiteracyMaterial, {
+  @Query(() => EthnoArt, {
```
CarsonF, line 22: I was trying to avoid breaking changes here. I think we should be able to return EthnoArt objects as "Songs" since they have the same shape. So these resolvers should be able to get by without any schema changes

src/components/literacy-material/literacy-material.resolver.ts
```diff
@@ -2,79 +2,73 @@ import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
...
+  EthnoArtListInput,
+  EthnoArtListOutput,
+  EthnoArtLoader,
+  EthnoArtService,
+  UpdateEthnoArtInput,
+  UpdateEthnoArtOutput,
+} from '../ethno-art';
+import { LiteracyMaterial } from './dto';
 
+/** @deprecated Use EthnoArtResolver instead */
```
CarsonF, line 17: The Query/Mutation/ResolveField decorators have a `deprecatedReason` property. Can you use this instead? This will communicate in the external schema that these are deprecated.

Issue #2204, opened 2021-10-11
Publish events even when mutating for a changeset


CarsonF:
We need to let the event handlers decide themselves if they should take action or not for changeset mutations.
It may just be the engagement events that do this right now?

The root goal is to fix https://github.com/SeedCompany/cord-field/issues/933

PR #2232, opened 2021-10-26
Engagement ceremony changeset

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2204 https://github.com/SeedCompany/cord-field/issues/933

### DIFF
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -366,10 +366,6 @@ export class EngagementService {
       view
     )) as LanguageEngagement;
 
-    if (changeset) {
-      return updated;
-    }
-
     const engagementUpdatedEvent = new EngagementUpdatedEvent(
       updated,
       object,
--- a/src/components/engagement/engagement.service.ts
+++ b/src/components/engagement/engagement.service.ts
@@ -127,9 +127,7 @@ export class EngagementService {
       session,
       viewOfChangeset(changeset)
     )) as LanguageEngagement;
-    if (changeset) {
-      return languageEngagement;
-    }
+
     const event = new EngagementCreatedEvent(
       languageEngagement,
       input,
--- a/src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts
+++ b/src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts
@@ -1,8 +1,10 @@
 import { node, relation } from 'cypher-query-builder';
-import { ID, ServerException } from '../../../common';
+import { union } from 'lodash';
+import { ID, ServerException, Session } from '../../../common';
 import {
   DatabaseService,
   EventsHandler,
+  IEventBus,
   IEventHandler,
   ILogger,
   Logger,
@@ -12,6 +14,8 @@ import { commitChangesetProps } from '../../changeset/commit-changeset-props.que
 import { rejectChangesetProps } from '../../changeset/reject-changeset-props.query';
 import { ProjectChangeRequestStatus } from '../../project-change-request/dto';
 import { ProjectChangesetFinalizedEvent } from '../../project-change-request/events';
+import { EngagementService } from '../engagement.service';
+import { EngagementUpdatedEvent } from '../events';
 
 type SubscribedEvent = ProjectChangesetFinalizedEvent;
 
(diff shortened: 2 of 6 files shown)


src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
```diff
@@ -40,6 +40,10 @@ export class SyncProgressReportToEngagementDateRange
       event: event.constructor.name,
     });
 
+    if (event instanceof EngagementCreatedEvent && event.engagement.changeset) {
+      return;
+    }
```
CarsonF, line 45: When I changeset is approved is an engagement updated event published to trigger progress reports to be created for the new engagement?

CarsonF (changes requested):
I rebased on develop for you. I was going to merge but then I noticed that engagement update events are still being skipped as well. Can you apply same changes there and confirm handlers are good or not?

src/components/engagement/engagement.service.ts
```diff
@@ -366,10 +366,6 @@ export class EngagementService {
       view
     )) as LanguageEngagement;
 
-    if (changeset) {
```
zkhin, line 371: > I rebased on develop for you. I was going to merge but then I noticed that engagement update events are still being skipped as well. Can you apply same changes there and confirm handlers are good or not?

src/components/engagement/handlers/set-initial-end-date.handler.ts
```diff
@@ -81,22 +94,22 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
...
+    changeset?: ID
   ) {
     const updateInput = {
-      id: engagement.id,
       initialEndDate: initialEndDate || null,
     };
-    // TODO: Refactor to call repository directly instead of engagementService methods
     if (engagement.__typename === 'LanguageEngagement') {
-      return await this.engagementService.updateLanguageEngagement(
+      await this.engagementRepo.updateLanguageProperties(
```
zkhin, line 103: Instead of using the update service method (prevent UpdateEngagementEvent triggering), I used the repo method.

src/components/periodic-report/handlers/sync-progress-report-to-engagement.handler.ts
zkhin: I skip processing progress reports in changeset mode for both Created/Updated events.

src/components/engagement/handlers/apply-finalized-changeset-to-engagement.handler.ts
```diff
@@ -78,13 +86,25 @@ export class ApplyFinalizedChangesetToEngagement
...
+          'collect(engagementId) as engagementIds'
         )
-        .return('project')
-        .run();
+        .first();
 
       // Remove deleting engagements
       await this.removeDeletingEngagements(changesetId);
+
+      // Trigger sync-progress-report-to-engagement handler
```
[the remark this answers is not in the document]
CarsonF: I think this is fine. Each handler should be idempotent right?

src/components/engagement/handlers/set-initial-end-date.handler.ts
CarsonF: Fine with me, less work to do 👍

src/components/engagement/handlers/set-initial-end-date.handler.ts
```diff
@@ -52,10 +63,15 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
...
         engagement,
         initialEndDate,
-        event.session
+        engagement.changeset
+      );
+      const updatedEngagement = await this.engagementService.readOne(
+        engagement.id,
+        event.session,
+        viewOfChangeset(engagement.changeset)
       );
```
CarsonF, line 75: Can't this read call be replaced by just setting the property directly?
```suggestion
      const updatedEngagement = {
        ...engagement,
        initialEndDate: {
           ...engagement.initialEndDate,
           value: initialEndDate,
        },
      };
```

src/components/engagement/handlers/set-initial-end-date.handler.ts
```diff
@@ -26,6 +34,9 @@ export class SetInitialEndDate implements IEventHandler<SubscribedEvent> {
 
     const engagement = 'engagement' in event ? event.engagement : event.updated;
 
+    if (event instanceof EngagementCreatedEvent && engagement.changeset) {
+      return;
+    }
```
CarsonF, line 39: I don't think there's a need to prevent this from happening in the changeset...It is just a property change like anything a user would directly change.

CarsonF (commented):
Looks good @zkhin. Just have a couple remaining notes.

CarsonF (approved):
🎉

PR #2254, opened 2021-11-02
Fix EthnoArt search

### LINKED ISSUE #2250: G&P | Can't create new `EthnoArt` Goals. These should be able to be created just like they can for `Story` Goals

### LINKED ISSUE #2253: EthnoArt search is not working

daniel-leod:
Fixes #2253 
Fixes #2250
Fixes [#996](https://github.com/SeedCompany/cord-field/issues/996)


CarsonF (changes requested):
We need to change the logic to be _only_ EthnoArt types. Otherwise other types could bypass security as well.

PR #2255, opened 2021-11-02
Allow otherGoals to be tracked if methodology is not selected

### LINKED ISSUE #2249: G&P | Goals of type `Other` can't be tracked if Methodology is not selected
When a user creates an `Other` Goal, selects the Progress Measurement and Step of Completed they are able to create the Goal. 

When the user visits the Goal Details to view/track progress they get a message that "Product does not have any steps".

If a user selecta a Methodology the app functions as expected and they can see the Step and record progress.

We need to allow users to view the step and track progress without a Methodology being selected.

daniel-leod:
Fixes #2249

### DIFF
--- a/src/components/product/product.service.ts
+++ b/src/components/product/product.service.ts
@@ -110,7 +110,9 @@ export class ProductService {
       type,
       methodology: input.methodology,
     });
-    const steps = input.methodology
+    const steps = type.includes('OtherProduct')
+      ? availableSteps
+      : input.methodology
       ? intersection(availableSteps, input.steps)
       : [];
 


CarsonF (commented):
Methodology is not a ProductStep, nor is it an array so this doesn't make sense.
The actual problem is Completed step is not selected, but it should be in the UI.
We'll leave that for https://github.com/SeedCompany/cord-field/issues/992

CarsonF:
Actually I’m looking again and it does make sense. I guess I needed more coffee, sorry about that.

PR #2256, opened 2021-11-02
Add Partner.projects

bryanjnelson:


src/components/partner/partner.service.ts
```diff
@@ -29,6 +35,7 @@ export class PartnerService {
   constructor(
     @Logger('partner:service') private readonly logger: ILogger,
     @Inject(forwardRef(() => AuthorizationService))
+    private readonly projectService: ProjectService,
```
CarsonF, line 38: You split the Inject decorator and the authorizationService lines here. So it's injecting AuthorizationService for the projectService property.

src/components/partner/partner.service.ts
```diff
@@ -186,6 +193,48 @@ export class PartnerService {
     return await mapListResults(results, (dto) => this.secure(dto, session));
   }
 
+  async listProjects(
+    partner: Partner,
+    input: ProjectListInput,
+    session: Session
+  ): Promise<SecuredProjectList> {
```
CarsonF, line 200: Let's forward to ProjectService.list here with a new partner id filter. This will allow us to load everything in one query and keep sorting & pagination working.

[Here's](https://github.com/SeedCompany/cord-api-v3/blob/66c8ee3447b8593f3f21ef4cf0f7cf97e245c0eb/src/components/project/project.service.ts#L453-L483) an example for listing ProjectMembers of a Project.

CarsonF (changes requested):

src/components/partner/partner.service.ts
bryanjnelson: Ugh.  Thanks.

src/components/partner/partner.service.ts
bryanjnelson: I modeled this after what is happening on the Language Detail page.  I'm unclear on what you mean here.  Do you mean that instead of even adding/calling this new listProjects method in the PartnerService I should instead call ProjectService.list?  Or that from within this PartnerService.listProjects I should call ProjectService.list?

CarsonF: The latter. The example I linked above does the latter as well.

src/components/partner/partner.service.ts
```diff
@@ -30,6 +36,8 @@ export class PartnerService {
     @Logger('partner:service') private readonly logger: ILogger,
     @Inject(forwardRef(() => AuthorizationService))
     private readonly authorizationService: AuthorizationService,
+    @Inject(forwardRef(() => ProjectService))
```
CarsonF, line 39: Only do this if it's necessary. If it'll inject without it, omit it.

src/components/partner/partner.service.ts
```diff
@@ -186,6 +194,28 @@ export class PartnerService {
...
+    input: ProjectListInput,
+    session: Session
+  ): Promise<SecuredProjectList> {
+    const projectListOutput = await this.projectService.list(
+      { ...input, filter: { ...input.filter, partnerId: partner.id } },
+      session
+    );
+
+    const perms = await this.authorizationService.getPermissions({
+      resource: IProject,
```
CarsonF, line 208:
```suggestion
      resource: Partner,
```

src/components/partner/partner.service.ts
```diff
@@ -186,6 +194,28 @@ export class PartnerService {
...
+      sessionOrUserId: session,
+    });
+
+    return {
+      ...projectListOutput,
+      canRead: perms.member.canRead,
```
CarsonF, line 214:
```suggestion
      canRead: perms.projects.canRead,
```
But there's still some confusion between who owns the permission etc. since there are multiple ways to access this data/relation.

src/components/project/dto/list-projects.dto.ts
```diff
@@ -87,6 +87,12 @@ export abstract class ProjectFilters {
     description: 'Only projects that are (not) in the "Preset Inventory"',
   })
   readonly presetInventory?: boolean;
+
+  @Field({
+    nullable: true,
+    description: 'Only project associated with a particular partner',
+  })
```
CarsonF, line 94:
```suggestion
```
We want this to be internal only so no decorator.

CarsonF (changes requested):
Still need to implement filter here https://github.com/SeedCompany/cord-api-v3/blob/66c8ee3447b8593f3f21ef4cf0f7cf97e245c0eb/src/components/project/query.helpers.ts#L17-L17

src/components/partner/partner.service.ts
bryanjnelson: Yeah, I get a circular dependency without it.  ☹️

src/components/partner/partner.service.ts
```diff
@@ -205,14 +204,14 @@ export class PartnerService {
...
-      resource: IProject,
+      resource: Partner,
       sessionOrUserId: session,
     });
 
     return {
       ...projectListOutput,
-      canRead: perms.member.canRead,
-      canCreate: perms.member.canEdit,
+      canRead: perms.projects.canRead,
```
bryanjnelson, line 213: @CarsonF - This doesn't seem to be working as the read permissions here are coming back as false.  😶   Any ideas here?

src/components/project/query.helpers.ts
```diff
@@ -80,6 +80,18 @@ export const projectListFilter = (filter: ProjectFilters) => (query: Query) => {
     ]);
   }
 
+  if (filter.partnerId != null) {
+    query
```
bryanjnelson, line 84: @CarsonF - Although this is working in my test cases, I'm not sure if this is the correct way to do this?

src/components/partner/partner.service.ts
CarsonF: Need to grant access in the role files
https://github.com/SeedCompany/cord-api-v3/blob/e618a443ed797a7583bd87704842baead638de37/src/components/authorization/roles

src/components/project/query.helpers.ts
CarsonF: I'd prefer to use a where clause path condition, like presetInventory above. Technically both work, but:
- using a where clause ensures rows don't expand.
   Row expansion happens if you like match multiple partners for a single project, so each row is now the combination of a unique project & partner leading to duplicate rows with the same project. This shouldn't happen here but it's something to keep in mind.
-  ensures clause siblings are ok.
    Like for example, you can't end a subquery and go straight into a where clause without having a `with` statement in-between.
https://github.com/SeedCompany/cord-api-v3/blob/e618a443ed797a7583bd87704842baead638de37/src/components/project/query.helpers.ts#L90-L92
   I think you can't have a where clause go straight into another match clause either. That was a problem here previously due to all of these being dynamic.
- Ultimately it's more idiomatic. We're trying to "filter data down based on these conditions" not "match related data".

src/components/project/query.helpers.ts
```diff
@@ -80,6 +80,18 @@ export const projectListFilter = (filter: ProjectFilters) => (query: Query) => {
...
+      .match([node('partner', 'Partner', { id: filter.partnerId })])
```
CarsonF, line 85: This doesn't need to be split from the reference below, it can be represented as a single path. Note that when using in where clause path condition you can't add variables to scope, so `node('', 'Partner', ...)`

src/components/partner/partner.service.ts
bryanjnelson: @CarsonF - I think I did this correctly.  Can you check it out?

CarsonF (commented):
LGTM! Would you mind rebasing on develop and squash down to a single commit with a message that represents the change? Something like "Expose partner's projects"

bryanjnelson:
> LGTM! Would you mind rebasing on develop and squash down to a single commit with a message that represents the change? Something like "Expose partner's projects"

Will do, when it's time.  Something isn't right with the query though...I'm not getting any results back.  ☹️

src/components/project/query.helpers.ts
```diff
@@ -80,6 +80,16 @@ export const projectListFilter = (filter: ProjectFilters) => (query: Query) => {
     ]);
   }
 
+  if (filter.partnerId != null) {
+    conditions.partnerId = path([
+      node('node'),
+      relation('out', '', 'partnership', ACTIVE),
+      node('', 'Partnership'),
+      relation('out', '', 'partner', ACTIVE),
+      node('', 'Partner', { value: filter.partnerId }),
```
CarsonF, line 89:
```suggestion
      node('', 'Partner', { id: filter.partnerId }),
```

bryanjnelson: 🤦🏼  - That was it; thanks!

src/components/partner/partner.service.ts
```diff
@@ -186,6 +193,28 @@ export class PartnerService {
...
+    session: Session
+  ): Promise<SecuredProjectList> {
+    const projectListOutput = await this.projectService.list(
+      { ...input, filter: { ...input.filter, partnerId: partner.id } },
+      session
+    );
+
+    const perms = await this.authorizationService.getPermissions({
+      resource: Partner,
+      sessionOrUserId: session,
```
CarsonF, line 208:
```suggestion
      sessionOrUserId: session,
      dto: partner,
```
Need to pass in partner object for it's sensitivity and scoped roles

bryanjnelson: @CarsonF - Makes sense, but I added that and it is still coming back false.

CarsonF (changes requested):
Since we didn't end up using new authorization logic would you mind removing those changes to the roles and model.db files?

CarsonF (approved):
Thanks @bryanjnelson! 🌮

PR #2269, opened 2021-11-08
Pass changeset internally to transitions logic

CarsonF:


CarsonF:
I don't think this is a relevant breaking changes. The changeset arg was never referenced.

PR #2285, opened 2021-11-16
Change Requests | Capture & Expose Status Change History

### LINKED ISSUE #2273: Change Requests | Capture & Expose Status Change History
Ok so one step at a time. I think we need this new schema:

type Mutation {
  transitionProject(input: ProjectTransitionInput!): ProjectTransitionOutput!
}
input ProjectTransitionInput {
  id: ID!
  changeset: ID
  step: ProjectStep!
  comment: String
}
type ProjectTransitionOutput {
  project: Project!
}

type SecuredProjectStep {
  history: [ProjectStepChange!]!
}
type ProjectStepChange {
  step: ProjectStep!
  by: User!
  comment: String
}
Does transitionProject() or changeProjectStep() make more sense?
These should fully work with changesets
Maybe we deprecate updateProject.step for this new mutation as well?
Once this is done we can talk about integrating changeset status and project step workflows.

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2273

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466919003) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/project/dto/project-transition.dto.ts
+++ b/src/components/project/dto/project-transition.dto.ts
@@ -1,0 +1,53 @@
+import { Field, InputType, ObjectType } from '@nestjs/graphql';
+import { keys as keysOf } from 'ts-transformer-keys';
+import {
+  DbLabel,
+  ID,
+  IdField,
+  parentIdMiddleware,
+  SecuredProps,
+} from '../../../common';
+import { User } from '../../user/dto';
+import { IProject, Project } from './project.dto';
+import { ProjectStep } from './step.enum';
+
+@InputType()
+export abstract class ProjectTransitionInput {
+  @IdField({ nullable: false })
+  id: ID;
+
+  @IdField({ nullable: true })
+  changeset?: ID;
+
+  @Field(() => ProjectStep)
+  step: ProjectStep;
+
+  @Field(() => String, { nullable: true })
+  comment?: string;
+}
+
+@ObjectType()
+export abstract class ProjectTransitionOutput {
+  @Field(() => IProject)
+  readonly project: Project;
+}
+
+export abstract class ProjectStepChange {
+  static readonly Props: string[] = keysOf<ProjectStepChange>();
+  static readonly SecuredProps: string[] =
+    keysOf<SecuredProps<ProjectStepChange>>();
+  static readonly Relations = {
+    by: User,
+  };
+
+  @Field({
+    middleware: [parentIdMiddleware],
+  })
+  @DbLabel('ProjectStep')
+  readonly step: ProjectStep;
+
+  @Field(() => String)
+  readonly comment: string | null;
+
+  readonly user: ID | null;
+}
--- a/src/components/project/dto/project-transition.dto.ts
+++ b/src/components/project/dto/project-transition.dto.ts
@@ -32,6 +32,7 @@ export abstract class ProjectTransitionOutput {
   readonly project: Project;
 }
 
+@ObjectType()
 export abstract class ProjectStepChange {
   static readonly Props: string[] = keysOf<ProjectStepChange>();
   static readonly SecuredProps: string[] =
--- a/src/components/project/dto/project-transition.dto.ts
+++ b/src/components/project/dto/project-transition.dto.ts
@@ -13,7 +13,7 @@ import { ProjectStep } from './step.enum';
 
 @InputType()
 export abstract class ProjectTransitionInput {
-  @IdField({ nullable: false })
+  @IdField()
   id: ID;
 
   @IdField({ nullable: true })
@@ -46,7 +46,7 @@ export abstract class ProjectStepChange {
   @DbLabel('ProjectStep')
   readonly step: ProjectStep;
 
-  @Field(() => String)
+  @Field(() => String, { nullable: true })
   readonly comment: string | null;
 
   readonly user: ID | null;
--- a/src/components/project/dto/project-transition.dto.ts
+++ b/src/components/project/dto/project-transition.dto.ts
@@ -1,12 +1,6 @@
 import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import { keys as keysOf } from 'ts-transformer-keys';
-import {
-  DbLabel,
-  ID,
-  IdField,
-  parentIdMiddleware,
-  SecuredProps,
-} from '../../../common';
+import { DbLabel, ID, IdField, Secured, SecuredProps } from '../../../common';
 import { User } from '../../user/dto';
 import { IProject, Project } from './project.dto';
 import { ProjectStep } from './step.enum';
@@ -41,14 +35,12 @@ export abstract class ProjectStepChange {
     by: User,
   };
 
-  @Field({
-    middleware: [parentIdMiddleware],
-  })
+  @Field(() => ProjectStep)
   @DbLabel('ProjectStep')
   readonly step: ProjectStep;
 
   @Field(() => String, { nullable: true })
-  readonly comment: string | null;
+  readonly comment?: string | null;
 
-  readonly user: ID | null;
+  readonly user?: Secured<ID | null>;
 }
--- a/src/components/project/dto/update-project.dto.ts
+++ b/src/components/project/dto/update-project.dto.ts
@@ -15,7 +15,6 @@ import {
 } from '../../../common';
 import { ReportPeriod } from '../../periodic-report/dto';
 import { IProject, Project } from './project.dto';
-import { ProjectStep } from './step.enum';
 
 @InputType()
 export abstract class UpdateProject {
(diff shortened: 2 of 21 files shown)


src/components/project/project.service.ts
```diff
@@ -380,6 +381,47 @@ export class ProjectService {
...
+    // verify new project step
+    await this.projectRules.verifyStepChange(
+      input.id,
+      session,
+      changes.step,
+      input.changeset
+    );
+
+    // update project step prop
+    const result = await this.repo.updateProperties(
```
zkhin, line 409: Need to replace with `this.repo. addProjectStep`

src/components/project/project.service.ts
```diff
@@ -380,6 +381,47 @@ export class ProjectService {
...
+      currentProject,
+      { step: changes.step },
+      input.changeset
+    );
+
+    const event = new ProjectUpdatedEvent(
```
zkhin, line 415: I am thinking it will be worth creating a new event only for Project Step update (`ex: ProjectStepUpdatedEvent`)

zkhin:
TODO list

1. Read ProjectStepChange history 
2. Read current ProjectStep

src/components/project/dto/project-transition.dto.ts
```diff
@@ -0,0 +1,53 @@
...
+  parentIdMiddleware,
+  SecuredProps,
+} from '../../../common';
+import { User } from '../../user/dto';
+import { IProject, Project } from './project.dto';
+import { ProjectStep } from './step.enum';
+
+@InputType()
+export abstract class ProjectTransitionInput {
+  @IdField({ nullable: false })
```
CarsonF, line 16:
```suggestion
  @IdField()
```

src/components/project/dto/project-transition.dto.ts
```diff
@@ -0,0 +1,53 @@
...
+    by: User,
+  };
+
+  @Field({
+    middleware: [parentIdMiddleware],
+  })
+  @DbLabel('ProjectStep')
+  readonly step: ProjectStep;
+
+  @Field(() => String)
```
CarsonF, line 49:
```suggestion
  @Field(() => String, { nullable: true })
```

src/components/project/project.service.ts
CarsonF: Would that actually simplify any of the handlers? I think not because they still have to check previous/current step to determine if they should run.

CarsonF (commented):
Looks good so far

src/components/project/dto/update-project.dto.ts
```diff
@@ -54,8 +53,8 @@ export abstract class UpdateProject {
   @DateField({ nullable: true })
   readonly estimatedSubmission?: CalendarDate | null;
 
-  @Field(() => ProjectStep, { nullable: true })
-  readonly step?: ProjectStep;
+  // @Field(() => ProjectStep, { nullable: true })
```
zkhin, line 56: What is the best way to deprecate the `UpdateProject.step`?

src/components/project/project.service.ts
```diff
@@ -637,6 +689,11 @@ export class ProjectService {
     };
   }
 
+  async listStepChangeHistory(_id: ID, _changeset?: ID) {
+    // TODO
```
zkhin, line 693: TODO

src/components/project/dto/update-project.dto.ts
CarsonF: Add `deprecatedReason` to the field options

src/components/project/dto/project-transition.dto.ts
```diff
@@ -0,0 +1,46 @@
+import { Field, InputType, ObjectType } from '@nestjs/graphql';
+import { keys as keysOf } from 'ts-transformer-keys';
+import { DbLabel, ID, IdField, Secured, SecuredProps } from '../../../common';
+import { User } from '../../user/dto';
+import { IProject, Project } from './project.dto';
+import { ProjectStep } from './step.enum';
+
+@InputType()
+export abstract class ProjectTransitionInput {
```
CarsonF, line 9: I was hoping to resolve the "transition" vs "step change" phrase before but it didn't happen.
I talked with @sethmcknight more and I think "transition" is too ambiguous.
Can we change all of those to StepChange? That makes it consistent with other half of the new types.

src/components/project/project-step-change.resolver.ts
```diff
@@ -0,0 +1,19 @@
...
+@Resolver(ProjectStepChange)
+export class ProjectStepTransitionResolver {
+  @ResolveField(() => SecuredUser)
+  async user(
+    @Parent() stepChange: ProjectStepChange,
+    @Loader(UserLoader) users: LoaderOf<UserLoader>
+  ): Promise<SecuredUser | null> {
+    if (!stepChange.user) {
+      return null;
+    }
```
CarsonF, line 16: Secured types should not be nullable. Exposing null when there is no user, means they know that I value is missing even if they aren't supposed to be able to read it. That security probably doesn't matter much here, but I'm just trying to be consistent.
```suggestion
  ): Promise<SecuredUser> {
```
`mapSecuredValue` handles the value being null too.

src/components/project/project-step.resolver.ts
```diff
@@ -26,6 +34,22 @@ export class ProjectStepResolver {
...
   }
 
+  @ResolveField(() => [ProjectStepChange], {
+    description: 'The history of project steps.',
+  })
+  async history(
+    @Parent() step: SecuredProjectStep & { parentId: ID; changeset?: ID },
+    @AnonSession() _session: Session
+  ): Promise<ProjectStepChange[]> {
+    if (!step.canRead || !step.canEdit || !step.value) {
```
CarsonF, line 44:
```suggestion
    if (!step.canRead || !step.value) {
```
You don't need edit permission to view the list

src/components/project/project.repository.ts
```diff
@@ -348,6 +376,90 @@ export class ProjectRepository extends CommonRepository {
...
+      .orderBy('stepChange.createdAt', 'DESC')
+      .return<ProjectStepChange>([
+        'step.value as step',
+        'comment',
+        'user.id as user',
+      ]);
+
+    const result = await query.run();
+    const changes: ProjectStepChange[] = [];
+    result.map((change) => changes.push(change));
```
CarsonF, line 459: What is this doing?

src/components/project/project.repository.ts
```diff
@@ -348,6 +376,90 @@ export class ProjectRepository extends CommonRepository {
...
+      .first();
+    if (!result) {
+      throw new ServerException('Failed to create Project Transition');
+    }
+    return result.id;
+  }
+
+  async listStepChangeHistory(
+    id: ID,
+    _changeset?: ID
```
CarsonF, line 425: `addProjectStep` handles changeset so this should probably too right?

src/components/project/project.service.ts
```diff
@@ -293,19 +293,12 @@ export class ProjectService {
...
-        changeset
-      );
-    }
-
     const {
       primaryLocationId,
       marketingLocationId,
       fieldRegionId,
+      step, // skip step, use this.updateStep
+      status, // skip status, use this.updateStep
```
CarsonF, line 301: change step via updateProject is deprecated but we still need to support it for a little bit.

src/components/project/dto/project-transition.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  @Field(() => ProjectStep)
+  @DbLabel('ProjectStep')
+  readonly step: ProjectStep;
+
+  @Field(() => String, { nullable: true })
+  readonly comment?: string | null;
+
+  readonly user?: Secured<ID | null>;
```
CarsonF, line 45:
```suggestion
  readonly user: Secured<ID | null>;
```

CarsonF (changes requested):

CarsonF:
@zkhin Can I scope creep this more? I forgot we need the datetime the changed happened at. Maybe the field is just called `at`? or maybe `createdAt` to be consistent (though doesn't feel as natural)?

zkhin:
> @zkhin Can I scope creep this more? I forgot we need the datetime the changed happened at. Maybe the field is just called `at`? or maybe `createdAt` to be consistent (though doesn't feel as natural)?

Added `createdAt` field to the `ProjectStepChange`

src/components/project/project.repository.ts
```diff
@@ -121,6 +124,23 @@ export class ProjectRepository extends CommonRepository {
           relation('out', '', 'owningOrganization', ACTIVE),
           node('organization', 'Organization'),
         ])
+        // match project transition step
+        .optionalMatch([
```
CarsonF, line 128: I believe both of these matches should be required right?

CarsonF (commented):
Since we no longer look at the Project->step property in the database, we need a migration to migrate at least the current property to a step change.

I haven't tested this yet, but I will next.

sethmcknight:
Too old

PR #2286, opened 2021-11-16
Tweak email validator

### LINKED ISSUE #2271: forgotPassword > Returned success response when user leave blank "Email" field
**Roles:** All

**Describe the bug**
forgotPassword > Returned success response when user leave blank `Email` field

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
`Error` or `Input` validation message should be returned

**Actual Result**
Returned success response when user leave blank `Email` field

See video -  

https://user-images.githubusercontent.com/5212585/140915954-6b5fa757-0c11-43bf-92d3-e8a7601cb028.mp4

**GraphQL Query:(forgotPassword)**
image

**GraphQL response**
```
{
    "data": {
        "forgotPassword": true
    }
}
```

**Query and Input File**
[BugForgotPasswordEmailBlank.txt](https://github.com/SeedCompany/cord-api-v3/files/7504287/BugForgotPasswordEmailBlank.txt)

CordAPIV3# 600901f

daniel-leod:
Fixes #2271

### DIFF
--- a/src/common/validators/email.validator.ts
+++ b/src/common/validators/email.validator.ts
@@ -12,7 +12,7 @@ export const IsEmail = (
       constraints: [options],
       validator: {
         validate: (value, args) =>
-          value !== null ||
+          value == null ||
           (typeof value === 'string' && isEmail(value, args?.constraints[0])),
         defaultMessage: () =>
           validationOptions?.each


daniel-leod:
@CarsonF I already double-check this one. `isEmail` validator is enough to verify that the input is not empty.

src/common/validators/email.validator.ts
```diff
@@ -12,8 +12,7 @@ export const IsEmail = (
       constraints: [options],
       validator: {
         validate: (value, args) =>
-          value !== null ||
```
CarsonF, line 15: Oh derp. We said anything not null is allowed. I'd like to continue checking and allowing null/undefined here as it's possible that those get passed in here and we don't want to enforce "required" only that the format if given is an email.
```suggestion
          value == null ||
          (typeof value === 'string' && isEmail(value, args?.constraints[0])),
```

CarsonF (changes requested):

CarsonF (approved):

Issue #1858, opened 2021-05-03
createSong > Getting "Could not create song" error while input field "scriptureReferences > end > book " have a "blank" value


iPiyushPatel:
**Describe the bug**
createSong > Getting `Could not create song` error while input field `scriptureReferences > end > book ` have a `blank` value

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
The valid error message should be returned.

**Actual Result**
Getting `Could not create song` error while input field `scriptureReferences > end > book ` have a `blank` value.

**Note:** Same issue is replicate in create LiteracyMaterial, Story, and Film module.

See video -  

https://user-images.githubusercontent.com/5212585/116870013-5f76e580-ac2f-11eb-9550-2f5018e08ee2.mp4

**GraphQL Query:(createSong)**
![image](https://user-images.githubusercontent.com/5212585/116871799-766b0700-ac32-11eb-8b05-08e2dfcfcd14.png)

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Could not create song",
            "extensions": {
                "codes": [
                    "Server"
                ],
                "status": 500,
                "code": "Server",
                "exception": {
                    "message": "Could not create song",
                    "stacktrace": [
                        "at SongService.create (components\\song\\song.service.ts:138:13)",
                        "at SongResolver.createSong (components\\song\\song.resolver.ts:47:18)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createSong"
            ]
        }
    ],
```

**Query and Input File**
[createSongBug.txt](https://github.com/SeedCompany/cord-api-v3/files/6414551/createSongBug.txt)

CordAPIV3# 84d680b4

ParkeBrown:
Referred to ASMs

parashara1:
Was able to successfully reproduce the bug after creating the API request with the mentioned notepad file. The output was the same error message, "Could not create song" with a status 500.

Issue #1261, opened 2020-09-24
updateLiteracyMaterial > Getting  " Book \"\" does not exist"  error while "scriptureReferences > end > book"  field is blank


iPiyushPatel:
**Describe the bug**
updateLiteracyMaterial > Getting  ` Book \"\" does not exist`  error while `scriptureReferences > end > book`  field is blank

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
The proper error message should be returned

**Actual Result**
 Getting  ` Book \"\" does not exist`  error while `scriptureReferences > end > book`  field is blank.
See video -  https://drive.google.com/file/d/1NmA3E7iPyKGDRYIEGYDqikDs_sC18N-8/view?usp=drivesdk

Note : I can found the same issue in product

**GraphQL Query:(updateLiteracyMaterial )**
![image](https://user-images.githubusercontent.com/5212585/94123590-9c952100-fe71-11ea-8995-b890392cbe7b.png)

**GraphQL response**

```
{
    "errors": [
        {
            "message": "Book \"\" does not exist",
            "extensions": {
                "codes": [
                    "NotFound",
                    "Input",
                    "Client"
                ],
                "status": 400,
                "code": "NotFound",
                "exception": {
                    "message": "Book \"\" does not exist",
                    "stacktrace": [
                        "at Function.find (components\\scripture\\books.ts:1272:11)",
                        "at Function.fromRef (components\\scripture\\books.ts:1276:17)",
                        "at Function.fromRef (components\\scripture\\books.ts:1389:17)",
                        "at Function.fromRef (components\\scripture\\books.ts:1478:20)",
                        "at D:\\Project\\Seedcompany\\projects\\cord-api-v3\\src\\components\\scripture\\dto\\scripture-range.dto.ts:64:41",
                        "at Object.exports.mapRange (components\\scripture\\dto\\scripture-range.dto.ts:26:8)",
                        "at Function.fromReferences (components\\scripture\\dto\\scripture-range.dto.ts:64:12)",
                        "at ScriptureReferenceService.update (components\\scripture\\scripture-reference.service.ts:96:33)",
                        "at LiteracyMaterialService.update (components\\literacy-material\\literacy-material.service.ts:192:5)"
                    ]
                }
            },
```

**Query and Input File**
[...]

PR #2295, opened 2021-11-23
Validate scriptureReference end bookName

### LINKED ISSUE #2292: createEthnoArt > Getting "Could not create ethno art" error while input field "scriptureReferences > end > book"  have a "blank" value
**Role:** All

**Describe the bug**
createEthnoArt > Getting `Could not create ethno art` error while input field `scriptureReferences > end > book ` have a `blank` value

**Steps to Reproduce**
On Graphql playground or Postman - create an API request as below snap & attached notepad file and execute the request

**Expected Result**
The valid error message should be returned.

**Actual Result**
Getting `Could not create ethno art` error while input field `scriptureReferences > end > book ` have a `blank` value.

See video -  

https://user-images.githubusercontent.com/5212585/142846147-b1ac61f3-7f71-4d00-bb4d-f5c2cdc938d9.mp4

**GraphQL Query:** (createEthnoArt)
image

**GraphQL response**
```
{
    "errors": [
        {
            "message": "Could not create ethno art",
            "extensions": {
                "codes": [
                    "Server"
                ],
                "status": 500,
                "code": "Server",
                "exception": {
                    "message": "Could not create ethno art",
                    "stacktrace": [
                        "at EthnoArtService.create (components/ethno-art/ethno-art.service.ts:64:13)"
                    ]
                }
            },
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createEthnoArt"
            ]
        }
    ],
    "data": null
}
```

**Query and Input File**
[createEthnoArtBug.txt](https://github.com/SeedCompany/cord-api-v3/files/7580182/createEthnoArtBug.txt)

CordAPIV3# cfbf82c

daniel-leod:
Fixes #2292 #1858 #1261

### DIFF
--- a/src/components/scripture/books.ts
+++ b/src/components/scripture/books.ts
@@ -599,7 +599,7 @@ export class Chapter implements Iterable<Verse> {
   }
 
   verse(verseNumber: number) {
-    if (verseNumber > this.totalVerses) {
+    if (verseNumber <= 0 || verseNumber > this.totalVerses) {
       throw new NotFoundException(
         `Verse ${verseNumber} of ${this.label} does not exist`
       );
--- a/src/components/scripture/dto/scripture-reference.transformer.ts
+++ b/src/components/scripture/dto/scripture-reference.transformer.ts
@@ -3,6 +3,7 @@ import { Type } from 'class-transformer';
 import { ValidateNested } from 'class-validator';
 import { Book } from '../books';
 import { ScriptureReferenceInput } from './scripture-reference.dto';
+import { IsValidBook } from './scripture-reference.validator';
 
 class ScriptureReferenceStartInput extends ScriptureReferenceInput {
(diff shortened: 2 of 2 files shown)


CarsonF (approved):
Thanks for getting this started @daniel-leod.

Putting the `@IsValidBook()` on the `bookName` property made the error path have "bookName" in it instead of "book" which is what our schema actually reflects. "bookName" is a private helper. Moving that to book setter fix that.

However, I also discovered that chapter & verse were not be validated here either. It turned out it was easier to just extend the parent class so its validation decorators applied here as well. I didn't do this previously because of a TS error, but it turned out that it didn't matter for this case so I ignored it.
Now all 3 properties are properly validated.

Also found that negative numbers for verses weren't validated either, so I added that too.

PR #2299, opened 2021-11-24
Product - Verse totals & verse equivalents

CarsonF:
```gql
type DirectScriptureProduct {
  """The total number of verses of the selected scripture in this product"""
  totalVerses: Int!

  """
  The total number of verse equivalents of the selected scripture in this product.
  Verse equivalents weight each verse based on its translation difficulty.
  """
  totalVerseEquivalents: Float!
}
```
I've opted to write these calculations to the DB. It's a little bit of work to calculate these. These will be referenced in progress  data, so I wanted them to be easily accessible.

----

I ended up adding formatting here as well. I opted to do it at the field level as this lets us reuse the already fetched data and just format it different. A field can be specified multiple times with different aliases.
```gql
fragment sp on StepProgress {
  completed(format: Percent) {
    value
  }
  verses: completed(format: Verses) {
    value
  }
}

fragment ps on ProgressSummary {
  planned(format: Percent)
  actual(format: Percent)
}
```
This seems to work well in testing adhoc gql queries, but I'd like to get some confirmation from UI implementation. It could be easy for them pass in a single format arg to the operation and just refetch from the API when user wants a different format. Another optional would be to ask for each format desired all at once with aliases and swap out props used based on selector.

### DIFF
--- a/src/components/ethno-art/dto/create-ethno-art.dto.ts
+++ b/src/components/ethno-art/dto/create-ethno-art.dto.ts
@@ -2,7 +2,7 @@ import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import { Type } from 'class-transformer';
 import { ValidateNested } from 'class-validator';
 import { NameField } from '../../../common';
-import { ScriptureRangeInput } from '../../scripture';
+import { ScriptureField, ScriptureRangeInput } from '../../scripture';
 import { EthnoArt } from './ethno-art.dto';
 
 @InputType()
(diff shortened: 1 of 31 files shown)


CarsonF:
@zkhin @tmpacifitech @daniel-leod Could I get one of you guys to implement the stubbed migration I have in this PR? It's the only thing remaining.

CarsonF:
Ok I'm merging for now. If UI work work easier with something different I'm open to changes.

Issue #2291, opened 2021-11-18
G&P | Capture planned complete date (fiscal year) for each Product


CarsonF:
Date field for all products
```gql
interface Product {
  plannedCompleteDate: CalendarDate!
}
```

CarsonF:
I made a mistake. I forgot that this date is _per step_. Not one per product.

I think the best way forward might be to create another object for product's steps. I think this could hold more fields in near future too.
```gql
interface Product {
  stepList: [ProductStepInfo!]!
  steps: [ProductStep!]! @deprecated(reason: "Use stepList instead")
}
type ProductStepInfo {
  id: ID! # Maybe?
  name: ProductStep!
  plannedCompleteDate: SecuredDate!
}
```

sherijennings:
Part of quarterly reporting project

CarsonF:
This is needed to completely stop using the PnP. This first "narrative report" pass will prob still use it.

PR #2308, opened 2021-12-08
Add planned complete date field to product

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2291

### DIFF
--- a/src/components/authorization/roles/administrator.role.ts
+++ b/src/components/authorization/roles/administrator.role.ts
@@ -335,6 +335,7 @@ export const Administrator = new DbRole({
         { propertyName: 'title', permission: { read, write, }, },
         { propertyName: 'description', permission: { read, write, }, },
         { propertyName: 'unspecifiedScripture', permission: { read, write, }, },
+        { propertyName: 'plannedCompleteDate', permission: { read, write, }, },
       ],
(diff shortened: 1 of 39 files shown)


zkhin:
@CarsonF Do we need to put the new field (`plannedCompleteDate `) in the CREATE/UPDATE DTO as well?

CarsonF:
> @CarsonF Do we need to put the new field (`plannedCompleteDate `) in the CREATE/UPDATE DTO as well?

Yes please.

CarsonF:
@zkhin can we also validate that the date is within the engagement date range?

zkhin:
> @zkhin can we also validate that the date is within the engagement date range?

Sure!

CarsonF:
Oh no. I've made a big mistake. I was looking at the PnP file to give you the column to extract from and realized that the date is _per step_ not product 🤦‍♂️ . This brings up larger schema changes, because it basically converts steps from a string to an object. I totally forgot this by the time I was making the ticket. I thought to myself "this is easy, why didn't I make this ticket sooner" OOF.

Still thinking more on best way forward...

CarsonF:
Closing because we are switching directions in #2291. Sorry for the lost work @zkhin.

PR #2339, opened 2022-01-10
Update CI tests to use Node 16

CarsonF:
16.9.0 bug was fixed a few days later with 16.9.1


CarsonF:
Passed enough to count

PR #2351, opened 2022-01-20
Add User.projects

daniel-leod:


bryanjnelson:
@daniel-leod - Does this include being able to filter by Role?

daniel-leod:
@bryanjnelson No, this is only filtered by userId. I'll update this to match the current query filtering.

src/components/project/list-filter.query.ts
```diff
@@ -38,6 +43,27 @@ export const projectListFilter = (input: ProjectListInput) =>
...
+              ]),
+              path([
+                node('node'),
+                relation('out', '', 'engagement', ACTIVE),
+                node('', 'Engagement'),
+                relation('out', '', 'intern', ACTIVE),
+                node('', 'User', { id: value }),
+              ]),
+            ])
+          : null,
```
CarsonF, line 66: You can re-use the single where clause the builder builds. The function won't be called if the value is nullable, so that check isn't necessary either.
```ts
({ value }) => ({
  userId: [
    path([...]),
    path([...]),          
  ],
})
```

CarsonF (changes requested):
We need to think about permissions too. Since this filter is already inside of the project list query, we can be sure they have access to the project. But we don't know if they have access to the project's team members. If they can view the project but cannot view its team members, then the project shouldn't show up in this list. Otherwise we would leak that they are on the project even though the requesting user cannot see the project's team members.

@sethmcknight is this a relevant use case? If not we can leave a TODO and circle back to it later.

CarsonF:
@sethmcknight says we can leave a TODO on this filter for now, and instead just disallow viewing this property for those use cases all together. So the requestor can see 0% or 100% of projects, but not the relevant X%. This covers the majority use case since most people that would use this feature can see all projects.

We can circle back to adding permissions to the roles.

CarsonF (approved):
:shipit:

PR #2368, opened 2022-02-07
Changes to Budget Record Permissions for FAs and LFAs.

bryanjnelson:
Fixes [#268 ](https://github.com/SeedCompany/management/issues/268).


src/components/project/project.service.ts
```diff
@@ -629,7 +629,9 @@ export class ProjectService {
     return {
       value: budgetToReturn,
       canRead: permsOfProject.budget.canRead,
-      canEdit: session.roles.includes('global:Administrator')
+      canEdit: this.budgetService.canEditFinalized(
+        session.roles.concat(project.scope)
+      )
         ? true
         : permsOfProject.budget.canEdit &&
           budgetToReturn?.status === BudgetStatus.Pending,
```
CarsonF, line 637: I know I wrote this but I think it would be better if we
```ts
permsOfProject.budget.canEdit && (
  budgetToReturn?.status === BudgetStatus.Pending || 
  this.budgetService.canEditFinalized(session.roles.concat(project.scope))
)
```
That way we are just extending the permissions instead of sometimes ignoring them.
However, in practice it shouldn't make a difference.

CarsonF (commented):
Looks good!

src/components/project/project.service.ts
bryanjnelson: I can get behind that.

CarsonF (approved):
Thanks @bryanjnelson! 🌮

PR #2372, opened 2022-02-08
Fix e2e tests

zkhin:


zkhin:
> Thanks for your work on this @zkhin!
> 
> It seems like the best workaround is to just split the test executions into different nodejs processes. That can be done however, without creating a separate github action job or jest config. I just applied a one-liner in my PR to split the tests up by 5 files at a time here [73f3270](https://github.com/SeedCompany/cord-api-v3/commit/73f3270cb8baae92a121d04cbfd224c65dbb2121)
> 
> It also seems like what I was trying to do with deleting projects wasn't working right. Here you just gave the user the Administrator role. I'm good with that and applied it to my PR as well, giving you credit. [376c16a](https://github.com/SeedCompany/cord-api-v3/commit/376c16aad9b549757b769d4f522b935375d726c9)
> 
> The last change you made to fix the engagement workflow, was to add a transition to Engagement Active. So now the engagement status goes
> 
> ```
> DiscussingChangeToPlan -> Active -> DiscussingChangeToPlan -> ActiveChangedPlan
> ```
> 
> I instead removed the explicit step to change to DiscussingChangeToPlan since that happens automatically now [ccdbcb8](https://github.com/SeedCompany/cord-api-v3/commit/ccdbcb88509e3cc63beb5553eed7ec5a7c8770a9)
> 
> ```
> DiscussingChangeToPlan -> ActiveChangedPlan
> ```
> 
> So we can just merge #2352 in favor of this one. Thanks again for your work here though!

Thank you!

PR #2378, opened 2022-02-10
Add constraint to progress extractor to only extract progress within project range

### LINKED ISSUE #2376: Product Progress Extraction should ignore products/steps outside of project range
When extracting products (from the _Planning_ sheet) we check and only accept products with steps that fall within the range of the project. ([source](https://github.com/SeedCompany/cord-api-v3/blob/99fb95880f867d2871e1dc643057fa0f7f34b7e7/src/components/product/product-extractor.service.ts#L53-L58), [source](https://github.com/SeedCompany/cord-api-v3/blob/99fb95880f867d2871e1dc643057fa0f7f34b7e7/src/components/product/product-extractor.service.ts#L27-L27))

**_Progress_ extraction also needs this constraint**

Here's an example, the gray cells represent progress that happened outside of the project range (the green are inside).
Currently we pull both the gray & green, but should only ignore the gray.
Screen Shot 2022-02-09 at 10 38 32 AM
[Adele 2 PnP FY22-Q1-lo.xlsm.zip](https://github.com/SeedCompany/cord-api-v3/files/8043205/Adele.2.PnP.FY22-Q1-lo.xlsm.zip)

daniel-leod:
Fixes #2376

### DIFF
--- a/src/components/product-progress/step-progress-extractor.service.ts
+++ b/src/components/product-progress/step-progress-extractor.service.ts
@@ -1,10 +1,10 @@
 import { Injectable } from '@nestjs/common';
 import { assert } from 'ts-essentials';
 import { MergeExclusive } from 'type-fest';
-import { entries } from '../../common';
+import { entries, fullFiscalYear } from '../../common';
 import { Cell, Column } from '../../common/xlsx.util';
 import { Downloadable } from '../file';
-import { findStepColumns, isGoalRow, Pnp, ProgressSheet } from '../pnp';
+import { findStepColumns, isGoalRow, PlanningSheet, Pnp } from '../pnp';
 import { ProductStep as Step } from '../product';
 import { Book } from '../scripture/books';
 import { StepProgressInput } from './dto';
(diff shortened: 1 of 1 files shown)


src/components/product-progress/step-progress-extractor.service.ts
```diff
@@ -60,12 +60,17 @@ const parseProgressRow =
       const percentDecimal = cell.asNumber;
       return percentDecimal ? percentDecimal * 100 : undefined;
     };
-    const steps = entries(stepColumns).map(
-      ([step, column]): StepProgressInput => ({
-        step,
-        completed: progress(column),
-      })
-    );
+    const steps = entries(stepColumns).flatMap(([step, column]) => {
```
CarsonF, line 63:
```suggestion
    const steps = entries(stepColumns).flatMap<StepProgressInput>(([step, column]) => {
```
Connect this type explicitly so we know this is the shape we are defining. Also helps with looking up usage & refactors.

src/components/product-progress/step-progress-extractor.service.ts
```diff
@@ -60,12 +60,17 @@ const parseProgressRow =
...
+      // To get the fiscal year is needed to check the next column
+      const fiscalYear = sheet.cell(column.next(), row).asNumber;
+      const fullFY = fiscalYear ? fullFiscalYear(fiscalYear) : undefined;
+      // Only include step if it references a fiscal year within the project
+      if (!fullFY) {
```
CarsonF, line 68: The comment is right but the code does not implement it

CarsonF (changes requested):

src/components/product-progress/step-progress-extractor.service.ts
```diff
@@ -60,12 +62,19 @@ const parseProgressRow =
...
-        step,
-        completed: progress(column),
-      })
+    const steps = entries(stepColumns).flatMap<StepProgressInput>(
+      ([step, column]) => {
+        // To get the fiscal year is needed to check the next column
+        const fiscalYear = sheet.cell(column.next(), row).asNumber;
+        const fullFY = fiscalYear ? fullFiscalYear(fiscalYear) : undefined;
+        // Only include step if it references a fiscal year within the project
+        if (!fullFY || !projectFiscalYears.intersection(fullFY)) {
```
CarsonF, line 71: This isn't exactly right either. If there's no fiscal year the product could still be in progress like `50%` or it could be blank meaning not started yet. I think really the only way to know if the step is planned is to look at the planning sheet. We can reuse the same logic from the product extractor to determine whether the step is planned or not for each row.

CarsonF (changes requested):

CarsonF:
Closing in favor of #2427

PR #2387, opened 2022-02-16
Fields Pipe & Id Shortcut

CarsonF:
- Created `Fields` pipe to grab info for fields requested out of GraphQL's resolve info.
- Created `IsOnlyId` pipe to simplify above to a boolean if only the object's `id` is requested.
- Skip going back to the DB for these relations if we don't need extra info:
  - `ProgressReport.engagement`
  - `Engagement.project`

### DIFF
--- a/src/components/periodic-report/progress-report-parent.resolver.ts
+++ b/src/components/periodic-report/progress-report-parent.resolver.ts
@@ -1,6 +1,7 @@
-import { Parent, ResolveField, Resolver } from '@nestjs/graphql';
+import { Info, Parent, ResolveField, Resolver } from '@nestjs/graphql';
+import { Fields, IsOnlyId } from '../../common';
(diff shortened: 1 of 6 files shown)


src/components/periodic-report/progress-report-parent.resolver.ts
```diff
@@ -1,16 +1,21 @@
...
   @ResolveField(() => LanguageEngagement)
   async parent(
+    @Info(Fields, IsOnlyId) onlyId: boolean,
     @Parent() report: ProgressReport,
     @Loader(EngagementLoader) engagements: LoaderOf<EngagementLoader>
-  ): Promise<Engagement> {
+  ) {
+    if (onlyId) {
+      return { id: report.parent.properties.id };
+    }
```
CarsonF, line 18: Here's an example

PR #2391, opened 2022-02-17
Fix comparing products of different types when extracting planning pnp

CarsonF:
When matching story (derivative) products by pnp index only consider derivative products

If other products of different types have been created from a different kind of pnp, they
should be ignored. That's still unexpected behavior and could have implications but this
fixes the server error.


bryanjnelson (commented):
Brilliant.

bryanjnelson (approved):

PR #2397, opened 2022-02-24
update leadership and staff role to allow uploading of narrative repo…

### LINKED ISSUE #2396: Update Permissions
update the permissions for the Leadership and Staff Member roles to allow access to Upload files & reports, and Edit the Received Date if the are Members of a Project.

michaelmarshall:
…rt and modifying the received date

closes #2396


CarsonF:
Also if the intent is to only allow these if they are a member of the project, then we'll need to make new role defs for the project membership scope instead of these global ones.

sethmcknight:
@michaelmarshall FYI I think to resolve Jim & Kathy's issue this needs to include all periodic reports (financial & progress), not just narrative.

PR #2385, opened 2022-02-15
Finish list refactor

elisabeth-sorrell:
closes #1787  and closes #1653 

Checked project change request, and all roles can read at least one property, so all roles can list at the moment. Also, no security group logic is used when listing CRs, so it should be good to go.

PR #2400, opened 2022-02-25
Remove SecurityGroups, Powers in DB, custom procedure

### LINKED ISSUE #2401: Clearout security group functionality from create functions and user functions

elisabeth-sorrell:
closes #2401 
To be merged after #2385


CarsonF:
Just FYI I started the powers refactor awhile ago with #1706. Totally happy to close that one, but wanted to give you a heads up in case there was something there still useful.

CarsonF (commented):
Yay so excited! 🎉 

Looks great so far, just need to wrap up the powers clean up in the test suite.

test/engagement-changeset-aware.e2e-spec.ts
```diff
@@ -120,11 +121,18 @@ const readProjectChangeset = (app: TestApp, id: string, changeset?: string) =>
...
+    await registerUser(app, { roles: [Role.Administrator] }); // only admin can create funding account for now
+    return await createFundingAccount(app);
+  });
+  const location = await runInIsolatedSession(app, async () => {
+    await registerUser(app, { roles: [Role.Administrator] }); // only admin can create funding account for now
+    return await createLocation(app, { fundingAccountId: fundingAccount.id });
+  });
+  const fieldRegion = await runInIsolatedSession(app, async () => {
+    await registerUser(app, { roles: [Role.Administrator] }); // only admin can create funding account for now
+    return await createRegion(app);
```
CarsonF, line 134: Seeing a lot of these. You can use `runAsAdmin` instead of `runInIsolatedSession`. It will login as the root admin instead of registering a new user every time.

Also merge these together when possible, so we don't waste session & login queries.

```suggestion
  // only admins can create these
  const [fundingAccount, location, region] = await runAsAdmin(app, async () => {
    const fundingAccount = await createFundingAccount(app);
    const location = await createLocation(app, { fundingAccountId: fundingAccount.id });
    const region = await createRegion(app);
    return [fundingAccount, location, region];
```

CarsonF (changes requested):
Can you

src/components/engagement/engagement.service.ts
```diff
@@ -607,7 +613,15 @@ export class EngagementService {
...
     }
+
     if (project.status !== ProjectStatus.InDevelopment) {
+      const changeRequests = await this.projectService.listChangeRequests(
+        project,
+        ProjectChangeRequestListInput.defaultVal,
+        session
+      );
+      if (project.status === ProjectStatus.Active && changeRequests.total > 0)
+        return;
```
CarsonF, line 624: I don't think this is necessary, unless I'm missing something. Having a `changeset` passed in means this change is for a changeset, which is where we would skip this check. Just because a project has a pending change-request/changeset does not mean an Engagement can be created on the Project _outside_ that changeset.

CarsonF: Also the business logic is duplicated [here](https://github.com/SeedCompany/cord-api-v3/blob/5275777d82f773be0f60000c0de39dba36cd02a9/src/components/project/project.service.ts#L452-L455) sadly, and needs to be updated as well.

src/components/authorization/dto/powers.ts
```diff
@@ -7,6 +7,7 @@ export enum Powers {
   CreateBudget = 'CreateBudget',
   CreateBudgetRecord = 'CreateBudgetRecord',
   CreateCeremony = 'CreateCeremony',
+  CreateChangeRequest = 'CreateChangeRequest',
```
CarsonF, line 10: This isn't necessary as it's not a root object. Instead we would want to use `project.changeRequests.canCreate`

src/components/engagement/engagement.service.ts
elisabeth-sorrell: I must've misunderstood what you were telling me then, because taking out the changeRequest logic fails a bunch of tests in the engagement-changeset-aware suite, and that's why I put that stuff in in the first place.

elisabeth-sorrell: You said "Yeah that needs to be updated to ignore if changing inside of a changeset". The stuff to be ignored being: if a project is any other status other than "Active", and the user adding the engagement is not an Admin, error out.

CarsonF: Just to be clear, the [if statement change above this](https://github.com/SeedCompany/cord-api-v3/pull/2400/files/332d87a56b6bf41228ee76c41fc7600d34aca14d..8acee685afac3880c31a49310ceaf413eb651670#diff-64ecdc9d362652a540d48b634cbc48b1758180b1836723f5b3926646b04452eaR604) (outside of this diff) is correct. Does that change your comment?

src/components/engagement/engagement.service.ts
```diff
@@ -593,9 +594,14 @@ export class EngagementService {
...
+   * [BUSINESS RULE] Only Projects with a Status of 'Active' and part of a changeset can have Engagements created or deleted.
    */
-  protected async verifyProjectStatus(projectId: ID, session: Session) {
-    if (session.roles.includes('global:Administrator')) {
+  protected async verifyProjectStatus(
+    projectId: ID,
+    session: Session,
+    changeset?: ID
+  ) {
+    if (changeset || session.roles.includes('global:Administrator')) {
```
elisabeth-sorrell, line 604: So this is needs to be taken out?

test/engagement-changeset-aware.e2e-spec.ts
elisabeth-sorrell: This suggested change also doesn't work

src/components/engagement/engagement.service.ts
CarsonF: That can stay. The query for the change request list is what should be taken out.

src/components/authorization/dto/powers.ts
CarsonF: Make sure that the roles are updated accordingly for `project.changeRequests.write = true`

CarsonF: Deferring this to a later PR

CarsonF (approved):

Issue #2424, opened 2022-04-22
Change direction of ownership on Directories


elisabeth-sorrell:
Refer to https://github.com/SeedCompany/cord-api-v3/issues/1753 Section `Ensure relationship direction matches owner correctly` for more information

Issue #1753, opened 2021-03-19
[RFC] Deletes


CarsonF:
Talking about deletes with @blakekrammes. Capturing take aways from our convo.

# Deleted At prop
Add `deletedAt` to all BaseNodes/Properties (respective to what's being deleted)

This is just good to have, and could be useful to identify a set of deletes. i.e. match anything deleted at this datetime.

# Set active false when?

Only set active=false on direct connections.
- If deleting property, set active false
- If deleting BaseNode
  - keep active true on all existing properties
  - Set active false on relationships to other things

# Labels on BaseNodes

Prefix all labels on BaseNodes with `Deleted_` so they aren't matched accidentally.

# Handling Uniqueness constraints 

We could leave properties untouched on BaseNode delete, because that's fine. However because some labels have a unique constraint this won't work. Our best solution is just set all labels on `Property` nodes to have a `Deleted_` prefix. This ensure constraints are not broken, and if we look up by label we don't accidentally match deleted nodes.

We can still determine if a property should be restored by the active relationship.

# Replacing properties with a label with a unique constraint 

All properties that have a label with a unique constraint need to have their labels adjusted as specified above.
ie. Old `ProjectName` properties are still constrained incorrectly currently.

# Remove active prop for relationships that have a BaseNode in between
Remove `active` prop on relations that have a BaseNode connection.

i.e.
```cql
(Project)-[]->(Engagement)<-[]-(Language)
```
active is not needed on these two relationships because we can check for Engagement label instead.

```cql
(FieldRegion)-[:director]-(User)
```
still needs an `active` prop because there's no BaseNode in between to check the label on.
Also all `Property` relationships still need it for same reason.

# Cascading deletes

For all _in_ relationships on `BaseNodes` add `cascadeDelete` if the `BaseNode` should be deleted when the other `BaseNode` is deleted.

```cql
(Project)-[:engagement { cascadeDelete: true }]->(Engagement)<-[:engagement { cascadeDelete: true }]-(Language)
```
So if either a `Project` or a `Language` is deleted, we know delete the `Engagement` as well.

This lets find all the `BaseNodes` to delete with a single query:
```cql
match (:BaseNode { id: "" })-[* { cascadeDelete: true }]->(:BaseNode)
```

# Ensure relationship direction matches owner correctly

For example, current we have
```cql
(:Directory)<-[:parent]-(:File)
```
instead we should have
```cql
(:Directory)-[:child]->(:File)
```

[...]

PR #2425, opened 2022-04-25
Switch relationship direction from parent FileNode to FileNode and rename from 'parent' to 'child'

elisabeth-sorrell:
Refer to https://github.com/SeedCompany/cord-api-v3/issues/1753 for more info. This help make sure all BaseNode to BaseNodes are the same relationship direction.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466553805) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/file/file.repository.ts
+++ b/src/components/file/file.repository.ts
@@ -90,7 +90,7 @@ export class FileRepository extends CommonRepository {
       .query()
       .match([
         node('parent', 'FileNode', { id: parentId }),
-        relation('in', '', 'parent', ACTIVE),
+        relation('out', '', 'child', ACTIVE),
         node('node', 'FileNode'),
         relation('out', '', 'name', ACTIVE),
         node('name', 'Property', { value: name }),
@@ -109,11 +109,11 @@ export class FileRepository extends CommonRepository {
       .query()
       .match([
         node('start', 'FileNode', { id }),
-        relation('out', 'parent', 'parent', ACTIVE, '*'),
+        relation('in', 'child', 'child', ACTIVE, '*'),
         node('node', 'FileNode'),
       ])
-      .with('node, parent')
-      .orderBy('size(parent)')
+      .with('node, child')
+      .orderBy('size(child)')
       // Using paginate to maintain order through hydration
       .apply(paginate({ page: 1, count: 100 }, this.hydrate()))
       .first();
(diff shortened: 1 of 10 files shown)


elisabeth-sorrell:
Closes https://github.com/SeedCompany/cord-api-v3/issues/2424

CarsonF (commented):
Previously migrations should not be changed as they would be working with the old schema.

CarsonF (changes requested):
See prev comment

src/components/file/file.repository.ts
```diff
@@ -109,11 +109,11 @@ export class FileRepository extends CommonRepository {
       .query()
       .match([
         node('start', 'FileNode', { id }),
-        relation('out', 'parent', 'parent', ACTIVE, '*'),
+        relation('in', 'child', 'child', ACTIVE, '*'),
```
CarsonF, line 112: I think the query variable should continue to be called `parent` as it is one of the _parents_ of the node.
```suggestion
        relation('in', 'parent', 'child', ACTIVE, '*'),
```

src/components/file/migrations/ChangeFileToFileRels.migration.ts
```diff
@@ -0,0 +1,44 @@
...
+      .query()
+      .match([
+        node('parent', 'FileNode'),
+        relation('out', 'p', 'parent'),
+        node('child'),
+      ])
+      .raw(
+        `
+        with collect(p) AS rels
+        CALL apoc.refactor.rename.type("parent", "child", rels )
```
CarsonF, line 33: Can we combine these two queries? Either one after the other or maybe both in sub queries after matching the appropriate relationships?

src/components/file/migrations/ChangeFileToFileRels.migration.ts
```diff
@@ -0,0 +1,44 @@
...
+  async up() {
+    // first, invert the relationship from incoming to outgoing
+    const res = await this.db
+      .query()
+      .match([
+        node('parent', 'FileNode'),
+        relation('in', 'rel', 'parent'),
+        node('child'),
+      ])
+      .raw(`CALL apoc.refactor.invert(rel) yield input, output`)
```
CarsonF, line 15: We also have around 500k of these relationships in prod. Any idea if either of these procedures will handle this? Or do we need to wrap them in an `apoc.periodic.iterate`?

elisabeth-sorrell:
@CarsonF back to you

elisabeth-sorrell:
@CarsonF Just making sure you know that the comments on the PR have been addressed

sethmcknight:
Too old

PR #2426, opened 2022-04-25
Fix security tests to use roles with correct powers instead of adding powers directly

elisabeth-sorrell:


CarsonF:
Can you drop the first ~9 commits which are unrelated to this?

CarsonF (approved):

PR #1630, opened 2021-02-09
Add transaction decorator to all action entry-points involving the DB

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3655589517) by [Unito](https://www.unito.io)

PR #2437, opened 2022-05-03
Automatically run all gql mutations in neo4j transactions

CarsonF:
Previously attempted this manually in #1630 but it was blocked and got stale.
This on the other hand is much cleaner.


bryanjnelson (approved):
Big win here!

PR #2438, opened 2022-05-03
Added check when parsing planning tab to verify viability of step based upon progress cell.

bryanjnelson:


src/components/product/product-extractor.service.ts
```diff
@@ -44,17 +50,26 @@ export class ProductExtractor {
...
+    const rowIndex = row.a1 - sheet.goals.start.row.a1;
+    const progressRow = pnp.progress.goals.start.row.a1 + rowIndex;
+
     const steps = entries(stepColumns).flatMap(([step, column]) => {
       const fiscalYear = sheet.cell(column, row).asNumber;
       const fullFY = fiscalYear ? fullFiscalYear(fiscalYear) : undefined;
       // only include step if it references a fiscal year within the project
       if (!fullFY || !sheet.projectFiscalYears.intersection(fullFY)) {
         return [];
       }
```
CarsonF, line 66: This logic is duplicated now in progress extraction. Would you mind abstracting to `isGoalStepPlannedInsideProject` function?

Also here we use `sheet.projectFiscalYears` but in progress we use `sheet.projectDateRange`. I believe this one is correct and we should standardize on that.

CarsonF (approved):
Otherwise looks good!

bryanjnelson:
@CarsonF - Once this gets your okay I'll squash these together.

bryanjnelson (commented):
Love it; something didn't sit well with me about how I had it either.

CarsonF (approved):

PR #2440, opened 2022-05-05
Allow DtoRepository to define concrete shape from interface DTO

CarsonF:
Allow DtoRepository to accept a different dto shape from the resource class

Useful for interface repos that work with concretes


elisabeth-sorrell (approved):
LGTM

PR #2441, opened 2022-05-05
Rollback S3 file creation on error

### LINKED ISSUE #2439: Rollback S3 file creation on error
Currently there's no rollback action for s3 backing file when our DB data gets rolled back.
The forward action happens here:
https://github.com/SeedCompany/cord-api-v3/blob/4c437b0a9636b83faaae83fc308f22d2d79cc98c/src/components/file/file.service.ts#L281

As a workaround for this we are calling a delete file mutation when the consumer gets the error here:
https://github.com/SeedCompany/cord-field/blob/8854b7414afc0f646c094ddd9b8bba403d62a02a/src/components/Upload/useUploadFile.ts#L35-L39

I'd like to handle this internally and remove this work around.

It's unclear without further investigation how this could technically work, since the neo4j transactions are all transparent to app business code.
Possibly there could hack on some kind of rollback callback logic. Or maybe it could happen independent of neo4j transactions.

CarsonF:
Fixes #2439

### DIFF
--- a/src/core/database/transactional-mutations.interceptor.ts
+++ b/src/core/database/transactional-mutations.interceptor.ts
@@ -7,7 +7,7 @@ import {
 import { GqlContextType, GqlExecutionContext } from '@nestjs/graphql';
 import { Connection } from 'cypher-query-builder';
 import { GraphQLResolveInfo } from 'graphql';
-import { lastValueFrom, of } from 'rxjs';
+import { from, lastValueFrom } from 'rxjs';
 
 /**
  * Run all mutations in a neo4j transaction.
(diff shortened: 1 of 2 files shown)


src/core/database/transactional-mutations.interceptor.ts
```diff
@@ -28,9 +28,8 @@ export class TransactionalMutationsInterceptor implements NestInterceptor {
       return next.handle();
     }
 
-    const res = await this.db.runInTransaction(
```
bryanjnelson, line 31: What's the effective difference between the `of` vs `from`?

bryanjnelson (approved):

src/core/database/transactional-mutations.interceptor.ts
CarsonF: The only one I can think of is this method will throw an exception instead of having the returned observable emit the error. The caller of this lifecycle hands this, so it wasn't a problem. This just felt cleaner.
```diff
- observable -> promise -> awaited -> observable / thrown error
+ observable -> promise -> observable
```

bryanjnelson: 👌🏼

PR #2445, opened 2022-05-12
Fixed fullFiscalQuarter function bug and added tests.

bryanjnelson:


src/common/fiscal-year.test.ts
```diff
@@ -0,0 +1,29 @@
+import { DateTime } from 'luxon';
+import { fiscalYear } from '.';
+import { fullFiscalQuarter } from './fiscal-year';
+
+describe('fiscal-year', () => {
```
CarsonF, line 5: Can you drop this grouping? It doesn't align with src code.
```suggestion
```

bryanjnelson: You don't want a file level grouping?  I'll admit it's odd; typically I'd have a class or component name there...

CarsonF (approved):

PR #2448, opened 2022-05-18
Stubbed stepCompleteDate tests

bryanjnelson:
@CarsonF - An example shell for @atGit2021 on how to lay out conditions to test a function.

### DIFF
--- a/src/components/pnp/isGoalStepPlannedInsideProject.spec.ts
+++ b/src/components/pnp/isGoalStepPlannedInsideProject.spec.ts
@@ -1,0 +1,13 @@
+describe('stepCompleteDate', () => {
+  it('should return null when fiscal quarter is nullish', () => {
+    //TODO
+  });
+
+  it('should return null when fiscal year is nullish', () => {
+    //TODO
+  });
+
+  it('should return end of full fiscal quarter when valid cell is passed in', () => {
+    //TODO
+  });
+});


src/components/pnp/isGoalStepPlannedInsideProject.spec.ts
```diff
@@ -0,0 +1,21 @@
+describe('stepCompleteDate', () => {
+  it('should return null when fiscal quarter is undefined', () => {
```
CarsonF, line 2: I'd combine these two, they are almost always handled the same way. Only exception is update patches where undefined is missing & null is set to null.
```suggestion
  it('should return null when fiscal quarter is nullish', () => {
```
```suggestion
  it('should return null when fiscal quarter is nil', () => {
```

CarsonF (approved):

src/components/pnp/isGoalStepPlannedInsideProject.spec.ts
bryanjnelson: @CarsonF - I waffled on that.  You mean use **either** the _nullish_ or _nil_ verbiage and combine the "undefined" and "null" conditions into one test?

CarsonF: Easy to test both within one test as well.

PR #2457, opened 2022-05-27
Fixed bug where incorrect progress cell was being checked during product extraction.

bryanjnelson:


src/components/product/product-extractor.service.ts
```diff
@@ -26,11 +26,12 @@ export class ProductExtractor {
     const sheet = pnp.planning;
 
     const stepColumns = findStepColumns(sheet, availableSteps);
+    const progressStepColumns = findStepColumns(pnp.progress);
```
CarsonF, line 29: I think availableSteps should be passed here too
```suggestion
    const progressStepColumns = findStepColumns(pnp.progress, availableSteps);
```

CarsonF (changes requested):

CarsonF (approved):
Thanks @bryanjnelson!

PR #2461, opened 2022-06-03
Fix onlyMultipleEngagements project list filter

### LINKED ISSUE #2458: ProjectList.onlyMultipleEngagements filter is broken.
Zero projects are returned when this filter is true. Obviously we want projects returned that have multiple engagements.

daniel-leod:
Fixes #2458

### DIFF
--- a/src/components/project/list-filter.query.ts
+++ b/src/components/project/list-filter.query.ts
@@ -6,17 +6,19 @@ export const projectListFilter = (input: ProjectListInput) =>
   filter.builder(input.filter, {
     type: filter.skip, // already applied
     status: filter.stringListProp(),
-    onlyMultipleEngagements: ({ value, query }) => {
-      if (!value) return null;
-      query
-        .match([
-          node('node'),
-          relation('out', '', 'engagement', ACTIVE),
-          node('engagement', 'Engagement'),
-        ])
-        .with('*, count(engagement) as engagementCount');
-      return { engagementCount: greaterThan(1) };
-    },
+    onlyMultipleEngagements:
+      ({ value, query }) =>
+      () =>
+        value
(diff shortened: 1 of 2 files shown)


CarsonF (changes requested):
I doubt this changes anything.
Changing `count()` to `size(collect())` is the same thing but probably less performant.
Also I changed `node` to `*` in ae6fad364751386b64c91a0dd7b61ead0dd239a1 to fix something.

src/components/project/list-filter.query.ts
```diff
@@ -6,17 +6,19 @@ export const projectListFilter = (input: ProjectListInput) =>
...
-          node('node'),
-          relation('out', '', 'engagement', ACTIVE),
-          node('engagement', 'Engagement'),
-        ])
-        .with('*, count(engagement) as engagementCount');
-      return { engagementCount: greaterThan(1) };
-    },
+    onlyMultipleEngagements:
+      ({ value, query }) =>
+      () =>
```
CarsonF, line 11: This applies the where clause after the others, as I'm sure you've figured out. But the `with node, engagementCount` will still drop the `requestingUser` variable. Have you tried running this query?

daniel-leod: Yes, I tested it out and works as expected even using the combination of other filters like `pinned` and `mine`.

CarsonF: Yes I just checked as well and `requestingUser` isn't re-used outside of filters. It's re-matched again, which is fine.

CarsonF (approved):
Thanks @daniel-leod! Tested locally and working! 🎉 

I tacked on another potential bug fix

PR #2463, opened 2022-06-08
Add ExactLength validator

daniel-leod:
Add `ExactLength` validator to ensure that properties like `code`, `provisionalCode` and `registryOfDialectsCode` have the proper length.


CarsonF (approved):
Thanks @daniel-leod!

Related to https://github.com/SeedCompany/cord-field/issues/502

PR #2468, opened 2022-06-14
Tweak partner update/create dto in order to make address clearable

### LINKED ISSUE #1747: Partner.address cannot be cleared
## Environment

| Tech      | Version / Name |
| --------- | -------------- |
| API       |                |
| FE        |                |
| OS/Device |                |
| Browser   |                |

daniel-leod:
Fixes #1747

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466529600) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/partner/dto/create-partner.dto.ts
+++ b/src/components/partner/dto/create-partner.dto.ts
@@ -34,7 +34,7 @@ export abstract class CreatePartner {
   readonly active?: boolean;
 
   @NameField({ nullable: true })
-  readonly address?: string;
+  readonly address?: string | null;
 }
 
 @InputType()
--- a/src/components/partner/dto/partner.dto.ts
+++ b/src/components/partner/dto/partner.dto.ts
@@ -13,6 +13,7 @@ import {
   SecuredProperty,
   SecuredProps,
   SecuredString,
+  SecuredStringNullable,
   Sensitivity,
   SensitivityField,
 } from '../../../common';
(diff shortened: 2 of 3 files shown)


CarsonF (changes requested):
I think previously we still accepted `null` which should clear it. The UI never sent a null value though. I'd prefer that rather than an empty string. I don't like using empty strings as the lack of a value, that's what null is for.

CarsonF:
You tried via the UI or GQL directly?

daniel-leod:
@CarsonF I tried via GQL directly

CarsonF:
Hmm I'd start with that then, the API should accept and save null. Also the types should be updated.
```diff
-  readonly address: SecuredString;
+  readonly address: SecuredStringNullable;
```
```diff
-  readonly address?: string;
+  readonly address?: string | null;
```

CarsonF (commented):
This breaking change is true, because the UI references the [field type explicitly with a fragment](https://github.com/SeedCompany/cord-field/blob/42af25d6fbd7b46c2dcf408937c509156c18463e/src/scenes/Partners/Detail/PartnerDetail.graphql#L47-L49).
Can solve by doing [this](https://github.com/SeedCompany/cord-field/blob/42af25d6fbd7b46c2dcf408937c509156c18463e/src/common/fragments/user.graphql#L4-L10).

CarsonF:
If
```gql
mutation {
  updatePartner(input: { partner: {
    id: "asdf"
    address: null
 }}) {
  partner {
    address { value }
  }
}
```
doesn't work then this we should fix whatever's happening in the service/repository file. Does this work?

CarsonF (approved):
I'm good with this GQL schema breaking change. As long as the UI is updated & deployed first to mitigate referencing this type directly. https://github.com/SeedCompany/cord-field/pull/1133

sethmcknight:

PR #2475, opened 2022-07-19
added experience-operations role functionality

atGit2021:
Bryan, please take a look at this initial draft pull request for enabling the ExperienceOperations role.  Fyi, I also had to edit two files in the cord-field repo to enable the ExperienceOperations role to be available in the UI dropdown selection, but those two files/directories are tagged in the git ignore settings.  
I have not yet addressed the renaming of Fundraising to RelationshipManager, but will target that after this functionality is good to go.


src/components/authorization/roles/experience-operations-global.role.ts
```diff
@@ -0,0 +1,473 @@
...
+// turned off prettier for role files to prevent auto-format making this file huge
+
+const read = true;
+const write = true;
+
+export const ExperienceOperations = new DbRole({
+  name: `global:${Role.ExperienceOperations}` as const,
+  powers: [
+    Powers.CreateEducation,
+    Powers.CreateProjectMember,
```
bryanjnelson, line 44: In looking at the spreadsheet Seth attached to the Issue, I'm unsure if the CreateProjectMember power should be allowed?

src/components/authorization/roles/experience-operations-on-project.role.ts
```diff
@@ -0,0 +1,473 @@
...
+
+const read = true;
+const write = true;
+
+export const ExperienceOperationsOnProject = new DbRole({
+  name: `project:${Role.ExperienceOperations}` as const,
+  powers: [
+    Powers.CreateEducation,
+    Powers.CreatePost,
+    Powers.CreateProjectMember,
```
bryanjnelson, line 45: Again, I'm unsure about these two from the spreadsheet...but I could be misreading.

bryanjnelson (commented):
Besides my two comments, we should get @CarsonF to take a look and make sure there aren't any _additional_ files that need touched when creating a new role like this.

src/components/authorization/roles/experience-operations-on-project.role.ts
atGit2021: The Powers.CreatePost already existed for Fundraising role, which is the base role/file used for ExperienceOperations. I then added the Powers.CreateProjectMember so this role has the ability to add a team member. We will need Seth or Carson to weigh in on whether this is the correct approach to enabling "the add team member" permission.

bryanjnelson: Okay.  I was basing my comments [off of this document](https://docs.google.com/spreadsheets/d/1Y5HPK3M2ECsreDqmR-Z9q7e2sJoL6WzD6low9N_aptc/edit#gid=0).  Specifically the rows at the very end which look like this:  

Cord_Schema_-_Properties_by_Base_Node_-_Google_Sheets

Perhaps that is not the proper place to look, however.

CarsonF: There's some confusion here. I intended Powers to only be used to for "global" actions. For example, creating Projects or Users, which don't exist inside another context. Project Members however are under/inside a Project, not a _global_ scope. So it was supposed to be authorized by the edit property of `Project.member`. You can see that's how we calculate the `canCreate` property of the team/member list, [here](https://github.com/SeedCompany/cord-api-v3/blob/ec33915ba0f9146d7968c05e73c9738c8cba1f2b/src/components/project/project.service.ts#L484-L484).
This is what the UI uses to determine whether the add team member button should be shown. However, on create the API currently doesn't use that same check and instead checks the power, [here](https://github.com/SeedCompany/cord-api-v3/blob/ec33915ba0f9146d7968c05e73c9738c8cba1f2b/src/components/project/project-member/project-member.service.ts#L91-L94). This doesn't really make sense as this action depends on which project is referenced.

In short that Power should be deleted, and the create method should be updated to check Project.member with the referenced project object. This is a common occurrence across the app that needs to be updated.

The Role definitions & AuthorizationService currently don't model this well. Every X object under project shouldn't need to be attached to project, but should be able to exist separately. I have a branch that drastically simplifies these role definitions, but naturally it's only 60% done.

atGit2021: Thank you Carson. Yeah, the permission/access functionality is still fuzzy for me, but I think I get your high level objective.  I'll delete the power and will look at the files you high lighted to see how it works and implement accordingly.

atGit2021: Updated files pertaining to ExperienceOperation Role permissions, per discussions with Carson and a call with Seth to reconcile the permissions on the xls.

bryanjnelson (approved):

CarsonF (commented):
LGTM

CarsonF (approved):

PR #2488, opened 2022-08-01
Update project and engagement workflow in changeset

### LINKED ISSUE #2480: Project/Engagement Workflow inside of Changesets
While inside of a changesets, the status/step workflow should be different.

These project steps:
- `DiscussingChangeToPlan`
- `PendingChangeToPlanApproval`
- `PendingChangeToPlanConfirmation`
- `DiscussingSuspension`
- `PendingSuspensionApproval`
- `DiscussingReactivation`
- `PendingReactivationApproval`
- `DiscussingTermination`
- `PendingTerminationApproval`

These engagement statuses:
- `DiscussingTermination`
- `DiscussingReactivation`
- `DiscussingChangeToPlan`
- `DiscussingSuspension`

These were/are temporary to help facilitate "change to plan" functionality until change requests could be delivered.

We need to update the available transitions for these workflows to skip these values and instead just go straight to the end result.
i.e.
```diff
- DiscussingSuspension -> PendingSuspensionApproval -> Suspended
+ Suspended
```

---

Obviously status change notifications shouldn't be triggered until the change request is approved.

--- 

Workflow around change request approval can happen separately.

zkhin:
Fixes #2480

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466524074) by [Unito](https://www.unito.io)


src/components/project/project.rules.ts
```diff
@@ -472,6 +472,11 @@ export class ProjectRules {
               type: TransitionType.Neutral,
               label: 'Will Not Change Plan',
             },
+            {
+              to: ProjectStep.ActiveChangedPlan,
+              type: TransitionType.Changeset,
```
CarsonF, line 477: It would probably be better to use a separate property here. The transitions inside of changesets still could have "types" so we don't want to block or overload that functionality. Maybe `changeset: true` or `inChangeset: true`.

CarsonF (changes requested):
I think there's some confusion here. The steps listed in the issue would be going away in the future. We need to add transitions to bypass those listed.
So for example, instead of `Active` to `DiscussingSuspension`, it would be replaced by `Active` to `Suspended`.

src/components/engagement/engagement.rules.ts
```diff
@@ -20,6 +20,7 @@ import {
 
 interface Transition extends EngagementStatusTransition {
   projectStepRequirements?: ProjectStep[];
+  inChangeset?: boolean;
```
zkhin, line 23: Add `inChangeset` prop and it will have three states now.

- `true`: transition will be available only in changeset mode
-  `false`: transition will be available not in changeset mode but in the normal workflow
-  `undefined`: transition will be available for both workflows.

CarsonF:
Hey @zkhin I made some changes here. Would you read through my commits and apply the same logic to the engagement workflow?

zkhin:
> Hey @zkhin I made some changes here. Would you read through my commits and apply the same logic to the engagement workflow?

Sure, thanks

sethmcknight:

PR #2489, opened 2022-08-02
Add `RichText` scalar type

### LINKED ISSUE #2484: RichText scalar type
I want to pursue rich text that leaves data in a JSON format (https://editorjs.io/). We need update our GQL API to support this. This just needs a Scalar that allows for JSON. There's several examples of prebuilt JSON fields, that can be used directly or used as an example. I just want to be specifically called `RichText` in schema so we are explicit with the JSON format.

I think the DB can handle JSON without any problems, so there's probably no change needed there. Also worth double checking our other code to see if it works, but I think it'll be fine.

The full text search functionality is really the only thing that's going to suffer. I think for now we can just ignore that. Maybe in future we can automatically save a plain text only version of the data that can be used for searching.

daniel-leod:
Closes #2484


CarsonF:
Superseded by #2494. Thanks for the effort @daniel-leod. I went with `graphql-scalars` library instead which has several scalars that could be used instead of just this single purpose one. Since you found out that neo4j can't handle JSON, I'm converting it to/from a string when going to DB. I also setup some dynamic resolvers to ensure these scalars are "used" in the schema to prevent compilation errors.

PR #2495, opened 2022-08-16
Add comments and comment threads

### LINKED ISSUE #2485: Comments & Comment Threads
These will be like Posts, except they won't have a type or shareability. The `body` will be [RichText](https://github.com/SeedCompany/cord-api-v3/issues/2484) type as well.
Functionality this should work similar to google docs comments.

Any `Resource` can made to be "Commentable" which will provide a list of comment threads, `[CommentThread]`. Each thread can have 1 or more comments.

This is different than Posts in that Comments are always attached to something else, where Posts can exists on their own (kinda). Posts might go away in the future.

daniel-leod:
Closes #2485


src/components/comments/dto/list-comment.dto.ts
```diff
@@ -0,0 +1,38 @@
...
+  PaginatedList,
+  SecuredList,
+  SortablePaginationInput,
+} from '../../../common';
+import { Comment } from './comment.dto';
+
+@InputType()
+export abstract class CommentFilters {
+  @Field()
+  readonly threadId: ID;
```
CarsonF, line 16: I think this filter should be internal, since we'll connect it manually in a resolver.

src/components/comments/comment-thread.repository.ts
```diff
@@ -0,0 +1,45 @@
...
+      .return<{ id: ID }>('node.id as id')
+      .first();
+  }
+
+  async list({ filter, ...input }: CommentThreadListInput, session: Session) {
+    const result = await this.db
+      .query()
+      .match(requestingUser(session))
+      .match(node('node', 'CommentThread'))
+      .apply(sorting(CommentThread, input))
```
CarsonF, line 40: Parent ID filter needs to be applied here.

src/components/comments/comment.service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  async update(input: UpdateComment, session: Session): Promise<Comment> {
+    const object = await this.readOne(input.id, session);
+    const changes = this.repo.getActualChanges(object, input);
+    await this.repo.updateProperties(object, changes);
+
+    return await this.readOne(input.id, session);
+  }
+
+  async delete(id: ID, session: Session): Promise<void> {
```
CarsonF, line 98: Let's delete the thread too if this is the only comment in the thread.

src/components/comments/comment.resolver.ts
```diff
@@ -0,0 +1,128 @@
...
+    @IdArg() id: ID
+  ): Promise<DeleteCommentOutput> {
+    await this.service.delete(id, session);
+    return { success: true };
+  }
+
+  @Mutation(() => CreateCommentThreadOutput, {
+    description: 'Create a comment thread',
+  })
+  async createCommentThread(
```
CarsonF, line 87: Instead of making the consumer call this only to turn around and call `createComment`. Let's merge the two. `createComment` can take a `resourceId` & optionally a `threadId`. If no threadId is given we can create the thread for them before creating the comment.
The output can be changed to have both the comment & the comment thread.

src/components/comments/comment.resolver.ts
```diff
@@ -0,0 +1,128 @@
...
+  CreateCommentOutput,
+  CreateCommentThreadInput,
+  CreateCommentThreadOutput,
+  DeleteCommentOutput,
+  UpdateCommentInput,
+  UpdateCommentOutput,
+} from './dto';
+
+@Resolver(Comment)
+export class CommentResolver {
```
CarsonF, line 38: I'm not sure we need all of these queries. I think really the only one that will be used in practice is `Commentable.commentThreads`, and we'll just traverse from there. What do you think?

src/components/comments/comment.repository.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  async list(input: CommentListInput, session: Session) {
+    const { threadId } = input.filter;
+    const result = await this.db
+      .query()
+      .match(requestingUser(session))
+      .match([
+        node('thread', 'CommentThread', { id: threadId }),
+        relation('out', '', 'comment'),
```
CarsonF, line 44:
```suggestion
        relation('out', '', 'comment', ACTIVE),
```

src/components/comments/dto/comment-thread.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+export class CommentThread extends Resource {
+  static readonly Props = keysOf<CommentThread>();
+  static readonly SecuredProps: string[] =
+    keysOf<SecuredProps<CommentThread>>();
+  static readonly Relations = {
+    comments: [Comment],
+  };
+
+  @Field(() => [Comment])
+  readonly comments: readonly Comment[];
```
CarsonF, line 18: I think it would be better if this was paginated as well.

src/components/comments/comment.resolver.ts
daniel-leod: Sounds good to me.

CarsonF (approved):
Alright I'm happy with this if you are @daniel-leod. I say we merge and start UI work.

daniel-leod:
LGTM!

PR #2502, opened 2022-08-22
Changeset editable & applied flags

CarsonF:
It's become apparent that whether a changeset is editable and/or applied are intrinsic properties of this interface. These low level flags help facilitate base functionality of changesets, regardless of specific implementation. 

I've also refactored the `ProjectChangesetFinalizedEvent` into `ChangesetFinalizingEvent`. Since the act of finalizing a changeset seems intrinsic as well. We currently don't have this as a flag, but maybe in the future?

With these changes the line is becoming clearer between what's the basic "changeset" functionality (staging of changes) and what's specific to ProjectChangeRequests (its own types & workflow/review process).


elisabeth-sorrell (approved):
LGTM

PR #2503, opened 2022-08-22
More robust changeset editability validation

CarsonF:
Replace changeset editability validation implementation

The [previous code](https://github.com/SeedCompany/cord-api-v3/blob/c9f4c419696b8c36aec57ec03b85abecde72426d/src/components/project-change-request/changeset-validation.plugin.ts#L30-L31) was guessing where a changeset argument would be,
and was defined by the consumer's operation. So easily defeated.

The new implementation validates based on what our API schema defines.
As long as input object uses the ChangesetIdField decorator it will be
validated in mutations regardless of key or depth.

This new implementation also uses the new interface property instead of
the ProjectChangeRequest concrete.


src/core/graphql/gql-execution-context.host.ts
```diff
@@ -0,0 +1,67 @@
...
+
+$ brew unlink node && brew install node@16 && brew link --overwrite node@16
+`);
+    }
+    throw new Error('The execution context has not been attached yet.');
+  }
+
+  intercept(context: ExecutionContext, next: CallHandler) {
+    const gqlContext = GqlExecutionContext.create(context);
+    return this.als.run(gqlContext, () => lastValueFrom(next.handle()));
```
CarsonF, line 61: This doesn't work because the interceptor only spans the query/mutation method, not the fields below it.
The info that this provides also doesn't makes sense across every field either as each would have different args, resolve info, etc.

So this needs to be re-thought.

CarsonF: Middleware spans the entire request, so I switched to that in conjunction with an apollo plugin. The middleware correctly scopes the async storage and holds a ref to an object which is hydrated "later" with the apollo plugin event.

PR #2506, opened 2022-08-25
Updates

CarsonF:
- Add path aliases for `~/common` & `~/core` folders
- TypeScript 4.5 → 4.7
- React 16 → 18 & new jsx runtime (no longer need to have react imports to use jsx)
- NestJS 8 → 9
- NestJS GraphQL 8 → 10
- Use `ts-patch` to unify TS transpiler plugins
- Switch to new maintained faker fork
- Import `gql` tag directly from `graphql-tag` instead of `apollo-server-core` re-export
  No need to depend directly on underlying server package.
- Add `yarn clean` script to drop temp/build files


CarsonF:
Too big. Splitting apart

PR #2507, opened 2022-08-25
parse and capture scripture references from book name or my notes

atGit2021:
Bryan, for your review. This is the code for capturing and parsing scripture references from the book column or my notes section when uploading the PnP planning and progress sheets.

### DIFF
--- a/src/components/product/product-extractor.service.ts
+++ b/src/components/product/product-extractor.service.ts
@@ -6,7 +6,6 @@ import { Cell, Column } from '../../common/xlsx.util';
 import { Downloadable } from '../file';
 import {
   findStepColumns,
-  getTotalVersesInRange,
   isGoalRow,
   isGoalStepPlannedInsideProject,
   isProgressCompletedOutsideProject,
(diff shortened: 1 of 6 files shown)


bryanjnelson (approved):
@atGit2021 - Looks good from my perspective; on to you @CarsonF!

(Note we still may need to handle the case where an existing line is updated.)

src/components/product/product-extractor.service.ts
```diff
@@ -114,14 +123,63 @@ const parseProductRow =
...
+        : (scriptureReferences = findScriptures(sheet.myNote(row)!))
+      : (scriptureReferences = findScriptures(bookName));
+
+    unspecifiedScripture =
+      scriptureReferences.length > 0 || !book
+        ? null
+        : {
+            book: book.name,
+            totalVerses: totalVersesToTranslate,
+          };
```
CarsonF, line 152: This seems a bit convoluted. We already know `parseScripture` will give us valid ranges otherwise we wouldn't be here. We can just sum up the count of the verses from the ranges and compare that to the total given.
```ts
const isKnown = sumedVerseCountFromRanges === sheet.totalVerses(row)!;
```
If known, we can just forward the scripture on plainly. Otherwise as unknown portion.
```ts
const scripture = isKnown ? parsedScripture : null;
const unknownPortion = isKnown ? null : { book: parsedScripture[0].start.book, totalVerses };
```

src/components/product/product-extractor.service.ts
```diff
@@ -13,9 +13,18 @@ import {
...
+  parseScripture,
+  ScriptureRange,
+  ScriptureRangeInput,
+} from '../scripture';
 import { ProductStep as Step } from './dto';
 
+type UnspecifiedScriptureType = {
+  book: string;
+  totalVerses: number;
+} | null;
```
CarsonF, line 27: Let's just import the existing `UnspecifiedScripturePortion` type instead

CarsonF (changes requested):
Currently this doesn't change the matching logic. I'm not sure if that was intended or not. I would be fine accepting this as a next step though.
To be clear `Gen 1:1-1:5` & `Gen 2:1-2:5` look like the same range with our current matching logic, which is obviously not true.

src/components/product/product-extractor.service.ts
atGit2021: Tried to resolve per the suggestion above, but in order to handle the additional check for scripture ranges in the myNotes column of the PnP more logic was needed.

bryanjnelson (approved):
Looks good to me; @CarsonF - wanna take a peek?

bryanjnelson (approved):
@CarsonF - You're turn!

CarsonF:
I made a couple changes.
- Moved the total verse function out of isGoalRow into a more appropriate spot.
- Refactored the scripture extraction logic. It's really hard to read logic that uses mutable variables with a bunch of nested conditions. The new logic only uses constants and has multiple returns to facilitate a flat structure of if/then logic.
- I replaced usage of lodash's `isEqual`. Using that feels dirty as it's comparing unknown shapes. It's definitely easy but I think it's just error prone. I replaced it with a comparison of the label (string) representing the ranges. This is more robust since it sorts and merges ranges before labelling, so as `[Matt 1, Matt 3, Matt 2] == [Matt 1-3]`

bryanjnelson (approved):
w00t!

atGit2021:
Thank you Carson/Bryan!

CarsonF (approved):

Issue #2479, opened 2022-07-26
Search by Eth Code and ROD #


sethmcknight:
Allow users with permissions to view `Eth Code` and `ROD #` to search by those values. Return `Languages` that match these values, and any `Projects` where those `Languages` have an `Engagement`

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466526968) by [Unito](https://www.unito.io)

PR #2518, opened 2022-09-01
Search by Eth Code and ROD

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2479

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466518087) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/language/language.repository.ts
+++ b/src/components/language/language.repository.ts
@@ -45,6 +45,19 @@ export class LanguageRepository extends DtoRepository<
   typeof Language,
   [session: Session, view?: ObjectView]
 >(Language) {
+  async languageIdByEthLangId(ethLangId: ID) {
+    const result = await this.db
+      .query()
+      .match([
+        node('node', 'EthnologueLanguage', { id: ethLangId }),
+        relation('in', '', 'ethnologue', ACTIVE),
+        node('language', 'Language'),
+      ])
+      .return<{ id: ID }>('language.id as id')
+      .first();
+    return result?.id;
+  }
+
   async create(input: CreateLanguage, ethnologueId: ID, session: Session) {
     const initialProps = {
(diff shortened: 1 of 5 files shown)


CarsonF (changes requested):
Hey @zkhin that's for your initiative here.

I believe if I remember correctly the current search can handle `EthnologueLanguage` BaseNodes. It's just that they aren't being included the BaseNode labels type filter or have a hydrator. This would be better than adding another GQL/DB query, as another query pushes the burden to the consumer for what's really an API problem.

I was able to quickly test and confirm that my theory. Here's the actions we should take:
- `EthnologueLanguage` could be added to [`searchableAbstracts`](https://github.com/SeedCompany/cord-api-v3/blob/27f448cf7b583dcc9d082aafaf1e678faacf8c49/src/components/search/dto/search-results.dto.ts#L71-L71) 
- Adjust [`types`](https://github.com/SeedCompany/cord-api-v3/blob/27f448cf7b583dcc9d082aafaf1e678faacf8c49/src/components/search/search.service.ts#L56-L56) to also include "EthnologueLanguage" if searching for "Language".
- Add to [`hydrators`](https://github.com/SeedCompany/cord-api-v3/blob/27f448cf7b583dcc9d082aafaf1e678faacf8c49/src/components/search/search.service.ts#L37-L37) a function to hydrate the `Language` object from an EthnologueLanguage BaseNode/id
- Adjust the [permission check logic](https://github.com/SeedCompany/cord-api-v3/blob/27f448cf7b583dcc9d082aafaf1e678faacf8c49/src/components/search/search.service.ts#L104-L113) to work with this case.

sethmcknight:

Issue #2526, opened 2022-09-19
Intern Project should allow Completion when at Finalizing Completion


bryanjnelson:
We had a Financial Analyst in a situation where an Intern Project was in Finalizing Completion, as well as the single Internship Engagement, but he did not have the option to Complete the Project.  The option should be available and should automatically advance the Internship Engagement to Complete as well.

CarsonF:
I think we should just return the engagement statuses from that DB query used and then exclude engagement's with `FinalizingCompletion` status from that check.

PR #2527, opened 2022-09-20
Adjust finalize completion

zkhin:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2526

### DIFF
--- a/src/components/project/project.rules.ts
+++ b/src/components/project/project.rules.ts
@@ -13,7 +13,7 @@ import {
 import { ConfigService, DatabaseService, ILogger, Logger } from '../../core';
 import { ACTIVE, INACTIVE } from '../../core/database/query';
 import { Role, withoutScope } from '../authorization';
-import { EngagementService } from '../engagement';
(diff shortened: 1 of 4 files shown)


CarsonF (changes requested):
Hey @zkhin the check should remain for other statuses. It's only the FinalizingCompletion status we want to exclude. Since that status is automatically changed to Completed when the project is. We'd still want this transition to be disabled if say the engagement was Active. Does that make sense?

src/components/project/project.rules.ts
```diff
@@ -746,7 +746,9 @@ export class ProjectRules {
           ],
         };
       case ProjectStep.FinalizingCompletion:
-        const disabled = await this.engagements.hasOngoing(id);
+        const disabled = await this.engagements.hasOngoing(id, [
+          EngagementStatus.FinalizingCompletion,
```
zkhin, line 750: Added `excludes` option to exclude `EngagementStatus.FinalizingCompletion`

CarsonF (approved):
I like it, thanks @zkhin!

PR #2529, opened 2022-09-26
Removed In Development origin restriction for engagement status updat…

bryanjnelson:
…es when reacting to an Active project status change.


CarsonF (approved):
yaya

PR #2530, opened 2022-09-26
Added validation for the existence of primary location when a project…

bryanjnelson:
… advances to Pending Finance Confirmation.


CarsonF (approved):
:shipit:

PR #2531, opened 2022-09-26
Authorization: Policies & Privileges

CarsonF:
# Old system: `DbRole`& `AuthorizationService`

Previously we only defined property level permissions with `DbRole` ([example](https://github.com/SeedCompany/cord-api-v3/blob/293ab488074bc1c7350282247875cdaa5a794ac9/src/components/authorization/roles/project-manager-global.role.ts)). For example, a role can read or write a `Project.name`. After some time with this system, we've experienced some shortcomings.
1. Every property had to be declared even though in most cases a given role had the same permissions for all properties
2. Similarly every new property added to the API need to be added all of the role definitions as well, almost always following what the other properties declared.
3. The trickiest is how do we know which objects to return in a list. We created a `canList` boolean that existed at the object level in the role definitions to solve this. But really what this indicated is whether you had permission to read the object, though it's only been applied in list queries.
4. To declare permissions that should apply when the role is a member of a project, we had an entirely separate role definition. This duplicated about 80% of what was already declared in the "global" definition.
5. These role definitions were basically one for one to roles. We didn't have a way to declare X role gets Y role permissions + these extra ones.
6. We have a `Power` enum that was originally meant to be used to let the UI know what permissions a user had when they couldn't be attached to other objects. So for example a user has the `CreateProject` power so they can create a project. However, `CreateEngagement` doesn't make sense because the UI should look to each project to define that `project.engagements.canCreate` since that could change based on which project you are referencing. There was also a bunch of other values that got muddled in as well, like `Grant*Role`. Previously these were individually assigned and saved in the DB, so they were an implementation detail and shouldn't have been exposed.
7. Delete permission has been in limbo for forever. A `canDelete` boolean was stubbed in role definitions, but it was never referenced by app logic. We also have a `DatabaseService.checkDeletePermission` which is also stubbed.
8. Every model/resource also defined a `Db*` version to use with these DbRole definitions which duplicated what the properties the API declared.

All in all the old system was verbose, not extendable, ridged, and not cohesive.  

# New system: `Policies` & `Privileges`

`Policies` (replacing `DbRoles`) are the system to declare permissions.
`Privileges` (replacing `AuthorizationService`) is the API that uses policies to determine permissions for resources.

Here's an example policy to use to show off new features.
```ts
import { member, Policy, Role, sensOnlyLow, sensMediumOrLower } from '../util';

@Policy([Role.FinancialAnalyst, Role.LeadFinancialAnalyst], (r) => [
  r.Budget.read.when(member).edit,
  r.Ceremony.read,
  r.Engagement.read.specifically((p) => [
    p.many('disbursementCompleteDate', 'status').when(member).edit,
  ]),
  r.Language.read.specifically((c) => c.locations.when(sensOnlyLow).read),
  r.Organization.whenAny(member, sensMediumOrLower).edit.or.create,
  r.Partner.edit.create
    .specifically((p) => [
      p.many('organization', 'pointOfContact').when(sensMediumOrLower).read,
    ])
    .children((c) => c.posts.edit),
])
export class ExamplePolicy {}
```

The biggest goal of this journey was to simplify permissions via a smarter, expressive system. Hopefully that's evident above.

Let's go through some of it.

## Imports/Types

At a high level this looks very different than the old role definitions. The new API does not require importing every object nor does it require a separate Db* class.

## Registering

```ts
@Policy(...)
export class ExamplePolicy {}
```
First we define a class with the Policy decorator. This class is then registered with the app via a module's `providers`, same as every thing else. Thus it can live in any part of the codebase, and not limited to the one folder previously. Currently all the ported policies are still in one spot, but it might make sense to have odd balls in future.

## Roles

```ts
[Role.FinancialAnalyst, Role.LeadFinancialAnalyst]
```
Next is the roles that this policy will apply to, in our example it applies to both Financial Analysts and Lead Financial Analysts.
As in the example, and our actual use case, it makes sense for the Lead FA to get all permissions the FA does.

Along these lines a role can be used in multiple policies and the new system will merge/union permissions for the role across all policies when checking, so there's no need to cram everything into one now.

Instead of specifying roles you can also say `"all"` so that the policy will apply to all roles.
[Here's an example](https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/authorization/policies/by-feature/read-util-objects.policy.ts), where we give read permission for Locations and Producibles (Stories, Songs, EthnoArt, etc) to everyone. This way we don't have to do this in every role's policy.

## Interface Permissions

That leads into the next feature, permissions for _interfaces_. Above we referenced `Producible` and in the example we reference `Engagement`. Both of these are interfaces for other concrete types. i.e. `LanguageEngagement` & `InternshipEngagement`. The new system is aware of these hierarchies and will apply the permissions given for interfaces to all of its implementations. This way we don't need to specify permissions for every implementation, if they are all the same, which is often the case for us.

Note that this intentionally only happens when the implementation is not specified in the policy. If it is, then this "inheritance" doesn't happen and all permissions need to specified manually.
I'm open to debate on this logic.

## Object level Permissions

```ts
r.Ceremony.read
```

The biggest change for what's available when reading these policies is object level permissions. As stated in # 3 above in old system, we didn't have this, only property level permissions. Now we have permissions for objects as well. This `read` replaces the `canList` flag from old system.

Aside from these object level permissions being available, we still support property level permissions. By default, any properties not specified will use the object level permissions. This saves us so much work when defining these policies.

## Property level Permissions

Property permissions can be declared with the `.specifically` feature.
```ts
r.Engagement.read.specifically((p) => [
  p.disbursementCompleteDate.edit,
]),
```
This says the policy can edit an Engagement's disbursementCompleteDate and read everything else.

In several of our polices we have 
```ts
r.Partner.read.specifically((p) => p.pointOfContact.none)
```
Meaning the policy cannot read pointOfContact but can read everything else.

## Relationships

Previously it if you wanted to declare read permission for a Project's engagements, you'd do that similar to properties on DbProject grants. This could be translated to what we have above. However, it becomes confusing with object level permissions. i.e. Should I say "Project.engagements.read" or "Engagement.read"? What does the latter mean?

Upon much reflection, there's actually 3 different kinds of relationships:

### Relationships to other "global" objects

These don't have objects being created or deleted, just a connection being made. An example of this is `Language.locations`. One can _read/edit_ this list/property but _creating/deleting_ is completely separate.
Because of this, these kinds of relationships are treated the same as properties; they are specified with `.specifically` and inherit from object level permissions.

These relationships, as all relationships, are made known via the `Relations` static property on the resource class.
https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/language/dto/language.dto.ts#L86-L90

### One-to-Many relationships to "child" objects

This is the most common. i.e.`Project.engagements`. This "list" is sourced from the existence of a child object. Thus these can be _created/deleted_, but an _edit_ action doesn't make sense.

A relationship is still declared in the `Relations` map, but additionally the owning resource declares it is a "child" or "not global" via a `Parent` static property.
https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/engagement/dto/engagement.dto.ts#L53-L53

These are declared via a separate `.children` feature in Policies.
```ts
r.Project.children(c => c.engagements.read.create.delete)
```
However, going back to the original question, how is this different than
```ts
r.Engagement.read.create.delete
```
Well I have logic in place to make these the same. Children relations will default to the object level permissions of their own type, _not_ the object level permission of the object owning the relation. So for example.
```ts
r.Engagement.read,
r.Project.edit,
```
One can read & edit the project properties, but it can only read its engagements, it cannot create or delete them (or "edit the list of them").

I encourage to keep the declarations flat, i.e. prefer `r.Engagement.read` vs `r.Project.children(c => c.engagements.read)`
So why have the `.children` feature at all?
This does makes sense for objects whose parent is dynamic. i.e. Posts, Comments, PrayerRequests.
So as an example, we want to be able to specify the one can create a comment/etc on a project, but not on a language.

### One-to-One relationships to "child" objects.

This is a bit of an edge case. i.e. `Project.budget`. There's potential for these to need an _edit_ action as the relationship could be swapped out, and there's potential for these to not need a _create/delete_ action. They work the same way as One-to-Many above, just with different allowed actions. I won't say too much more here as it's subject to discovery/change/iteration.

## Declaring Create Power

Previously powers were grant in the old system. Now I envision these powers only being referenced by GQL consumers. 
The only powers really used were "create X object" & "grant x role" (more on that later).

Instead of using a `CreateX` enum, which has to be manually updated for every resource, we just use an object level permission.
```ts
r.Partner.create
```
When checking for this, it is done the same way as a read action. So this is more cohesive and simpler to maintain.

## Conditions

Conditions are a huge part of this new system. The two conditions we currently have are a sensitivity level, and whether there is membership. These were an after thought for the old system. Declaring permissions for only membership was a pain, and the `AuthorizationService` was aware of both these conditions and manually handled them.

With the new system we have a conditions interface and every action (read/edit/etc) can be conditional based on a given condition. The new system does not know about membership or sensitivity, but instead leaves those conditions to be defined independently. This gives way to other future conditions, like ownership or certain statuses for objects. 

Conditions are attached in policies via a `.when` call. Any actions following that condition will have the condition applied.
```ts
r.Budget.read.when(member).edit,
```
So here we say the budget can always be read, but can only be _edited_ if one is a member.

The `.or` call will reset any previous conditions.
```ts
r.Organization.when(member).edit.or.create,
```
So this means one can _edit_ when a member but they can always _create_.

Multiple conditions can be attached with AND boolean logic or OR boolean logic.
```ts
r.Organization.whenAny(member, sensMediumOrLower).edit
```
Here one can _edit_ when they are member or if they aren't a member if the sensitivity is Medium or Low.

```ts
r.Partnership.whenAll(member, sensOnlyLow).read
```
Here a partnership can only be read when one is a member and the sensitivity is Low. ([source](https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/authorization/policies/by-role/consultant.policy.ts#L18-L18))

# Privileges

`Privileges` is the new service for reading these policies.

This service is just a launching point for the other classes. To know what privileges/permissions are available we need to know several things: the user requesting via `session`, the resource type, and optionally a specific instance of the resource for use with conditions.
Instead of asking for all of those at once there are provided classes to give partial info. So for example these all do the same thing.
```ts
Privileges.forUser(session).forResource(Project)
Privileges.forResource(Project).forUser(session)
Privileges.for(session, Project)
```

Here are some common tasks:

Can a user create a project?
```ts
Privileges.for(session, Project).can('create') // => boolean
```

I also created a shortcut to throw an `UnauthorizedException` if `can` is false, since this is a common use case.
```ts
Privileges.for(session, Project).verifyCan('create') // => void / throws error
```

Can a user read projects?
```ts
Privileges.for(session, Project).filterToReadable()
```
this provides a DB query fragment to inject a where clause that executes conditions based on what's needed for user.
In use it's
```ts
.query()
// match & user filter
.apply(this.privileges.for(session, Project).filterToReadable())
// then sorting, paginate
```

Can a user read a project's engagements?
While this technically works the same as above, specifying the "edge" is more precise.
```ts
Privileges.forEdge(Project, 'engagements').forUser(session).filterToReadable()
```

Can a user create a project's engagements?
```ts
Privileges.forUser(session).forEdge(Project, 'engagement', project).can('create') // verifyCan is available too
// another way:
Privileges.for(session, Project, project).can('create', 'engagement')
```

There's also an "object like view" of the permissions accessed via the `.all` getter.
```ts
const perms = Privileges.for(session, Project, project).all;
perms.engagement.create; // => boolean
perms.name.read; // => boolean
perms.name.edit; // => boolean
```
This doesn't yet have the object level permissions though. It mainly exists to support compatibility with the old system.

# Securing Objects

Old way
```ts
const securedProps = await this.authorizationService.secureProperties(
  Unavailability,
  dto,
  session
);

return {
  ...dto,
  ...securedProps,
  canDelete: await this.repo.checkDeletePermission(dto.id, session),
};
```
New way
```ts
return this.privileges.for(session, Unavailability).secure(dto);
```

# Assignable Roles

Granular authorization for assigning roles now exists. Previously this was half implemented with `Power.Grant*Role` but not checked and thus never used.

Now policies can specify of which roles they can apply to other users.
```ts
@Policy(Role.ProjectManager, (r) => [
  ...

  Role.assignable(r, [
    Role.Intern,
    Role.Liaison,
    Role.Mentor,
    Role.RegionalCommunicationsCoordinator,
    Role.Translator,
  ]),

])
```
This is accessed in GQL via 
```gql
user {
  roles {
    assignableRoles # list of available roles that could be assigned
    value # currently assigned roles
  }
}
```

Connecting logic [here](https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/user/user.service.ts#L184-L204).

# What's next?

- Confirm the new policies match business expectations
- Remove usage of `AuthorizationService`
- ~~There's a `secureProps` privilege method that is supports the existing "secure object" logic.
   I'd like to create & migrate to a new one that incorporates `{ canDelete: privileges.can('delete') }`.~~ ✅ 
- ~~An owner condition should be created. It could at least be specific to `User` to cover [this](https://github.com/SeedCompany/cord-api-v3/blob/refactor%2Fauthorization/src/components/user/user.service.ts#L119-L128).~~ ✅ 
- ...
- Profit?


CarsonF:
:shipit:

PR #2536, opened 2022-10-07
Various policy refactoring and updates.

bryanjnelson:
- Solidified policies with the help of Seth
  - Removed unused Liaison & Regional Communication Coordinator roles
- Overall refactoring
  - Removed top level Post & FileNode object entries since they make no sense in isolation
  - Alphabetized all entries
  - Removed all ChangeRequest object entries (until functionality is fully rolled out)
  - Tweaked e2e tests

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466510928) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3466510928


src/components/authorization/policies/by-role/project-manager.policy.ts
```diff
@@ -1,57 +1,59 @@
...
+      .specifically((p) => p.pmcEntityCode.none)
+      .children((c) => c.posts.read.create),
+    r.Partnership.whenAny(member, sensMediumOrLower).read.create.specifically(
+      (p) => [
+        p.many('agreement', 'agreementStatus', 'types', 'partner', 'primary')
+          .edit,
+      ]
+    ),
+    r.PeriodicReport.read.when(member).edit,
+    r.Producible.when(member).edit.create,
```
CarsonF, line 43:
```suggestion
    r.Producible.edit.create,
```
Producibles are global objects. They are only related via `Product.produces`.

src/components/partnership/partnership.module.ts
```diff
@@ -13,8 +16,11 @@ import { PartnershipService } from './partnership.service';
 @Module({
   imports: [
     forwardRef(() => AuthorizationModule),
-    FileModule,
     forwardRef(() => BudgetModule),
+    forwardRef(() => EngagementModule),
+    FileModule,
+    forwardRef(() => LocationModule),
+    ProjectChangeRequestModule,
```
CarsonF, line 23: I don't think should need any of this

src/components/project/project.module.ts
```diff
@@ -51,6 +51,11 @@ import { ProjectUserConnectionResolver } from './user-connection.resolver';
     ...Object.values(handlers),
     ...Object.values(migrations),
   ],
-  exports: [ProjectService, ProjectMemberModule, ProjectRules],
+  exports: [
+    ProjectService,
+    ProjectMemberModule,
+    ProjectRules,
+    ProjectRepository,
```
CarsonF, line 58: Where is `ProjectRepository` being used outside of this module?

src/components/authorization/policies/by-feature/user-can-manage-own-comments.policy.ts
```diff
@@ -1,8 +1,8 @@
 import { owner, Policy } from '../util';
 
 @Policy('all', (r) => [
-  r.Post.when(owner).edit,
+  r.Post.when(owner).edit.delete,
   r.CommentThread.when(owner).edit,
   r.Comment.when(owner).edit,
```
CarsonF, line 6: Want to add delete here too?

src/components/authorization/policies/by-role/fundraising.policy.ts
```diff
@@ -2,46 +2,29 @@ import { member, Policy, Role, sensMediumOrLower } from '../util';
...
-  r.Engagement.read,
   r.Language.read.specifically((p) => [
     p.registryOfDialectsCode.whenAny(member, sensMediumOrLower).read,
     p.signLanguageCode.whenAny(member, sensMediumOrLower).read,
     p.locations.whenAny(member, sensMediumOrLower).read,
   ]),
-  r.Organization.whenAny(member, sensMediumOrLower).read.specifically((p) => [
+  r.Organization.whenAny(sensMediumOrLower).read.specifically((p) => [
+    //TODO - this none can be removed when policy executor is refactored to check for address.none in parent policy
     p.address.none,
```
CarsonF, line 13: This is tricky. With more thought, I don't think this should just be allowed without noting something in the other policy. Otherwise we break the guideline of not knowing if X line will actually apply and not be overridden by another.

Currently if not specified all props fallback to resource level perms. I think this should still be the default as it's probably the more common case. So we need a flag to turn off this default.
At first I thought of something like
```ts
r.Organization.noProps
```
which would set all properties to `none`. 
The same as
```ts
r.Organization.specifically(p => p.many(...keys(p)).none
```
But I don't think this is entirely what we want, as we still want to fallback to resource level if no other policy specifies any specific prop permissions.

So instead we need a flag and an update to the executor logic.
```ts
r.Organization.deferProps
```
And Executor does this:
1. Policy A - checks for prop, sees this skips
2. Policy B - checks for prop. If exists, uses that; else, continue
3. Policy A - uses resource level

Thoughts? Thoughts on `deferProps` as name of flag?

src/components/engagement/engagement.service.ts
```diff
@@ -620,4 +623,9 @@ export class EngagementService {
       );
     }
   }
+
+  async privilegesFor(id: ID, session: Session) {
+    const engagement = await this.repo.readOne(id, session);
+    return this.privileges.for(session, IEngagement, engagement);
+  }
```
CarsonF, line 630: I'm not sold on this. It's clunky.
The only reason this is here is to access `repo.readOne`, which gives us an unsecured version of the object to provide to the privileges.

This thing is called "object" in lots of places, but I've started transitioning to call it "context". The actual shape of it is very unclear. Its only purpose is to provide info to conditions. 

Currently these conditions don't use any secured info, like `scope`, `sensitivity`.
There's potential for usage like `Project.step` condition, which is secured. So if the condition needed it but the user couldn't read it then what happens. Admittedly this is a super edge case. 
And could be handled like user cannot read, then condition is false regardless of value. That would probably be the easiest, and I don't think we limit read access to this prop anyways.
Another option would be to provide the actual value internally even when publicly the user has no read access.

There's also more and more situations where the context needs to be customized for edge cases, using the `with*` helpers. I think that needs some refinement as well.

So the need for this context to be the unsecured object is questionable and I'm not even sure it should be typed that strictly.

---

So we could do something like
```ts
const engagement = await ResourceLoader.load(IEngagement, id);
const privileges = Privileges.for(session, IEngagement, engagement);
```
Which is already setup and doesn't need another boilerplate method.
(Given the type is adjusted in `Privileges`).

src/components/project/project-member/project-member.service.ts
```diff
@@ -85,17 +90,23 @@ export class ProjectMemberService {
...
+    const projectId = isIdLike(projectOrId) ? projectOrId : projectOrId.id;
+    const project = isIdLike(projectOrId)
+      ? await this.projectService.readOneUnsecured(projectOrId, session)
+      : projectOrId;
+
+    this.authorizationService.privileges
+      .for(session, IProject, project)
+      .verifyCan('create', 'member');
+
+    // session.roles
```
CarsonF, line 105:
```suggestion
```

CarsonF (changes requested):
I gotta fairly far in this. Made some notes for you and myself.

src/components/authorization/policies/by-role/fundraising.policy.ts
```diff
@@ -2,46 +2,29 @@ import { member, Policy, Role, sensMediumOrLower } from '../util';
...
+    p.locations.when(sensMediumOrLower).read,
```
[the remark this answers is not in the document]
bryanjnelson: So in this case then it is defaults to the object level Organization object, is what you're saying.

But what about in investor-common.policy.ts - https://github.com/SeedCompany/cord-api-v3/blob/69d60d0dbb3f4e70e5ae41ada4b0b9332b0ff6c1/src/components/authorization/policies/by-role/investor-common.ts#L29-L32 ?  Is line 31 necessary here, or is that inherited from the Organization object?  If it is, then wouldn't it inherit the `member` restriction as well?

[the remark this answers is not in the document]
CarsonF: Yes it defaults; it does not merge. So line 31 is necessary to remove the member condition from `locations`.

src/components/partnership/partnership.module.ts
```diff
@@ -13,8 +13,8 @@ import { PartnershipService } from './partnership.service';
 @Module({
   imports: [
     forwardRef(() => AuthorizationModule),
-    FileModule,
     forwardRef(() => BudgetModule),
+    FileModule,
```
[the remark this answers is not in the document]
bryanjnelson: I left it alphabetical, since it wasn't before.  But it still isn't anyway 'cause the last two are switched.  I'll revert it.

src/components/authorization/policies/by-role/fundraising.policy.ts
CarsonF: Upon further reflection, I think this is the edge case. Instead we should be explicit when we want all resource level permission to apply to all properties. The flag could be something like `.andAllProps`?

bryanjnelson:
@CarsonF - Okay, at this point the e2es are fixed and the items you suggested I change are completed.  The 3 comments you have above that are left (I believe) are on your plate for design whenever you get a chance, and then I can implement.

src/components/authorization/policies/by-role/fundraising.policy.ts
bryanjnelson: Handled via [2562](https://github.com/SeedCompany/cord-api-v3/pull/2562)

CarsonF (approved):
Looks great @bryanjnelson!! 🎉

PR #2539, opened 2022-10-12
Change `member` condition to ignore roles

CarsonF:
The logic in the old system was a user with X role "scoped" to project membership. As the new auth system unwound this logic, it makes more sense to completely separate these concepts. Policies are only considered if the user has the role globally and then if they are a member at all they get X permissions. I've confirmed this with @sethmcknight so here it is.

To facilitate this I added a `member:true` to the `ScopedRole` type and returned it in [`matchProjectScopedRoles`](https://github.com/SeedCompany/cord-api-v3/compare/improve/member-condition#diff-2a99be5f26d1bb73f22d61738715b3c13760d0601500a7f14f1074fa05e0f7a3).
When executing the condition in cypher it's easier too since we can just check for a path existence.

---

I also fixed the stack trace for trailing condition errors when using `.when` wrappers.

---

I also allowed policies to return multilevel arrays to help facilitate abstractions.
This now works without having to spread each line:
```ts
@Policy('all', r => [
  inherit(...),
  Object.values(r),
])
```
The less syntax the better in these policies.

### DIFF
--- a/src/components/authorization/dto/role.dto.ts
+++ b/src/components/authorization/dto/role.dto.ts
@@ -43,7 +43,7 @@ export type AuthScope = GlobalScope | ProjectScope;
 export type ProjectScopedRole = `${ProjectScope}:${Role}`;
 export type GlobalScopedRole = `${GlobalScope}:${Role}`;
 
-export type ScopedRole = `${AuthScope}:${Role}`;
+export type ScopedRole = `${AuthScope}:${Role}` | 'member:true';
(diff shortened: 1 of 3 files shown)


CarsonF:
This is actually really interesting. If the `memberWith` condition (which checks for membership + roles) continues to not be used, then membership roles no longer play a part in authorization. They merely serve as a title or connection for other logic like notifications.

PR #2542, opened 2022-10-14
added parent to comments list

andrewmurraydavid:
When working on the RichText feature for the UI (https://github.com/SeedCompany/cord-field/issues/1150) we ended up needing to pass the parent of a comments list.

This change will enable proper caching in the UI application.


CarsonF (approved):
LGTM! Thanks @andrewmurraydavid

PR #2545, opened 2022-10-18
Attempt to handle large product queries more gracefully.

bryanjnelson:
- Create/handle new Session Expired Error
- Lower max batch size of product lower to 25 from 100

### DIFF
--- a/src/components/product/product.loader.ts
+++ b/src/components/product/product.loader.ts
@@ -30,8 +30,7 @@ export class ProductLoader extends OrderedNestDataLoader<AnyProduct> {
 
   getOptions(): OrderedNestDataLoaderOptions<AnyProduct> {
     return {
-      // Increase the batching timeframe from the same nodejs frame to 10ms
-      batchScheduleFn: (cb) => setTimeout(cb, 10),
+      ...super.getOptions(),
       maxBatchSize: 25,
     };
   }
--- a/src/components/product/product.loader.ts
+++ b/src/components/product/product.loader.ts
@@ -1,5 +1,9 @@
 import { ID } from '../../common';
-import { LoaderFactory, OrderedNestDataLoader } from '../../core';
+import {
+  LoaderFactory,
+  OrderedNestDataLoader,
+  OrderedNestDataLoaderOptions,
+} from '../../core';
 import {
   AnyProduct,
   DerivativeScriptureProduct,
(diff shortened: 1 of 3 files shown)


src/components/product/product.loader.ts
```diff
@@ -23,4 +27,12 @@ export class ProductLoader extends OrderedNestDataLoader<AnyProduct> {
   async loadMany(ids: readonly ID[]) {
     return await this.products.readMany(ids, this.session);
   }
+
+  getOptions(): OrderedNestDataLoaderOptions<AnyProduct> {
+    return {
+      // Increase the batching timeframe from the same nodejs frame to 10ms
+      batchScheduleFn: (cb) => setTimeout(cb, 10),
```
CarsonF, line 34:
```suggestion
      ...super.getOptions(),
```

src/core/database/errors.ts
```diff
@@ -55,6 +55,27 @@ export class ServiceUnavailableError extends Neo4jError {
   }
 }
 
+export class SessionExpiredError extends Neo4jError {
+  static readonly code = 'ServiceUnavailable' as const;
```
CarsonF, line 59:
```suggestion
  static readonly code = 'SessionExpired' as const;
```
Also need to call `fromNeo` in the `cast` function below.

src/core/exception.filter.ts
```diff
@@ -142,6 +146,12 @@ export class ExceptionFilter implements GqlExceptionFilter {
         message: 'Failed to retrieve data from CORD database',
       };
     }
+    if (exs.some((e) => e instanceof SessionExpiredError)) {
+      return {
+        codes: ['SessionExpired', 'Transient', 'Database', 'Server'],
+        message: 'The CORD database session has expired',
```
CarsonF, line 152:
```suggestion
        message: 'The query to the database has expired',
```
maybe this would be better?

CarsonF (changes requested):

bryanjnelson:
@CarsonF - I'll squash after your review.

src/core/database/errors.ts
```diff
@@ -55,6 +55,27 @@ export class ServiceUnavailableError extends Neo4jError {
   }
 }
 
+export class SessionExpiredError extends Neo4jError {
+  static readonly code = 'SessionExpiredError' as const;
```
CarsonF, line 59:
```suggestion
  static readonly code = 'SessionExpired' as const;
```

CarsonF (commented):
Close ;)

src/core/database/errors.ts
bryanjnelson: 🤦🏼‍♂️

CarsonF (approved):
Thanks man!

bryanjnelson:
@CarsonF - it's forcing another review 😕

CarsonF (approved):

PR #2548, opened 2022-10-25
Added new progress report beta feature

andrewmurraydavid:
Create NewProgressReports beta feature for the new Progress/Quarterly reports for the UI


CarsonF (approved):
Looks great. I'm not sure who we want to show this too at this point. But FYI the Administrator role automatically gets all beta features. Obv this would be better by user by env, maybe in future we can do real feature flags.

PR #2550, opened 2022-10-26
Commentable GQL Query + more

CarsonF:
- Add GQL query `commentable`, to load a `Commentable` resource
  This is better for UI instead of `commentThreads` query as the the list will always be automatically associated with the parent object; instead of having to manually sync changes to the cached queries for `commentThreads`.
- Throw a new client error (`InvalidIdForType`) when resource id given is not a `Commentable` resource.
   Previously, a server schema error was thrown, but this is a client problem.

Other changes:
- Migrate `CommentService` to use `Privileges`
- Update `ResourcesHost.getInterfaces` to use GQL schema when it's available (web server, not CLI/REPL)
- Change `ResourcesHost.getByName` to return an `EnhancedResource`

### DIFF
--- a/src/common/exceptions/index.ts
+++ b/src/common/exceptions/index.ts
@@ -6,3 +6,4 @@ export * from './not-implemented.exception';
 export * from './unauthenticated.exception';
 export * from './unauthorized.exception';
 export * from './service-unavailable.exception';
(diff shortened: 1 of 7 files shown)


andrewmurraydavid (approved):
LGTM 👍

PR #2551, opened 2022-10-26
ProgressReport Workflow & prep for other new functionality

CarsonF:


CarsonF:
This is a soft breaking change. In that the typenames changed but the fields have not. The UI does not reference these names as of https://github.com/SeedCompany/cord-field/commit/958f2a38

Issue #2547, opened 2022-10-21
Issue in GQL Users Query


bryanjnelson:
When executing the following GQL Query:

![Explorer!UNITO-UNDERSCORE!!UNITO-UNDERSCORE!!UNITO-UNDERSCORE!CORD!UNITO-UNDERSCORE!prod!UNITO-UNDERSCORE!!UNITO-UNDERSCORE!!UNITO-UNDERSCORE!Studio](https://user-images.githubusercontent.com/1914742/197270452-8face9c3-bac3-4623-a316-29cbac9a22a6.png)

this happens:

```
TypeError: Cannot read properties of null (reading 'constructor')
    at src/core/database/query/filters.ts:41:26
    at Query.apply (src/core/database/query-augmentation/apply.ts:39:15)
    at UserRepository.list (src/components/user/user.repository.ts:204:8)
    at UserService.list (src/components/user/user.service.ts:216:43)
    at UserResolver.users (src/components/user/user.resolver.ts:130:18)
```

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466509824) by [Unito](https://www.unito.io)

PR #2556, opened 2022-11-03
Nullable user filters

zkhin:
https://github.com/SeedCompany/cord-api-v3/issues/2547

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3472816967) by [Unito](https://www.unito.io)


CarsonF (approved):
Kinda annoying but I see why this is necessary. The framework doesn't know if we want to keep the explicit null or use the default.

I like the transform here as it is right next to where the default value is applied (in schema).

Would you apply this logic to all of our other filters?

CarsonF:
Superseded by #2709

PR #2557, opened 2022-11-07
Allow secured objects to be used as context for Privileges

CarsonF:
This limitation was artificial and in practice irrelevant.
Our current usage of context is mainly `sensitivity`, `scope`, which are not secured.
If we needed other secured properties in the future, we could create an internal property to house this value to pass around and use internally.
Odds are even the secured properties that could be needed, like `Project.step`, will still be readable in every place that matters.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3489703307) by [Unito](https://www.unito.io)
┆Link To Item: https://seed-company-squad.monday.com/boards/3451697530/pulses/3489703307


bryanjnelson (approved):
Thanks!

PR #2562, opened 2022-11-10
Rethink Policy Props Logic

CarsonF:
# Problem Scenario
Currently we have this edge case scenario:
```ts
// Policy A
r.Organization.read
// Policy B
r.Organization.specifically(p => p.address.none)
```
If policy `A` is executed first then the user has `read` permission for `address`.
If policy `B` is executed first then the user has **no** permission for `address`.
So that's no good, we obviously what these to be deterministic.

This brings up what do we mean by the policy `A` line.
Currently it means the user can `read` `Organizations` _and_ all of its properties (since they aren't specified).
But perhaps in this case we wanted to defer to other policies so that policy `B`'s line applies.

# Logical Changes
So this PR changes a few things, in attempt to address these problems.
1. We now exhaustive check all props first, before falling back to resource level.
    - Before we'd check and use in this order:
      - Policy A `Organization.address.read`
      - Policy A `Organization.read`
      - Policy B `Organization.address.read`
      - Policy B `Organization.read`
    - Now the order is:
      - Policy A `Organization.address.read`
      - Policy B `Organization.address.read`
      - Policy A `Organization.read`
      - Policy B `Organization.read`
1. Add `ResourceGranter.andAllProps`
    - The first point has big implications. It means another policy with a conditional permission on a property takes precedence over the current policy's resource level permission.
    - This could not be desired, so now there's an additional flag (`andAllProps`) to explicitly communicate that the current policy's resource level permissions apply to all the properties as well and no other policies should "get in the way".
    - Other property level permissions are still considered. (See Example 4)
3. `PropGranter.none` is actually a _DENY_ action
    - This _deny_ as precedence over other actions only within the current policy, but not outside of the current policy.

# Examples

Let's run through some examples for clarity.

## Example 1
```ts
// Policy A
r.Organization.read
// Policy B
r.Organization.read.specifically(p => p.address.when(member).read)
```
Assuming both apply, the user will only have `read` permission if a member for `address`.

## Example 2
```ts
// Policy A
r.Organization.read
// Policy B
r.Organization.specifically(p => p.address.none)
```
Assuming both apply, the user will not have any permissions for `address`.

## Example 3
```ts
// Policy A
r.Organization.read.andAllProps
// Policy B
r.Organization.specifically(p => p.address.none)
```
Assuming both apply, the user will have `read` permissions for `address`.

## Example 4
```ts
// Policy A
r.Organization.read.andAllProps
// Policy B
r.Organization.specifically(p => p.address.edit)
```
Assuming both apply, the user will have `read` & `edit` permissions for `address`.

## Example 5
```ts
r.Organization.specifically(p => [
  p.address.none,
  p.address.read,
  // or any other form, etc
  p.address.read.none,
])
```
The user will not have any permissions for `address`.

# Conclusion

Policies are deterministic now for these edge cases.
For the most part, things should behave more intuitively. 

It's unclear where `.andAllProps` should apply to existing policies. I've already applied them to Administrator and Leadership  where we are expecting unfettered/read access to everything.

Hopefully this all makes sense and is the right direction.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3509457875) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/authorization/policies/by-role/financial-analyst-lead.policy.ts
+++ b/src/components/authorization/policies/by-role/financial-analyst-lead.policy.ts
@@ -8,8 +8,8 @@ import { member, Policy, Role, sensMediumOrLower } from '../util';
   r.Language.read.specifically((c) => [
     c.locations.whenAny(member, sensMediumOrLower).read,
   ]),
-  r.Organization.edit,
-  r.Partner.edit,
+  r.Organization.edit.andAllProps,
+  r.Partner.edit.andAllProps,
   r.Partnership.read.create.delete.specifically((p) => [
     p.many('mouStartOverride', 'mouEndOverride', 'mou', 'mouStatus').edit,
   ]),
--- a/src/components/authorization/policy/builder/allow-all.helper.ts
+++ b/src/components/authorization/policy/builder/allow-all.helper.ts
@@ -3,6 +3,7 @@ import { ResourceAction } from '../actions';
 import { ResourcesGranter } from '../granters';
 import { asNormalized } from './as-normalized.helper';
 import { action } from './perm-granter';
+import { andAllProps } from './resource-granter';
 
 /**
  * A helper to allow these actions for all resources
@@ -13,9 +14,9 @@ export const allowAll =
     Object.values(r).map((res) => allowActions(res, ...actions));
 
 /**
- * A helper to allow these actions for this resource.
+ * A helper to allow these actions for this resource & and all its props.
  */
 export const allowActions = (
   granter: ValueOf<ResourcesGranter>,
   ...actions: ResourceAction[]
-) => asNormalized(granter, (g) => g[action](...actions));
+) => asNormalized(granter, (g) => g[action](...actions)[andAllProps]);
--- a/src/components/authorization/policy/builder/perm-granter.ts
+++ b/src/components/authorization/policy/builder/perm-granter.ts
@@ -12,6 +12,11 @@ export type Permission = Condition<any> | boolean;
 export const action = Symbol('PermGranter.action');
 
 /**
+ * Add to the perms list.
+ */
+export const withPerms = Symbol('PermGranter.withPerms');
+
+/**
  * Extract permissions from granter.
  */
 export const extract = Symbol('PermGranter.extract');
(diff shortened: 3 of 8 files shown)


CarsonF:
After a long talk with @bryanjnelson, we've decided against this direction 👎 

There's two situations affected by this edge case.

# FA / LFA Policy

```ts
// FA, LFA Policy
    r.Partner.read.create
      .specifically((p) => [
        p.many('organization', 'pointOfContact').when(sensMediumOrLower).read,
      ])

// LFA Policy
  r.Partner.edit,
```
This works exactly like you'd expect _without_ this PR change.
Each policy conveys that the user has _at least_ this much permission.
The LFA "override" policy then widens those permissions.

With this PR, the LFA "override" policy would need `.andAllProps` as my [last commit](https://github.com/SeedCompany/cord-api-v3/pull/2562/commits/f7d6883af5a81912ef07b24099ea7d8f8849bf6d) does. This is only because the FA policy specified a reduced permission set that was getting in the way of these broader permissions. It is confusing that when looking at the LFA policy, there are _constraints_ that are specified else where.

So this example, works exactly like we want now and this PR hurts that.

# Investor Organization Policies
```ts
// Common Policy
  r.Organization.whenAny(member, sensOnlyLow).read.specifically((p) => [
    p.address.none,
    p.locations.when(sensOnlyLow).read,
  ]),

// Marketing Policy
// Omits as it's the same as above

// Fundraising Policy
  r.Organization.whenAny(sensMediumOrLower).read.specifically((p) => [
    p.address.none,
    p.locations.read,
  ]),

// ExperienceOperations Policy
  r.Organization.whenAny(sensMediumOrLower).read.specifically((p) => [
    p.address.none,
  ]),
```

With this PR we could leave the `address.none` outside of the last two policies because it applies in the common one. However, that isn't intuitive because again there are _constraints_ on the permissions that are defined else where.

The better solution would be to move the declaration from the Common policy to the Marketing one. Then we can be sure that what we are looking at _at least_ applies regardless of any other policies. While this was initially thought of as the common permission set between all 3 roles, it actually is not because of the relationship/interaction between resource & property level permissions.

---

The non-determinism mentioned in OP is resolved by just not writing policies in a way that's not deterministic lol.
Maybe in the future we can have a lint rule/test to confirm this.

PR #2564, opened 2022-11-14
Fix PnP to identify products from progress sheet the same way planning sheet does

atGit2021:
…erence notes section does not exist on the progress sheet

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3523040122) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/pnp/extract-scripture.ts
+++ b/src/components/pnp/extract-scripture.ts
@@ -1,0 +1,49 @@
+import { Row } from '~/common/xlsx.util';
+import {
+  parseScripture,
+  ScriptureRange,
+  tryParseScripture,
+} from '../scripture';
+import { WrittenScripturePlanningSheet } from './planning-sheet';
+
+export const extractScripture = (row: Row<WrittenScripturePlanningSheet>) => {
+  const sheet = row.sheet;
+  const bookName = sheet.bookName(row)!;
+  const totalVerses = sheet.totalVerses(row)!;
+  const scriptureFromBookCol = parseScripture(bookName);
+
+  const common = {
+    bookName,
+    totalVerses,
+  };
+
+  // If scripture from book column matches total count use it.
+  if (ScriptureRange.totalVerses(...scriptureFromBookCol) === totalVerses) {
+    return {
+      ...common,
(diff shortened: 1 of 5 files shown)


bryanjnelson (approved):
@CarsonF - We know we haven't dove into the world of index checking yet, but we couldn't think of another way to handle this scenario.  If there's another, or better way, please let us know.

bryanjnelson (dismissed):
@CarsonF - All yours!

CarsonF:
@atGit2021 I DRY'd up the logic for extract so this "complicated" business logic isn't in two spots. I also refactored the matching logic to use immutable variables with different names instead of the same mutable one. I think this makes it easier to read. Will you take a look and see if I've made any mistakes?

atGit2021:
> @atGit2021 I DRY'd up the logic for extract so this "complicated" business logic isn't in two spots. I also refactored the matching logic to use immutable variables with different names instead of the same mutable one. I think this makes it easier to read. Will you take a look and see if I've made any mistakes?

@CarsonF I pulled down and tested again, and all looks good both on the planning as well as the progress extraction.  I discovered the reason why I saw some of the entries not getting picked up correctly, and it's not due to the code, but due to the fact that some of the PnP's I was using to test with had incorrect scripture references in the notes column.  For example, Psalms 4:1-9 in the notes section would throw an exception because Psalms 4 only has 8 verses in the chapter, so once the xls notes column and the lifetime verses to translate column were corrected to Psalms 4:1-8 and 8 total verses to translate then all worked appropriately.  The only thing is the user doesn't see any message other than upload was successful, so they may spend some time wondering why it was pulled in as an unspecified scripture reference until they realize they put an incorrect scripture count and/or reference.  Other than that, all the code is working well, so it should be good to go.

CarsonF:
Thanks for digging into this @atGit2021. Glad it's all working. I agree that we are really vague with users in this feature. We've never been able to prioritize anything here since we are just trying to replace this excel file all together.

CarsonF (approved):
Targeting master for quicker release

CarsonF (approved):

PR #2573, opened 2022-12-02
Fix CommunityStory & Highlight not allowing sensitivity conditions in policies

CarsonF:
Here we add a static prop to DTOs to manually mark them as valid for `SensitivityCondition`.

In some cases we dynamically add sensitivity to objects that don't have those props, like `withEffectiveSensitivity`.
But the policies didn't know that the code calling the privileges service were passing this info (sensitivity) since it was dynamic.

With this flag we can communicate that.

There is no check between this flag and the actual caller passing this info, which isn't ideal. But this does allow us to keep using the sensitivity condition in policies only on resources that allow it. In practice the policies reference this manually more times than the service code. So favoring strictness for them vs the service code is a worthy trade.


bryanjnelson (approved):
Thank ye!

PR #2684, opened 2022-12-05
added LFA role to the EngagementStatus.FinalizingCompletion approvers list

atGit2021:
This is a fix for the following HALP ticket, https://seedcompany.halp.com/tickets/7400


CarsonF (approved):
LGTM. If LA has permission so should LFA.

bryanjnelson (approved):
Thanks @atGit2021.  This will solve that Prod bug!  🚢

PR #2689, opened 2022-12-06
Field Partner Policy Updates

bryanjnelson:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3629288915) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/authorization/policies/by-role/field-partner.policy.ts
+++ b/src/components/authorization/policies/by-role/field-partner.policy.ts
@@ -1,6 +1,17 @@
 import { member, Policy, Role, variant } from '../util';
 
 @Policy(Role.FieldPartner, (r) => [
+  r.Ceremony.read,
+  r.Engagement.when(member).read,
+  r.EthnologueLanguage.when(member).read,
+  r.FieldRegion.read,
+  r.FieldZone.read,
+  r.Language.when(member).read,
+  r.Organization.when(member).read.specifically((p) => p.address.none),
+  r.Partner.when(member).read.specifically((p) => p.pointOfContact.none),
+  r.Partnership.when(member).read,
+  r.PeriodicReport.read,
+  r.Product.read,
   r.ProgressReport.when(member).read.specifically((p) => p.reportFile.none),
   [r.ProgressReportCommunityStory, r.ProgressReportHighlight].flatMap((it) => [
     it.when(member).create.read,
@@ -9,6 +20,10 @@ import { member, Policy, Role, variant } from '../util';
       p.responses.whenAll(member, variant('draft')).edit,
     ]),
   ]),
+  r.Project.when(member).read.specifically((p) => p.rootDirectory.none),
+  r.ProjectMember.when(member).read,
   r.StepProgress.whenAll(member, variant('partner')).edit,
+  r.Unavailability.read,
+  r.User.read,
 ])
 export class FieldPartnerPolicy {}


src/components/authorization/policies/by-role/field-partner.policy.ts
```diff
@@ -1,6 +1,29 @@
...
+  ),
+  r.EthnologueLanguage.when(member).read,
+  r.FieldRegion.read,
+  r.FieldZone.read,
+  r.FinancialReport.read,
+  r.Language.when(member).read,
+  r.Organization.when(member).read.specifically((p) => p.address.none),
+  r.Partner.when(member).read.specifically((p) => p.pointOfContact.none),
+  r.Partnership.when(member).read,
+  r.PeriodicReport.read,
```
CarsonF, line 20: Unless we are trying to also grant read access to `NarrativeReports` this isn't needed here as the two other concretes are specified. Or we could leave this and remove FinancialReport entry.

src/components/authorization/policies/by-role/field-partner.policy.ts
```diff
@@ -1,6 +1,29 @@
...
 
 @Policy(Role.FieldPartner, (r) => [
+  r.Ceremony.read,
+  inherit(
+    r.Engagement.read,
+    r.InternshipEngagement.when(member).edit.specifically(
+      (p) => p.ceremony.none
+    ),
+    r.LanguageEngagement.when(member).read
+  ),
```
CarsonF, line 11: They can read LanguageEngagement ceremonies but not Internship ones?

src/components/authorization/policies/by-role/field-partner.policy.ts
```diff
@@ -1,6 +1,29 @@
...
+  r.FinancialReport.read,
+  r.Language.when(member).read,
+  r.Organization.when(member).read.specifically((p) => p.address.none),
+  r.Partner.when(member).read.specifically((p) => p.pointOfContact.none),
+  r.Partnership.when(member).read,
+  r.PeriodicReport.read,
+  r.Product.read,
+  r.Project.when(member).read.specifically((p) => p.rootDirectory.none),
+  r.ProjectMember.when(member).read,
+  r.StepProgress.read,
```
CarsonF, line 24: I doubt this is actually intentional compared to the StepProgress entry below.
```suggestion
```

CarsonF (changes requested):

src/components/authorization/policies/by-role/field-partner.policy.ts
bryanjnelson: Nice catch...removing Financial Report.

src/components/authorization/policies/by-role/field-partner.policy.ts
bryanjnelson: Right...cut/paste from the other policy.

src/components/authorization/policies/by-role/field-partner.policy.ts
bryanjnelson: Should be both; a miss.  👍🏼

bryanjnelson: Talked to business...both should be read only.

CarsonF (approved):
🚀

PR #2699, opened 2022-12-12
Impersonate roles/users via new HTTP headers

### LINKED ISSUE #2682: Role Impersonation
I'd like to implement a way for a user to call the API logged in as X user (with X roles) but act as a user with Y roles.
This will simplify API service calls allowing authorization tokens to be decoupled from business logic.

In order to prevent authorized role escalation we should limit this functionality to only the Administrator role. Maybe in the future we can properly grant a permission subset based on user role input & their actual roles.

CarsonF:
Closes #2682

Use this header so impersonate a specific user
```http
X-CORD-Impersonate-User: w2mBuPVnkbw
```

Use this header to Impersonate one or more roles
```http
X-CORD-Impersonate-Role: StaffMember, Controller
```
If used with user header the user's roles are added to the ones given here.

In both cases, these are only allowed if the requesting user has permission to "assign" all of the impersonatee's & given roles.
I believe `AssignableRoles` already communicates the restrictions we want here, so I used that.
It's probably worth renaming `assignable` to something else, like maybe `admin`.

### DIFF
--- a/src/common/role.dto.ts
+++ b/src/common/role.dto.ts
@@ -44,6 +44,8 @@ export namespace Role {
     resources: ResourcesGranter,
     roles: readonly Role[]
   ) => resources.AssignableRoles.grant(roles);
-
   Object.defineProperty(Role, 'assignable', { enumerable: false });
+
+  export const all = new Set(Object.keys(Role)) as ReadonlySet<Role>;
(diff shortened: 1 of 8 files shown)


bryanjnelson (approved):
:shipit:

PR #2703, opened 2022-12-17
Refactor ExceptionFilter

CarsonF:
- Split out determining the raw-ish json of the error into a separate `ExceptionNormalizer` class
- Handle responses for HTTP (not GQL) requests ourselves instead of forwarding to Nest's default
  - The structure of the response will change but it's more consistent with our GQL output.
- Just close requests that are not found routes in prod instead of returning 404

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3679997804) by [Unito](https://www.unito.io)


src/core/exception/is-from-hack-attempt.ts
```diff
@@ -0,0 +1,16 @@
+import { ArgumentsHost } from '@nestjs/common';
+import { NotFoundException as NestjsNotFoundException } from '@nestjs/common/exceptions/not-found.exception';
+import { IncomingMessage } from 'http';
+
+export const isFromHackAttempt = (error: Error, args: ArgumentsHost) => {
+  if (
+    // Any route not found is a hack attempt
+    error instanceof NestjsNotFoundException &&
+    // Let devs make errors, so only check deployed
+    process.env.NODE_ENV === 'production' &&
```
bryanjnelson, line 10: Smart.

bryanjnelson (approved):

bryanjnelson:
Kick it!  🚢

PR #2727, opened 2023-01-19
Fix notifications for DidNotDevelop/Rejected

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3840320754) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/project/project.rules.ts
+++ b/src/components/project/project.rules.ts
@@ -2,6 +2,7 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
 import { node, relation } from 'cypher-query-builder';
 import { first, intersection, Many, uniq } from 'lodash';
+import { UnreachableCaseError } from 'ts-essentials';
 import { Promisable } from 'type-fest';
 import {
   ID,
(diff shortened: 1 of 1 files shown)


src/components/project/project.rules.ts
```diff
@@ -804,12 +806,21 @@ export class ProjectRules {
...
         return {
-          approvers: [Role.Administrator],
+          approvers: [
+            Role.Administrator,
+            Role.ProjectManager,
+            Role.RegionalDirector,
+            Role.FieldOperationsDirector,
+          ],
           transitions: [],
+          getNotifiers: async () => [...(await this.getProjectTeamUserIds(id))],
```
CarsonF, line 820: I don't think the approvers does anything here as there are no transitions anyways. Also the notifiers could be simplified.
```suggestion
          approvers: [],
          transitions: [],
          getNotifiers: () => this.getProjectTeamUserIds(id),
```

atGit2021: ok, I wasn't sure who could approve to these statuses so I pinged Seth to get his feedback, but will revise per your feedback.

CarsonF (approved):
Thanks @atGit2021!

PR #2747, opened 2023-02-06
Fix progress summary percent/verse equivalent check for numbers like 1.00000001

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3939152217) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/progress-summary/progress-summary.extractor.ts
+++ b/src/components/progress-summary/progress-summary.extractor.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@nestjs/common';
-import { round } from 'lodash';
+import { clamp, round } from 'lodash';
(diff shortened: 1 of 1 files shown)


bryanjnelson (approved):

CarsonF (changes requested):
This is only 2 decimal places for percents. I suggest we keep the precision given to us.

I would just round when doing the conversion check.
```ts
if (round(planned, 4) > 1) {
```
After we know whether we need to divide by the total verse equivalents, we could clamp to make sure we don't go over 100%.
```ts
planned = Math.min(planned, 1)
...
```

CarsonF (approved):

PR #2751, opened 2023-02-07
revised isScriptureEqual to capture empty arrays

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3947139909) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/scripture/is-equal.ts
+++ b/src/components/scripture/is-equal.ts
@@ -9,7 +9,8 @@ export const isScriptureEqual = (
 ) => {
   if (
     (a.length !== 0 && b.length === 0) ||
-    (a.length === 0 && b.length !== 0)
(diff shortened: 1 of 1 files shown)


CarsonF (commented):
If both arrays are empty they are not equal? 🤔

atGit2021:
It resolves the issue, but revise as you see fit.

Sent via the Samsung Galaxy A51, an AT&T 4G LTE smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

CarsonF:
Superseded by #2753

We see now how this fixes the problem with matching the scripture products. #2753 moves that check up a frame, so this method isn't called if the scripture is empty.
Good sleuthing again!

PR #2764, opened 2023-03-03
Feature: Email notifcations for qtrly report status changes

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4084609702) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/progress-report/workflow/events/workflow-updated.event.ts
+++ b/src/components/progress-report/workflow/events/workflow-updated.event.ts
@@ -1,0 +1,13 @@
+import { ID, UnsecuredDto } from '~/common';
+import { ProgressReportStatus as Status } from '../../dto';
+import { ProgressReportWorkflowEvent as WorkflowEvent } from '../dto/workflow-event.dto';
+import { InternalTransition } from '../transitions';
+
+export class WorkflowUpdatedEvent {
+  constructor(
+    readonly reportId: ID,
+    readonly previousStatus: Status,
+    readonly next: InternalTransition | Status,
+    readonly workflowEvent: UnsecuredDto<WorkflowEvent>,
+  ) {}
+}
--- a/src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
+++ b/src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
@@ -1,0 +1,199 @@
+import { EmailService } from '@seedcompany/nestjs-email';
+import { RequireExactlyOne } from 'type-fest';
+import { ID, mapFromList, Role, UnsecuredDto } from '~/common';
+import {
+  ConfigService,
+  EventsHandler,
+  IEventHandler,
+  ILogger,
+  Logger,
+} from '~/core';
+import {
+  ProgressReportStatusChangedProps as EmailReportStatusNotification,
+  ProgressReportStatusChanged,
+} from '~/core/email/templates/progress-report-status-changed.template';
+import { AuthenticationService } from '../../../authentication';
+import { LanguageService } from '../../../language';
+import { PeriodicReportService } from '../../../periodic-report';
+import { ProjectService } from '../../../project';
+import { UserService } from '../../../user';
+import { ProgressReportStatus as Status } from '../../dto';
+import { ProgressReportWorkflowEvent } from '../dto/workflow-event.dto';
+import { WorkflowUpdatedEvent } from '../events/workflow-updated.event';
+import { ProgressReportWorkflowRepository } from '../progress-report-workflow.repository';
+import { ProgressReportWorkflowService } from '../progress-report-workflow.service';
+import { InternalTransition } from '../transitions';
+
+const rolesToAlwaysNotify = [
+  Role.ProjectManager,
+  Role.RegionalDirector,
+  Role.FieldOperationsDirector,
+];
+
+@EventsHandler(WorkflowUpdatedEvent)
+export class ProgressReportWorkflowNotificationHandler
+  implements IEventHandler<WorkflowUpdatedEvent>
+{
+  constructor(
+    private readonly auth: AuthenticationService,
+    private readonly repo: ProgressReportWorkflowRepository,
+    private readonly configService: ConfigService,
+    private readonly userService: UserService,
+    private readonly projectService: ProjectService,
+    private readonly languageService: LanguageService,
+    private readonly reportService: PeriodicReportService,
+    private readonly emailService: EmailService,
+    private readonly workflowService: ProgressReportWorkflowService,
+    @Logger('progress-report:status-change-notifier')
+    private readonly logger: ILogger,
+  ) {}
+
+  async handle({
+    reportId,
+    previousStatus,
+    next,
+    workflowEvent,
+  }: WorkflowUpdatedEvent) {
+    const { enabled } = this.configService.progressReportStatusChange;
+    if (!enabled) {
+      return;
+    }
+    const { projectId, languageId } = await this.repo.getProjectInfoByReportId(
+      reportId,
+    );
+
+    const userIdByEmail: Record<string, ID | undefined> = mapFromList(
+      [
+        ...(await this.getEnvNotifyees(next)),
+        ...(await this.getProjectNotifyees(reportId, next)),
+      ],
+      ({ id, email }) => [email, id],
+    );
+
+    const notifications = await Promise.all(
+      Object.entries(userIdByEmail).map(([email, userId]) =>
+        this.prepareNotificationObject(
+          reportId,
+          previousStatus,
+          userId ? { userId } : { email },
+          workflowEvent,
+          projectId,
+          languageId,
+        ),
+      ),
+    );
+
+    this.logger.info('Notifying', {
+      emails: notifications.map((r) => r.recipient.email.value),
+      projectId: notifications[0]?.project.id ?? undefined,
+      languageId: notifications[0]?.language.id ?? undefined,
+      reportId: reportId,
+      newStatusVal: notifications[0]?.newStatusVal ?? undefined,
+      previousStatusVal: notifications[0]?.previousStatusVal ?? undefined,
+    });
+
+    for (const notification of notifications) {
+      if (notification.recipient.email.value) {
+        await this.emailService.send(
+          notification.recipient.email.value,
+          ProgressReportStatusChanged,
+          notification,
+        );
+      }
+    }
+  }
+
+  private async prepareNotificationObject(
+    reportId: ID,
+    previousStatus: Status,
+    receiver: RequireExactlyOne<{ userId: ID; email: string }>,
+    unsecuredEvent: UnsecuredDto<ProgressReportWorkflowEvent>,
+    projectId: ID,
+    languageId: ID,
+  ): Promise<EmailReportStatusNotification> {
+    const recipientId = receiver.userId
+      ? receiver.userId
+      : this.configService.rootAdmin.id;
+    const recipientSession = await this.auth.sessionForUser(recipientId);
+
+    const recipient = receiver.userId
+      ? await this.userService.readOne(recipientId, recipientSession)
+      : this.fakeUserFromEmailAddress(receiver.email!);
+
+    const project = await this.projectService.readOne(
+      projectId,
+      recipientSession,
+    );
+    const language = await this.languageService.readOne(
+      languageId,
+      recipientSession,
+    );
+    const report = await this.reportService.readOne(reportId, recipientSession);
+    const changedBy = await this.userService.readOne(
+      unsecuredEvent.who,
+      recipientSession,
+    );
+    const workflowEvent = this.workflowService.secure(
+      unsecuredEvent,
+      recipientSession,
+    );
+
+    return {
+      changedBy,
+      recipient,
+      project,
+      language,
+      report,
+      newStatusVal: report.status?.value,
+      previousStatusVal: report.status?.value ? previousStatus : undefined,
+      workflowEvent: workflowEvent,
+    };
+  }
+
+  private fakeUserFromEmailAddress(
+    receiver: string,
+  ): EmailReportStatusNotification['recipient'] {
+    return {
+      email: { value: receiver, canRead: true, canEdit: false },
+      displayFirstName: {
+        value: receiver.split('@')[0],
+        canRead: true,
+        canEdit: false,
+      },
+      displayLastName: { value: '', canRead: true, canEdit: false },
+      timezone: {
+        value: this.configService.defaultTimeZone,
+        canRead: true,
+        canEdit: false,
+      },
+    };
+  }
+
+  private async getEnvNotifyees(next: Status | InternalTransition) {
+    const { forTransitions, forBypasses } =
+      this.configService.progressReportStatusChange.notifyExtraEmails;
+    const envEmailList =
+      typeof next !== 'string'
+        ? forTransitions.get(next.name)
+        : forBypasses.get(next);
+    return [
+      ...(envEmailList?.map((email) => ({ id: undefined, email })) ?? []),
+      ...(envEmailList ? await this.repo.getUserIdByEmails(envEmailList) : []),
+    ];
+  }
+
+  private async getProjectNotifyees(
+    reportId: ID,
+    next: Status | InternalTransition,
+  ) {
+    const roles = [
+      ...rolesToAlwaysNotify,
+      ...(typeof next !== 'string' ? next.notify?.membersWithRoles ?? [] : []),
+    ];
+
+    const members = await this.repo.getProjectMemberInfoByReportId(reportId);
+    return members.filter((mbr) =>
+      mbr.roles.some((role) => roles.includes(role)),
+    );
+  }
+}
--- a/src/components/progress-report/workflow/progress-report-workflow.repository.ts
+++ b/src/components/progress-report/workflow/progress-report-workflow.repository.ts
@@ -5,6 +5,8 @@ import {
   NotFoundException,
   Order,
   RichTextDocument,
+  Role,
+  ServerException,
   Session,
   UnsecuredDto,
 } from '~/common';
@@ -85,7 +87,11 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
     session: Session,
     notes?: RichTextDocument,
   ) {
-    await this.recordEvent(report, { status, transition, notes }, session);
+    return await this.recordEvent(
+      report,
+      { status, transition, notes },
+      session,
+    );
   }
 
   async recordBypass(
@@ -94,7 +100,7 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
     session: Session,
     notes?: RichTextDocument,
   ) {
-    await this.recordEvent(report, { status, notes }, session);
+    return await this.recordEvent(report, { status, notes }, session);
   }
 
   private async recordEvent(
@@ -102,7 +108,7 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
     props: Record<string, any>,
     session: Session,
   ) {
-    await this.db
+    const result = await this.db
       .query()
       .apply(
         await createNode(WorkflowEvent, {
@@ -115,8 +121,9 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
           out: { who: ['User', session.userId] },
         }),
       )
-      .return('*')
-      .run();
+      .apply(this.hydrate())
+      .first();
+    return result!.dto;
   }
 
   async currentStatus(reportId: ID): Promise<Status> {
@@ -145,5 +152,100 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
       object: { id: report },
       changes: { status },
     });
+  }
+
+  async getProjectMemberInfoByReportId(reportId: ID) {
+    const query = this.db
+      .query()
+      .match([
+        node('report', 'ProgressReport', { id: reportId }),
+        relation('in', '', ACTIVE),
+        node('engagement', 'Engagement'),
+        relation('in', '', 'engagement', ACTIVE),
+        node('project', 'Project'),
+        relation('out', '', 'member', ACTIVE),
+        node('member', 'ProjectMember'),
+        relation('out', '', 'user', ACTIVE),
+        node('user', 'User'),
+      ])
+      .optionalMatch([
+        node('user'),
+        relation('out', '', 'email', ACTIVE),
+        node('email', 'EmailAddress'),
+      ])
+      .match([
+        node('member'),
+        relation('out', '', 'roles', ACTIVE),
+        node('role', 'Property'),
+      ])
+      .return<{
+        id: ID;
+        email: string;
+        roles: Role[];
+      }>([
+        'user.id as id',
+        'email.value as email',
+        'coalesce(role.value, []) as roles',
+      ]);
+    return await query.run();
+  }
+
+  async getUserInfoByEmail(email: string) {
+    const query = this.db
+      .query()
+      .match([
+        node('email', 'EmailAddress', { value: email }),
+        relation('in', '', 'email', ACTIVE),
+        node('user', 'User'),
+      ])
+      .return<{
+        id?: ID;
+        email?: string;
+      }>(['user.id as id', 'email.value as email']);
+    return await query.run();
+  }
+
+  async getUserIdByEmails(emails: readonly string[]) {
+    const query = this.db
+      .query()
+      .match([
+        node('email', 'EmailAddress'),
+        relation('in', '', 'email', ACTIVE),
+        node('user', 'User'),
+      ])
+      .where({ 'email.value': inArray(emails) })
+      .return<{
+        id: ID;
+        email: string;
+      }>(['user.id as id, email.value as email']);
+    return await query.run();
+  }
+
+  async getProjectInfoByReportId(reportId: ID) {
+    const query = this.db
+      .query()
+      .match([
+        node('report', 'ProgressReport', { id: reportId }),
+        relation('in', '', ACTIVE),
+        node('engagement', 'Engagement'),
+        relation('in', '', 'engagement', ACTIVE),
+        node('project', 'Project'),
+      ])
+      .match([
+        node('engagement'),
+        relation('out', '', ACTIVE),
+        node('language', 'Language'),
+      ])
+      .return<{
+        projectId: ID;
+        languageId: ID;
+      }>(['project.id as projectId', 'language.id as languageId']);
+    const result = await query.first();
+    if (!result) {
+      throw new ServerException(
+        `Unable to retrieve project and language information for reportId ${reportId}`,
+      );
+    }
+    return result;
   }
 }
--- a/src/components/progress-report/workflow/progress-report-workflow.service.ts
+++ b/src/components/progress-report/workflow/progress-report-workflow.service.ts
@@ -6,11 +6,12 @@ import {
   UnauthorizedException,
   UnsecuredDto,
 } from '~/common';
-import { ResourceLoader } from '~/core';
+import { IEventBus } from '~/core';
 import { Privileges } from '../../authorization';
 import { ProgressReport, ProgressReportStatus as Status } from '../dto';
 import { ExecuteProgressReportTransitionInput } from './dto/execute-progress-report-transition.input';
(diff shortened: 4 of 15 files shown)


CarsonF:
The lint errors are from a recent formatting change affecting the entire codebase #2763

If you just rebase and lint it should grab them.

bryanjnelson (approved):
@CarsonF - On to you.

src/components/progress-report/workflow/transitions.ts
```diff
@@ -23,7 +23,7 @@ export const Transitions = defineTransitions({
         Role.RegionalDirector,
         Role.FieldOperationsDirector,
       ],
-      Emails: [''],
+      Emails: ['MONDAY_EMAIL_DISTRO', 'andre_turner@tsco.org'],
```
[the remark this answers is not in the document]
atGit2021: BTW, I'll remove my hardcoded email address and move this MONDAY distro to the correct "status" change once the refactoring process/code changes are done.  I have them in this status just to test.

src/components/progress-report/workflow/progress-report-workflow.service.ts
```diff
@@ -1,25 +1,51 @@
...
 import { Privileges } from '../../authorization';
+import { Project, ProjectService } from '../../project';
+import { User, UserService } from '../../user';
 import { ProgressReport, ProgressReportStatus as Status } from '../dto';
 import { ExecuteProgressReportTransitionInput } from './dto/execute-progress-report-transition.input';
 import { ProgressReportWorkflowEvent as WorkflowEvent } from './dto/workflow-event.dto';
 import { ProgressReportWorkflowRepository } from './progress-report-workflow.repository';
 import { Transitions } from './transitions';
 
+export interface EmailReportStatusNotification {
```
CarsonF, line 24: I'd move this to the right above where the template is defined. This is the input shape needed for that template so it makes sense to define it there.

I know the other place we don't

src/components/progress-report/workflow/progress-report-workflow.service.ts
```diff
@@ -100,6 +126,84 @@ export class ProgressReportWorkflowService {
...
 
-    // TODO(transition.notify);
+    const notificationInfo = await this.repo.getNotificationInfoByReportId(
+      reportId,
+    );
+    const projectId = notificationInfo[0]?.projectId ?? undefined;
+    const languageDisplayName = notificationInfo[0]?.languageDisplayName;
+    const reportQtrLabel = `Q${fiscalQuarter(
+      notificationInfo[0]?.start,
+    )} FY${fiscalYear(notificationInfo[0]?.start)}`;
```
CarsonF, line 136: This is "UI" logic, so I would move it into the template.

src/components/progress-report/workflow/progress-report-workflow.service.ts
```diff
@@ -100,6 +126,84 @@ export class ProgressReportWorkflowService {
...
+          if (item.email) emailList.push(item.email);
+        }
+      });
+    });
+
+    transition.notify.Emails.forEach((element: string) => {
+      if (element.includes('@')) {
+        emailList.push(element);
+      } else if (element === 'MONDAY_EMAIL_DISTRO') {
+        emailList.push(this.configService.mondayEmailDistro);
```
CarsonF, line 151: This is a magic string. The idea with having `notify` be an object with unique properties was so we didn't have to introspect the values of the properties. Instead this could be a check `transition.notifiy.monday`. Maybe it's just a boolean. 
However since this monday is just an email address, I would instead avoid the word "monday" all together.
Instead the env key could be something like `PROGRESS_REPORT_TRANSITION_NOTIFY_EMAILS`.
It could be dynamic and allow for mapping any transition to any number of emails.

```
Start=address@monday.com,carson_full@tsco.org;Translation Done=carson_full@tsco.org
```
Which could easily be parsed into
```ts
Record<TransitionName, string[]>
```

This allows the specific env to give email addresses for any transition and keeps the codebase here unaware of Monday.com.

src/components/progress-report/workflow/progress-report-workflow.service.ts
```diff
@@ -100,6 +126,84 @@ export class ProgressReportWorkflowService {
...
+      }
+    });
+
+    emailList = uniq(emailList);
+    const recipients = emailList.map((email) =>
+      this.prepareRecipientObject(email),
+    );
+
+    const changedBy = await this.userService.readOne(session.userId, session);
+    const project = await this.projectService.readOne(projectId, session);
```
CarsonF, line 161: If the recipient is a specific user we should use a session corresponding to them instead of the user that is making the mutation.
That's what we do [here](https://github.com/SeedCompany/cord-api-v3/blob/cc00184941ece5480fc301c9bfbec308e27dac93/src/components/project/project.rules.ts#L1052-L1071).

src/components/progress-report/workflow/progress-report-workflow.service.ts
```diff
@@ -100,6 +126,84 @@ export class ProgressReportWorkflowService {
...
+
+    const notifications: EmailReportStatusNotification[] = [];
+    recipients.forEach((recipient) =>
+      notifications.push({
```
CarsonF, line 165: There's a lot of "loop over X, map, and push result to Y" here. Instead these could just be `.map` calls and skip mutating an existing variable.

```suggestion
    const notifications = recipients.map(
      (recipient): EmailReportStatusNotification => ({
```

src/components/progress-report/workflow/transitions.ts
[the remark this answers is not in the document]
CarsonF: The direction I was pushing Andre in was to keep these transitions declarations declarative. And let the service handle any logic to fulfill the configuration.
I gave most of my comments on this specific thing on another comment.

src/components/progress-report/workflow/transitions.ts
```diff
@@ -17,36 +17,71 @@ export const Transitions = defineTransitions({
     to: Status.InProgress,
     label: 'Start',
     type: Type.Approve,
-    notify: 'FPM',
+    notify: {
+      TeamMembersWithRoles: [
+        Role.ProjectManager,
+        Role.RegionalDirector,
+        Role.FieldOperationsDirector,
+      ],
```
CarsonF, line 25: If every transition is going to have these recipients I would just implicitly apply them in the service instead of copying them for every transition here.

src/components/progress-report/workflow/transitions.ts
```diff
@@ -17,36 +17,71 @@ export const Transitions = defineTransitions({
...
     label: 'Submit for Translation',
     type: Type.Approve,
-    notify: ['Next', 'Translator'],
+    notify: {
+      TeamMembersWithRoles: [
+        Role.ProjectManager,
+        Role.RegionalDirector,
+        Role.FieldOperationsDirector,
+      ],
+      Emails: [''],
```
CarsonF, line 40: Empty strings should be avoided to declare a lack of a value. nulls or undefined should be used instead where it's a single value. Where it's a list, an empty list works better. In this case the property could just be optional and omitted where not needed.

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+  project,
+  languageEngagement,
+  report,
+  reportQtrLabel,
+  newStatusVal,
+  previousStatusVal,
+}: EmailReportStatusNotification) {
+  const projectUrl = project?.id
+    ? useFrontendUrl(`/projects/${project.id}`)
+    : useFrontendUrl(`/projects/no-project-found`);
```
CarsonF, line 26: I don't think there should ever be a case where we don't have a project. I would enforce that a project is required for this template. Ensuring that upstream caller has good data is better than forwarding a dead link to users.

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+    : 'Quarterly Report';
+
+  const changerUrl = useFrontendUrl(`/users/${changedBy.id}`);
+  const changerName = fullName(changedBy);
+
+  const oldStatus = startCase(previousStatusVal) || undefined;
+  const newStatus = startCase(newStatusVal) || undefined;
+
+  const changedAt = DateTime.now()
+    .setZone(changedBy.timezone.value ?? 'America/Chicago')
```
CarsonF, line 43: It would be better to inject this default value with a hook context.
```suggestion
    .setZone(changedBy.timezone.value ?? useDefaultTimezone())
```

Just like we do for frontend urls.
https://github.com/SeedCompany/cord-api-v3/blob/master/src/core/email/templates/frontend-url.tsx

https://github.com/SeedCompany/cord-api-v3/blob/cc00184941ece5480fc301c9bfbec308e27dac93/src/core/config/config.service.ts#L105-L105
https://github.com/SeedCompany/cord-api-v3/blob/cc00184941ece5480fc301c9bfbec308e27dac93/src/core/config/config.service.ts#L120-L122

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+    ? `Quarterly Report - ${reportQtrLabel}`
+    : 'Quarterly Report';
+
+  const changerUrl = useFrontendUrl(`/users/${changedBy.id}`);
+  const changerName = fullName(changedBy);
+
+  const oldStatus = startCase(previousStatusVal) || undefined;
+  const newStatus = startCase(newStatusVal) || undefined;
+
+  const changedAt = DateTime.now()
```
CarsonF, line 42: It would be better to keep this function pure and have the `changedAt` time passed in, instead of assuming "now".
Ideally this would reflect the same time has the `Event.at` stored in the DB.

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    from <em>{oldStatus}</em>{' '}
+                  </>
+                ) : null}
+                to <em>{newStatus}</em>{' '}
+              </>
+            ) : null}
+            for <a href={projectUrl}>{projectName}</a> - {languageEngagement} on{' '}
+            {changedAt}
+          </Text>
+          <HideInText>
```
CarsonF, line 82: I don't think there's a need to hide this button in text. It will be converted by a convention.
```suggestion
```

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const newStatus = startCase(newStatusVal) || undefined;
+
+  const changedAt = DateTime.now()
+    .setZone(changedBy.timezone.value ?? 'America/Chicago')
+    .toLocaleString(DateTime.DATETIME_FULL);
+
+  return (
+    <EmailTemplate
+      title={
+        reportLabel && oldStatus && newStatus
```
CarsonF, line 49: report label is never falsy.
```suggestion
        oldStatus && newStatus
```

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+          ? `${reportLabel} changed from ${oldStatus} to ${newStatus}`
+          : `${reportLabel} Status Change`
+      }
+    >
+      <Heading>
+        {reportLabel && newStatus ? (
```
CarsonF, line 55:
```suggestion
        {newStatus ? (
```

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+              <>
+                {oldStatus ? (
+                  <>
+                    from <em>{oldStatus}</em>{' '}
+                  </>
+                ) : null}
+                to <em>{newStatus}</em>{' '}
+              </>
+            ) : null}
+            for <a href={projectUrl}>{projectName}</a> - {languageEngagement} on{' '}
```
CarsonF, line 79: I'd mention the engagement name first as it's the most important and identifiable piece of information we have.
User's will be spammed with these notifications when the end of the quarter rolls around and they will all be for the same quarter.

In fact I would put that label before the quarter label in the subject as well.

Otherwise the email box is
```
Q1 FY2023 changed from Not Started to In Progress
Q1 FY2023 changed from Not Started to In Progress
Q1 FY2023 changed from Not Started to In Progress
Q1 FY2023 changed from Not Started to In Progress
Q1 FY2023 changed from Not Started to In Progress
Q1 FY2023 changed from Not Started to In Progress
```
GIven that we really only have one project for a language going at once too, I'd drop the project name in the subject for terseness.
```
{language} {quarter} now {newStatus}
```
```
Mankon Q1 FY2023 now In Progress
```

Though maybe the project name would help. Might be worth trying/asking both and see which one is better.
```
{language} ({project}) {quarter} now {newStatus}
```
```
Mankon (Abakwa Oral Cluster) Q1 FY2023 now In Progress
```

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -146,4 +148,57 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
+        languageDisplayName: string;
+        start: CalendarDate;
+      }>([
+        'email.value as email',
+        'role.value as role',
+        'project.id as projectId',
+        'languageDisplayName.value as languageDisplayName',
+        'start.value as start',
+      ])
+      .logIt();
```
CarsonF, line 201: We don't want to commit `logIt`
```suggestion
```

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -146,4 +148,57 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
+        projectId: ID;
+        languageDisplayName: string;
+        start: CalendarDate;
+      }>([
+        'email.value as email',
+        'role.value as role',
+        'project.id as projectId',
+        'languageDisplayName.value as languageDisplayName',
+        'start.value as start',
+      ])
```
CarsonF, line 200: This query is confused. It doesn't make sense to return N rows for each user email/role and the also ask for a single project id, language name, and start date. It would be easiest to just split these into two different queries. Or possibly just asking for each service.

Ideally all of this extra info could just come from the `ResourceLoader`
```tsx
resources.load(User, userId);
resources.load(Project, projectId);
resources.load(ProgressReport, reportId);
```
That's much easier than importing each module and calling the service.readOne methods.
However, this doesn't yet work with session for different users, which is why I didn't suggest it.
Maybe this is something I can change with the dataloader library migration.

CarsonF (changes requested):
Progress @atGit2021! 💪

src/components/progress-report/workflow/progress-report-workflow.service.ts
atGit2021: for my own learning, why is it necessary to get a session for each recipient?

CarsonF: Basically these emails are UI for the user. All UI (or even data at a lower level) should be given under the lens of that user. So maybe a certain recipient can't see the project name, even though the user triggering the notification can. We wouldn't want to show the project name to that user just because another unrelated user is taking an action.
I know that's a bit contrived, but that's the idea. Also it's unlikely that this changes the actual data shown in the email for the users but the principle is important. And the more we do it, the more patterns will emerge to help form an easier way to do these kinds of things.

bryanjnelson (changes requested):

bryanjnelson (approved):
Those changes look good, sir!

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    const projectMemberInfo = (await this.repo.getProjectMemberInfoByReportId(
+      reportId,
+    )) as Member[];
+    return projectMemberInfo
+      .filter((mbr) =>
+        mbr.role?.some((role) => allRolesToNotify.includes(role)),
+      )
+      .map((mbr) => {
+        delete mbr.role;
```
CarsonF, line 206: Why?

src/components/progress-report/workflow/events/workflow-updated.event.ts
```diff
@@ -0,0 +1,14 @@
+import { ID, Session, UnsecuredDto } from '~/common';
+import { ProgressReportStatus as Status } from '../../dto';
+import { ProgressReportWorkflowEvent as WorkflowEvent } from '../dto/workflow-event.dto';
+import { InternalTransition } from '../transitions';
+
+export class WorkflowUpdatedEvent {
+  constructor(
+    readonly reportId: ID,
+    readonly previousStatus: Status,
+    readonly next: InternalTransition | Status,
```
CarsonF, line 10:
```suggestion
```
Grab from 
```ts
workflowEvent.transition
workflowEvent.status
```

src/core/config/config.service.ts
```diff
@@ -117,6 +117,18 @@ export class ConfigService implements EmailOptionsFactory {
     };
   }
 
+  sendProgressReportNotifications = this.env
+    .boolean('PROGRESS_REPORT_NOTIFY_EMAILS')
+    .optional(false);
```
CarsonF, line 122: Default to `EMAIL_SEND` so it doesn't need to be manually specified in most cases.
```suggestion
    .optional(this.createEmailOptions().send);
```

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -146,4 +152,79 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
+      .return<{
+        projectId: ID;
+        languageId: ID;
+      }>(['project.id as projectId', 'language.id as languageId']);
+    const result = await query.first();
+    if (!result) {
+      throw new NotFoundException(
+        'Could not find project or language by report id',
+      );
+    }
```
CarsonF, line 227: This will never be false. I'd probably just not null assert it. Otherwise I'd throw a ServerException, NotFound is a user error.
```suggestion
    const result = await query.first()!;
```

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -146,4 +152,79 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
+    const query = this.db
+      .query()
+      .match([
+        node('email', 'EmailAddress', { value: email }),
+        relation('in', '', 'email', ACTIVE),
+        node('user', 'User'),
+      ])
+      .return<{
+        id?: ID;
+        email?: string;
```
CarsonF, line 198: I don't think either of these are nullable.
```suggestion
        id: ID;
        email: string;
```
Email is being matched by so it definitely exists based on input.
`User.id` always exists, if the user/email is not found there is just no row returned.

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -146,4 +152,79 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
+      ])
+      .match([
+        node('member'),
+        relation('out', '', 'roles', ACTIVE),
+        node('role', 'Property'),
+      ])
+      .return<{
+        id: ID;
+        email?: string;
+        role?: Role[];
```
CarsonF, line 183: This is a nit, but var names for lists should be plural.

I don't the property value will ever be null, but you could also coalesce it to be sure.
```cql
coalesce(roles.value, []) as roles
```

src/components/progress-report/workflow/transitions.ts
```diff
@@ -11,78 +11,106 @@ import {
 // Therefore, these should generally flow down.
 // "back transactions" should come before/above "forward transactions".
 
+/*
+  The following Roles are always notifed for each transition and are handled within
+  progress-report-workflow-notification.handler:
+    Role.ProjectManager,
+    Role.RegionalDirector,
+    Role.FieldOperationsDirector,
+*/
```
CarsonF, line 20: I get that this is trying to be helpful, but this is a prime comment to become stale in the future, when that actual code changes and this doesn't.
I'd say something like
```ts
/**
 * There could be roles that are implicitly always notified.
 * @see import('./handlers/progress-report-workflow-notification.handler').rolesToAlwaysNotify
 */
```
The jsdoc syntax allows you to jump right to the defined list in your IDE.

src/components/progress-report/workflow/transitions.ts
```diff
@@ -11,78 +11,106 @@ import {
...
 export const Transitions = defineTransitions({
   Start: {
     from: Status.NotStarted,
     to: Status.InProgress,
     label: 'Start',
     type: Type.Approve,
-    notify: 'FPM',
+    notify: {
+      TeamMembersWithRoles: [], //add additional roles desired to the array, e.g. Role.Marketing
+    },
```
CarsonF, line 29: `notify` is still typed as `any`. I'd change the types and document there instead.
https://github.com/SeedCompany/cord-api-v3/blob/master/src/components/progress-report/workflow/transitions.ts#L134-L134
```ts
notify?: {
  /**
   * Notify project members with these roles
   */
  membersWithRoles?: readonly Role[];
}
```

src/components/progress-report/workflow/transitions.ts
```diff
@@ -11,78 +11,106 @@ import {
...
   },
   'In Progress -> Pending Translation': {
     from: Status.InProgress,
     to: Status.PendingTranslation,
     label: 'Submit for Translation',
     type: Type.Approve,
-    notify: ['Next', 'Translator'],
+    notify: {
+      TeamMembersWithRoles: [],
+    },
```
CarsonF, line 38: No need to pollute this file with optional values.
```suggestion
```

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+    }
+  }
+
+  private async prepareNotificationObject(
+    session: Session,
+    reportId: ID,
+    previousStatus: Status,
+    receiver: Recipient,
+    unsecuredEvent: UnsecuredDto<ProgressReportWorkflowEvent>,
```
CarsonF, line 122: Just pass the event.
```suggestion
    receiver: Recipient,
    event: WorkflowUpdatedEvent,
```

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -0,0 +1,231 @@
...
+  ): Promise<EmailReportStatusNotification> {
+    const { projectId, languageId } = await this.repo.getProjectInfoByReportId(
+      reportId,
+    );
```
CarsonF, line 126: This will be called for every recipient even though it's the same info for all.
Call this once and pass it in here.

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -0,0 +1,231 @@
...
+      project,
+      language,
+      report,
+      newStatusVal: report.status?.value,
+      previousStatusVal: previousStatus,
+      workflowEvent: workflowEvent,
+    };
+  }
+
+  private mapRecipient(receiver: Recipient) {
```
CarsonF, line 167: This is kinda a meaningless name and doesn't apply to every recipient. Explicit return type here will help make the connection to where it's going and help your IDE help you and provide better type errors.
```suggestion
  private fakeUserFromEmailAddress(
    receiver: Recipient,
  ): EmailReportStatusNotification['recipient'] {
```

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    const membersWithValidRoles = await this.getTeamMembersWithValidRoles(
+      reportId,
+      next,
+    );
+
+    const envEmailList = this.getEnvList(next) as [];
+    const usersInEnvList: Recipient[] = await Promise.all(
+      envEmailList.map(async (item) => {
+        const user = await this.repo.getUserInfoByEmail(item);
```
CarsonF, line 71: It would be better grab all the users from email addresses at once.
```ts
const users = await this.repo.getUserIdByEmails(envEmailList);
const userIdByEmail: Record<string, ID> = mapFromList([
  ...envEmailList.map(email => ({ id: undefined, email }), // defaults for missing users
  ...users,
], (({ id, email }) => [email, id]);
```

CarsonF (changes requested):

src/components/progress-report/workflow/events/workflow-updated.event.ts
CarsonF: Which is why `status` is also available.

atGit2021: Since `next` contained the actual transition object which contains the `notify` property, I'm left with only a transitionId inside the workflowEvent, `workflowEvent.transition`. Which API (or other method) should I now use to obtain a transition object by transitionId in order to obtain the notify property in order to get the valid team member roles to notify.

CarsonF: Mmm yes you're right. I apologize. It could be possible to use the normal secured version of WorkflowEvent here which would have the transitions object. Or you could look it up easily like [this](https://github.com/SeedCompany/cord-api-v3/blob/3793f29ef7598c8cc9a075827d6c082995e119b9/src/components/progress-report/workflow/progress-report-workflow.service.ts#L46-L46). However, I probably wouldn't recommend those over what you have now.

Feel free to leave this as is 💪

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
atGit2021: I'm having trouble figuring out the new Neo4j cypher query for the new query by array that you prefer to use.  I've tried several different ways but none have worked so far.  This is the latest that I've tried so far, but still no success; please let me know what the correct query should be:
 `  async getUserIdByEmails(emails: string[]) {
    const query = this.db
      .query()
      .match([
        node('email', 'EmailAddress'),
        relation('in', '', 'email', ACTIVE),
        node('user', 'User'),
      ])
      .where({ 'user.email': inArray(emails) })
      .return<{
        users?: User[];
      }>(['user as users']);
    return await query.run();
  }`

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
atGit2021: I eliminated the session because I can retrieve from the workflowEvent, but I need the previousStatus since its not located anywhere else, and I need the reportId (and just tweaked the code to pass in the projectId and languageId per your other feedback) so I can generate the pertinent objects using the recipientSession.

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
CarsonF: That's close!
```diff
- .where({ 'user.email': inArray(emails) })
+ .where({ 'email.value': inArray(emails) })
```
Then
```ts
.return<{ id: ID, email: string }>([
	'user.id as id',
	'email.value as email',
])
```

src/core/email/templates/progress-report-status-changed.template.tsx
```diff
@@ -0,0 +1,91 @@
...
+          </>
+        ) : (
+          `${reportLabel} has a new status`
+        )}
+      </Heading>
+
+      <Section>
+        <Column>
+          <Text paddingBottom={16}>
+            {changerName ? <a href={changerUrl}>{changerName}</a> : 'Someone'}{' '}
```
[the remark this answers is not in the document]
atGit2021: When you say, "break out into their own components", are you saying the above code suggestion should be in a different file and imported, or just added to the top of this template.tsx?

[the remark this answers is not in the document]
atGit2021: never mind, I think I figured it out.

CarsonF (approved):
Looks great @atGit2021! I just did some last 10% polish. This was a huge task, great job 🎉

Issue #2345, opened 2022-01-14
Collapse long scripture product labels


CarsonF:
Currently we have labels like:
> 1 Samuel 1:12–1:18, 1:21–1:23, 2:1–3:2, 3:20–7:17, 9:4–9:14, 9:17–9:22, 10:1–10:8, 10:12, 10:25–12:15, 13:15–14:46, 14:49–14:52, 15:3–15:6, 15:12, 15:24–15:26, 15:30–15:33, 16:8–16:9, 16:14–18:4, 18:6, 18:10–18:24, 19–20, 21:7–25:44

If there are more than X ranges (maybe 3?) then we should convert to something like
> 1 Samuel 1:12–1:18, 1:21–1:23, and 19 other portions

Maybe this collapse number can be a graphql augment with a default value so it can be customized a bit more based on what the UI needs.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466928294) by [Unito](https://www.unito.io)

PR #2779, opened 2023-03-22
Configurable scripture range label collapsing

elisabeth-sorrell:
Fixes https://github.com/SeedCompany/cord-api-v3/issues/2345

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4185236236) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -110,7 +110,18 @@ export class ProductResolver {
       could return null.
     `,
   })
-  label(@Parent() product: AnyProduct): string | null {
+  label(
+    @Parent() product: AnyProduct,
+    @Args('collapseAfter', {
+      nullable: true,
+      defaultValue: 3,
+      description: stripIndent`
+        Collapses ranges after a Scripture book after showing the specified
+        number of ranges then shows how many more ranges are after that number.
+        Default is 3.`,
+    })
+    collapseAfter: number,
+  ): string | null {
     if (product.placeholderDescription.value) {
       return product.placeholderDescription.value;
     }
@@ -130,7 +141,10 @@ export class ProductResolver {
         const totalVerses = Book.find(book).totalVerses;
         return `${book} (${verses} / ${totalVerses} verses)`;
       }
-      return labelOfScriptureRanges(product.scriptureReferences.value);
+      return labelOfScriptureRanges(
+        product.scriptureReferences.value,
+        collapseAfter,
+      );
     }
     if (!product.produces.value) {
       return null;
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -113,6 +113,7 @@ export class ProductResolver {
   label(
     @Parent() product: AnyProduct,
     @Args('collapseAfter', {
+      nullable: true,
       defaultValue: 3,
       description: stripIndent`
         Collapses ranges after a Scripture book after showing the specified
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -120,6 +120,12 @@ export class ProductResolver {
         number of ranges then shows how many more ranges are after that number.
         A value of <= 0  collapses all Scripture ranges of that book.
         Default is 0 (collapse all ranges).
+
+        For example, with a \`collapseAfter\` value of 2:
+        \`Genesis 1:2, 1:5, 1:7, 1:9\` becomes \`Genesis 1:2, 1:5 and 2 other portions\`
+        with a \`collapseAfter\` value of 0 or less:
+        \`Genesis 1:2, 1:5, 1:7, 1:9\` remains the same.
+
       `,
       type: () => Int,
       defaultValue: 0,
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -1,6 +1,7 @@
 import {
   Args,
   Info,
+  Int,
   Mutation,
   Parent,
   Query,
@@ -113,13 +114,15 @@ export class ProductResolver {
   label(
     @Parent() product: AnyProduct,
     @Args('collapseAfter', {
-      nullable: true,
-      defaultValue: 3,
+      nullable: false,
       description: stripIndent`
         Collapses ranges after a Scripture book after showing the specified
         number of ranges then shows how many more ranges are after that number.
-        Default is 3.
+        A value of <= 0  collapses all Scripture ranges of that book.
+        Default is 0 (collapse all ranges).
       `,
+      type: () => Int,
+      defaultValue: 0,
     })
     collapseAfter: number,
   ): string | null {
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -118,7 +118,8 @@ export class ProductResolver {
       description: stripIndent`
         Collapses ranges after a Scripture book after showing the specified
         number of ranges then shows how many more ranges are after that number.
-        Default is 3.`,
+        Default is 3.
+      `,
     })
     collapseAfter: number,
   ): string | null {
--- a/src/components/product/product.resolver.ts
+++ b/src/components/product/product.resolver.ts
@@ -113,7 +113,6 @@ export class ProductResolver {
   label(
     @Parent() product: AnyProduct,
     @Args('collapseAfter', {
-      nullable: true,
       defaultValue: 3,
       description: stripIndent`
         Collapses ranges after a Scripture book after showing the specified
--- a/src/components/scripture/labels.ts
+++ b/src/components/scripture/labels.ts
@@ -91,13 +91,14 @@ export const labelOfScriptureRanges = (
   const hasSame = (key: keyof ScriptureReference) =>
     uniq(refs.flatMap((ref) => [ref.start[key], ref.end[key]])).length === 1;
   const same = hasSame('book') ? 'book' : undefined;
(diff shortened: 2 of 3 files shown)


src/components/product/product.resolver.ts
```diff
@@ -110,7 +110,18 @@ export class ProductResolver {
...
-  label(@Parent() product: AnyProduct): string | null {
+  label(
+    @Parent() product: AnyProduct,
+    @Args('collapseAfter', {
+      nullable: true,
+      defaultValue: 3,
+      description: stripIndent`
+        Collapses ranges after a Scripture book after showing the specified
+        number of ranges then shows how many more ranges are after that number.
+        Default is 3.`,
```
CarsonF, line 121:
```suggestion
        Default is 3.
     `,
```

src/components/scripture/labels.ts
```diff
@@ -88,7 +91,16 @@ export const labelOfScriptureRanges = (refs: readonly ScriptureRange[]) => {
   const hasSame = (key: keyof ScriptureReference) =>
     uniq(refs.flatMap((ref) => [ref.start[key], ref.end[key]])).length === 1;
   const same = hasSame('book') ? 'book' : undefined;
-  const labels = refs.map((ref) => labelOfScriptureRange(ref, same)).join(', ');
+  const labels = refs.map((ref) => labelOfScriptureRange(ref, same));
```
CarsonF, line 94: No need to create labels if we are going to drop them. I'd do the slice here.
```suggestion
  const labels = refs.slice(0, collapseAfter).map((ref) => labelOfScriptureRange(ref, same));
```

src/components/scripture/labels.ts
```diff
@@ -88,7 +91,16 @@ export const labelOfScriptureRanges = (refs: readonly ScriptureRange[]) => {
...
   const prefix = same === 'book' ? `${refs[0].start.book} ` : '';
-  return `${prefix}${labels}`;
+  const collapseLabels = (labels: string[]) => {
+    if (collapseAfter && labels.length > collapseAfter) {
+      const collapsed = labels.slice(0, collapseAfter);
+      collapsed.push(`and ${labels.length - collapseAfter} other portions`);
+      return collapsed.join(', ');
+    }
+    return labels.join(', ');
```
CarsonF, line 102: I found out about the Intl.ListFormat API recently. It could help format this string.
```tsx
const labelOutput = new Intl.ListFormat().format(labels)
```
Would still need to throw in the last segment.
```ts
[
   ...labels,
   ...(labels.length > collapseAfter ? [`${labels.length - collapseAfter} other portions`] : []),
]
```

CarsonF (changes requested):
Woah an unexpected PR! Welcome back 😄

src/components/scripture/labels.ts
elisabeth-sorrell: Just noting that I made a const to keep the original length of the refs because we lose that on the split

elisabeth-sorrell:
Made  `collapseAfter` not nullable. For one thing, it makes all the ranges disappear, which is fixable, but I was also figuring that making it nullable is just not needed right now. Don't see the necessity of doing it. Probably missed something though.

CarsonF:
> Made `collapseAfter` not nullable. For one thing, it makes all the ranges disappear, which is fixable, but I was also figuring that making it nullable is just not needed right now. Don't see the necessity of doing it. Probably missed something though.

Looks nullable still to me, and I think that's fine and better!

src/components/scripture/labels.ts
```diff
@@ -88,7 +91,16 @@ export const labelOfScriptureRanges = (refs: readonly ScriptureRange[]) => {
...
     uniq(refs.flatMap((ref) => [ref.start[key], ref.end[key]])).length === 1;
   const same = hasSame('book') ? 'book' : undefined;
-  const labels = refs.map((ref) => labelOfScriptureRange(ref, same)).join(', ');
+  const totalRefs = refs.length;
+  const labels = refs
+    .slice(0, collapseAfter)
+    .map((ref) => labelOfScriptureRange(ref, same));
   const prefix = same === 'book' ? `${refs[0].start.book} ` : '';
-  return `${prefix}${labels}`;
+  const labelOutput = new Intl.ListFormat().format([
```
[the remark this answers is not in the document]
elisabeth-sorrell: Wait. What about if it's already truncated? we'll have the output 
```
Genesis 1:1–1:13, 2:2–2:20, 3:1–3:20, 3 other portions
```
maybe, if the prefix is the same and it's truncated, we keep the default style so we have
```
Genesis 1:1–1:13, 2:2–2:20, 3:1–3:20, and 3 other portions
```

src/components/scripture/labels.ts
```diff
@@ -88,7 +91,16 @@ export const labelOfScriptureRanges = (refs: readonly ScriptureRange[]) => {
   const hasSame = (key: keyof ScriptureReference) =>
     uniq(refs.flatMap((ref) => [ref.start[key], ref.end[key]])).length === 1;
   const same = hasSame('book') ? 'book' : undefined;
-  const labels = refs.map((ref) => labelOfScriptureRange(ref, same)).join(', ');
+  const totalRefs = refs.length;
```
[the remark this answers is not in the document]
elisabeth-sorrell: Yup. I like using that better, for sure. I'm going to use it for some other stuff too. let me know what you think.

src/components/scripture/labels.ts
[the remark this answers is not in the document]
CarsonF: Oh yeah. I agree to keep it long with the truncated message

CarsonF (commented):
Looks great @elisabeth-sorrell! Would mind adding a test or two for the new use case?

elisabeth-sorrell:
> Looks great @elisabeth-sorrell! Would mind adding a test or two for the new use case?

Will do. You should see something for that tomorrow :)

CarsonF (commented):
I'm thinking about this more, and I've got a few more thoughts. 
- I don't think we should default to 3 at GQL layer. I think it should be opt-in for consumers. 
- The GQL type right now is `Float`, but should be an `Int` instead.
- Negative numbers and 0 are not handled right now.
  I think 0 could just mean collapse after none.. thoughts?
  Then we could just treat all negatives as 0.
- I'm realising when you said nullable you meant at the GQL layer.
  I think there's two options:
  - allow nullable and require `collapseAfter > 0`
  - force a number (not nullable) and coerce `collapseAfter <= 0` to 0 meaning _all_, and then 0 could be the default value.

elisabeth-sorrell:
>   I think 0 could just mean collapse after none.. thoughts?
>   Then we could just treat all negatives as 0.
What would that look like? at the moment it would probably treat it as
```
Genesis and 19 other portions
```
I don't really like this because it is not clear that the 19 portions are from Genesis. Maybe if it were something like
```
19 portions from Genesis
```
it would be clearer, but... the grammar fanatic in me isn't really liking it. However I like this output better than the first. What do you think? 

> * I'm realising when you said nullable you meant at the GQL layer.
>   I think there's two options:
>   
>   * allow nullable and require `collapseAfter > 0`
>   * force a number (not nullable) and coerce `collapseAfter <= 0` to 0 meaning _all_, and then 0 could be the default value.
I like the second option better. I would rather deal with numbers than `null`.

CarsonF:
> What do you think?

I don't think we are on the same page. I was saying zero would display all and truncate none.
Your test could change to this:
```diff
- ['Matthew 1:1, 1:4, 1:20, 2:1', undefined, 'Matthew 1:1, 1:4, 1:20, 2:1'],
+ ['Matthew 1:1, 1:4, 1:20, 2:1', 0, 'Matthew 1:1, 1:4, 1:20, 2:1'],
```

> I like the second option better. I would rather deal with numbers than null.

I think I agree with that, though I typically say null is should be used to represent the absence of a value not some magic string/number.

elisabeth-sorrell:
> I was saying zero would display all and truncate none. 

Ah, I'm really glad that I misunderstood that! This makes a lot more sense.

elisabeth-sorrell:
You know what? I'm wondering if whether I should rename `collapseAfter` to `truncate`. The main reasoning is because it makes more sense that `truncate = 0` would not truncate anything, but `collapseAfter = 0` isn't very clear that it would do the same thing. 
You think `truncate` would be a better name?

elisabeth-sorrell:
> Your test could change to this:
> 
> ```diff
> - ['Matthew 1:1, 1:4, 1:20, 2:1', undefined, 'Matthew 1:1, 1:4, 1:20, 2:1'],
> + ['Matthew 1:1, 1:4, 1:20, 2:1', 0, 'Matthew 1:1, 1:4, 1:20, 2:1'],
> ```

I added in the test for zero, and kept the test for `undefined`. The main reasoning was because `labelOfScriptureRanges` appears in the `extract-products-from-pnp` handler, and I had thought that keeping the `collapseAfter` param in `labelOfScriptureRanges` optional would help keep that piece from breaking.

CarsonF:
> You think `truncate` would be a better name?

I like `collapseAfter`. Truncate means just dropping, and we're not doing that we're grouping them together into a count.
`After` conveys it's the count to keep and not collapse. 
So really the positive name would be something like `enumerateAtLeast`.
I don't think what you have is any less clear than that.
You could give an example in the arg description. That would probably be more helpful than debating over a name.

CarsonF (approved):
Thanks @elisabeth-sorrell!!

PR #2783, opened 2023-03-30
Migrate db to add location map image nodes

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4234346779) by [Unito](https://www.unito.io)


willdch (approved):
looks good to me.

PR #2786, opened 2023-04-05
Sync S3 signing expiration with http cache

CarsonF:
This fixes a fresh cache url pointing to an invalid s3 url.

This moves the signing expiration from the Bucket layer to the FileService layer. This makes more sense as it's a business decision vs the actual persistence logic.

I also took the opportunity to refactor to use aws-sdk v3 command classes.
Doing it this way with generics is more elegant, as we don't need to explicitly reference actions and their input options in bucket abstraction.

---

In the future, maybe we can set the S3 files to be public if they are in cord, and skip url signing for them. That would allow us to increase our caching duration. Another option would be to put a CDN in front, I think that would allow the CDN to serve the fresh cached file directly instead of redirecting to a stale pre-signed url.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4263443216) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/file/bucket/file-bucket.ts
+++ b/src/components/file/bucket/file-bucket.ts
@@ -1,51 +1,30 @@
 import {
   GetObjectOutput as AwsGetObjectOutput,
-  GetObjectCommandInput,
   HeadObjectOutput,
-  PutObjectCommandInput,
 } from '@aws-sdk/client-s3';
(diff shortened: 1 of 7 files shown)


bryanjnelson (approved):
:shipit:

PR #2790, opened 2023-04-10
removed organization property

atGit2021:
@CarsonF - We think that this change will solve the issue of an FA not being able to view a Partner that is not attached to a project yet.  This is problematic currently when an FA creates a new Partner and then can't see it.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4285173590) by [Unito](https://www.unito.io)


bryanjnelson (approved):

CarsonF (approved):
🤷‍♂️

PR #2792, opened 2023-04-21
Ability to swap a project member for another across projects

### LINKED ISSUE #2490: Ability to Replace Project Members in Response to Staff Transitions
Although not super common, we often receive requests to replace a project member with a different project member.   A common example of this would be an FPM transitions out of a role and another FPM takes over.  The need would be to swap the project member out of about 20 - 40 projects.  There are also use cases where an even greater number of projects are affected.

There is currently no easy way to achieve this functionality.  Something that could achieve this would be a way through the API to mass update in bulk, likely with some options for only updating active projects, etc.  Potentially an Admin UI piece could even exist to allow non-developers to perform the task.

elisabeth-sorrell:
closes https://github.com/SeedCompany/cord-api-v3/issues/2490


sethmcknight:

PR #2804, opened 2023-05-12
Match by start date instead of end date on periodic report by date query.

bryanjnelson:


CarsonF:
If instead of adjusting the input value, we just change the condition for reports within the value, then the final report could also be queried if needed. Start >= input >= end

bryanjnelson:
Well, I wasn't changing the input...I was just comparing the input against something different (start instead of end).  But yes, your check will allow us to query for the final report as well if I'm following.  But if you send in the end date of the project, won't you still get both the "last regular" report and the "final" report with your logic?

CarsonF:
Yeah true. Technically periodic reports are not constrainted to a quarter. Some financial are monthly for example. So "start of quarter" could an incorrect assumption. Within the report range is safer.

You are right though that trying to get final with end date would match both.

Maybe this should be changed to a list output?

CarsonF (approved):

PR #2810, opened 2023-06-28
EdgeDB Foundation

CarsonF:
This is the foundation of our EdgeDB integration.

# What's here?

## NestJS module

A bare bones `EdgeDB` module, has been created. It does initialize and expose a client.
```ts
@Injectable()
class Foo {
  constructor(private db: EdgeDB) {}
  one() {
    return this.db.querySingle('select 1;');
  }
}
```

## Initial Schema
A small initial schema mostly just to have _something_.
- `type User` (minus links to other schema not yet created)
- `abstract type Resource`
- `currentUserId` global

## Code Generation
- Their TS query builder is generated which provides strong types based on our schema
- Our EdgeDB schema is generated as TS types/interfaces
- `*.edgeql` query files generate `*.edgeql.ts` files containing strongly typed queries to use in app code.
     Just like how `*.graphql` files work in other repos.
- These files are ignored from git & excluded from linting

## Custom Scalars
- A seemly trivial, but critical part of integrating with the existing codebase. 
- Any EdgeDB `uuid`s are typed as our `ID` type (instead of raw `string`)
- EdgeDB's `std::datetime` & `cal::local_date` use Luxon's `DateTime` & our `CalendarDate` respectively. This applies to both type generation and runtime value conversion.
- Code is expandable to other custom scalars needed in future

## CI/CD
- Our GitHub Actions workflows have been updated to setup EdgeDB server, migrate to our schema, and generate the above TS files.
- Our Dockerfile now generates the above TS files at build via an internal EdgeDB server & our current schema
- Our Dockerfile now exposes their `edgedb` cli tool in the final image.
     This allows the image to be called migrate any connected DB to the schema defined by the image.

# What's Next?

- Merge infra PRs to setup EdgeDB instances via CDK & run migrations on deploy
- Connecting current `Session` to the `currentUserId` global
- Implicitly having GQL mutations to run inside of an EdgeDB transaction
- More of our actual schema
- Determine how "auth context data" will be passed down our schema trees; namely: sensitivity & project membership.
- Create EdgeDB Access Policies dynamically based on our Auth Policies. At least for object-level reads, so our list queries can work without having to worry about this.
- Consider inline queries with template tag, a la `gql` code gen. Having inline code blocks for small queries seems nice compared to a required _individual_ edgeql file.
- Port queries/repositories

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4721319851) by [Unito](https://www.unito.io)


bryanjnelson (approved):
Foundational, indeed!  Thanks @CarsonF!

PR #2816, opened 2023-07-13
Bump xlsx 18.5 -> 20.0

CarsonF:
Idk about this ⬇️ 
> Use UTC interpretation of Date objects for date cells (potentially breaking)

https://git.sheetjs.com/sheetjs/sheetjs/src/branch/master/CHANGELOG.md

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4806959056) by [Unito](https://www.unito.io)

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "source-map-support": "^0.5.21",
     "ts-essentials": "^9.3.2",
     "winston": "^3.10.0",
-    "xlsx": "^0.18.5",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.0/xlsx-0.20.0.tgz",
     "yaml": "^2.3.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


atGit2021 (approved):
LGTM!

atGit2021:
@CarsonF , I read some of the change log files in sheetjs repo and looked for anything that seemed significant for how we use the xlsx package.  I don't believe there will be any impacts.  We use that package for extraction purposes but our date/time calculations are being done using the luxon package.  I pulled your branch down and tested a few PnP planning and progress extractions/uploads to see if there were any negative effects, but all looks ok.  This is not a thorough analysis but not sure what else to test to see if I can break the code.

CarsonF:
Thanks for digging in @atGit2021. I believe you are right. Also our date logic already was using the native JS date.
https://github.com/SeedCompany/cord-api-v3/blob/8bbbae287d11685f8520b4ebe59275762a5a3d73/src/common/xlsx.util.ts#L246-L247
So it's not like we we accounting for anything special there with regard to timezones.

PR #2817, opened 2023-07-13
Change timezone data caching strategy

### LINKED ISSUE #664: Resiliency for timezone file
Currently our timezone API's will break in production if GitHub API is unavailable or if the repo changes its structure.

To combat this, whenever we successfully retrieve the file we should write it to a disk cache.
This file can then be used when the API is unavailable or even as a _cache-then-refresh_ policy.

The file should be grabbed and written to disk as apart of the docker image creation. This ensures that a file is always available for that image regardless of external events. If the file fails to download we should error out on the build so we know that we need to make a change to the process.

- - -
This is a low priority. But if you want to do it talk to me first.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/3466973603) by [Unito](https://www.unito.io)

CarsonF:
Fixes #664

Previously, we only cached the data for 5 minutes. This data actually only changes around once a month.

Instead, we will grab it at build and save it to the image. This way it will be automatically fresh as of build time, and we won't rely on GitHub in production.

For dev mode, we will just grab it anytime it doesn't exist on disk.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4807021557) by [Unito](https://www.unito.io)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,9 @@ RUN echo GIT_BRANCH=$GIT_BRANCH >> .env
 # This stage can run everything
 FROM node as dev
 
+# Grab latest timezone data
+RUN mkdir -p .cache && curl -o .cache/timezones https://raw.githubusercontent.com/moment/moment-timezone/master/data/meta/latest.json
+
 # Install dependencies (in separate docker layer from app code)
 COPY .yarn .yarn
 COPY patches patches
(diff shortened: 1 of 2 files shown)


andrewmurraydavid (approved):
A lot cleaner! Nice job, @CarsonF! 🚀

PR #2818, opened 2023-07-13
updates consultant roles

brentkulwicki:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4807365392) by [Unito](https://www.unito.io)

PR for 4558680508

### DIFF
--- a/src/components/authorization/policies/by-role/consultant-manager.policy.ts
+++ b/src/components/authorization/policies/by-role/consultant-manager.policy.ts
@@ -26,7 +26,9 @@ import { member, Policy, Role, sensMediumOrLower, sensOnlyLow } from '../util';
   r.Partnership.read.specifically((p) => [
     p.many('partner', 'organization').whenAny(member, sensOnlyLow).read,
   ]),
-  r.ProgressReport.whenAny(member, sensMediumOrLower).read,
+  r.ProgressReport.whenAny(member, sensMediumOrLower).read.children((c) =>
+    [c.teamNews, c.communityStories, c.highlights].flatMap((it) => it.read),
+  ),
   r.Project.read
     .specifically((p) => [
       p.many('step', 'stepChangedAt').edit,
--- a/src/components/authorization/policies/by-role/consultant-manager.policy.ts
+++ b/src/components/authorization/policies/by-role/consultant-manager.policy.ts
@@ -26,14 +26,7 @@ import { member, Policy, Role, sensMediumOrLower, sensOnlyLow } from '../util';
   r.Partnership.read.specifically((p) => [
     p.many('partner', 'organization').whenAny(member, sensOnlyLow).read,
   ]),
-  r.ProgressReport.whenAny(member, sensMediumOrLower).specifically(
-    (p) => p.status.read,
-  ),
-  [
-    r.ProgressReportCommunityStory,
-    r.ProgressReportHighlight,
-    r.ProgressReportTeamNews,
-  ].flatMap((it) => [it.read.specifically((p) => [p.responses.read])]),
+  r.ProgressReport.whenAny(member, sensMediumOrLower).read,
   r.Project.read
     .specifically((p) => [
       p.many('step', 'stepChangedAt').edit,
--- a/src/components/authorization/policies/by-role/consultant-manager.policy.ts
+++ b/src/components/authorization/policies/by-role/consultant-manager.policy.ts
@@ -26,6 +26,14 @@ import { member, Policy, Role, sensMediumOrLower, sensOnlyLow } from '../util';
   r.Partnership.read.specifically((p) => [
     p.many('partner', 'organization').whenAny(member, sensOnlyLow).read,
   ]),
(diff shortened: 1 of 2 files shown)


src/components/authorization/policies/by-role/consultant-manager.policy.ts
```diff
@@ -26,6 +26,14 @@ import { member, Policy, Role, sensMediumOrLower, sensOnlyLow } from '../util';
   r.Partnership.read.specifically((p) => [
     p.many('partner', 'organization').whenAny(member, sensOnlyLow).read,
   ]),
+  r.ProgressReport.whenAny(member, sensMediumOrLower).specifically(
+    (p) => p.status.read,
+  ),
```
CarsonF, line 31: I think we want to be able to read the object/any property, right? In that cast it would be this:
```suggestion
  r.ProgressReport.whenAny(member, sensMediumOrLower).read,
```

src/components/authorization/policies/by-role/consultant-manager.policy.ts
```diff
@@ -26,6 +26,14 @@ import { member, Policy, Role, sensMediumOrLower, sensOnlyLow } from '../util';
...
+  [
+    r.ProgressReportCommunityStory,
+    r.ProgressReportHighlight,
+    r.ProgressReportTeamNews,
+  ].flatMap((it) => [it.read.specifically((p) => [p.responses.read])]),
```
CarsonF, line 36: I believe the call out for _read_ perm on `responses` is redundant as that's the object has _read_ perm as well.
```suggestion
  ].flatMap((it) => [it.read]),
```

brentkulwicki: I believe we do need this because of the way it is nested. Otherwise, the sections show but the responses themselves don't. Bryan and I paired on this yesterday afternoon.

src/components/authorization/policies/by-role/consultant.policy.ts
```diff
@@ -22,6 +22,14 @@ import { inherit, member, Policy, Role } from '../util';
...
   r.PeriodicReport.read,
   r.Product.read,
+  r.ProgressReport.when(member).read,
+  r.ProgressReportVarianceExplanation.when(member).read,
+  r.ProgressReportWorkflowEvent.when(member).read,
+  [
+    r.ProgressReportCommunityStory,
+    r.ProgressReportHighlight,
+    r.ProgressReportTeamNews,
+  ].flatMap((it) => [it.read.specifically((p) => [p.responses.read])]),
```
CarsonF, line 32: I think the idea is to allow CM's to see these as well for low/medium sens projects that that are _NOT_ a member of, right? If so, we need to call this out in CM too with that _or_ condition.

src/components/authorization/policies/by-role/consultant-manager.policy.ts
CarsonF: I Dont Believe You Will Ferrell - 6JB4v4xPTAQFi

CarsonF (commented):
Your test failures are unrelated FYI

brentkulwicki:
I think this has it all squared away now. 

Consultant + Consultant manager can see progress reports including the team updates, story and progress as members of a project

Consultant manager can see any of these things for any Low or Medium sensitivity projects

src/components/authorization/policies/by-role/consultant.policy.ts
```diff
@@ -22,6 +22,13 @@ import { inherit, member, Policy, Role } from '../util';
   r.Partnership.when(member).read,
   r.PeriodicReport.read,
   r.Product.read,
+  r.ProgressReport.when(member).read.children((c) =>
+    [c.teamNews, c.communityStories, c.highlights].flatMap((it) => it.read),
+  ),
+  r.ProgressReportVarianceExplanation.when(member).read,
+  r.ProgressReportTeamNews.read.specifically((p) => p.responses.read),
+  r.ProgressReportCommunityStory.read.specifically((p) => p.responses.read),
+  r.ProgressReportHighlight.read.specifically((p) => p.responses.read),
```
CarsonF, line 31: I believe what we want here is
```ts
[
  r.ProgressReportCommunityStory,
  r.ProgressReportHighlight,
  r.ProgressReportTeamNews,
  r.ProgressReportVarianceExplanation,
].flatMap((it) => it.when(member).read),
```

Your lines 26 is trying to be the same thing as 29-31. So we don't need both, and we prefer the non-children syntax for static parent cases.

The other difference is that 29-31 allows reading globally, instead of only if a member. Which I don't think is the intent.

Also I know we talked before about this, but the
```ts
.specifically((p) => p.responses.read),
```
shouldn't be needed as it's the same action permission as what the object level has.

CarsonF (changes requested):

brentkulwicki:
Fixed here: https://github.com/SeedCompany/cord-api-v3/pull/2853#pullrequestreview-1559317546

PR #2823, opened 2023-07-19
Fix policies incorrectly defaulting to object level action when a property is specifically called out

CarsonF:
Here's an example of something being ignored.
https://github.com/SeedCompany/cord-api-v3/blob/ad02252edfda9eb300e3c15282aceabac3fb661e/src/components/authorization/policies/by-role/consultant-manager.policy.ts#L20-L20
It was falling back to organization object level _read_ instead of using the absence of actions explicitly called out on the `address` property.

### DIFF
--- a/src/components/authorization/policies/by-role/financial-analyst.policy.ts
+++ b/src/components/authorization/policies/by-role/financial-analyst.policy.ts
@@ -20,7 +20,7 @@ import {
       r.Engagement.read
         .when(member)
         .create.delete.specifically((p) => [
-          p.many('disbursementCompleteDate', 'status').when(member).edit,
+          p.many('disbursementCompleteDate', 'status').edit,
         ]),
       r.LanguageEngagement.specifically((p) => [p.paratextRegistryId.none]),
     ),
--- a/src/components/authorization/policies/by-role/financial-analyst.policy.ts
+++ b/src/components/authorization/policies/by-role/financial-analyst.policy.ts
@@ -59,7 +59,7 @@ import {
             'financialReportPeriod',
             'financialReportReceivedAt',
           )
-          .when(member).edit,
+          .read.when(member).edit,
       ])
       .children((c) => c.posts.edit),
     r.ProjectMember.read.when(member).edit.create.delete,
--- a/src/components/authorization/policies/by-role/regional-director.policy.ts
+++ b/src/components/authorization/policies/by-role/regional-director.policy.ts
@@ -6,7 +6,7 @@ import { member, Policy, Role, sensMediumOrLower } from '../util';
 
   r.Partnership.read,
   r.Project.when(member).edit.specifically(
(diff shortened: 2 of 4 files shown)


bryanjnelson (approved):
🚢

src/components/authorization/policies/by-role/financial-analyst.policy.ts
```diff
@@ -59,7 +59,7 @@ import {
             'financialReportPeriod',
             'financialReportReceivedAt',
           )
-          .when(member).edit,
+          .read.when(member).edit,
```
bryanjnelson, line 62: Edit implies read, correct?  But in this case we want no member restrictions on reads so that's why the change?

src/components/authorization/policies/by-role/regional-director.policy.ts
```diff
@@ -6,7 +6,7 @@ import { member, Policy, Role, sensMediumOrLower } from '../util';
 
   r.Partnership.read,
   r.Project.when(member).edit.specifically(
-    (p) => p.rootDirectory.when(sensMediumOrLower).read,
+    (p) => p.rootDirectory.edit.when(sensMediumOrLower).read,
```
bryanjnelson, line 9: So here the root directory was not editable, but now you are making it editable?

bryanjnelson (approved):

src/components/authorization/policies/by-role/financial-analyst.policy.ts
CarsonF: Exactly. We call out these properties because we want to _add_ **edit** access in certain cases. But because they are called out, the object level **read** is ignored.
So we need to "restore" that before widening the scope.

src/components/authorization/policies/by-role/regional-director.policy.ts
CarsonF: Logically there's no change here. This was added to restore previous functionality with this bug fix.

We want `rootDirectory` to be editable when you are a member.
Which the `.edit` does since it takes the closest previously defined condition.

Like above, we are additionally adding read access for low/med sens projects even if you are not a member.
But because we called that out we also need to restore the object-level permission on this property.

CarsonF:
As two different use cases are revealing, it could be worth it to have another method.
```ts
// edit everything, but this one prop.
X.edit.narrowProps(p => p.prop.read)
```
```ts
// read everything including prop, and also allow editing prop when member.
X.read.widenProps(p => p.prop.when(member).edit)
```

bryanjnelson:
> As two different use cases are revealing, it could be worth it to have another method.
> 
> ```ts
> // edit everything, but this one prop.
> X.edit.narrowProps(p => p.prop.read)
> ```
> 
> ```ts
> // read everything including prop, and also allow editing prop when member.
> X.read.widenProps(p => p.prop.when(member).edit)
> ```

This would probably make things more clear, yes.

PR #2830, opened 2023-07-21
Refactor how we adjust the neo4j retry logic

CarsonF:
In a lot of places we blindly throw a nicer exception wrapping the original, which gives the nicer message to the API caller.
Here's an example:
https://github.com/SeedCompany/cord-api-v3/blob/71d27115f7bcdbbe4e6d5d2f9088c3f8a2317117/src/components/project/project.service.ts#L183-L183

This is a problem when the error is transient database error, which the transaction executor should retry.
To work around this, we need to also check if the exception thrown wraps a _retriable_ database error.

We previously did this by digging into the neo4j lib and monkey patched the "should retry" check to look at our "previous" exceptions.
This is invasive and doesn't work with ESM where module exports are immutable.

Instead, we can change our transaction handling logic.
Now when an error is thrown, before we hand it to neo4j's `TransactionExecutor` to check for "should retry", we look for the first previous exception that's a `Neo4jError`.
We give that error to `TransactionExecutor`, and associate the original in a map.
Finally, after `TransactionExecutor` deems the error is not retryable, we unwrap it from our previously saved mapping.

This is much better and assumes less about the Neo4j driver's inner workings.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4853819962) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/database/cypher.factory.ts
+++ b/src/core/database/cypher.factory.ts
@@ -7,7 +7,7 @@ import type { Driver, Config as DriverConfig, Session } from 'neo4j-driver';
 import type { LoggerFunction } from 'neo4j-driver-core/types/types';
 import type QueryRunner from 'neo4j-driver/types/query-runner';
 import { Merge } from 'type-fest';
-import { csv, getPreviousList } from '~/common';
+import { csv } from '~/common';
 import { dropSecrets } from '~/common/mask-secrets';
 import { ConfigService } from '../config/config.service';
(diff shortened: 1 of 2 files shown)


adam-soltech (approved):
LGTM. I was going to suggest not using a nested try/catch but it looks like we need the outer try/catch to unwrap the original error

bryanjnelson:
YES.

PR #2834, opened 2023-07-24
Ability to have engagements where the Intern/Language is uncertain.

### LINKED ISSUE #1121: Anticipated Engagements Count
Often when a Project is in Early Conversations and Concept Development, we have an idea of how many Engagements might be a part of the Project, but don't know the exact Languages, People, or Partners who will be engaged.

Without this number, Leadership is blind to what may be coming in future years, making it difficult to forecast financial and personnel needs. We need a way to capture the number of anticipated engagements in order to provide these projections.

elisabeth-sorrell:
Closes https://github.com/SeedCompany/cord-field/issues/1121

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4863922671) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/engagement/dto/create-engagement.dto.ts
+++ b/src/components/engagement/dto/create-engagement.dto.ts
@@ -32,14 +32,23 @@ export abstract class CreateEngagement {
 
   @Field(() => EngagementStatus, { nullable: true })
   readonly status?: EngagementStatus;
+
+  @Field({
+    nullable: true,
+    description: stripIndent`
+      User can set when the language or intern for the 
+      engagement is unknown as of yet.
+    `,
+  })
+  readonly nameWhenUnknown?: string;
 }
 
 @InputType()
 export abstract class CreateLanguageEngagement extends CreateEngagement {
   static readonly Props = keysOf<CreateLanguageEngagement>();
 
-  @IdField()
-  readonly languageId: ID;
+  @IdField({ nullable: true })
+  readonly languageId?: ID;
 
   @Field({ nullable: true })
   readonly firstScripture?: boolean;
@@ -76,8 +85,8 @@ export abstract class CreateLanguageEngagement extends CreateEngagement {
 export abstract class CreateInternshipEngagement extends CreateEngagement {
   static readonly Props = keysOf<CreateInternshipEngagement>();
 
-  @IdField()
-  readonly internId: ID;
+  @IdField({ nullable: true })
+  readonly internId?: ID;
 
   @IdField({ nullable: true })
   readonly mentorId?: ID;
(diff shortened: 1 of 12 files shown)


elisabeth-sorrell:
Requirements: Need to be able to add a Language or Intern engagement without specifying the language or intern because there are times in a project when other engagement details are developing but the language or intern itself if unknown. There are also times when an incorrect language or intern is inadvertently added, and right now the only way to correct it is either deleting the engagement and creating a new one with the correct language or intern, or by going straight to the db to establish the correct connections.

The solution to these problems is by letting a user create an engagement without necessarily specifying the intern or language, and then letting the user update the engagement later on in the engagement’s detail page. When asked about how to distinguish between multiple unknown engagements, Seth said he wanted some sort of arbitrarily generated name for each unknown language/intern (perhaps Language 1,2,3, Intern 1,2,3 or Language A,B,C etc). 

- [ ] Make language and intern ids optional for engagements
Not so much of a big deal, just change the dtos. I also thought of maybe creating some newer abstract dtos like `Unknown[Language|Intern]Engagement` and then extending those with concrete classes `[Language|Intern]Engagement`. I wasn't really sure though about how much that really brought to the table other than confusion. So far, just making the ids optional seemed the most straightforward way for now.
- [ ] Once they are optional, each of the CRUD operations need to be double-checked and tweaked to handle an optional Language/Intern id
For instance, if a [Language|Intern] engagement is being updated, we need to make sure that we aren't updating to a duplicated [Language|Intern]. Also the queries need double checking to make sure that just adding a `?` to the ids would do the trick
- [ ] Find a way to identify each unknown engagement, I decided to go with the third idea below, because it doesn't seem to be very confusing. I don't really like the name `nameWhenUnknown` though... :\
I thought of a few ways we could do this...
  - We could make the [language|intern]ids a unioned type of `ID|Unknown_ID`, where the `ID` is the usual generated id that we do for neo4j and the Unknown_ID looks like `[Language|Intern] ${SOME_DIGIT}`. Some caveats I thought of doing that would mean that we have to make a parser to distinguish between known/unknown ids, we also don't _ever_ do ids like that, also matching would just turn into a mess because the unknown ids would actually return a result when we don't want to have one, not to mention the fact that we use that id to match a [Language|Intern] node.
  - We could let users define their own unknown Language/Intern engagement names. Not sure how that would look like graphically, and doesn't seem right to have a prop hanging around that's going to be dead half the time. And, it kind of defeats Seth's idea of auto-generating them.
  - We could add a calculated prop, `nameWhenUnknown`, that's generated when a user doesn't specify a language\intern. I don't really like the name, but I think we do need to somehow persist an unknown name in the db, and this seems to get the job done
  - Could have the front-end just generate names on the fly, without persisting the value at all. Not really into that, because for one thing, I don't think it's good practice to have the front end responsible for generating data, and for another thing, it's going to get confusing once users start going from unknown to known engagements. 

- [ ] identify and implement other business rules when it comes to having an unknown engagement (eg, what steps are allowed/ not allowed to have unknown engagements, etc)

sethmcknight:

PR #2835, opened 2023-07-24
Legacy ResourceMap → `@RegisterResource`

brentkulwicki:
This removes the LegacyResourceMap by using the `RegisterResource()` deocorator instead of keeping everything needing to be references in a big, static array.

Also, I noticed a mispelling of the `hightlights.dto.ts` => `highlights.dto.ts` so some of these changes are because of that change. 

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4864417402) by [Unito](https://www.unito.io)


src/components/authorization/model/resource-map.ts
```diff
@@ -1,100 +1,6 @@
...
-  InternshipProject,
-  ProjectMember,
-  PeriodicReport,
-  FinancialReport,
-  NarrativeReport,
-  ProjectChangeRequest,
-  Story,
-  Unavailability,
-  User,
   AssignableRoles,
```
brentkulwicki, line 5: @CarsonF - I am confused a little bit by this AssignableRoles. It does not fit the Resource shape (however, some of its properties fit the resource shape) - So if I try to move this to the decorator approach, it blows up a lot of stuff. But I am curious why this is actually a part of the Legacy Map in the first place with how different it is than the rest of the items I have removed.

CarsonF: What blows up?

brentkulwicki: it doesn't recognize that the decorator is a function when removing this specific resource. 

image

brentkulwicki: Trying to trace the threads of how this relates to all the other things that fail (because for science, I went and undid some of the changes for these resources that start throwing these errors) and essentially none of them will work once this one is removed.

CarsonF: Oh this is circular dependency problem with imports. These are long standing rough edges.
Try importing from `~/core/resources` so `core` as a whole is not referenced.

Also if you rebase for ESM that's now in develop, the ESM loader calls out explicitly where the import is a problem.

CarsonF (approved):
🎉 :shipit:

PR #2839, opened 2023-07-26
Replace cypher CLI syntax highlighting library

CarsonF:
The main reason for this is that the old library includes the TS DOM lib. This added those global types to our codebase, and we have no way to override that.
This is one of two libraries adding the DOM lib. I'll fully address this in another PR but this was big enough to be on its own.

In general, though, this new library is much better at calling the syntax. Though I did take some time to tweak it.

It's worth noting that it uses chalk which requires ESM. So, wouldn't have been easy to use this before now.

# Before #1975
Before
# After
After

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4877496982) by [Unito](https://www.unito.io)


bryanjnelson (approved):
🖊️

PR #2848, opened 2023-07-29
faker.js dummy data updates

brentkulwicki:
When running e2e tests locally, there are a ton of faker.js methods we use that are showing console warnings because they were deprecated and will be removed in the next version (Couple examples of these below). This PR goes ahead and moves those to the newer, recommended methods so that those warnings are gone. 
Screenshot 2023-07-29 at 3 06 09 PM
Screenshot 2023-07-29 at 3 02 30 PM

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4895252332) by [Unito](https://www.unito.io)


CarsonF (approved):
Appreciate it! 🤝

PR #2851, opened 2023-08-01
Finish migration to Privileges from AuthorizationService

brentkulwicki:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4913620532) by [Unito](https://www.unito.io)


src/components/product/migrations/SplitScriptureSpanningBooks.migration.ts
```diff
@@ -130,7 +130,11 @@ export class SplitScriptureSpanningBooksMigration extends BaseMigration {
       // refs: labelOfScriptureRanges(product.scriptureReferences),
       product,
     });
-    return ix(this.splitRefsByBook(product.scriptureReferences))
+    return ix(
+      this.splitRefsByBook(
+        product.scriptureReferences as unknown as DbScriptureReferences,
+      ),
+    )
```
brentkulwicki, line 137: I want to call this typing out specifically here @CarsonF - It feels like there is something downstream like the others we removed, but I couldn't find it. This doesn't feel correct, but outside of this, I wasn't sure where to go with a migration.

src/components/budget/budget.service.ts
```diff
@@ -168,20 +163,16 @@ export class BudgetService {
...
-      dto: result as ResourceShape<Budget>['prototype'],
-    });
+    const privs = this.privileges.for(session, Budget, result).all;
 
     const securedProps = parseSecuredProperties(
       result,
-      perms as PermissionsOf<Budget>,
+      privs,
       Budget.SecuredProps,
     );
```
CarsonF, line 172: Can we drop this for
```ts
privileges.secure(result)
```
It could replace the final return spread (except for `status` and `records`)

src/components/budget/budget.service.ts
```diff
@@ -168,20 +163,16 @@ export class BudgetService {
...
 
     let records = null;
-    if (perms.records.canRead) {
+    if (privs.records.read) {
```
CarsonF, line 175: FYI this could be
```ts
privileges.can('read, 'records')
```
You don't have to use the `.all` proxy.

src/components/budget/budget.service.ts
```diff
@@ -234,11 +225,9 @@ export class BudgetService {
...
     const result = await this.budgetRecordsRepo.readOne(id, { session, view });
 
-    const securedProps = await this.authorizationService.secureProperties(
-      BudgetRecord,
-      result,
-      session,
-    );
+    const securedProps = this.privileges
+      .for(session, BudgetRecord)
+      .secure(result);
```
CarsonF, line 230: Same here, `securedProps` could just be returned straight up.

CarsonF (changes requested):
I'll stop here for now as this covers a lot of tweaks needed.

In general, the `secure()` encompasses:
- unsecured props, `...dto`
- `securedProps`
- `canDelete`

So this kind of thing can be dropped
```ts
return {
  ...dto,
  ...secured,
  canDelete,
};
```
for 
```ts
return secure(dto)
```

Sorry I wasn't able to look at this sooner for you 😞

src/components/product/migrations/SplitScriptureSpanningBooks.migration.ts
CarsonF: We should probably just delete all our existing migrations at this point. They've all been ran where needed and we shouldn't need to maintain them anymore.

src/components/budget/budget.service.ts
brentkulwicki: Yea. I think I learned more as I went and forgot to come back and check these 🤦

src/components/product/migrations/SplitScriptureSpanningBooks.migration.ts
brentkulwicki: Neo4j ignorance here, but guessing we don't need these migrations for new dev setup? I know in the SQL world you have to keep migrations forever. Is that not the case in the graph world?

src/components/budget/budget.service.ts
brentkulwicki: Yea. I just saw securedProps being deprecated this morning as I was doing docs. Will re-factor this as well

src/components/product/migrations/SplitScriptureSpanningBooks.migration.ts
CarsonF: Both SQL/EdgeDB have schemas. So the migrations _mainly_ handled changes to those. Neo4j doesn't have a schema (on write), so these migrations only handled moving data around to the new format. As new DB's don't have this old data format to pull forward, the migrations are just ignored.

src/components/engagement/engagement.service.ts
```diff
@@ -258,35 +254,24 @@ export class EngagementService {
...
-        false
-      >;
-
       return {
         ...(dto as UnsecuredDto<LanguageEngagement>),
-        ...secured,
+        ...this.privileges
+          .for(session, LanguageEngagement)
+          .secure(dto as UnsecuredDto<LanguageEngagement>),
         canDelete,
```
brentkulwicki, line 268: @CarsonF - I know we talked about the newer privileges having everything we need (including delete permissions) but this feels a little bit different? Can this canDelete still be inferred from the privileges here making the entire `canDelete` variable removable?

src/components/story/story.service.ts
```diff
@@ -86,27 +89,14 @@ export class StoryService {
...
-    const securedProps = await this.authorizationService.secureProperties(
-      Story,
-      {
+    return {
+      ...dto,
+      ...this.privileges.for(session, Story).secure({
         ...dto,
         scriptureReferences: this.scriptureRefs.parseList(
-          dto.scriptureReferences,
+          dto.scriptureReferences as unknown as DbScriptureReferences,
```
CarsonF, line 97: These should be changed with the same strategy we used in products. Since we changed the Producible DTO, that affects all these as well.

brentkulwicki: Let me work on it again. This threw lots of null errors in the tests with that change and I had to revert on these. Maybe I need to update the tests? But it looked as if the structure changed for what GraphQL is looking for, so I reverted the changes.

src/components/product/product.service.ts
```diff
@@ -275,101 +273,19 @@ export class ProductService {
...
-        steps: {
-          ...securedProps.steps,
-          value: securedProps.steps.value ?? [],
-        },
-        canDelete,
-      };
-      return derivative;
+      return this.privileges
+        .for(session, DerivativeScriptureProduct)
+        .secure(dto);
```
CarsonF, line 279: Could your try a unified return like this: 
```ts
      return this.privileges
        .for(session, resolveProductType(dto))
        .secure(dto);
```

src/components/product/product.service.ts
```diff
@@ -382,12 +298,9 @@ export class ProductService {
...
     const changes = this.getDirectProductChanges(input, currentProduct);
 
-    await this.authorizationService.verifyCanEditChanges(
-      Product,
-      await this.secure(currentProduct, session),
-      changes,
-      'product',
-    );
+    this.privileges
+      .for(session, Product, currentProduct)
```
CarsonF, line 302:
```suggestion
      .for(session, DirectScriptureProduct, currentProduct)
```

src/components/product/product.service.ts
```diff
@@ -465,13 +378,9 @@ export class ProductService {
...
     const changes = this.getDerivativeProductChanges(input, currentProduct);
-
-    await this.authorizationService.verifyCanEditChanges(
-      Product,
-      await this.secure(currentProduct, session),
-      changes,
-      'product',
-    );
+    this.privileges
+      .for(session, Product, currentProduct)
```
CarsonF, line 382:
```suggestion
      .for(session, DerivativeScriptureProduct, currentProduct)
```

src/components/product/product.service.ts
```diff
@@ -569,12 +478,9 @@ export class ProductService {
...
     };
 
-    await this.authorizationService.verifyCanEditChanges(
-      Product,
-      await this.secure(currentProduct, session),
-      changes,
-      null,
-    );
+    this.privileges
+      .for(session, Product, currentProduct)
```
CarsonF, line 482:
```suggestion
      .for(session, OtherProduct, currentProduct)
```

src/components/project/project.service.ts
```diff
@@ -235,29 +235,9 @@ export class ProjectService {
...
-      },
-      tags: {
-        ...securedProps.tags,
-        value: securedProps.tags.canRead ? securedProps.tags.value : [],
-      },
-      canDelete: isIdLike(session)
-        ? false // Assume email workflow that doesn't need to know this. Skip lookup.
-        : session.roles.includes('global:Administrator'),
-      __typename: `${project.type}Project`,
+      ...this.privileges.for(session, IProject, project).secure(project),
```
CarsonF, line 240:
```ts
return this.privileges.for(session, IProject).secure(project);
```

src/components/user/education/education.service.ts
```diff
@@ -92,11 +83,7 @@ export class EducationService {
     }
     const changes = this.repo.getActualChanges(ed, input);
     if (result.id !== session.userId) {
-      await this.authorizationService.verifyCanEditChanges(
-        Education,
-        ed,
-        changes,
-      );
+      this.privileges.for(session, Education).verifyChanges(changes);
```
CarsonF, line 86: I hate to call this out as this feature isn't used, but it sets a bad precedent. We should have a policy declaring that owners can edit educations.
Here we've hacked around the old system that didn't support owner conditions.
At least we should leave a TODO saying this is bad and should be moved into the policies.

src/components/user/user.service.ts
```diff
@@ -202,22 +191,38 @@ export class UserService {
...
   async permissionsForListProp(
-    prop: keyof PermissionsOf<SecuredResource<typeof User>>,
-    session: Session,
-  ) {
-    const perms = await this.authorizationService.getPermissions({
-      resource: User,
-      sessionOrUserId: session,
-    });
-    return { ...perms[prop], canCreate: perms[prop].canEdit };
+    privs: Partial<Record<ResourceAction, boolean>>,
```
CarsonF, line 194: Let's drop this method. It's not adding any value.

CarsonF (changes requested):
Closer and closer!

src/components/user/education/education.service.ts
brentkulwicki: Going to add this as TODO and will create a ticket to the backlog to remove this pattern from all of these as this gets used in a lot of places

src/core/database/results/parse-secured.ts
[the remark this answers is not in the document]
brentkulwicki: Looks like it's not used anywhere now so deleting the whole thing

src/components/project/project.service.ts
```diff
@@ -559,12 +546,9 @@ export class ProjectService {
...
-    const perms = await this.authorizationService.getPermissions({
-      resource: User,
-      sessionOrUserId: session,
-    });
+    const perms = this.privileges.for(session, User).all.projects;
 
-    if (!perms.projects.canRead) {
+    if (!perms.read) {
       return SecuredList.Redacted;
     }
```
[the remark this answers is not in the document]
brentkulwicki: So really we can drop perms here alltogether then as this little check is the only reason it gets used anyway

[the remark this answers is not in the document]
brentkulwicki: Actually with the comment above this, I think there is a broader issue with the DB like you indicated
```
    // Instead of trying to handle which subset of projects should be included,
    // based on doing the work of seeing which project teams they can view,
    // we'll use this course all/nothing check. This, assuming role permissions
    // are set correctly, allows the users which can view all projects & their members
    // to use this feature.
```

src/components/language/language.service.ts
```diff
@@ -49,18 +49,15 @@ export class LanguageService {
...
 
   async create(input: CreateLanguage, session: Session): Promise<Language> {
-    await this.authorizationService.checkPower(Powers.CreateLanguage, session);
+    this.privileges.for(session, Language).verifyCan('create');
 
-    await this.authorizationService.checkPower(
-      Powers.CreateEthnologueLanguage,
-      session,
-    );
+    this.privileges.for(session, EthnologueLanguage).verifyCan('create');
```
CarsonF, line 60: Let's drop this. EthnologueLanguage's are always created as sub-objects of Language. They are never directly user created. The first check is sufficient.
```suggestion
```

src/components/language/language.service.ts
```diff
@@ -136,17 +125,8 @@ export class LanguageService {
 
     return {
       ...dto,
```
CarsonF, line 127:
```suggestion
```

src/components/location/location.service.ts
```diff
@@ -87,28 +86,17 @@ export class LocationService {
...
-      Location,
-      dto,
-      session,
-    );
-
     return {
       ...dto,
-      ...securedProps,
-      canDelete: await this.repo.checkDeletePermission(dto.id, session),
+      ...this.privileges.for(session, Location).secure(dto),
```
CarsonF, line 91: No spread

src/components/location/location.service.ts
```diff
@@ -209,11 +197,9 @@ export class LocationService {
...
     session: Session,
   ): Promise<SecuredLocationList> {
-    const perms = await this.authorizationService.getPermissions({
-      resource: label,
-      sessionOrUserId: session,
-      dto,
-    });
+    const perms = this.privileges
+      .for(session, Location, dto)
+      .forEdge(label.name.toLowerCase() as ChildListsKey<any>);
```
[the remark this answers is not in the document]
CarsonF: Still recommend this

src/components/organization/organization.service.ts
```diff
@@ -93,16 +88,9 @@ export class OrganizationService {
...
-      Organization,
-      dto,
-      session,
-    );
-
     return {
       ...dto,
-      ...securedProps,
-      canDelete: await this.repo.checkDeletePermission(dto.id, session),
+      ...this.privileges.for(session, Organization).secure(dto),
```
CarsonF, line 93: No spread

src/components/partner/partner.service.ts
```diff
@@ -99,24 +95,9 @@ export class PartnerService {
...
-      types: {
-        ...securedProps.types,
-        value: securedProps.types.value || [],
-      },
-      financialReportingTypes: {
-        ...securedProps.financialReportingTypes,
-        value: securedProps.financialReportingTypes.value || [],
-      },
-      canDelete: await this.repo.checkDeletePermission(dto.id, session),
+      ...this.privileges.for(session, Partner).secure(dto),
```
CarsonF, line 100: No spread

src/components/partnership-producing-medium/partnership-producing-medium.service.ts
```diff
@@ -30,13 +34,11 @@ export class PartnershipProducingMediumService {
...
-      resource: IProject,
-      sessionOrUserId: session,
-      sensitivity: engagement.sensitivity,
-      otherRoles: engagement.scope,
-    });
-    if (!perms.partnership.canRead) {
+    const perms = withEffectiveSensitivity(
+      withScope(this.privileges.for(session, IProject), engagement.scope),
+      engagement.sensitivity,
+    );
```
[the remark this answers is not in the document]
CarsonF: Still need this

src/components/partnership/partnership.service.ts
```diff
@@ -171,28 +168,9 @@ export class PartnershipService {
...
-      mouEnd: {
-        ...securedProps.mouEnd,
-        canEdit: false, // edit the project mou or edit the partnership mou override
-      },
-      types: {
-        ...securedProps.types,
-        value: securedProps.types.value ?? [],
-      },
-      canDelete: await this.repo.checkDeletePermission(dto.id, session),
+      ...this.privileges.for(session, Partnership).secure(dto),
```
CarsonF, line 173: No spread

src/components/product/product.service.ts
```diff
@@ -337,18 +331,14 @@ export class ProductService {
...
-      session,
-    );
+    const securedProps = this.privileges
+      .for(session, DirectScriptureProduct)
+      .secure(asProductType(DirectScriptureProduct)(dto));
+
     const direct: DirectScriptureProduct = {
       ...dto,
-      totalVerses: dto.totalVerses ?? 0,
-      totalVerseEquivalents: dto.totalVerseEquivalents ?? 0,
```
CarsonF, line 351: I think we still need this.

src/components/project/project.service.ts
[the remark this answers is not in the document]
CarsonF: > What's not covered in the DB query is whether the user can list project members. So if the can see the project but not it's members and then they filter on a certain user and a project, the correlation is leaked.
Let me see if I can adjust the DB query.

I looked into this. Currently all of our roles have the same perms for projects & project members. So the check currently in place is sufficient. Though it would be worth revisiting after DB migration.


---

As I've been scrolling through I've seen this pattern in several spots. So what I think we should do is leave this here as you have it. And then do another pass later, to ensure our all of our lists work with these "subsets" / conditional items.
That would be ensuring the DB queries reference privileges and removing the check from the service files.

src/components/user/unavailability/unavailability.service.ts
```diff
@@ -77,16 +72,9 @@ export class UnavailabilityService {
...
-      Unavailability,
-      dto,
-      session,
-    );
-
     return {
       ...dto,
-      ...securedProps,
-      canDelete: await this.repo.checkDeletePermission(dto.id, session), // TODO
+      ...this.privileges.for(session, Unavailability).secure(dto),
```
CarsonF, line 77: No spread

src/components/user/user.service.ts
```diff
@@ -112,14 +105,9 @@ export class UserService {
...
   async secure(user: UnsecuredDto<User>, session: Session): Promise<User> {
-    const securedProps = this.privileges
-      .for(session, User, user)
-      .secureProps(user);
-
     return {
       ...user,
-      ...securedProps,
-      canDelete: await this.userRepo.checkDeletePermission(user.id, session),
+      ...this.privileges.for(session, User).secure(user),
```
CarsonF, line 110: No spread

CarsonF (changes requested):

src/components/project/project.service.ts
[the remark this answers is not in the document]
brentkulwicki: Added a ticket for this: https://seed-company-squad.monday.com/boards/3451697530/pulses/5008840819

src/components/partnership-producing-medium/partnership-producing-medium.service.ts
```diff
@@ -30,13 +35,13 @@ export class PartnershipProducingMediumService {
...
-      resource: IProject,
-      sessionOrUserId: session,
-      sensitivity: engagement.sensitivity,
-      otherRoles: engagement.scope,
-    });
-    if (!perms.partnership.canRead) {
+    const adjustedContext = withEffectiveSensitivity(
+      withScope(engagement, ['member:true']),
+      Sensitivity.High,
+    ) as any;
```
CarsonF, line 41: Sorry this was just an example I was giving in the previous comment.
We don't want to hardcode these context values.
```suggestion
    const adjustedContext = engagement as any;
```

src/components/partnership-producing-medium/partnership-producing-medium.service.ts
```diff
@@ -69,14 +74,14 @@ export class PartnershipProducingMediumService {
       engagementId,
       session,
     );
+    const engagementContext = withEffectiveSensitivity(
+      withScope(engagement, engagement.scope),
+      engagement.sensitivity,
+    ) as any;
```
CarsonF, line 80: These function calls replace the default values with the same values.
```suggestion
    const engagementContext = engagement as any;
```

CarsonF (approved):
Donald Trump - W3GsOiBLzYyYfJPjMQ

PR #2852, opened 2023-08-01
Tweak local neo4j config to ignore auth

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4913662664) by [Unito](https://www.unito.io)


atGit2021 (approved):
Adam and I just pulled and tested with the additional password length variable in the docker-compose override file and all works well.

PR #2853, opened 2023-08-02
Consultant (Manager) Policy - Add Progress Report Reads

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4918890424) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/authorization/policies/by-role/consultant.policy.ts
+++ b/src/components/authorization/policies/by-role/consultant.policy.ts
@@ -2,16 +2,19 @@ import { member, Policy, Role } from '../util';
 
 // NOTE: There could be other permissions for this role from other policies
 @Policy([Role.Consultant, Role.ConsultantManager], (r) => [
-  r.Ceremony.read,
-  r.Engagement.read,
   [
+    r.Ceremony,
+    r.Engagement,
     r.EthnologueLanguage,
     r.Language,
     r.Organization,
     r.Partner,
     r.Partnership,
+    r.Product,
     r.Project,
     r.ProjectMember,
+    r.PeriodicReport,
+    r.StepProgress,
   ].map((it) => it.when(member).read),
 
   r.InternshipEngagement.when(member).edit.specifically((p) => [
@@ -25,12 +28,9 @@ import { member, Policy, Role } from '../util';
   r.NarrativeReport.when(member).edit.create,
   r.Organization.specifically((p) => p.address.none),
   r.Partner.specifically((p) => p.pointOfContact.none),
-  r.PeriodicReport.read,
-  r.Product.read,
   r.Project.when(member).specifically((p) => [
     p.many('step', 'stepChangedAt', 'rootDirectory').edit,
   ]),
-  r.StepProgress.read,
   r.Unavailability.read,
   r.User.read.create,
 ])
--- a/src/components/authorization/policies/by-role/consultant.policy.ts
+++ b/src/components/authorization/policies/by-role/consultant.policy.ts
@@ -14,6 +14,10 @@ import { member, Policy, Role } from '../util';
     r.Project,
     r.ProjectMember,
     r.PeriodicReport,
+    r.ProgressReportCommunityStory,
+    r.ProgressReportHighlight,
+    r.ProgressReportTeamNews,
+    r.ProgressReportVarianceExplanation,
     r.StepProgress,
   ].map((it) => it.when(member).read),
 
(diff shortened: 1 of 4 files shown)


brentkulwicki (approved):
Pulled and tested locally. ✅

src/components/authorization/policies/by-role/consultant.policy.ts
```diff
@@ -1,32 +1,40 @@
...
-  r.Partner.when(member).read.specifically((p) => p.pointOfContact.none),
-  r.Partnership.when(member).read,
-  r.PeriodicReport.read,
-  r.Product.read,
-  r.Project.when(member).read.specifically(
-    (p) => p.many('step', 'stepChangedAt', 'rootDirectory').edit,
-  ),
-  r.ProjectMember.when(member).read,
-  r.StepProgress.read,
+  r.Organization.specifically((p) => p.address.none),
```
bryanjnelson, line 33: So we're loosening the permissions on org and partner?

bryanjnelson (approved):

src/components/authorization/policies/by-role/consultant.policy.ts
bryanjnelson: Yeah, before Consultants were only able to see as orgs/partners as members, but now they can see them globally?

CarsonF: There's no `read` action declared here.

bryanjnelson: I must not understand then.  To me it looks like they could previously read an org (except the address) when they were a member of a project.  And now we're just calling out that they specifically can't see a the address on a org.  So I guess it isn't loosening, but tightening.

However, why would the `none` need called out if we don't give them any permissions on an org anyway?  I remember having this discussion before, but I can't remember the answer.

CarsonF: Yeah it's because properties only inherit/default-to the object-level permissions when they are not called out specifically. So since this `address` is called out, it doesn't matter what the object-level permissions are here.

Within the policy both the object-level and property-level actions are union'ed together.

All that to say there's no functional change here.
```ts
r.Organization.when(member).read.specifically((p) => p.address.none),
```
is the same as
```ts
r.Organization.when(member).read,
r.Organization.specifically((p) => p.address.none),
```
(though it's worth pointing out the first snippet would use the member condition, and the second doesn't have one - but it doesn't matter for this case because we are saying `.none` instead of an actual action)

That two line snippet is what we have here:
https://github.com/SeedCompany/cord-api-v3/pull/2853/files#diff-d7364e8ffa2f74e8997bbaaae929f24a123bf3f60f7fe947c9f27d6d4872436eR10
https://github.com/SeedCompany/cord-api-v3/pull/2853/files#diff-d7364e8ffa2f74e8997bbaaae929f24a123bf3f60f7fe947c9f27d6d4872436eR22
https://github.com/SeedCompany/cord-api-v3/pull/2853/files#diff-d7364e8ffa2f74e8997bbaaae929f24a123bf3f60f7fe947c9f27d6d4872436eR33

Still open to your thoughts here though. Maybe you don't like this.

bryanjnelson: Two things...first, thanks for the reminder on the `.none`.  Second, I missed that you had the Organization as a member up in the top list.  I wasn't even thinking to look there, tbh, because I haven't seen us split permissions for a single object across a file before.  I'm not sure I like that approach.  I feel like they should be grouped together I think.  🤔

CarsonF: Yeah I was back and forth on that too. Really I think what I was trying to accomplish was "this role can basically read all project related info if they are a member", so I was trying to group that intent together. Then below we tweak specific things.

bryanjnelson: I can see that approach...it's just we haven't done it that way yet so it seems like a one off to me.  Honestly, though, if I think about coming back in here looking for what a particular permission is on something, I'd want to have it all grouped together in one spot personally.

I'll defer to you, sire.

bryanjnelson: I can see your point(s) there.  And if we _could_ get to the place of having a UI (just readonly) that would be huge.

PR #2855, opened 2023-08-03
Migrate to our new scripture library

CarsonF:
I took the opportunity to do some refactoring when I was extracting the scripture code to the library.
So a lot of changes here are just name changes.

I didn't include the _random range generation_ nor the _book difficulty_ in the library as they seemed fairly specific/narrow use cases. So those are the heaviest changes here, since they needed to be extracted out.

Latest library code, also handles parsing `Mark 3 & 4` and `1st John`, etc. Which is the only logical change that this PR represents.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4927700061) by [Unito](https://www.unito.io)


bryanjnelson (approved):
Forward!

PR #2857, opened 2023-08-07
Add Organization.type and Organization.reach

bryanjnelson:
Links to [0509](https://seed-company-squad.monday.com/boards/3451697530/pulses/4788454420).


src/components/organization/dto/organization-type.dto.ts
```diff
@@ -0,0 +1,25 @@
...
+  Church = 'Church',
+  Parachurch = 'Parachurch',
+  Mission = 'Mission',
+  TranslationOrganization = 'TranslationOrganization',
+  Alliance = 'Alliance',
+}
+
+registerEnumType(OrganizationType, {
+  name: 'OrganizationType',
+  description: 'The type of organization',
```
CarsonF, line 17: Customize the label like this:
```suggestion
  description: 'The type of organization',
  valuesMap: {
    Parachurch: {
      description: `@label Para-church`,
    },
  },
```

CarsonF (changes requested):
From the designs it looks like both of these are lists.
Screenshot 2023-08-09 at 8 34 28 AM

src/components/organization/dto/organization-type.dto.ts
bryanjnelson: So this would be used only if we wanted the label to look different than the value?

CarsonF: Yeah if it doesn't follow the standard convention you can customize like this. This is just put in the GQL schema. The UI repo looks for this and uses it as an override instead of doing the standard transformation.
https://github.com/SeedCompany/cord-field/blob/develop/src/api/schema/enumLists/enumLists.codegen.ts#L44-L44
https://github.com/SeedCompany/cord-field/blob/develop/src/api/schema/enumLists/enumLists.codegen.ts#L57-L59

bryanjnelson: Cool.  In this case we don't need it, because the actual spelling of the word here is "Parachurch" with no hyphen; it's just wrong on the mock-up.

CarsonF: Ah cool.

bryanjnelson:
@CarsonF - The business has confirmed that these are indeed both lists; this review can continue as you're able!

CarsonF (approved):
I rebased and:
- dropped the custom db labels, we don't need those.
- switched to the new enum API.

bryanjnelson:
> I rebased and:
> 
> * dropped the custom db labels, we don't need those.
> * switched to the new enum API.

Roger...so is this good to go then?

PR #2858, opened 2023-08-07
Allow conditions to declare how they should be unioned/intersected with other conditions of their type

CarsonF:
This is a step to simplify conditions as they get combined.

Previously there were situations where we'd have something like:
```sql
Member OR Sens Medium/Low OR Sens Low
```
Now the two sibling sensitivities will get simplified:
```sql
Member OR Sens Medium/Low
```

# Future Work

This doesn't handle more complex situations like:
```sql
Member OR (Member AND Variant(s) { draft, translated, fpm })
```
Obviously the variants don't matter here, and this could be simplified to
```
Member
```

Another is:
```sql
(Member AND Variant(s) { official }) OR (Member AND Variant(s) { partner })
```
Which instead could be written as:
```sql
Member AND Variant(s) { official, partner }
```

Both of these cases are much harder to simplify programmatically.


src/components/authorization/policy/conditions/aggregate.condition.ts
```diff
@@ -71,27 +74,82 @@ export class AndConditions<
...
+    }
+
+    const merged = [...groupBy(conditions, byType).values()].flatMap(
+      (sames) => {
+        const same = sames[0]!;
+        return same.intersect ? same.intersect(sames) : sames;
+      },
+    );
+
+    if (merged.length === 1) {
```
brentkulwicki, line 92: Seems to me like this should only ever be 0 or 1, so you should be able to simplify to just `if (merged.length)` right?

brentkulwicki (approved):
1 comment, but this looks good as is too 

thanks for detailed pr description + thorough code comments

src/components/authorization/policy/conditions/aggregate.condition.ts
CarsonF: It shouldn't ever be zero. We do handle no conditions better in the policy builder.
The one length is just to unwrap the aggregate
So
```
any(member, member)
```
Just becomes
```
member
```
Instead of
```
any(member)
```

PR #2860, opened 2023-08-11
Media User & Detected Metadata

CarsonF:
Currently this only serves to enhance the existing file functionality.

With this scope I think the functionality is complete, save for:
- [ ] Allow user metadata to be updated after upload
- [ ] Migration to detect metadata for existing files


src/core/database/highlight-cypher.util.ts
```diff
@@ -9,7 +9,15 @@ export function highlight(query: string) {
...
   });
-  pretty = pretty.split('\n').slice(0, -1).join('\n');
+  pretty = pretty
+    .split('\n')
+    .slice(0, -1)
+    // strip off trailing spaces meant for a "code block background."
+    // we don't set a background color, and this makes copy/pasting into
+    // neo4j editor painful.
+    // eslint-disable-next-line no-control-regex
+    .map((line) => line.replace(/( )+\u001B\[39m$/, ''))
```
brentkulwicki, line 19: What is this pattern matching for here that a .trim() wouldn't work for?

brentkulwicki (approved):
One question but looks good overall.

src/core/database/highlight-cypher.util.ts
CarsonF: Trim doesn't work because the spaces are followed by that terminal escape sequence. I tried lol

src/components/file/media/media-detector.service.ts
```diff
@@ -0,0 +1,83 @@
...
+        '-show_streams',
+        '-i',
+        'pipe:0',
+      ],
+      {
+        reject: false,
+        input: stream,
+      },
+    );
+    return JSON.parse(stdout);
```
adam-soltech, line 81: Since reject is set to false should there be some error handling here in case stdout ends up not being valid json?

src/components/file/media/media-detector.service.ts
```diff
@@ -0,0 +1,83 @@
...
+      return {
+        __typename: 'Image',
+        dimensions: {
+          width: size.width ?? 0,
+          height: size.height ?? 0,
+        },
+      };
+    }
+
+    const isAudio = file.mimeType.startsWith('audio/');
```
adam-soltech, line 36: Will this sort of audio, video, image strategy be repeated anywhere else? IF so it might be worth assigning 'audio/', 'video/' etc to an enum rather than hardcoding them

src/components/file/media/media.repository.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    const result = await query.first();
+    if (!result) {
+      throw new ServerException('Failed to save media info');
+    }
+    return result.dto;
+  }
+}
+
+const toDbShape = (input: Partial<AnyMedia>) => ({
```
adam-soltech, line 131: I think I groked this but it was a bit hard to follow....

Probably a bit of a nit but something like this might be easier to read:
```

const getDimensions = (typename: string | undefined, dimensions: { width?: number | null, height?: number | null }) => {
  return (typename === 'Image' || typename === 'Video') 
    ? dimensions 
    : { width: null, height: null };
};

const getDuration = (typename: string | undefined, duration: number | undefined) => {
  return { duration: typename === 'Audio' ? duration : null };
};

const toDbShape = (input: Partial<AnyMedia>) => {
  return {
    type: input.__typename,
    altText: input.altText,
    caption: input.caption,
    mimeType: input.mimeType,
    ...getDimensions(input.__typename, input.dimensions),
    ...getDuration(input.__typename, input.duration),
  };
};
```

adam-soltech (commented):
Looks great overall. I had a few questions and suggestions but nothing blocking

src/components/file/media/media-detector.service.ts
CarsonF: Yeah that's a good idea.

src/components/file/media/media-detector.service.ts
CarsonF: I doubt it. This is the only spot we have logic based on mime types. But definitely something to keep in mind for future.
Depending on the future need, the media type could be check instead, which is a string literal union.
```ts
media.__typename === 'Audio'
```

src/components/file/media/media.repository.ts
CarsonF: Part of why this is syntax heavy is the nested check for null `__typename`. I wanted this to be used for multiple purposes:
1. Saving detected metadata, and if for whatever reason type changes we want to clear out previous/stale values
2. Saving user metadata. Which should just need the `altText`, `caption` as input. So we'd want to ignore typename logic there and not clear out the existing values.

That additional check could certainly be added to your helper functions above, or some combination. But IMO they don't add enough value to warrant the additional typing needed for the function inputs.

CarsonF:
Thanks for the reviews guys. The biggest concern with new features is the schema, so that's where I'd like the most consideration (just FYI). Internal bits can always be tweaked.

I'm not sure I'm thrilled about the naming here.
https://github.com/SeedCompany/cord-api-v3/blob/5eebbe18f6cefd72bb6594c8d1fde8f60f012079/src/components/file/dto/upload.dto.ts#L40-L40
Both the property name and the type name.

Same for property name here
```gql
interface FileNode {
  media: Media
```
Sometimes I'm too terse for my own good. I think I'm ok with this case. I don't think changing to `mediaInfo` for example would add more clarity.

Open to suggestions on all the additional naming added

PR #2864, opened 2023-08-15
Separate DB Type from Unsecured Type / Refactor Product Scripture Parsing

CarsonF:
Now we have
```
DB Type -> Unsecured Type -> Type
```
This let's us handle DB results separately from the securing/authorization logic.

As we now do with the product's scripture references.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/4995334605) by [Unito](https://www.unito.io)


brentkulwicki (approved):
Thorough and wonderful. Thank you, sir!!

PR #2869, opened 2023-08-23
Fix Apollo Server v4 error handling

CarsonF:
- [Fix ExceptionNormalizer dependency on ~/core](https://github.com/SeedCompany/cord-api-v3/commit/eb49b602744ef931253b8655dc9e5c05e456d84f) 
  Part of this is allowing `ConfigService` to be accessed by a string instead of the class ref
- [Unify ApolloServer error handling with our global exception filter](https://github.com/SeedCompany/cord-api-v3/commit/926e813e1241e96800b73f796a731416558e4a05) 
	Previously, some paths were not running through our exception filter.
	This may have been from the Apollo Server v4 upgrade.
	I noticed this with error through from a field resolver.
	
	All error -> output is handled in a unified spot.
	Now we check codes instead of regex matching messages, win.
- [Remove error logging from GraphqlLoggingPlugin](https://github.com/SeedCompany/cord-api-v3/commit/d8eff902c567f1f216ca31959d38d8667f85b68c) 
	This code path stopped working with Apollo Server v4.
	Their errors now have `extensions`, so it was hitting the early return.
	
	These errors are now handled in our normal code path with previous commit,
	so there's no need to restore this logic.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5034007362) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/config/config.module.ts
+++ b/src/core/config/config.module.ts
@@ -11,9 +11,13 @@ import { VersionService } from './version.service';
       useFactory: (env: EnvironmentService) =>
         Object.assign(new ConfigService(), new (makeConfig(env))()),
     },
+    {
+      provide: 'CONFIG',
+      useExisting: ConfigService,
+    },
     EnvironmentService,
     VersionService,
   ],
-  exports: [ConfigService, VersionService],
+  exports: [ConfigService, 'CONFIG', VersionService],
 })
 export class ConfigModule {}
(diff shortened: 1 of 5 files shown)


bryanjnelson:
Check codes over RegEx === Big Win!   :shipit:

bryanjnelson (approved):

PR #2870, opened 2023-08-23
New `makeEnum` abstraction for better enums

CarsonF:
- Uses string literals instead of TS enums, avoiding the pitfalls of those.
   But individual values can still be referenced via `Enum.X` syntax.
- Integrated with GraphQL schema
- Extra enum value info is declared with the value, instead of in a separate `valueMap`
- Has helper methods built in
- Other custom helpers can easily be added for specific use cases
- Since this is runtime code, enum values can be programmatically made

I've changed several of our enums over to this new abstraction as proof of work. I tackled the most complicated ones.
Hopefully the diff of this branch gives a good idea/example of the new functionality.

This should be a good first step to having enums sourced from EdgeDB schema as well.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5041142409) by [Unito](https://www.unito.io)


bryanjnelson (approved):
Gold!!  🏅

brentkulwicki (approved):
I think we are moving away from a very standard way of doing things here, however, I definitely see the purpose in this and think you chose the correct way forward here.

PR #2873, opened 2023-08-27
Fix `Postable.posts` resolution error

CarsonF:
- [Give a better error message when ResourceHost encounters null input](https://github.com/SeedCompany/cord-api-v3/commit/11caf17d7daeec8a460c5d12ec72c00b470236d0)

- [Change how Postable asks for __typename](https://github.com/SeedCompany/cord-api-v3/commit/c95fa372005ba6bf9c3fb51889eda2d3b28ba5fd) 
	Stop requiring implementations to declare the `__typename` field.
	This apparently was tricky for TS as it was missed in the _Project_ shapes.
	This string can also be gotten from GraphQL or `ResourceLoader`,
	so it's not always necessary to specify manually.
	
	Instead, we'll change types to only ask for `__typename`
	when the `PostableRef` is an actual `Postable` object.

- [Fix regression](https://github.com/SeedCompany/cord-api-v3/commit/6220c2894664d090425468f361c17b14038824d6) from d6790f0 where `Postable.posts` errored out because typename was missing
  Previous commit helps reveal this mistake.

### DIFF
--- a/src/components/post/postable.resolver.ts
+++ b/src/components/post/postable.resolver.ts
@@ -1,4 +1,5 @@
-import { Parent, ResolveField, Resolver } from '@nestjs/graphql';
+import { Info, Parent, ResolveField, Resolver } from '@nestjs/graphql';
+import { GraphQLResolveInfo } from 'graphql';
 import { ListArg, LoggedInSession, Resource, Session } from '~/common';
 import { Loader, LoaderOf } from '~/core';
 import { Postable } from './dto';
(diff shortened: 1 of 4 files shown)


CarsonF:
FYI @brentkulwicki

src/components/post/postable.resolver.ts
```diff
@@ -14,13 +15,17 @@ export class PostableResolver {
     description: 'List of posts belonging to the parent node.',
   })
   async posts(
+    @Info() info: GraphQLResolveInfo,
```
bryanjnelson, line 18: Didn't know about this!!

PR #2876, opened 2023-09-01
revised logic for progress summary

atGit2021:
Revised the logic so the progress summary will only bypass from being saved to the database if both the planned and actuals are zero.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5092410564) by [Unito](https://www.unito.io)


CarsonF (changes requested):

CarsonF (approved):
LGTM!

PR #2884, opened 2023-09-07
Add `Partner.countries`

willdch:
https://seed-company-squad.monday.com/boards/3451697530/pulses/5037868084

@bryanjnelson here is that draft PR for you to look at.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5124319112) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/location/dto/location.dto.ts
+++ b/src/components/location/dto/location.dto.ts
@@ -10,6 +10,7 @@ import {
   Secured,
   SecuredEnum,
   SecuredProperty,
+  SecuredPropertyList,
   SecuredProps,
   SecuredString,
   SecuredStringNullable,
@@ -53,6 +54,11 @@ export class Location extends Resource {
   description: SecuredProperty.descriptionFor('a location'),
 })
 export class SecuredLocation extends SecuredProperty(Location) {}
+
+@ObjectType({
+  description: SecuredPropertyList.descriptionFor('a list of locations'),
+})
+export class SecuredLocations extends SecuredPropertyList(Location) {}
 
 declare module '~/core/resources/map' {
   interface ResourceMap {
--- a/src/components/partner/dto/create-partner.dto.ts
+++ b/src/components/partner/dto/create-partner.dto.ts
@@ -3,6 +3,7 @@ import { Transform, Type } from 'class-transformer';
 import { Matches, ValidateNested } from 'class-validator';
 import { uniq } from 'lodash';
 import { ID, IdField, IdOf, IsId, NameField } from '../../../common';
+import { Location } from '../../../components/location';
 import { FieldRegion } from '../../field-region';
 import type { Language } from '../../language';
 import { FinancialReportingType } from '../../partnership/dto/financial-reporting-type';
@@ -42,6 +43,11 @@ export abstract class CreatePartner {
 
   @IdField({ nullable: true })
   readonly languageOfWiderCommunicationId?: IdOf<Language> | null;
+
+  @Field(() => [IDType], { nullable: true })
+  @IsId({ each: true })
+  @Transform(({ value }) => uniq(value))
+  readonly countries?: ReadonlyArray<IdOf<Location>> = [];
 
   @Field(() => [IDType], { nullable: true })
   @IsId({ each: true })
--- a/src/components/partner/dto/update-partner.dto.ts
+++ b/src/components/partner/dto/update-partner.dto.ts
@@ -3,6 +3,7 @@ import { Transform, Type } from 'class-transformer';
 import { Matches, ValidateNested } from 'class-validator';
 import { uniq } from 'lodash';
 import { ID, IdField, IdOf, IsId, NameField } from '../../../common';
+import { Location } from '../../../components/location';
 import { FieldRegion } from '../../field-region';
 import type { Language } from '../../language';
 import { FinancialReportingType } from '../../partnership/dto/financial-reporting-type';
@@ -42,6 +43,11 @@ export abstract class UpdatePartner {
 
   @IdField({ nullable: true })
   readonly languageOfWiderCommunicationId?: IdOf<Language> | null;
+
+  @Field(() => [IDType], { nullable: true })
+  @IsId({ each: true })
+  @Transform(({ value }) => (value ? uniq(value) : undefined))
+  readonly countries?: ReadonlyArray<IdOf<Location>>;
 
   @Field(() => [IDType], { nullable: true })
   @IsId({ each: true })
--- a/src/components/partner/partner.repository.ts
+++ b/src/components/partner/partner.repository.ts
@@ -64,6 +64,7 @@ export class PartnerRepository extends DtoRepository<
             input.languageOfWiderCommunicationId,
           ],
           fieldRegions: ['FieldRegion', input.fieldRegions],
+          countries: ['Location', input.countries],
         }),
       )
       .return<{ id: ID }>('node.id as id')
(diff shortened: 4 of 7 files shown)


CarsonF:
parter?

src/components/partner/partner.resolver.ts
```diff
@@ -93,6 +95,28 @@ export class PartnerResolver {
...
+  async countries(
+    @Parent() partner: Partner,
+    @Loader(LocationLoader) loader: LoaderOf<LocationLoader>,
+  ): Promise<SecuredLocations> {
+    const countries = (await loader.loadMany(partner.countries.value)).flatMap(
+      (country) => {
+        if (country instanceof NotFoundException) {
+          return [];
+        } else if (country instanceof Error) {
+          return country;
```
bryanjnelson, line 108: In this case `country` is actually an `Error`, so you want to throw it instead of return it:

```typescript
throw country;
```

This will take care of your build error.

src/components/location/dto/location.dto.ts
```diff
@@ -54,6 +55,11 @@ export class Location extends Resource {
 })
 export class SecuredLocation extends SecuredProperty(Location) {}
 
+@ObjectType({
+  description: SecuredPropertyList.descriptionFor('a location or null'),
```
bryanjnelson, line 59:
```suggestion
  description: SecuredPropertyList.descriptionFor('a list of locations'),
```

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -1,8 +1,9 @@
-import { Field, InputType, ObjectType } from '@nestjs/graphql';
+import { Field, ID, InputType, ObjectType } from '@nestjs/graphql';
```
bryanjnelson, line 1: Carson corrected me here...let's alias this instead of the other `ID`.

```suggestion
import { Field, ID as IDType, InputType, ObjectType } from '@nestjs/graphql';
```

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -1,8 +1,9 @@
...
 import { Transform, Type } from 'class-transformer';
 import { Matches, ValidateNested } from 'class-validator';
 import { uniq } from 'lodash';
-import { ID, IdField, IdOf, NameField } from '../../../common';
+import { ID as CordID, IdField, IdOf, NameField } from '../../../common';
```
bryanjnelson, line 5: Let's keep this as the simple `ID` name.

```suggestion
import { ID, IdField, IdOf, NameField } from '../../../common';
```

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -41,6 +42,10 @@ export abstract class CreatePartner {
 
   @IdField({ nullable: true })
   readonly languageOfWiderCommunicationId?: IdOf<Language> | null;
+
+  @Field(() => [ID], { nullable: true })
+  @Transform(({ value }) => uniq(value))
+  readonly countries?: ReadonlyArray<IdOf<Location>> = [];
```
bryanjnelson, line 48:
```suggestion
  @Field(() => [ID], { nullable: true })
  @IsId({ each: true })
  @Transform(({ value }) => uniq(value))
  readonly countries?: ReadonlyArray<IdOf<Location>> = [];
```

src/components/partner/dto/update-partner.dto.ts
bryanjnelson, line 1: You can apply the same changes here to `UpdatePartner` as shown above for `CreatePartner`.

src/components/partner/partner.module.ts
bryanjnelson, line 1: You can revert these changes...we no longer need them since we'll be 🪓 ing `listCountries` in the service.

src/components/partner/partner.repository.ts
```diff
@@ -103,6 +105,15 @@ export class PartnerRepository extends DtoRepository<
             .raw('WHERE size(projList) = 0')
             .return(`'High' as sensitivity`),
         )
+        .subQuery('node', (sub) =>
+          sub
+            .optionalMatch([
```
bryanjnelson, line 110:
```suggestion
            .match([
```

src/components/partner/partner.repository.ts
```diff
@@ -103,6 +105,15 @@ export class PartnerRepository extends DtoRepository<
...
+              node('node'),
+              relation('out', '', 'countries', ACTIVE),
```
bryanjnelson, line 112:
```suggestion
              relation('out', '', 'countries'),
```

src/components/partner/partner.service.ts
```diff
@@ -33,6 +40,7 @@ import { PartnerRepository } from './partner.repository';
 @Injectable()
 export class PartnerService {
   constructor(
+    private readonly countriesService: LocationService,
```
bryanjnelson, line 43: 🪓

src/components/partner/partner.service.ts
```diff
@@ -192,6 +236,23 @@ export class PartnerService {
...
+      session,
+    );
+
+    return {
+      ...countriesListOutput,
+      canRead: true,
+      canCreate: this.privileges.for(session, IProject).can('create'),
+    };
+  }
+
```
bryanjnelson, line 255: 🪓 - no longer needed

src/components/partner/partner.service.ts
```diff
@@ -136,14 +145,49 @@ export class PartnerService {
...
+          return;
+        }
+        throw new NotFoundException(
+          errMsg,
+          `partner.${resourceField}${Array.isArray(ids) ? `[${index}]` : ''}`,
+        );
+      }),
+    );
+  }
+
```
bryanjnelson, line 190: For this whole section you'll be using an abstraction now that handles most of this; you're code will be much simpler.  You can model it after [the example here](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/partner/partner.service.ts#L146C2-L158C2).  Note that you'll have to rebase on `develop` to get Carson's abstraction code before you can make these changes.

bryanjnelson (changes requested):
Sorry for all the churn here; this was basically all errors that I made that you inherited.  😬 

Thanks for the flexibility @willdch!

bryanjnelson (approved):
Thanks @willdch!  Look'n good!

@CarsonF - on to you for a final pass.

bryanjnelson:
> Looks good!
> 
> I wonder if we should add a restriction to only allow Locations that are type == Country?

Sounds reasonable.  At what point would that be?  A custom input validation in the Resolver?  I would think not, because we'd have to make another call to the db to check there.  Perhaps rather in the service layer, on the create and update?

CarsonF:
Yeah the service layer. Could do something like
```ts
const loader = await resourceLoader.getLoader(Location);
const locations = await loadManyIgnoreMissingThrowAny(loader, ids);
const invalid = locations.flatMap(l => l.type !== 'Country' ? l : []);
if (invalid.length) {
	const ex = new InvalidReferencesException('Only locations that are countries can be used', invalid);
	throw ex.withField('partner.countries');
}
```

bryanjnelson:
Cool cool.  Do we really want `InvalidReferencesException` though?  In it's current form it only takes one param.  Unless you want us to alter or overload it to take a second?  I thought perhaps maybe you were meaning a different Exception altogether?

CarsonF:
On second thought we should have a custom exception class for this specific use case. Since classes correspond to codes, that's how a consumer should programmatically check for this error.
You can copy paste the InvalidReferencesException and just change the class name and message.

bryanjnelson (approved):
LGTM!

@CarsonF ?

CarsonF (approved):

PR #2886, opened 2023-09-07
Add the root node and resource attached to root node to hydrated FileNode

CarsonF:
Part of work for #2883, pulled out as this logic can stand on its own.

- `FileNode`'s now return the `root` `FileNode` from the DB
- `FileNode`'s now return the _resource_ and _relation_ the _root_ `FileNode` is attached to.

This allows other parts of the application to start doing more complex logic with these files.
For example, authorization has mostly been stubbed. This is the first step in knowing where the file "lives" under, in order to facilitate more fine grain checks.

This change requires `FileNode`'s to not be orphaned. This affects how project's root directories were created. So that has been refactored, and now is one less DB call too.

I also added an event for `AfterFileUpload`, which doesn't relate at all other than being in the file service. It will allow other parts of the app to hook into file uploads for verification.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5125562810) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/file/dto/node.ts
+++ b/src/components/file/dto/node.ts
@@ -4,6 +4,7 @@ import { DateTime } from 'luxon';
 import { Readable } from 'stream';
 import { keys as keysOf } from 'ts-transformer-keys';
 import { MergeExclusive, Opaque } from 'type-fest';
+import { BaseNode } from '~/core/database/results';
 import { RegisterResource } from '~/core/resources';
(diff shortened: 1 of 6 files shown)


bryanjnelson (approved):
🦾

brentkulwicki (approved):

PR #2887, opened 2023-09-08
Media 1.1 / Polls / Event Injection

CarsonF:
Part of work for #2883, pulled out as this logic can stand on its own.

The biggest goal of this PR is to allow authorization to the `updateMediaMetadata` mutation to be configurable.
This is a generic mutation, but its objects are nested within other objects whose access is more fine tuned. We've basically ignored this concept so far in the application and assumed good actors...so here's the first step into fixing that.

I tried a few things here, that I'm not completely sure about. Just trying to see how it played out. Happy to receive feedback about it.

# Polls
The first is [polls](https://github.com/SeedCompany/cord-api-v3/blob/bf2dc3d86341fa53a4410cbb630e8181ceee3782/src/common/poll.ts). We needed event handlers to be communicate values back to the event source. 
My starting point was that multiple handlers could weigh in to this event. They should be able to do so without conflicting with, or being aware of, other handlers. Though we have [event handler prioritization](https://github.com/SeedCompany/cord-api-v3/blob/bf2dc3d86341fa53a4410cbb630e8181ceee3782/src/components/project/handlers/step-changed-notification.handler.ts#L13-L13), it's kinda a bad thing to rely on, since it needs to make some loose assumptions about handlers.

It made sense to make this it's own thing separate from events, so multiple events could use it.
It's generic so votes don't have to be only boolean, but can be enum values or anything.

The goal of `Polls` is to provide a standard way for the application to weigh in on X decision.
 
```ts
const whichColor = new Poll<Color>();
whichColor.vote('red');
whichColor.vote('green');
whichColor.vote('blue');
whichColor.vote('red');

whichColor.plurality // 'red'
whichColor.majority // undefined
whichColor.vote('red');
whichColor.majority // 'red' (3/5)
```

Admittedly this is possibly over engineered.
I take solace in the fact that this is well encapsulated, can be dropped in as single liners anywhere, and is flexible in result aggregation.

```ts
const allowUpdate = new Poll();
const event = new MyEvent(
  // Only allow event handlers to vote, not read results.
  allowUpdate.voter
);
// true if plurality voted true and no one vetoed.
const isAllowed = allowUpdate.plurality && !allowUpdate.vetoed;

// in handlers
event.allowUpdate.vote(true);
```

Again though it could simply be
```ts
const event = new MyEvent();
event.allowUpdate++;
const isAllowed = event.allowUpdate > 0;
```
However this does expose the current value, so handlers could base logic off it which is bad, since handlers run non deterministically.
The current value could be made private and an increment/decrement methods exposed. But that is more work for each event, which is what I was trying to avoid.
This also only works with boolean values, voting with anything else would have to be done differently.
This also doesn't capture ties.

Part of me also thinks that even just being in a situation where multiple places are voting on functionality sounds like spaghetti. In several places we have multiple event handlers for an event, but they only act on certain properties of the event. Leading to still a 1-1 situation for individual events, just decoupled.

It's probably too early to know if this is a good direction or not, since we don't have other use cases. 

This PR's concrete use case:
https://github.com/SeedCompany/cord-api-v3/blob/ea7f2b664adfb26179c73c01e447d21553b12ddc/src/components/file/media/media.service.ts#L38-L49
https://github.com/SeedCompany/cord-api-v3/blob/ea7f2b664adfb26179c73c01e447d21553b12ddc/src/components/file/media/events/can-update-event.ts#L4-L14
Vote happens in other PR:
https://github.com/SeedCompany/cord-api-v3/blob/d1a5870ffc691a0503e5493d0b1cc48f0b9af906/src/components/progress-report/media/handlers/update-media-metadata-check.handler.ts#L23-L27

# Partial Injection
This may be unique to events, since they are some of the only places we actually create instances ourselves.
And that's because each time they have different properties, obviously.
But what would it look like if they could also inject services as well? Is this even a good idea?
[Here](https://github.com/SeedCompany/cord-api-v3/commit/56990e898f3198e83ae01e9cf9fc6c09395cb610) I setup the event to be injectable, and added a thin helper to do that creation merging.
[Here](https://github.com/SeedCompany/cord-api-v3/commit/9c65a1b646e8daf907f70942a60ae56897d863b5) I injected service to discover more information about the event properties.
This could be in the service, and then passed into the event. That's how we do it everywhere else. This was an attempt to do it differently to see what we think.
A possible advantage of this is the logic could be skipped if no event handlers need it, though that's probably not likely.

### DIFF
--- a/src/common/create-and-inject.ts
+++ b/src/common/create-and-inject.ts
@@ -1,0 +1,18 @@
+import { ModuleRef } from '@nestjs/core';
+import { pickBy } from 'lodash';
+import { Class } from 'type-fest';
+
+/**
+ * A helper to create an instance of a class and inject dependencies.
+ */
+export async function createAndInject<T extends Class<any>>(
+  moduleRef: ModuleRef,
+  type: T,
+  ...input: ConstructorParameters<T>
+): Promise<InstanceType<T>> {
+  const injection = await moduleRef.resolve(type);
+  const injectionProps = pickBy(injection);
+  const object = new type(...input);
+  Object.assign(object, injectionProps);
+  return object;
+}
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -26,6 +26,7 @@ export * from './pagination.input';
 export * from './pagination-list';
 export * from './parent-id.middleware';
 export * from './parent-types';
+export * from './poll';
 export * from './resource.dto';
 export * from './role.dto';
 export * from './secured-list';
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -3,6 +3,7 @@ export { Many, many, maybeMany, JsonSet, ArrayItem } from '@seedcompany/common';
 export * from './temporal';
 export * from './calculated.decorator';
 export * from './context.type';
+export * from './create-and-inject';
 export * from './data-object';
 export * from './date-filter.input';
 export { DbLabel } from './db-label.decorator';
--- a/src/common/poll.ts
+++ b/src/common/poll.ts
@@ -1,0 +1,124 @@
+import { setOf } from '@seedcompany/common';
+
+export class PollResults<T> {
+  constructor(protected readonly data: PollData<T>) {}
+
+  /** Returns true if there were any votes */
+  get anyVotes() {
+    return this.numberOfVotes > 0;
+  }
+
+  /** Returns true if there were no votes */
+  get noVotes() {
+    return this.numberOfVotes === 0;
+  }
+
+  get numberOfVotes() {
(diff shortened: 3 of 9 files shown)


bryanjnelson (approved):
👍🏼

brentkulwicki (approved):
Test is failing, and I think I agree with your point about this "might be over engineering"

HOWEVER

I don't have any great suggestions for simplifcation AND I think this makes the code easy to read and understand, so I am all for it

CarsonF:
Thanks guys. Finally figured out test failures are fixed in #2886. Will merge that first and rebase this.

PR #2900, opened 2023-09-27
Add Partner.languagesOfConsulting

adam-soltech:
https://seed-company-squad.monday.com/boards/3451697530/pulses/4788464835
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5233960579) by [Unito](https://www.unito.io)


src/components/partner/partner.resolver.ts
```diff
@@ -93,6 +98,29 @@ export class PartnerResolver {
     );
   }
 
+  @ResolveField(() => SecuredLanguages)
+  async languagesOfConsulting(
+    @Parent() partner: Partner,
+    @Loader(LanguageLoader) loader: LoaderOf<LanguageLoader>,
+  ): Promise<SecuredLanguages> {
+    const languages = (
+      await loader.loadMany(partner.languagesOfConsulting.value)
```
adam-soltech, line 107: The value passed to the loadMany here gives:

Screenshot 2023-09-27 at 8 42 13 AM

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -41,6 +41,10 @@ export abstract class CreatePartner {
 
   @IdField({ nullable: true })
   readonly languageOfWiderCommunicationId?: IdOf<Language> | null;
+
+  @Field(() => [Language], { name: 'languagesOfConsulting', nullable: true })
```
CarsonF, line 45: This is a list of IDs, not Language objects.
```suggestion
import { ID as IDType } from '@nestjs/graphql';
  @Field(() => [IDType], { nullable: true })
  @IsId({ each: true })
```

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -3,7 +3,7 @@ import { Transform, Type } from 'class-transformer';
 import { Matches, ValidateNested } from 'class-validator';
 import { uniq } from 'lodash';
 import { ID, IdField, IdOf, NameField } from '../../../common';
-import type { Language } from '../../language';
+import { Language } from '../../language';
```
CarsonF, line 6: This change is what is causing your initialization problem.

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -41,6 +41,10 @@ export abstract class UpdatePartner {
 
   @IdField({ nullable: true })
   readonly languageOfWiderCommunicationId?: IdOf<Language> | null;
+
+  @Field(() => [Language], { name: 'languagesOfConsulting', nullable: true })
```
CarsonF, line 45: Same here

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -41,6 +41,10 @@ export abstract class UpdatePartner {
...
+  @Transform(({ value }) => uniq(value))
```
CarsonF, line 46: Otherwise the an `undefined` will get converted to `[]` which means clear all instead of ignore with no changes.
```suggestion
  @Transform(({ value }) => value ? uniq(value) : undefined)
```

CarsonF (changes requested):

src/components/partner/partner.service.ts
```diff
@@ -141,6 +142,15 @@ export class PartnerService {
       );
     }
 
+    if (languagesOfConsulting) {
+      await this.repo.updateRelationList({
+        id: object.id,
+        relation: 'languagesOfConsulting',
+        label: 'Language',
```
bryanjnelson, line 149: I'll defer to @CarsonF, but I'm not sure this `label` is needed?

bryanjnelson (approved):
I can not speak to the changes in `PartnerRepository` so I will be let @CarsonF weigh in there, but the rest looks good to me.

Thanks @adam-soltech!

src/components/partner/partner.repository.ts
```diff
@@ -119,12 +120,34 @@ export class PartnerRepository extends DtoRepository<
...
+        .with([
+          'node',
+          'languagesOfConsultingIds',
+          'props',
+          'sensitivity',
+          'organization',
+          'pointOfContact',
+          'languageOfWiderCommunication',
+          'scopedRoles',
+        ])
```
CarsonF, line 143: We do the subquery above to avoid having to do this.
```suggestion
```

src/components/partner/partner.service.ts
CarsonF: No and it's actually the other label. So it'll be `Partner` here by default, which is what we want. I didn't allow giving the label of the other side. I just assumed `BaseNode` would be fine. Technically the caller could give a valid id for a different resource, that would be accepted and saved here. So that's probably not the best. But that would have to be from API directly, UI won't allow. I wasn't too worried about it because EdgeDB will validate this better anyways.

CarsonF (changes requested):

src/components/partner/partner.service.ts
```diff
@@ -141,6 +142,15 @@ export class PartnerService {
...
+        newList: languagesOfConsulting,
+      });
+    }
```
bryanjnelson, line 152: Now that Carson is handling the validation within `updateRelationList`, you can handle the exceptions as shown [in this example](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/partner/partner.service.ts#L153C2-L158C2).

src/components/partner/partner.repository.ts
adam-soltech: You're correct that this is not needed for the API response. However, without these definitions the e2e tests breaks with 'undefined var' errors. I'm definitely not a cypher expert but after a bit of research this 'with' solved those. Open to other ideas...

CarsonF: What was the error exactly?

src/components/partner/partner.resolver.ts
```diff
@@ -112,6 +117,37 @@ export class PartnerResolver {
...
+  ): Promise<SecuredLanguages> {
+    const languagesOfConsultingForLoader =
+      partner.languagesOfConsulting.value.map(
+        (id) =>
+          ({
+            id,
+            view: { active: true },
+          } as const),
+      );
+    const languages = (
```
CarsonF, line 133: I added a helper to make this easier: `loadSecuredIds`. Check it out.

src/components/partner/partner.repository.ts
```diff
@@ -140,6 +141,27 @@ export class PartnerRepository extends DtoRepository<
...
+            .match([
+              node('node'),
+              relation('out', '', 'languagesOfConsulting', ACTIVE),
+              node('languagesOfConsulting', 'Language'),
+            ])
+            .return(
+              'collect(languagesOfConsulting.id) as languagesOfConsultingIds',
+            ),
+        )
+        .with([
```
CarsonF, line 155: This is the reason the test suite is failing. Since you started this, the query above was modified to add an additional variable that is not included in this list.
I still stand behind the fact that this clause should not be needed.
If you are having trouble without, perhaps moving above the `matchProps` clause.

CarsonF (commented):
Other than these two things, this looks good!

src/components/partner/partner.resolver.ts
adam-soltech: The helper does not work in this case because language loader requires a view property. Do you know what that property is for? Should 'active' always be true? If so I could probably just make another helper

CarsonF: It's for the changesets feature, which is still WIP. For this use case it'll always be the active version.
It wraps another helper: `loadManyIgnoreMissingThrowAny`. You could use that instead.
Like this:
```ts
const { value: ids, ...rest } = partner.languagesOfConsulting;
const value = await loadManyIgnoreMissingThrowAny(
  loader,
  ids.map((id) => ({ id, view: { active: true } } as const)),
);
return { ...rest, value };
```

.gitignore
```diff
@@ -2,6 +2,7 @@
 /dist
 *.gql
 *.graphql
+/data
```
CarsonF, line 5: This is a local setting for you. This path is only mentioned in one example comment.

I'd suggest adding a convention to your global gitignore like `*.local` and then this could be `.db.local`.
Or you could move it outside the project dir.

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -53,6 +53,10 @@ export abstract class UpdatePartner {
   @IsId({ each: true })
   @Transform(({ value }) => (value ? uniq(value) : undefined))
   readonly fieldRegions?: ReadonlyArray<IdOf<FieldRegion>>;
+
+  @Field(() => [IDType], { name: 'languagesOfConsulting', nullable: true })
+  @Transform(({ value }) => (value ? uniq(value) : undefined))
+  readonly languagesOfConsulting?: ReadonlyArray<IdOf<Language>> = [];
```
CarsonF, line 59: We want to assume a missing prop here means unchanged, not clear existing value.
```suggestion
  readonly languagesOfConsulting?: ReadonlyArray<IdOf<Language>>;
```

CarsonF (changes requested):
Just a couple small tweaks!

.gitignore
adam-soltech: I saw these settings in our gitignore:

.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json

which are also local.... nevertheless I have done this

src/components/partner/dto/update-partner.dto.ts
adam-soltech: Good catch... I think I copied this from the list dto and didn't think of the implication here

.gitignore
CarsonF: Yeah i go back and forth on whether these should be there. On one hand, yes they are a local thing and probably apply to every repo so they should be in your global gitignore. On the other hand, it's really standard to have these in the project's gitignore. Most repos have them, so it doesn't seem to be hurting anything.
In this case, `/data` could be anything, so that's what makes me want to push it exclusively to local. Also, this will be going away with edgedb migration, so that further's the thought of punting it instead of a better DX.

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.gql
 *.graphql
 
+
```
CarsonF, line 6:
```suggestion
```

CarsonF (approved):
Thanks @adam-soltech!

.gitignore
adam-soltech: Yeah if it's going away we don't need it here. Good call then

PR #2905, opened 2023-10-03
Move invalid ID validation into updateRelationList method

CarsonF:
We'll always want to do this, so do it once here.

Using a custom exception class as well, so the consumer can get a list of invalid IDs back programmatically.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5270307212) by [Unito](https://www.unito.io)


bryanjnelson (approved):
❤️ !

PR #2914, opened 2023-10-12
Update Variance Reason Options (and add deprecations)

wes-cutting:
To be considered with [Cord-Field PR](https://github.com/SeedCompany/cord-field/pull/1486)
[Ticket explaining changes](https://seed-company-squad.monday.com/boards/3451697530/pulses/5123136460)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5320368981) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/progress-report/variance-explanation/migrations/rename.migration.ts
+++ b/src/components/progress-report/variance-explanation/migrations/rename.migration.ts
@@ -1,0 +1,26 @@
+import { node, relation } from 'cypher-query-builder';
+import { BaseMigration, Migration } from '~/core';
+import { ACTIVE } from '~/core/database/query';
+
+@Migration('2023-10-12T11:00:00')
+export class RenameReasonOptionMigration extends BaseMigration {
+  async up() {
+    await this.db
+      .query()
+      .match([
+        node('', 'ProgressReportVarianceExplanation'),
+        relation('out', '', 'reasons', ACTIVE),
+        node('reason', 'Property'),
+      ])
+      .raw('where any(x in reason.value where x = $old)', {
+        old: 'Partner organization issues currently being addressed.',
+      })
+      .setValues({
+        'reason.value': [
+          'Partner organization issues currently being addressed',
+        ],
+      })
+      .return('count(reason) as count')
+      .executeAndLogStats();
+  }
+}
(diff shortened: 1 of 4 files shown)


CarsonF (changes requested):
Noting that we talked about adding a `deprecated` field

src/components/progress-report/variance-explanation/variance-explanation.dto.ts
```diff
@@ -26,6 +26,9 @@ export abstract class ProgressReportVarianceExplanation {
 
   readonly report: ProgressReport & SetUnsecuredType<ID>;
 
+  @Field()
+  readonly deprecated: SecuredStringList;
```
CarsonF, line 30: This isn't needed. The other one is sufficient

CarsonF: [Here](https://github.com/SeedCompany/cord-api-v3/pull/2914/files#diff-3c01125b0bcbed569795854d33632f57f8f35e2dac24bd6090c2ee8015376b94R19). This DTO shape is for the actual saved value for each report, we don't need to track the options here or if they are deprecated.

wes-cutting: Oh I see

src/components/progress-report/variance-explanation/variance-explanation.repository.ts
```diff
@@ -39,6 +39,7 @@ export class ProgressReportVarianceExplanationRepository extends DtoRepository(
   protected hydrate() {
     const defaults: UnsecuredDto<VarianceExplanation> & ExpressionInput = {
       report: 'report.id' as ID,
+      deprecated: [],
```
CarsonF, line 42: Again not needed. This deprecated list is only for the options, not for actual saved values for reports.

CarsonF (approved):
Thanks @wes-cutting!

I fixed a few things:
- I forgot to import the migration into the nestjs module.
- I moved the note about the deprecated list into the description field, so it comes across in the GQL schema.
- I also added a check to ensure that deprecated reasons cannot be selected (if they are not currently selected).

PR #2917, opened 2023-10-18
Migrate to new common utils / Less lodash

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5352246144) by [Unito](https://www.unito.io)


bryanjnelson (approved):
In most cases these changes shorten the footprint of the code, and I think even in the minority cases of where they don't it is just as readable, if not more readable.  I'm in.

🚀   Thanks @CarsonF!

PR #2918, opened 2023-10-18
Fix NaN from missing difficulty of book

CarsonF:
@atGit2021 did all the hard work of narrowing the `NaN` down to this function returning `undefined`.
```ts
undefined * x == NaN
```

I fixed the Psalms name from [the name change](https://github.com/SeedCompany/libs/commit/3bf0f68fdb98b3f3d47aa4bb7422f6fbdcb4d362), added tests to ensure our book map is exhaustive, and am returning `0` as factor when book cannot be found as a last chance safety check.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5353627293) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/scripture/book-difficulty-factor.test.ts
+++ b/src/components/scripture/book-difficulty-factor.test.ts
@@ -1,0 +1,6 @@
+import { Book } from '@seedcompany/scripture';
+import { difficultyOfBook } from './book-difficulty-factor';
+
+test.each([...Book])(`difficultyOfBook is defined: %s`, (book) => {
+  expect(difficultyOfBook(book)).toBeDefined();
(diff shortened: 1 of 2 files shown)


atGit2021:
@CarsonF why didn't it pick up the book name since both names, Psalm and Psalms, are listed, just in reverse order?

CarsonF:
> @CarsonF why didn't it pick up the book name since both names, Psalm and Psalms, are listed, just in reverse order?

Those other alternate names are only used when fuzzy matching with `Book.named()`. This is using `book.name` which is the always the "canonical" name.

PR #2920, opened 2023-10-25
Recalc total verse equivalents saved as NaN

atGit2021:
@CarsonF could you help me with the cypher query?  I'm getting back an empty array, so not sure how to pull ScriptureRange for both the scriptureReferences (Direct Scripture Products) and the scriptureReferencesOverride (Derivative Scripture Products) as a collection.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5394965702) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/product/migrations/fix-nan-total-verse-equivalents.migration.ts
+++ b/src/components/product/migrations/fix-nan-total-verse-equivalents.migration.ts
@@ -1,0 +1,52 @@
+import { node, relation } from 'cypher-query-builder';
+import { ID } from '~/common';
+import { BaseMigration, Migration } from '~/core/database';
+import { ACTIVE } from '~/core/database/query';
+import { getTotalVerseEquivalents } from '../../scripture/verse-equivalents';
+import { ProductService } from '../product.service';
+
+@Migration('2023-10-27T09:47:07')
+export class FixNaNTotalVerseEquivalentsMigration extends BaseMigration {
+  constructor(private readonly productService: ProductService) {
+    super();
+  }
+
+  async up() {
+    const session = this.fakeAdminSession;
+    const ids = await this.db
+      .query()
+      .match([
+        node('product', 'Product'),
+        relation('out', '', 'totalVerseEquivalents', ACTIVE),
+        node('tve', 'Property'),
+      ])
+      .raw('WHERE tve.hadNaN = true or isNaN(tve.value)')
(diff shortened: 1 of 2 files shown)


CarsonF (commented):
I wouldn't really bother trying to duplicate that. You already pull it again anyways with the `readOneUnsecured`
I'd just:
```ts
const ids = await this.db.query()
  .match([
    node('product', 'Product', { active: true }),
    relation('out', '', 'totalVerseEquivalents', ACTIVE),
    node('tve', 'Property'),
  ])
  .raw('WHERE tve.hadNaN = true or isNaN(tve.value)')
  .return<{ id: ID }>('product.id as id')
  .map('id')
  .run();
const products = await this.productService.readManyUnsecured(ids, this.fakeAdminSession);
for (const product of products) {
  const totalVerseEquivalents = // TODO recalc
  await this.productService.updateDirect({ id: product.id, totalVerseEquivalents });
  ...
}
```

src/components/product/migrations/fix-nan-total-verse-equivalents.migration.ts
```diff
@@ -0,0 +1,52 @@
...
+      ids,
+      this.fakeAdminSession,
+    );
+
+    for (const p of products) {
+      const correctTotalVerseEquivalent = getTotalVerseEquivalents(
+        ...p.scriptureReferences,
+      );
+
+      if (p.__typename === 'DirectScriptureProduct') {
```
bryanjnelson, line 38: Should anything be logged or done if the type is something other than one of these two?

bryanjnelson (approved):
@atGit2021 - Looks good outside of my comment (that I don't know the answer to).

@CarsonF - Over to you for a Cypher check.

src/components/product/migrations/fix-nan-total-verse-equivalents.migration.ts
atGit2021: I was going to add a logging statement, but then I noticed that out of the 60 products 24 are DirectScriptureProducts and the remaining 36 are DerivativeScriptureProducts. Do you think I should add one anyways for "safety" sake?

CarsonF: Nah. These are the only two types with this property.

CarsonF (approved):
LGTM!

PR #2921, opened 2023-10-27
removed check for undefined email address

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5407756515) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -126,19 +126,17 @@ export class UserService {
     await this.userRepo.updateProperties(user, simpleChanges);
 
     // Update email
-    if (email !== undefined) {
-      try {
-        await this.userRepo.updateEmail(user, email);
-      } catch (e) {
-        if (e instanceof UniquenessError && e.label === 'EmailAddress') {
-          throw new DuplicateException(
-            'person.email',
-            'Email address is already in use',
-            e,
-          );
-        }
-        throw new ServerException('Failed to create user', e);
+    try {
+      await this.userRepo.updateEmail(user, email);
+    } catch (e) {
+      if (e instanceof UniquenessError && e.label === 'EmailAddress') {
+        throw new DuplicateException(
+          'person.email',
+          'Email address is already in use',
+          e,
+        );
       }
+      throw new ServerException('Failed to create user', e);
     }
 
     // Update roles


src/components/user/user.service.ts
```diff
@@ -126,19 +126,17 @@ export class UserService {
...
-      } catch (e) {
-        if (e instanceof UniquenessError && e.label === 'EmailAddress') {
-          throw new DuplicateException(
-            'person.email',
-            'Email address is already in use',
-            e,
-          );
-        }
-        throw new ServerException('Failed to create user', e);
+    try {
```
bryanjnelson, line 129: @CarsonF - I've talked about this with Andre and if you look at Task 0736 in Monday you'll see that the issue is dealing with being able to clear out an email once it exists.  I do think that @atGit2021's solution will work, however we wanted to make sure this is an okay thing to do here.  If we make this change, there's a potential that a user (or someone with access) could remove an email and then be locked out (or lock someone else) out of the system if they tried to login.  The fix could be an easy help desk ticket and an admin could add it back in, but it could spiral.  The user could try to register a new account with the old email address, which would work, but the new account obviously wouldn't be tied to the old one and then activity could be split on them...etc.

Any thoughts on this?

atGit2021: @CarsonF @bryanjnelson, another issue that would need to be discussed;  this will also impact the reset password functionality.

CarsonF:
The undefined check is fine. We want missing/omitted properties to be left unchanged, not cleared.
The problem is null is not being respected.
[Here's](https://github.com/SeedCompany/cord-api-v3/compare/bugfix/email-null?expand=1) a proper fix that doesn't clear emails unintentionally.
Even with this though the UI is not sending nulls for empty values from that form field, but instead omitting that value as if it were to be unchanged. So [here's](https://github.com/SeedCompany/cord-field/compare/bugfix/email-null?expand=1) a fix for that logic too.

As far as auth consequences, I hear you and probably should be talked about with product. Part of the issue here is we have Users representing People entities and Users which log in. Maybe the solution is to distinguish them a bit more, like a conversion process to change from a managed person to a log-in-able user.

bryanjnelson:
@CarsonF - That's a good potential solution to help with other issue here.  Perhaps we want a User of the system to sort of extend a Person, or something like that.  You're giving the User more capabilities because they actually are going to use Cord.  I'm pretty sure this model could be set up easily in EdgeDB.  😄

CarsonF:
Closing in favor of https://github.com/SeedCompany/cord-api-v3/pull/2923 https://github.com/SeedCompany/cord-field/pull/1492

Auth implications can happen elsewhere with product team

PR #2930, opened 2023-11-02
Generate typed inline queries using new `edgeql()` wrapper

CarsonF:
I know the edgedb team pushes for the query builder, and they want it to be as close to EdgeQL syntax as possible. It's certainly closer than what we have with our cypher query builder, but it's not quite there.
This is why I went through the effort of supporting `*.edgeql` query files which could be used in app code.

However, I fear that the query builder will still be favored over these files because small queries can be contained in a block of code rather than a whole exclusive file. This is a bit concerning, because the QB does have rumors of being hard on TS and the aforementioned cypher QB headache.

This is why I'm introducing support for typed query strings inline in TS files...

# `edgeql()`
```ts
import { edgeql, EdgeDB } from '~/core/edgedb';

class Repository {
  constructor(private db: EdgeDB) {}

  listUsers() {
    const query = edgeql(`
      select User {*}
    `);

    const users = await this.db.run(query);

    users[0].realFirstName // fully typed.
  }
}
```

This does require running `yarn edgedb:gen` after every query change, just like the query files do.
So that's still a pro for the QB.
I'd like to support watching for changes in a future iteration.

# `EdgeDB` service

I've adjusted the `EdgeDB` service to support a unified signature for running all types of queries.

```ts
import { users as usersFromQueryFile } from './users.edgeql';

const inline = edgeql(`
  select User {*}
`);

const qb = e.select(e.Project.Member, (pm) => pm['*']);

const a = await this.db.run(usersFromQueryFile);
const b = await this.db.run(inline);
const c = await this.db.run(qb);
```

The previously exported Executor methods are no longer exposed. Will keep it that way until need arises, as this should be sufficient for app usage. The `Client` service is still available as well for special cases.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5441919616) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/edgedb/edgedb.module.ts
+++ b/src/core/edgedb/edgedb.module.ts
@@ -3,7 +3,8 @@ import { createClient } from 'edgedb';
 import { KNOWN_TYPENAMES } from 'edgedb/dist/codecs/consts.js';
 import { ScalarCodec } from 'edgedb/dist/codecs/ifaces.js';
 import { Class } from 'type-fest';
-import { Client, EdgeDB } from './reexports';
+import { EdgeDB } from './edgedb.service';
+import { Client } from './reexports';
 import { LuxonCalendarDateCodec, LuxonDateTimeCodec } from './temporal.codecs';
 
 @Module({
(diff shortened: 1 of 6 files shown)


bryanjnelson (approved):
Amazing.  🎉

PR #2922, opened 2023-10-30
EdgeDB CORD Schema Part 1

CarsonF:
Most parts of these objects are done:
- User
- Project
- Engagement
- Language
- Ceremony
- Project Member

Key points have been worked out:
- How sensitivity & (project) membership can be declared across schema objects for authorization system
- How sensitivity is calculated for translation projects (triggers written)
- Good direction on some conventions, like module organization.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5419271928) by [Unito](https://www.unito.io)

PR #2931, opened 2023-11-07
EdgeDB Project Context Schema

CarsonF:
# Why
This builds upon #2922. It established the pattern for how a project's sensitivity and current user's membership is shared for all types under a project, engagement, products, partnerships, etc.

However, we still have the use case of Languages, Partners, etc. These also use their related project's sensitivity and membership to determine authorization. It's just that they need aggregate multiple projects instead of a single one.

I had `isMember` and `effectiveSensitivity` defined on `Language`, and this is useful when we are looking at a language directly.
```
select Language {*}
filter .effectiveSensitivity <= Sens.Medium
```
This would indeed work for language queries and access policies. So I thought I had the problem solved, though admittedly a bit clumsy for these handful of types that were "project aware but outside projects". Since they would need to declare their own `sensitivity` and `isMember` fields themselves.

I started working on `File` schema, and wanted something that also works for `Posts/Comments` which are embedded into other resources. Now if I wanted to say "this directory can only be read if it's parent's sensitivity is medium or lower" it was impossible.
Every type that had a computed sensitivity property would need to be enumerated.
Something like
```
select Object {
  sensitivity := (
    [is Project].sensitivity ??
    [is Language].sensitivity ??
    [is Partner].sensitivity ??
    [is Organization].sensitivity
  )
}
```
This is not good.
We needed a way to refer to a single `sensitivity` field.

To do this we needed a unified _**stored**_ field that all project based queries could point back to.

# Project Context (What)

This adds a `Project::Context` type which holds a list of projects.
All project (context) aware objects will now have a reference to one of these:
- For Project `English 1`, its project context list would just be itself:
   `{ English 1 }`
- For Language `English`, its project context list would be all projects engaged with it:
   `{ English 1, English 2, English 3 }`
- For Partner `Seed Co`, its project context list would be all projects its partnered with:
   `{ English 1, Spanish 1 }`

Now `Language`, `Project`, `Engagement` all extend from `Project::ContextAware`. This abstract type is now able to provide a _single_ computed field for `sensitivity` that all 3 of these concrete types can use.
Looking back at our query above, it can now be
```
select Object {
  [is Project::ContextAware].sensitivity
}
```

## Draw backs

I've prioritized reading back these values, as that is the main need. For app queries, access policies, and adhoc reads from DB UI.
In order to make this happen though, this project context list is (unfortunately) denormalized from the actual relations that declare these connections.
So it's up to the _write process_ to declare how these different relations should affect project contexts.
For languages (and partners in future), this is done with triggers (which you can see I've added [here](https://github.com/SeedCompany/cord-api-v3/pull/2931/files#diff-5d11b98de213d001e2229978d3f171f85951a44d37a1fe2dba45d0dc1b91fd18R104-R111)). So queries do not need to worry about it.
You can also see that `Ethnologue::Language` now actually supports these as well. It shares the project context from the related `default::Language` so it gets those changes automatically.

`Projects` is the main hiccup. Currently edgedb cannot "link to self" with a single insert operation. Since a project's context is a list of itself, this is a problem. So for now, when creating projects, the project needs to be created first, and the separately, secondly updated to add itself to its context. But this will be done once in app code, and will just have to be something to remember if doing adhoc project creation.
This is by far the biggest rough edge, and I'll keep an eye out for a better (schema declared) workaround.

# `ownSensitivity`

With sensitivity being the single computed field on all of these types, we needed a different stored field to be the source.
I called this `ownSensitivity`, aka a Language's own sensitivity. Where as the "effective sensitivity" for authorization is now pulled from `sensitivity`.
This, I think, is more clear too. `ownSensitivity` is always writable, and `sensitivity` is always computed.
`Project::ContextAware` declares this field as optional, but `Project` & `Language` overload to make it required.

# `Project::Child`

I renamed `Project::Resource` -> `Project::Child` to hopefully be a bit more clear.
Also `Engagement::Resource` -> `Engagement::Child`.
Still not 100% sold, but could be better...maybe.

# EdgeDB 4.0

Also bumped to v4 lol

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5471944036) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -1,5 +1,5 @@
 module default {
-  abstract type Project extending Resource, Mixin::Pinnable, Mixin::Taggable, Project::HasContext {
+  abstract type Project extending Resource, Project::ContextAware, Mixin::Pinnable, Mixin::Taggable {
     required name: str {
       constraint exclusive;
     };
@@ -48,6 +48,10 @@ module default {
 #       link marketingLocation: Location;
 #       link fieldRegion: FieldRegion;
 #       link rootDirectory: Directory;
+    
+    overloaded link projectContext: Project::Context {
+      default := (insert Project::Context);
+    }
   }
   
   type TranslationProject extending Project {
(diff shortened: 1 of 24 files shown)


dbschema/project.esdl
```diff
@@ -57,7 +57,7 @@ module default {
     
     trigger confirmProjectSens after update for each do (
```
bryanjnelson, line 61: Personal pet peeve...sensitivity is spelled out 90% of the places...why not make it consistent and spell it out everywhere?  It doesn't add much space to do so.

bryanjnelson (approved):
Soooo much progress.

dbschema/project.esdl
CarsonF: Does that mean you don't like this either? 🙈 
https://github.com/SeedCompany/cord-api-v3/blob/e2df0a2c28559b1250e99c06d44255ed74dd1106/dbschema/sensitivity.esdl#L5-L5

I can spell out.

bryanjnelson: I don't, but there's more of a reason to do it there.  Feel free to ignore.

PR #2933, opened 2023-11-10
Move Language/Project name to `Mixin::Named`

CarsonF:
This makes it easy to grab a name from a polymorphic query.
I'm hoping we can rework user to also use this. Producibles & Files can also use this.
```
select Resource {
  id,
  [is Mixin::Named].name
}
```
I think this will come in handy for Search as well.

Also added some constraints for a trimmed string, so "Foo " becomes "Foo" and "   " becomes null & constraint error.
It's as a function so we could choose to apply it to more places.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5488033422) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/language.esdl
+++ b/dbschema/language.esdl
@@ -1,8 +1,5 @@
 module default {
-  type Language extending Resource, Project::ContextAware, Mixin::Pinnable, Mixin::Taggable {
-    required name: str;
-    index on (str_sortable(.name));
-    
+  type Language extending Resource, Project::ContextAware, Mixin::Named, Mixin::Pinnable, Mixin::Taggable {
     required displayName: str {
       default := .name;
     }
--- a/dbschema/z.named.esdl
+++ b/dbschema/z.named.esdl
@@ -1,0 +1,24 @@
+module Mixin {
+  abstract type Named {
+    required name: str {
+      rewrite insert, update using (default::str_clean(.name));
+    };
+    
+    index on (default::str_sortable(.name));
+    
+    index fts::index on (
+      fts::with_options(
+        .name,
+        language := fts::Language.eng,
+      )
+    );
+  } 
+}
+ 
+module default {
+  function str_clean(string: str) -> optional str
+    using(
+      with trimmed := str_trim(string, " \t\r\n")
+      select if len(trimmed) > 0 then trimmed else <str>{}
+    );
+}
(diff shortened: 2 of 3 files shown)


dbschema/z.named.esdl
```diff
@@ -0,0 +1,24 @@
...
+        language := fts::Language.eng,
+      )
+    );
+  } 
+}
+ 
+module default {
+  function str_clean(string: str) -> optional str
+    using(
+      with t := str_trim(string, " \t\r\n")
```
bryanjnelson, line 21: `t`?   🤔

dbschema/language.esdl
```diff
@@ -1,8 +1,5 @@
 module default {
-  type Language extending Resource, Project::ContextAware, Mixin::Pinnable, Mixin::Taggable {
-    required name: str;
```
bryanjnelson, line 3: Why doesn't this still exist (but as `overloaded`) like it does on `Project`?

dbschema/z.named.esdl
CarsonF: Random var name referenced on the next line. this probably goes back to the `sens` debate

dbschema/language.esdl
CarsonF: Project names are exclusive. Language names are not...for reasons...probably bad ones but @sethmcknight has convinced me of it before.

dbschema/z.named.esdl
bryanjnelson: That was my point.  I'm not real big on one letter variables.  We're not exactly trying to save bytes here.  Why not make it more clear what it is?  🤪

dbschema/language.esdl
bryanjnelson: Ah, got it now!

bryanjnelson (approved):
🚀

dbschema/z.named.esdl
CarsonF: Yeah I guess my perspective is it is the output of the str_trim function, so that makes it obvious. And this is a generic function so we can't give it a more clear name.
I could change it to `trimmed`...does that really add clarity? 🤔

bryanjnelson: I think it does, yes.  It's style and preference, so whatevz though.

CarsonF: Roger, updated! 👍

PR #2934, opened 2023-11-10
[EdgeDB] Create `Location` schema

bryanjnelson:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5488035870) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/location.esdl
+++ b/dbschema/location.esdl
@@ -1,0 +1,32 @@
+module default {
+  type Location extending Resource, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
+    }
+    required type: Location::Type;
+
+    isoAlpha3: Location::IsoAlpha3Code {
+      constraint exclusive;
+    };
+
+    #TODO - links
+    #fundingAccount: FundingAccount;
+    #defaultFieldRegion: FieldRegion;
+    #mapImage: File;
+  }
+}
+
+module Location {
+  scalar type Type extending enum<
+    Country,
+    City,
+    County,
+    Region,
+    State,
+    CrossBorderArea
+  >;
+
+  scalar type IsoAlpha3Code extending str {
+    constraint regexp(r'^[A-Z]{3}$');
+  }
+}


dbschema/location.esdl
```diff
@@ -0,0 +1,28 @@
+module default {
+  abstract type Location extending Resource {
```
CarsonF, line 2:
```suggestion
  type Location extending Resource {
```

dbschema/location.esdl
```diff
@@ -0,0 +1,28 @@
...
+    required name: str {
```
CarsonF, line 3: It's grab this from Named from the PR I just put up.
https://github.com/SeedCompany/cord-api-v3/pull/2933/files#diff-038ac5fc7d17f701e97927ce062fb044c9dd954bdbbf3566134674f04bb87194L2-R3

dbschema/location.esdl
```diff
@@ -0,0 +1,28 @@
...
+      constraint exclusive;
+    }
+    required type: Location::Type;
+
+    isoAlpha3: Ethnologue::code {
```
CarsonF, line 8: Hmm yeah this is a 3 digit alpha lower cased code....I hate to reference ethnologue here though. Two unrelated standards. I think I'd be fine copy pasting that here to declare this as an `IsoAlpha3Code`.
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3
Wikipedia actually uppercases them...Oh actually so do we. So it's a unique regex anyways.

dbschema/location.esdl
```diff
@@ -0,0 +1,28 @@
...
+      constraint exclusive;
+    };
+
+    #TODO - links
+    #fundingAccountId: FundingAccount;
+    #defaultFieldRegionId: FieldRegion;
```
CarsonF, line 14: I know these are just comments, but I feel like it's worth pointing out that this would be links to the object so
```suggestion
    #fundingAccount: FundingAccount;
    #defaultFieldRegion: FieldRegion;
```

dbschema/location.esdl
```diff
@@ -0,0 +1,28 @@
...
+module Location {
+  scalar type Type extending enum<
+    Country,
+    City,
+    County,
+    Region,
+    State,
+    CrossBorderArea
+  >;
+}
```
CarsonF, line 28: Get those trailing line breaks, as God intended 😛

CarsonF (changes requested):

dbschema/location.esdl
bryanjnelson: That's what I was getting at.  It wasn't exactly the same regex.

CarsonF: Yep! My b if I miscommunicated that earlier today

dbschema/location.esdl
```diff
@@ -0,0 +1,32 @@
+module default {
+  type Location extending Resource, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
+    }
+    required type: Location::Type;
+
+    isoAlpha3: IsoAlpha3Code {
```
CarsonF, line 8:
```suggestion
    isoAlpha3: Location::IsoAlpha3Code {
```

CarsonF (approved):

PR #2940, opened 2023-11-13
Implicitly run GQL mutations in EdgeDB transactions

CarsonF:
This is the same thing as what we do for Neo4j.
It's worth noting that both of these interceptors are live right now, which could have unintended consequences with retry/rollback logic between the two DBs. This is only a temporary migration concern, so I'm not going to give it much thought until I need to.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5500377488) by [Unito](https://www.unito.io)


bryanjnelson (approved):
🚀

PR #2957, opened 2023-11-14
[EdgeDB] Create `FundingAccount` schema

bryanjnelson:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5508396647) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/funding-account.esdl
+++ b/dbschema/funding-account.esdl
@@ -1,0 +1,12 @@
+module default {
+  type FundingAccount extending Resource, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
+    }
+
+    required accountNumber: int16 {
+      constraint min_value(0);
+      constraint max_value(9);
+    }
+  }
+}


dbschema/funding-account.esdl
```diff
@@ -0,0 +1,16 @@
...
+    overloaded name {
+      constraint exclusive;
+    }
+    
+    required accountNumber: FundingAccount::accountNumber;
+  }
+}
+
+module FundingAccount {
+  scalar type accountNumber {
```
CarsonF, line 12: Idk how necessary a separate type is. In other places we kinda justified the guess by saying it could be used in more places. `population` could be used in multiple spots, etc.
It seems really unlikely that this will ever be used anywhere else. We could just inline the constraints.
Thoughts?

bryanjnelson: I'm totally fine with that.

CarsonF (approved):
:shipit:

PR #2962, opened 2023-11-14
[EdgeDB] Create `Budget` and `BudgetRecord` schemas and associated triggers

bryanjnelson:
[Monday story](https://seed-company-squad.monday.com/boards/3451697530/pulses/5452828843)

NOTE:  Wait to merge until Org/Partner schemas are complete.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5512960148) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/budget.esdl
+++ b/dbschema/budget.esdl
@@ -1,0 +1,40 @@
+module default {
+  type Budget extending Project::Child {
+    required status: Budget::Status {
+      default := Budget::Status.Pending;
+    };
+    
+    universalTemplate: File;
+    
+    records := .<budget[is Budget::Record];
+  }
+}
+  
+module Budget {
+  type Record extending Project::Child {
+    constraint exclusive on ((.budget, .fiscalYear, .organization));
+    
+    required fiscalYear: int16 {
+      readonly := true;
+    };
+    
+    amount: float32;
+    
+    required budget: default::Budget {
+      readonly := true;
+      on target delete delete source;
+    };
+    
+    required organization: default::Organization {
+      readonly := true;
+      on target delete delete source;
+    };
+  }
+  
+  scalar type Status extending enum<
+    Pending,
+    Current,
+    Superceded,
+    Rejected
+  >;
+}
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -60,6 +60,14 @@ module default {
         message := "A Project's own context should be itself (no more or less)"
       )
     );
+    
+    trigger createBudgetOnInsert after insert for each do (
+      insert default::Budget {
+        createdAt := datetime_of_statement(),
+        project := __new__,
+        projectContext := __new__.projectContext,
+      }
+    );
   }
   
   type TranslationProject extending Project {
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -64,6 +64,7 @@ module default {
     trigger createBudgetOnInsert after insert for each do (
       insert default::Budget {
         createdAt := datetime_of_statement(),
+        modifiedAt := datetime_of_statement(),
         project := __new__,
         projectContext := __new__.projectContext,
       }
--- a/dbschema/engagement.esdl
+++ b/dbschema/engagement.esdl
@@ -75,6 +75,7 @@ module default {
     trigger connectDedicationCeremony after insert for each do (
       insert Engagement::DedicationCeremony {
         createdAt := datetime_of_statement(),
+        modifiedAt := datetime_of_statement(),
         engagement := __new__,
         project := __new__.project,
         projectContext := __new__.projectContext,
@@ -128,6 +129,7 @@ module default {
     trigger connectCertificationCeremony after insert for each do (
       insert Engagement::CertificationCeremony {
         createdAt := datetime_of_statement(),
+        modifiedAt := datetime_of_statement(),
         engagement := __new__,
         project := __new__.project,
         projectContext := __new__.projectContext,
--- a/dbschema/migrations/00030.edgeql
+++ b/dbschema/migrations/00030.edgeql
@@ -1,0 +1,43 @@
+CREATE MIGRATION m1f6w2tfwomhiwhtzwnpraqfo24x3rnfcpnntt7nxe5f5l7utyjzpa
+    ONTO m1t7atw76jdjhcnrngoj5euzbz7qxve7dokvbhh4x5333yxm5klkwa
+{
+  CREATE MODULE Budget IF NOT EXISTS;
+  CREATE TYPE Budget::Record EXTENDING Project::Child {
+      CREATE REQUIRED LINK organization: default::Organization {
+          ON TARGET DELETE DELETE SOURCE;
+          SET readonly := true;
+      };
+      CREATE REQUIRED PROPERTY fiscalYear: std::int16 {
+          SET readonly := true;
+      };
+      CREATE PROPERTY amount: std::float32;
+  };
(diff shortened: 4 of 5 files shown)


dbschema/budget-record.esdl
```diff
@@ -0,0 +1,13 @@
+module default {
+  type BudgetRecord extending Project::ContextAware {
```
CarsonF, line 2: Technically these are project children too, so they should be represented as such.
```suggestion
  type BudgetRecord extending Project::Child {
```

dbschema/budget-record.esdl
```diff
@@ -0,0 +1,13 @@
...
+    required fiscalYear: int16;
+    required amount: float32 {
+      default := 0;
+    };
```
CarsonF, line 6: Let's make this optional. We auto create these, so the presence of a value here is the only way we know if it's been touched or not. And I think 0 is a valid value people enter.

Also matches app code: https://github.com/SeedCompany/cord-api-v3/blob/e9007a81ac14a3a6269e42fa0f4d2c8c6dc87529/src/components/budget/dto/budget-record.dto.ts#L38-L38
```suggestion
    amount: float32;
```

dbschema/budget-record.esdl
```diff
@@ -0,0 +1,13 @@
+module default {
+  type BudgetRecord extending Project::ContextAware {
+    required fiscalYear: int16;
```
CarsonF, line 3: Let's make this readonly. These don't change.

dbschema/budget-record.esdl
```diff
@@ -0,0 +1,13 @@
...
+    required amount: float32 {
+      default := 0;
+    };
+
+    required budget: Budget;
+
+    #TODO
+    # required organization: Organization;
```
CarsonF, line 11: This should be readonly as well. Let's wait to merge this until org/partners are done. I forgot about this dependency.

Records should be exclusive for budget/project, org, and fiscal year as well.

dbschema/budget.esdl
```diff
@@ -0,0 +1,19 @@
+module default {
+  type Budget extending Project::Child, Project::ContextAware {
```
CarsonF, line 2: Project Child is already context aware
```suggestion
  type Budget extending Project::Child {
```

dbschema/budget.esdl
```diff
@@ -0,0 +1,19 @@
...
+      default := Budget::status.Pending;
+    };
+
+    records: BudgetRecord;
+    single universalTemplate: File;
+  }
+}
+
+module Budget {
+  scalar type status extending enum<
```
CarsonF, line 13: Noting that we confirmed in DM that enums will be PascalCase.

dbschema/budget-record.esdl
CarsonF, line 1: I'd just put this in the budget file. Both are small enough to fit nicely together and they both are useless without each other.

dbschema/budget.esdl
```diff
@@ -0,0 +1,19 @@
+module default {
+  type Budget extending Project::Child, Project::ContextAware {
+    required status: Budget::status {
+      default := Budget::status.Pending;
+    };
+
+    records: BudgetRecord;
```
CarsonF, line 7: Ok so with this and above, there are two points of truth for the relationship.
We want this to be either
```
multi records: BudgetRecord;
```
or
```
records := .<budget[is BudgetRecord]
```
(same holds true for `BudgetRecord.budget` as well)

I don't know off the top of my head which is better, it's obv use case dependent.

dbschema/budget-record.esdl
CarsonF, line 2: Maybe `Budget::Record` instead?
```suggestion
module Budget {
  type Record extending Project::ContextAware {
```

CarsonF (changes requested):

dbschema/budget-record.esdl
bryanjnelson: I was waffling on splitting these two into their separate files, but ultimately I think that's better than combining them if they are small.  _However_, since the use case in this instance is the complete dependency of one on the other (i.e. it can't exist by itself), I think your solution here is the best.  Nice one.

dbschema/budget-record.esdl
bryanjnelson: I'm a little unsure what you mean by

> Records should be exclusive for budget/project

Like this?

```
type Record extending Project::Child {
    constraint exclusive on (.project);

Comment (CarsonF) [reply dbschema/budget-record.esdl:11]: Something like
```
constraint exclusive on ((.budget, .organization, .fiscalYear))
```
The `budget` is the question there. Idk if that's a stored field or a computed one. Referencing the other unresolved comment below.

File: dbschema/budget.esdl (hunk 7 above)
Comment (bryanjnelson) [reply dbschema/budget.esdl:7]: Let me put this out there as a proposed solution:

```
type Budget extending Project::Child {
    records := .<budget[is Budget::Record];
}

type Record extending Project::Child {
    required budget: Budget {
      on target delete delete source;
    };
}
```

Thoughts?

Comment (CarsonF) [reply dbschema/budget.esdl:7]: Yeah makes sense to me! 

I'm trying to find the use case for the other way around. Where the multi link is the stored truth instead of a computed back link.
I guess `User.pins` is a good example

Comment (bryanjnelson) [reply dbschema/budget.esdl:7]: In that situation is there even a backlink at all?

**Review by CarsonF — APPROVED** [review]
This looks good! I still think it would be better to wait to merge until Org/Partner is done since there's a dependency here.

The only other thing I can think of is moving the app event handlers to triggers here.
https://github.com/SeedCompany/cord-api-v3/tree/develop/src/components/budget/handlers
- create a budget when a project is created
- change budget status based on project status changes
- syncing record creation/deletion based on funding partnerships is harder so we can wait until later for that. It obviously needs the partnerships schema done first as well.

File: dbschema/project.esdl [hunk 8]
```diff
@@ -60,6 +60,24 @@ module default {
...
     );
+
+    trigger createBudgetOnInsert after insert for each do (
+      insert default::Budget {
+        project := __new__.project,
+        projectContext := __new__.projectContext,
+      }
+    );
+
+    trigger updateBudgetSensitivityOnSensitivityChange after update for each do (
```
CarsonF, line 71: No need. The `projectContext` given on insert calculates the `sensitivity` on read.
https://github.com/SeedCompany/cord-api-v3/blob/9f47163ab06438265a7e8dff549522ab74bcc596/dbschema/project.esdl#L115-L117

CarsonF (approved):

dbschema/project.esdl
```diff
@@ -60,6 +60,13 @@ module default {
         message := "A Project's own context should be itself (no more or less)"
       )
     );
+
+    trigger createBudgetOnInsert after insert for each do (
+      insert default::Budget {
+        project := __new__.project,
```
CarsonF, line 66: Ah CI caught this
```suggestion
        project := __new__,
```

CarsonF (changes requested):

dbschema/project.esdl
```diff
@@ -60,6 +60,13 @@ module default {
         message := "A Project's own context should be itself (no more or less)"
       )
     );
+
+    trigger createBudgetOnInsert after insert for each do (
+      insert default::Budget {
+        project := __new__,
```
CarsonF, line 66: Triggers appear to have a hard time with default expressions.
I had to add it here too
https://github.com/SeedCompany/cord-api-v3/blob/9f47163ab06438265a7e8dff549522ab74bcc596/dbschema/engagement.esdl#L130-L130

CarsonF (approved):

PR #2965, opened 2023-11-15
0702 renamed marketingLocation to marketingCountryOverride

willdch:
Rename marketingLocation to ~~MarketingLocationOverride~~ MarketingCountryOverride.

NOTE:  This is dependent upon other work and should not be merged independently. 

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5518296450) by [Unito](https://www.unito.io)


bryanjnelson (approved):
Assuming you found all the references, I think this looks good @willdch.

It's possible the business will want to alter some of the permissions since we're changing the model here a little, but we can probably check and do that as a separate task/PR.

@CarsonF - on to you.

CarsonF:
Let's merge this into #3047

PR #2967, opened 2023-11-17
[EdgeDB] Create `Organization` and `Partner` schemas

bryanjnelson:
[Monday story](https://seed-company-squad.monday.com/boards/3451697530/pulses/5452830216)

### DIFF
--- a/dbschema/organization.esdl
+++ b/dbschema/organization.esdl
@@ -1,0 +1,29 @@
+module default {
+  type Organization extending Resource, Project::ContextAware, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
+    }
+    
+    acronym: str;
+    #address: str; #TODO - this needs figured out - needed on here and Partner?
+    multi types: Organization::Type;
+    multi reach: Organization::Reach;
+  }
+}
+  
+module Organization {
+  scalar type Type extending enum<
+    Church,
+    Parachurch,
+    Mission,
+    TranslationOrganization,
+    Alliance
+  >;
+  
+  scalar type Reach extending enum<
+    Local,
+    Regional,
+    National,
+    Global
+  >;
+}
--- a/dbschema/organization.esdl
+++ b/dbschema/organization.esdl
@@ -24,6 +24,6 @@ module Organization {
     Local,
     Regional,
     National,
-    Global
+    `Global`
   >;
 }
--- a/dbschema/partner.esdl
+++ b/dbschema/partner.esdl
@@ -1,36 +1,37 @@
 module default {
-  type Partner extending Resource, Project::ContextAware, Mixin::Pinnable, Mixin::Taggable, Mixin::Named {
+  type Partner extending Resource, Project::ContextAware, Mixin::Named, Mixin::Pinnable, Mixin::Taggable {
     overloaded name {
       constraint exclusive;
     }
-
-    required active: bool;
-    required globalInnovationsClient: bool;
-
+    
+    required active: bool {
+      default := true;
+    };
+    required globalInnovationsClient: bool {
+      default := false;
+    };
+    
     pmcEntityCode: str {
       constraint regexp(r'^[A-Z]{3}$');
     }
-
+    
     #address: str; #TODO - this needs figured out - needed on here and Organization?
-    types: array<Partner::Type>;
-    financialReportingTypes: array<Partner::FinancialReportingType>;
-
-    pointOfContact: Person;
+    multi types: Partner::Type;
+    multi financialReportingTypes: Partner::FinancialReportingType;
+    
+    pointOfContact: User;
     languageOfWiderCommunication: Language;
-
+    
     required organization: Organization {
+      readonly := true;
       constraint exclusive;
     };
-    required multi languagesOfConsulting: Language;
-    required multi fieldRegions: FieldRegion;
-    required multi countries: Location;
-
-    overloaded link projectContext: Project::Context {
-      default := (insert Project::Context);
-    }
+    multi languagesOfConsulting: Language;
+    multi fieldRegions: FieldRegion;
+    multi countries: Location;
   }
 }
-
+  
 module Partner {
   scalar type Type extending enum<
     Managing,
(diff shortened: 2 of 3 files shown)


dbschema/organization.esdl
```diff
@@ -0,0 +1,29 @@
+module default {
+  type Organization extending Resource, Project::ContextAware, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
+    }
+
+    acronym: str;
+    #address: str; #TODO - this needs figured out - needed on here and Partner?
+    types: array<Organization::Type>;
```
CarsonF, line 9: I _think_ I determined that this is better
```suggestion
    multi types: Organization::Type;
```

> If order is important, use an [array](https://www.edgedb.com/docs/datamodel/primitives#ref-datamodel-arrays). Otherwise, multi properties are recommended.

dbschema/organization.esdl
```diff
@@ -0,0 +1,29 @@
...
+    reach: array<Organization::Reach>;
+  }
+}
+
+module {
```
CarsonF, line 14:
```suggestion
module Organization {
```

dbschema/partner.esdl
```diff
@@ -0,0 +1,38 @@
+module default {
+  type Partner extending Resource, Project::ContextAware, Mixin::Pinnable, Mixin::Taggable {
```
CarsonF, line 2: I think it would be better if this was `Named`. It's really annoying to not have this. I think it will most immediately affect is search. You can see all the references to "PartnerByOrg" in https://github.com/SeedCompany/cord-api-v3/blob/9f47163ab06438265a7e8dff549522ab74bcc596/src/components/search/search.service.ts where we work around this.

We could just have a trigger on Organization to update this on the related Partner.

Speaking of which I don't even see organization linked here....on nvm it's buried in the other links.

dbschema/partner.esdl
```diff
@@ -0,0 +1,38 @@
...
+
+    pmcEntityCode: str {
+      constraint regexp(r'^[A-Z]{3}$');
+    }
+
+    #address: str; #TODO - this needs figured out - needed on here and Organization?
+    types: array<Partner::Type>;
+    financialReportingTypes: array<Partner::FinancialReportingType>;
+
+    pointOfContact: default::Person;
```
CarsonF, line 14:
```suggestion
    pointOfContact: User;
```
The other explicit `default::`s below aren't really needed either...

dbschema/partner.esdl
```diff
@@ -0,0 +1,38 @@
...
+    single languageOfWiderCommunication: default::Language;
+    required organization: default::Organization;
+    required languagesOfConsulting: default::Language;
+    required fieldRegions: default::FieldRegion;
+    required countries: default::Location;
```
CarsonF, line 19:
```suggestion
    required multi languagesOfConsulting: default::Language;
    required multi fieldRegions: default::FieldRegion;
    required multi countries: default::Location;
```

dbschema/partner.esdl
CarsonF, line 2: We should automatically create a project context for this. Just like we do here
https://github.com/SeedCompany/cord-api-v3/blob/9f47163ab06438265a7e8dff549522ab74bcc596/dbschema/language.esdl#L82-L84

dbschema/partner.esdl
```diff
@@ -0,0 +1,38 @@
...
+      constraint regexp(r'^[A-Z]{3}$');
+    }
+
+    #address: str; #TODO - this needs figured out - needed on here and Organization?
+    types: array<Partner::Type>;
+    financialReportingTypes: array<Partner::FinancialReportingType>;
+
+    pointOfContact: default::Person;
+    single languageOfWiderCommunication: default::Language;
+    required organization: default::Organization;
```
CarsonF, line 16: We should probably create this automatically with a trigger. It can be given the same project context for this object.
Same as this
https://github.com/SeedCompany/cord-api-v3/blob/9f47163ab06438265a7e8dff549522ab74bcc596/dbschema/language.esdl#L26-L32
Though there's no `ownSensitivity` here.


I wonder if this should be exclusive too at least for now, so it's a 1-1 relationship.

CarsonF (changes requested):

dbschema/organization.esdl
bryanjnelson: 🤦🏼

dbschema/partner.esdl
bryanjnelson: Yeah, I was going to put it on there, but I was following the current model which doesn't have `name` on `Partner`.  It does seem odd not to have it.

The trigger idea is nice.

dbschema/partner.esdl
bryanjnelson: I thought `multi` is implicit on links?

CarsonF: _Stored_ links are `single` by default. _Backlinks_ are `multi` by default.

dbschema/partner.esdl
CarsonF: I think I just didn't want to bother with Neo4j. Here it's much easier.

bryanjnelson: So we're saying that a `Partner` has a sensitivity based upon all the `Project`s it's connected to?  I'm not sure I realized that.

Also, what does the `isMember` gain us here for a `Partner` from a business perspective?  Is this just so that on a Partner's page we can see which Projects they are a part of?

dbschema/partner.esdl
bryanjnelson: I'm not following here.  In the case of your example I understand that we want to create an `Ethnologue` based on a new `Language`.  Are we saying here that we want to automatically create a `Partner` when we create an `Organization`?

dbschema/partner.esdl
bryanjnelson: The `isMember` piece is a little fuzzy, but I get the idea; thanks.

dbschema/partner.esdl
bryanjnelson: After a Slack discussion, the solution here for now is to do this:

```
required organization: Organization {
      constraint exclusive;
    };
```

which denotes a 1-to-1 relationship between `Partner` and `Organization`.

PR #2968, opened 2023-11-17
EdgeDB QB implicit default type or named module 

CarsonF:
Essentially this allows
```ts
e.User
e.User.Status
```
Which is closer to what EdgeQL allows

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5534240837) by [Unito](https://www.unito.io)


bryanjnelson (approved):
This is great!!  Thanks @CarsonF!

PR #2970, opened 2023-11-22
[EdgeDB] Create `Producible`, `Film`, `EthnoArt`, and `Story` schemas

bryanjnelson:
@CarsonF - I'm not necessarily sold on the splitting out of all of these into their own `esdl` files now that I see the end result.  A case could be made to place them all in the `producible.esdl`.  Or perhaps even include them in the `Producible` module even?  Thoughts appreciated.

[Monday Task](https://seed-company-squad.monday.com/boards/3451697530/pulses/5452828634)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5553618688) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/producible.esdl
+++ b/dbschema/producible.esdl
@@ -1,0 +1,13 @@
+module default {
+  abstract type Producible extending Resource, Mixin::Named {
+    overloaded name {
+      delegated constraint exclusive;
+    };
+    
+    scriptureReferences: multirange<int32>;
+  }
+  
+  type EthnoArt extending Producible;
+  type Film extending Producible;
+  type Story extending Producible;
+}
--- a/dbschema/migrations/00028.edgeql
+++ b/dbschema/migrations/00028.edgeql
@@ -1,0 +1,14 @@
+CREATE MIGRATION m1rob2znvgmu7myqon4wzkdxur6765udhhucohw6apoxhtrbbnxova
+    ONTO m1aqhugpxmkst3yhkmyd5avjuqvbxk7ekgroiipkkhz3ockszv2rka
+{
+  CREATE ABSTRACT TYPE default::Producible EXTENDING default::Resource, Mixin::Named {
+      ALTER PROPERTY name {
+          SET OWNED;
+          CREATE DELEGATED CONSTRAINT std::exclusive;
+      };
+      CREATE PROPERTY scriptureReferences: multirange<std::int32>;
+  };
+  CREATE TYPE default::EthnoArt EXTENDING default::Producible;
+  CREATE TYPE default::Film EXTENDING default::Producible;
+  CREATE TYPE default::Story EXTENDING default::Producible;
+};


dbschema/ethno-art.esdl
```diff
@@ -0,0 +1,7 @@
+module default {
+  type EthnoArt extending Mixin::Named, Producible {
```
CarsonF, line 2: I'd probably put mixins after "stronger"(?) parents.
```suggestion
  type EthnoArt extending Producible, Mixin::Named {
```

dbschema/producible.esdl
```diff
@@ -0,0 +1,12 @@
+module default {
+  abstract type Producible extending Resource {
+    scriptureReferences: multirange<int32>;
+  }
+}
+
+module Producible {
+  function sortScriptureReferences(scriptureReferences: multirange<int32>) -> multirange<int32>
```
CarsonF, line 8: Doesn't this happen automatically on `set`/save?

CarsonF (commented):
> I'm not necessarily sold on the splitting out of all of these into their own `esdl` files now that I see the end result. A case could be made to place them all in the `producible.esdl`. Or perhaps even include them in the `Producible` module even? Thoughts appreciated.

Yeah I could go either way. I'd probably lean towards one file. Idk about moving them all into the `Producible` module. That would be a non-trivial name change compared to app code, but I definitely like the idea. Just not sure about how much we should be deviating at this point...

Speaking of deviation though, I wonder if we should just make `Producible` extend `Named`. My thought years ago was to not enforce this field, as producibles could be labeled differently, e.g. film title vs something else. This hasn't exactly played out well though. They all have a single name field right now, so it's kinda silly we don't just have it on the producible.
That certainly makes it easier:
```
abstract type Producible extending Resource, Mixin::Named {
  overloaded name {
    delegated constraint exclusive;
  }
  scriptureReferences: multirange<int32>;
}
type EthnoArt extending Producible;
type Film extending Producible;
type Story extending Producible;
```

dbschema/ethno-art.esdl
bryanjnelson: Yeah, I can see that.  Sounds like a fine standard.  🤪

bryanjnelson:
> Speaking of deviation though, I wonder if we should just make Producible extend Named. My thought years ago was to not enforce this field, as producibles could be labeled differently, e.g. film title vs something else. This hasn't exactly played out well though. They all have a single name field right now, so it's kinda silly we don't just have it on the producible.
> That certainly makes it easier:
> 
> abstract type Producible extending Resource, Mixin::Named {
>   overloaded name {
>     delegated constraint exclusive;
>   }
>   scriptureReferences: multirange<int32>;
> }
> type EthnoArt extending Producible;
> type Film extending Producible;
> type Story extending Producible;

Yeah, I was wondering if you were going to say something about that.  I like it...and will make it so!

dbschema/producible.esdl
CarsonF: Doesn't the sorting/minimizing happen when the multirange is created? Aka there's no such thing as an unsorted multirange?

bryanjnelson: Hmm...it does.  So we have no need of a function at all then?

dbschema/producible.esdl
```diff
@@ -0,0 +1,13 @@
+module default {
+  abstract type Producible extending Resource, Mixin::Named {
+    overloaded name {
+      constraint exclusive;
```
CarsonF, line 4:
```suggestion
      delegated constraint exclusive;
```
So it's unique per concrete type

bryanjnelson: Sorry...missed that.  Makes sense.

CarsonF (approved):

PR #2971, opened 2023-11-22
[EdgeDB] Create `Education` & `Unavailability` types in `User` schema

bryanjnelson:
[Monday Story](https://seed-company-squad.monday.com/boards/3451697530/pulses/5452830669)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5558073015) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/user.esdl
+++ b/dbschema/user.esdl
@@ -24,6 +24,12 @@ module default {
     multi link pins: Mixin::Pinnable {
       on target delete allow;
     }
+    multi link education: User::Education {
+      on target delete allow;
+    }
+    multi link unavailabilities: User::Unavailability {
+      on target delete allow;
+    }
   }
   
   alias RootUser := (
@@ -33,5 +39,25 @@ module default {
 }
  
 module User {
+  type Education extending default::Resource {
+    required degree: Degree;
+    required major: str;
+    required institution: str;
+  }
+  
+  type Unavailability extending default::Resource {
+    required description: str;
+    required dates: range<cal::local_date>;
+  }
+  
   scalar type Status extending enum<Active, Disabled>;
+  
+  scalar type Degree extending enum<
+    Primary,
+    Secondary,
+    Associates,
+    Bachelors,
+    Masters,
+    Doctorate
+  >;
 }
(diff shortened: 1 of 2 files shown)


dbschema/user.esdl
```diff
@@ -33,5 +33,26 @@ module default {
...
+  type Education extending default::Resource {
+    required degree: Degree;
+    required major: str;
+    required institution: str;
+  }
+
+  type Unavailability extending default::Resource {
+    required description: str;
+    required start: cal::local_date;
+    required end: cal::local_date;
```
CarsonF, line 45: Looks like start/end are reserved words. So...
```suggestion
    required `start`: cal::local_date;
    required `end`: cal::local_date;
```
These backticks will be required for use in EdgeQL as well. I don't think this is that big of a deal. Data Explorer and Query Builder handle these normally. EdgeQL will probably use `{*}` for select and only need to specify on insert query.

I don't think it's worth changing the property names. I'd stick with these and use backticks 👍 

---

Another thought here is to use a range instead.
```
range<cal::local_date>
```
This would verify we have a valid range automatically, otherwise we'd need to [verify this constraint ourselves](https://github.com/SeedCompany/cord-api-v3/blob/cf9b844877dfed414124754ee369f06242af5180/dbschema/project.esdl#L26-L26).

Not sure about which way would be more idiomatic.

CarsonF (changes requested):
Need to add links to User
```
  type User {
    multi link unavailabilities: User::Unavailability {
      on target delete allow;
    }
  }
```

dbschema/user.esdl
bryanjnelson: Honestly, I really like the idea of `range<cal::local_date>`.  It's really exactly what we need here, built-in constraint and all.  Good idea.

dbschema/user.esdl
```diff
@@ -24,6 +24,9 @@ module default {
     multi link pins: Mixin::Pinnable {
       on target delete allow;
     }
+    multi link unavailabilities: User::Unavailability {
+      on target delete allow;
+    }
```
CarsonF, line 29: Education too. Sorry for not being there explicit on last comment

CarsonF (changes requested):

dbschema/user.esdl
```diff
@@ -33,5 +39,25 @@ module default {
...
 module User {
+  type Education extending default::Resource {
+    required degree: Degree;
+    required major: str;
+    required institution: str;
+  }
+
+  type Unavailability extending default::Resource {
+    required description: str;
+    required dateRange: range<cal::local_date>;
```
bryanjnelson, line 50: @CarsonF - not sold on the name of this property...open to suggestions.

bryanjnelson: Yeah, thought about that.  Don't love it.  But I've got nothing better!

CarsonF (approved):

PR #2980, opened 2023-11-30
Refactor users tests

CarsonF:
- Only create one org to test with
- Don't do log in logic for every user created
- Combine some tests for performance
- Fix email tests to actually test clearing user's email
(undefined -> null)


bryanjnelson (approved):
🚀

PR #2983, opened 2023-12-06
Agl/add is language of consultiung boolean

adam-soltech:
[ticket](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/4788464835?filter=XQAAAALKAAAAAAAAAABBqQqHk62Wwa8yt9cbbXI48HpL2ei1XixEiNVOhiWmf4IrTn4lrcVxz0d8EzlqDpxL1dc7i9JjI1nNW0HEFgqaXhsDsHWTHjMEE6-qlUKWgWTaunFWUdBXUlo9GrL7-Dxs3yXtVsCpk7bU6tq4ASH03XfQsurYW6oz6SfWXIHjSrPNZIVE3__87GiYdcf5oUuHV6IpVDC18IDUOyTd3hSlyn4xwV6TGJGJ____cnOAAA)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5640489530) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/migrations/00033.edgeql
+++ b/dbschema/migrations/00033.edgeql
@@ -1,7 +1,9 @@
-CREATE MIGRATION m1ijgeabkfengfe2lpxjtq7xxqjiulhehay7tuutdoa5g3scsenwvq
+CREATE MIGRATION m1cwdkj3hjcsprjurukh7aaofw63m5avm52nkzekwwdmliytgx4wja
     ONTO m1hb7zty3d4ekb5ftznc4tv4h5mgkxy6db42zbxylaw2rmrpveqsxq
 {
-  ALTER TYPE Project::Child {
-      ALTER TRIGGER enforceCorrectProjectContext USING (std::assert((__new__.projectContext = __new__.project.projectContext), message := "Given project context must match given project's context"));
+  ALTER TYPE default::Language {
+      CREATE REQUIRED PROPERTY isLanguageOfConsulting: std::bool {
+          SET default := false;
+      };
   };
 };
--- a/dbschema/migrations/00033.edgeql
+++ b/dbschema/migrations/00033.edgeql
@@ -1,9 +1,7 @@
-CREATE MIGRATION m1cwdkj3hjcsprjurukh7aaofw63m5avm52nkzekwwdmliytgx4wja
+CREATE MIGRATION m1ijgeabkfengfe2lpxjtq7xxqjiulhehay7tuutdoa5g3scsenwvq
     ONTO m1hb7zty3d4ekb5ftznc4tv4h5mgkxy6db42zbxylaw2rmrpveqsxq
 {
-  ALTER TYPE default::Language {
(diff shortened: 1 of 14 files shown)


dbschema/migrations/00033.edgeql
bryanjnelson, line 1: You wouldn't have known this, but we're not actually checking in EdgeDB Migrations at this point.  Carson is handling this himself, and grouping them together as makes sense.  Eventually that will change after we're fully migrated, but for now don't worry about doing the migration.

bryanjnelson (changes requested):
Thanks @adam-soltech!

CarsonF:
`Language.isLanguageOfConsulting` seems not the best name since it has the object name in it again.
- `isUsedForConsulting`?
- `canBeUsedForConsulting`?
- `allowedForConsulting`?
- `consulting`?

maybe an enum list would be better to encompass all of these?
```ts
Language.allowedUse = ['WiderCommunication', 'Consulting', 'X'];
```
Though I guess it would be better to describe what this language _is_ vs how the language is allowed to be used in other places.
If we wanted different roles/conditions to be able to assign individual values, then a single list isn't the best way.

dbschema/migrations/00033.edgeql
adam-soltech: Without the migration the [e2e tests failed](https://github.com/SeedCompany/cord-api-v3/actions/runs/7134600634/job/19429741481)

bryanjnelson: Ah yes.  Then for now just eliminate the EdgeDB additions completely.  We'll attack that in a separate PR altogether.

adam-soltech: That also breaks the build. I sent the commit without edgedb and broken build to the dev channel

CarsonF:
`isUsed` seems like the least offensive prefix. I wonder if `isUsedWithConsultants` would be better than `ForConsulting`?

bryanjnelson:
> `isUsed` seems like the least offensive prefix. I wonder if `isUsedWithConsultants` would be better than `ForConsulting`?

If the UI is going to say "for consulting" then I'd just as soon match that.  Not much is gained otherwise imo.

CarsonF:
I guess that's a question for product.

adam-soltech:
> > `isUsed` seems like the least offensive prefix. I wonder if `isUsedWithConsultants` would be better than `ForConsulting`?
> 
> If the UI is going to say "for consulting" then I'd just as soon match that. Not much is gained otherwise imo.

I agree but will run it by product in standup tomorrow

dbschema/migrations/00033.edgeql
bryanjnelson: You'll still want to take both the .edgql and .esdl files out of this PR.  Edge stuff is being done separately.

sethmcknight:

PR #2985, opened 2023-12-14
[EdgeDB] Exception Handling/Formatting

CarsonF:
This removes the boilerplate of wrapping db errors 1) to convert to generic failures to nicer error messages, 2) convert exclusivity violations to duplicate input errors.

Transactions will now also retry retryable errors, even when they are wrapped in other ones. Not as useful now, but doesn't hurt anything and matches what we do with Neo4j.

Fixed auth not throwing `NoSession`, when it should be.

Have a custom `ExclusivityViolationError` error class now with can be used to programmatically check the object FQN & property name that failed. As stated above, this is converted to a `DuplicateException` on the way out.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5688525797) by [Unito](https://www.unito.io)

### DIFF
--- a/src/common/exceptions/duplicate.exception.ts
+++ b/src/common/exceptions/duplicate.exception.ts
@@ -1,3 +1,10 @@
+import { ArgumentsHost } from '@nestjs/common';
+import {
+  GqlContextType as ContextKey,
+  GqlExecutionContext,
+} from '@nestjs/graphql';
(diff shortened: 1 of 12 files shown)


bryanjnelson (approved):
A yuge step forward in simplicity.

PR #2986, opened 2023-12-14
[EdgeDB] Abstract Common & DTO Repositories

CarsonF:
This eases the migration to EdgeDB repositories by providing some of the common functions that were declared on the neo4j side.
Additionally it provides all of these common methods out of the box: `readOne, readMany, list, create, update, delete`.

Usage:
```ts
class MyRepo extends RepoFor(Foo).withDefaults() {}
```
If one of the default methods signature doesn't match what you want you can customize it.
```ts
class MyRepo extends RepoFor(Foo).customize(cls => {
  return class extends cls {
    async create(color: string) {
      ...
    }
  }
}) {}
```
Obviously more additional methods can be added on as well, like the User Repo already does.

By default the _hydrated_ object is `*` aka `select User {*}`. This can be customized with a `hydrate` option, and then later accessed via class prop. It has to be like this for TS generics to work.
```ts
class UserRepo extends RepoFor(User, {
  hydrate: (user) => ({
    ...user['*'],
    educationIds: user.education.id,
    // etc
  })
}).withDefaults() {

  listActive() {
    const query = e.select(e.User, user => ({
      ...this.hydrate(user),
      filter: e.op(user.status, '=', e.User.Status.Active),
    });
    return this.db.run(query);
  }

}
```

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5688686531) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/project/project-member/project-member.repository.ts
+++ b/src/components/project/project-member/project-member.repository.ts
@@ -117,7 +117,9 @@ export class ProjectMemberRepository extends DtoRepository<
           node('user', 'User'),
         ])
         .subQuery('user', (sub) =>
-          sub.with('user as node').apply(this.users.hydrate(session.userId)),
+          sub
+            .with('user as node')
+            .apply(this.users.hydrateAsNeo4j(session.userId)),
         )
         .return<{ dto: UnsecuredDto<ProjectMember> }>(
           merge('props', { user: 'dto' }).as('dto'),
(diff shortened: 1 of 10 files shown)


bryanjnelson (approved):
🙌🏼  Nice work, @CarsonF!

PR #2990, opened 2023-12-21
Add type filtering in `PeriodicReportRepository.list` query

bryanjnelson:
Addresses [Monday Task 0915](https://seed-company-squad.monday.com/boards/3451697530/pulses/5731437304)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5731531635) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/periodic-report/periodic-report.repository.ts
+++ b/src/components/periodic-report/periodic-report.repository.ts
@@ -182,7 +182,7 @@ export class PeriodicReportRepository extends DtoRepository<
     const filters = { type, parent, start, end };
     const result = await this.db
       .query()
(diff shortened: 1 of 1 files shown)


bryanjnelson:
@CarsonF - This is likely not the best way to fix this issue, but it was a big enough deal that we needed to get something out there.  We can fix it more cleanly later.  It appears something is not working correctly with the `type` clause on the `filter.builder` function, but we couldn't figure it out.

atGit2021 (approved):
LGTM!

CarsonF (approved):

Issue #612, opened 2020-06-26
Project.deptId (departmentId) needs to be calculated and set when project status/step changes to 'Pending Finance Confirmation'


CarsonF:
Rename the property to `departmentId` while you're at it.

Department ID is a numeric ID that's auto incrementing. We treat it as a string here though because that's how we treat all IDs.
It doesn't just auto increment though because it has other logic it follows. Like year I think. We need to confirm with @sethmcknight @kjones140 or finance. We need to support this ID for new projects for finance until we can migrate them to our new IDs.

candleshine:
This ticket is blocked until we have the scheme by which departmentId is coded; what is the algorithm for nextDepartmentId.

I imagine there's a cypher query for lastDeptId created. 
then a formula for the next one
then a way to move the pointer to the newly created One.  We need a lock for concurrent creates, hard to imagine in our app, but possible.

sethmcknight:
Aaahhh Department ID's... where to begin.

Ok so a Department ID is a five digit number where the first digit corresponds to a specific Funding Account and the other four digits are a unique identifier within that Funding Account.

The Funding Account that a Project should be assigned to is based on the Location assigned to the Project. Each Location is assigned to one of 6 Funding Accounts. Each Funding Account has a unique number defined in the Funding Account data.

Within the Funding Account's range of Department ID's we would start with attempting to assign X0000 as the Project's Department ID. If X0000 is already assigned to a Project we iterate through until we find the first available Department ID within that Funding Account, in this example let's assume it is X0123.

If y'all want to we could then store X0124 as the next Department ID to try. However, because of the way these ID's used to be assigned, we are not guaranteed that X0124 - X9999 are unassigned. The next time that we need to assign a Department ID to a Project with the Funding Account of X, we will need to check if X0124 is available or if we need to try X0125, etc until we find the next number that has not been assigned.

Does that make sense @CarsonF & @candleshine?

The Financial Accounts should be in the database but here they are for your reference. Each Location has been linked to one of these in data (Incidentally @CarsonF we also need to surface the Location's Funding Account on the Language Page)

Asia - South (4)
Asia - Islands (3)
Pacific (7)
Africa (2)
Americas, Eurasia (5)
Asia - Mainland (1)

CarsonF:
I'm glad I have you to remember all this stuff @sethmcknight ❤️ 

---

> However, because of the way these ID's used to be assigned, we are not guaranteed that X0124 - X9999 are unassigned.

[...]

PR #3023, opened 2024-01-04
[EdgeDB] departmentId

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5786036475) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -10,7 +10,11 @@ module default {
       default := Sensitivity.High;
     };
     
-    departmentId: str;
+    departmentId: int32 {
+      constraint exclusive;
+      constraint min_value(10000);
+      constraint max_value(99999);
+    };
     
     required step: Project::Step {
       default := Project::Step.EarlyConversations;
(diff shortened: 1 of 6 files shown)


dbschema/project.esdl
```diff
@@ -44,9 +69,17 @@ module default {
 #     multi link engagements := .<project[is Engagement];
     property engagementTotal := count(.<project[is Engagement]);
     
-#       link primaryLocation: Location;
-#       link marketingLocation: Location;
-#       link fieldRegion: FieldRegion;
+    primaryLocation: Location;
+    trigger enforceFundingAccount after update for each do (
+      assert(
+        any(__new__.primaryLocation.fundingAccount.accountNumber > 0)
```
CarsonF, line 75: This `any()` converts null/empty set to false instead of ignoring.

CarsonF:
History in #612. Re-reading that, I said that it should be a string because it's an ID, like all of our other IDs. I thought it was because of a leading zero formatting concern, but I think I was mixing that up with `registryOfDialects`. So maybe I should convert this back to a str...but I'm not really convinced. An int seems easier for this. We can always cast to str in the query too. Thoughts?

bryanjnelson (approved):
:shipit:

PR #3039, opened 2024-01-09
[EdgeDB] Build out scripture collection schema, for easier querying & inspection

CarsonF:
There's a trade off here in simplicity of app code using this type vs information available.
I favored having information available, aka reading/select queries.
I'm exploring abstractions in app code currently to support this schema while reducing LoC.

- ✅ The `labels` here allow easy inspection on what the scripture is during adhoc exploration. They will be computed on write by the app.
- ✅ The `start/end.book/chapter/verse` fields allow the API to serve this data without having to compute from verse IDs. Good for reads.
- ✅ The `verseId` supports answering complex analytical questions. i.e. Does this engagement/product include _Matthew 1_?

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5822486982) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/producible.esdl
+++ b/dbschema/producible.esdl
@@ -4,7 +4,20 @@ module default {
       delegated constraint exclusive;
     };
     
-    scriptureReferences: multirange<int32>;
+    scripture: Scripture::Collection {
+      on source delete delete target if orphan;
+      # https://github.com/edgedb/edgedb/issues/5827
+      # rewrite insert, update using (
+      #   if exists .scripture.verses then .scripture else <Scripture::Collection>{}
+      # );
+    }
+    # Enforce no empty collections for this type's use-case. Use null/empty-set instead.
+    trigger denyEmptyScriptureCollection after insert, update for each do (
+      assert(
+        exists __new__.scripture.verses,
+        message := "`Producible.scripture` should have a `Scripture::Collection` with verses or be null/empty-set"
+      )
+    );
   }
   
   type EthnoArt extending Producible;
--- a/dbschema/scripture.esdl
+++ b/dbschema/scripture.esdl
@@ -1,0 +1,54 @@
+module Scripture {
+  type Collection {
+    required label: str {
+      readonly := true;
+    }
+    # Want to allow optional for now for DerivativeScriptureProduct's scripture override use-case
+    multi verses: VerseRange {
+      readonly := true;
+      on source delete delete target if orphan;
+    }
+    ids := multirange(array_agg(.verses.ids));
+  }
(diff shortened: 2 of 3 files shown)


dbschema/producible.esdl
```diff
@@ -4,7 +4,20 @@ module default {
       delegated constraint exclusive;
     };
     
-    scriptureReferences: multirange<int32>;
+    scripture: Scripture::Collection {
```
CarsonF, line 7: Renamed so that queries can do something like
```edgeql
select Film {
  scriptureReferences := (select .scripture.verses {
    `start`: { book, chapter, verse },
    `end`: { book, chapter, verse }
  })
}
```
This free shape being different than a `Scripture::Collection` means the same field/column name cannot be used.

dbschema/producible.esdl
```diff
@@ -4,7 +4,20 @@ module default {
...
+      on source delete delete target if orphan;
+      # https://github.com/edgedb/edgedb/issues/5827
+      # rewrite insert, update using (
+      #   if exists .scripture.verses then .scripture else <Scripture::Collection>{}
+      # );
```
CarsonF, line 12: This would be my ideal, but since it's bugged I used the trigger below to at least error on empty scripture collections.

DerivativeScriptureProducts have the option to override the scripture defined on their producible.
I wanted to leave scripture collections empty-able, at least for now, to leave the option to support an empty collection serving as an override meaning the product produces no scripture, even though it's attached to a producible that does.

There's probably a cleaner way, but I think products will be re-thought shortly after migration, so I'm trying not to change too much right now.

dbschema/scripture.esdl
```diff
@@ -0,0 +1,48 @@
...
+  }
+  
+  type Verse {
+    label := .book ++ " " ++ <str>.chapter ++ ":" ++ <str>.verse;
+    required book: str {
+      readonly := true;
+    }
+    required chapter: int16 {
+      readonly := true;
+      constraint min_value(1);
```
bryanjnelson, line 36: Maybe a max constraint at 150 (for Psalms)?

dbschema/scripture.esdl
```diff
@@ -0,0 +1,48 @@
...
+    }
+    required verse: int16 {
+      readonly := true;
+      constraint min_value(1);
```
bryanjnelson, line 40: Maybe a max constraint at 176 (for Psalm 119)?

bryanjnelson (approved):
Forward!  🚢

PR #3042, opened 2024-01-10
Project sorting status, step, engagement count

adam-soltech:
[ticket](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5675458057?filter=XQAAAALKAAAAAAAAAABBqQqHk62Wwa8yt9cbbXI48HpL2ei1XixEiNVOhiWmf4IrTn4lrcVxz0d8EzlqDpxL1dc7i9JjI1nNW0HEFgqaXhsDsHWTHjMEE6-qlUKWgWTaunFWUdBXUlo9GrL7-Dxs3yXtVsCpk7bU6tq4ASH03XfQsurYW6oz6SfWXIHjSrPNZIVE3__87GiYdcf5oUuHV6IpVDC18IDUOyTd3hSlyn4xwV6TGJGJ____cnOAAA)

This adds sorting on projects entity for status (by status type rather than alphabetically) and engagements (by engagement count)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5825418833) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/project/dto/project.dto.ts
+++ b/src/components/project/dto/project.dto.ts
@@ -4,12 +4,14 @@ import { stripIndent } from 'common-tags';
 import { DateTime } from 'luxon';
 import { keys as keysOf } from 'ts-transformer-keys';
 import { MergeExclusive } from 'type-fest';
+import { sortingForEnumIndex } from '~/core/database/query';
 import { abstractType, e } from '~/core/edgedb';
 import { RegisterResource } from '~/core/resources';
 import {
   DateInterval,
   DateTimeField,
   DbLabel,
+  DbSort,
   DbUnique,
   ID,
   IntersectionType,
@@ -44,7 +46,7 @@ import { Postable } from '../../post/dto';
 import { ProjectChangeRequest } from '../../project-change-request/dto';
 import { ProjectMember } from '../project-member/dto';
 import { ProjectStatus } from './project-status.enum';
-import { SecuredProjectStep } from './project-step.enum';
+import { ProjectStep, SecuredProjectStep } from './project-step.enum';
 import { ProjectType } from './project-type.enum';
 
 type AnyProject = MergeExclusive<TranslationProject, InternshipProject>;
@@ -114,10 +116,12 @@ class Project extends Interfaces {
     middleware: [parentIdMiddleware],
   })
   @DbLabel('ProjectStep')
+  @DbSort(sortingForEnumIndex(ProjectStep))
   readonly step: SecuredProjectStep;
 
   @Field(() => ProjectStatus)
   @DbLabel('ProjectStatus')
+  @DbSort(sortingForEnumIndex(ProjectStatus))
   readonly status: ProjectStatus;
 
   readonly primaryLocation: Secured<ID | null>;
--- a/src/components/project/project.repository.ts
+++ b/src/components/project/project.repository.ts
@@ -286,6 +286,14 @@ export class ProjectRepository extends CommonRepository {
             query
               .apply(matchProjectSens('node'))
               .return<{ sortValue: string }>('sensitivity as sortValue'),
+          engagements: (query) =>
+            query
+              .match([
+                node('node'),
+                relation('out', '', 'engagement'),
+                node('engagement', 'LanguageEngagement'),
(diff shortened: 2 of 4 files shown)


dbschema/default.esdl
```diff
@@ -0,0 +1,3 @@
+module default {
```
CarsonF, line 1: You can delete this file

src/components/project/dto/project.dto.ts
```diff
@@ -77,7 +79,7 @@ export const resolveProjectType = (val: Pick<AnyProject, 'type'>) =>
 })
 class Project extends Interfaces {
   static readonly DB = abstractType(e.Project);
-  static readonly Props: string[] = keysOf<Project>();
+  static readonly Props: string[] = [...keysOf<Project>(), 'projectStatus'];
```
CarsonF, line 82: Not a canonical prop, revert

src/core/database/query/lists.ts
CarsonF, line 1: This file can be reverted, please.

src/core/database/query/match-project-based-props.ts
```diff
@@ -187,3 +187,23 @@ export const rankSens = (variable: string) => oneLine`
...
+        .return('count(engagement) as engagementTotal'),
+    );
+};
+
+export const getStatus = (nodeName: string) => {
+  return (query: Query) =>
+    query.subQuery(nodeName, (sub) =>
+      sub.match([node(nodeName)]).return(nodeName + '.status as projectStatus'),
+    );
+};
```
CarsonF, line 209: Not used
```suggestion
```

src/core/database/query/match-project-based-props.ts
```diff
@@ -187,3 +187,23 @@ export const rankSens = (variable: string) => oneLine`
...
+    query.subQuery(nodeName, (sub) =>
+      sub
+        .match([
+          node(nodeName),
+          relation('out', '', 'engagement'),
+          node('engagement', 'LanguageEngagement'),
+        ])
+        .return('count(engagement) as engagementTotal'),
+    );
+};
```
CarsonF, line 202: I would probably just inline this in the sorting spot. I don't see too many use cases for this, and it's not abstracting that much.
```ts
engagements: (query) =>
  query
    .match([
      node('node'),
      relation('out', '', 'engagement'),
      node('engagement', 'LanguageEngagement'),
    ])
    .return<{ sortValue: number }>('count(engagement) as sortValue'),
```

src/components/project/project.repository.ts
```diff
@@ -282,10 +282,10 @@ export class ProjectRepository extends CommonRepository {
       .apply(this.privileges.for(session, IProject).filterToReadable())
       .apply(
         sorting(IProject, input, {
-          sensitivity: (query) =>
+          engagements: (query) =>
             query
-              .apply(matchProjectSens('node'))
-              .return<{ sortValue: string }>('sensitivity as sortValue'),
```
CarsonF, line 288: We still need to keep this custom sorting for sensitivity. However, I'm realizing this doesn't actually sort correctly.
We need to call convert to a sens rank
```ts
.return<{ sortValue: string }>(`${rankSens('sensitivity')} as sortValue`),
```

CarsonF (changes requested):

src/core/database/query/lists.ts
adam-soltech: right - thanks for catching

src/components/project/project.repository.ts
CarsonF: I was actually wrong here.
This query just customized how the sensitivity value was retrieved.
The [field still had a DbSort transformer](https://github.com/SeedCompany/cord-api-v3/blob/f67c7cc8f7c9e121b76852071f7aa013e4b302ca/src/common/sensitivity.enum.ts#L18-L18) which was applied in `sorting()`'s next step.
So no change actually needed here.

CarsonF (approved):
Thanks @adam-soltech!

I just moved the enum index sorting to a helper function, since it would be valuable in any enum sorting.

PR #3043, opened 2024-01-10
[EdgeDB] Meaningful seed execution logs / Remove placeholder entries

CarsonF:
Now the current scripts communicate what was added if any.
Screenshot 2024-01-10 at 8 54 07 AM


bryanjnelson (approved):
Great idea, @CarsonF!

PR #3044, opened 2024-01-11
0701 default marketing region migration

willdch:
https://seed-company-squad.monday.com/boards/3451697530/pulses/5186744629

Created migration file for defaultMarketingRegion relationship.  
* There is one country in the DB with a name of "Not Applicable" which has a fieldRegion name of "any".  Not sure if we handled this use case correctly.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5842212514) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/location/migrations/default-marketing-region.migration.ts
+++ b/src/components/location/migrations/default-marketing-region.migration.ts
@@ -1,0 +1,101 @@
+import { node, relation } from 'cypher-query-builder';
+import { ID, IdOf } from '~/common';
+import { BaseMigration, Migration } from '~/core/database';
+import { ACTIVE } from '~/core/database/query';
+import { Location } from '../dto';
+import { UpdateLocation } from '../dto/update-location.dto';
+import { LocationService } from '../location.service';
+
+@Migration('2024-02-06T06:22:15')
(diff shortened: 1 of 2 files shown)


src/components/location/migrations/default-marketing-region.migration.ts
```diff
@@ -0,0 +1,101 @@
...
+      'Asia - Mainland': 'Asia',
+      'Asia - South': 'Asia',
+      any: 'Not Specified',
+    };
+
+    const marketingRegionIdMap: { [key: string]: IdOf<Location> } = {};
+
+    mRegions.forEach((m) => {
+      marketingRegionIdMap[m.locationName] = m.id;
+    });
```
bryanjnelson, line 79: @CarsonF - @willdch, @atGit2021, and I were trying to figure out how to eliminate these lines of code by simply returning this map structure directly as the structure of `mRegions` from the cypher query above on lines 16-33?

CarsonF: Not worth it. Much easier in app code, and no diff in performance.

PR #3046, opened 2024-01-12
[EdgeDB] Posts & Embedded Schema

CarsonF:
# Embedded Opener
I wanted a single thing we could use for "dynamic parents". This is now represented as an `Embedded` object which has a `container` link to another `Resource`.
```edgeql
abstract type Embedded {
  required single link container: default::Resource;
}
```
I wanted to avoid using word "parent" as the file tree schema will have "parents" but these aren't the same thing.
It's a small abstraction, but I'll show why this could be helpful further down.

# Post

I used Posts as a concrete type to prove out the embedded idea. The only hard thing here is the polymorphism of their "parent/container". Well and the RichText JSON scalar which I'll do in a separate PR.
This polymorphism made it hard for current app queries to answer authorization questions.
Is this post attached-to/contained-in a high sensitivity project? As discussed ad nauseam, determining sensitivity in the previous DB was very hard and unique for each object type.

Well based on the previous work done for `ProjectContext` here in EdgeDB land, `Post` can easily check for project context aware objects.
```edgeql
select Post {
  sensitivity := .container[is Project::ContextAware].sensitivity
}
```
This works across our different types as well. Projects, Languages, Partners, all compute this value differently, but Post doesn't need to worry about it.

Naturally I put this as a computed in the schema, so queries (and UI explorer) can just
```edgeql
select Post {*}
```
And see if a post has a sensitivity and what it is.

Currently the app code, as to pull the parent/container for each post and confirm permissions before allowing access to do something with the post.
So this change (along with the to be implemented access policies), should allow the app code to seamlessly query posts without having to do a bunch of work with the parent.

# Embedded Closer

As demonstrated above, the `ProjectContext` work is proving to make things much simpler when dealing with polymorphism. "If X is aware, grab its sensitivity".
But there's another case, what "if X isn't aware, but it has a container that could be".

This is what `Embedded` aims to answer. Though, I suspect it will be less useful than `ProjectContext`.

Perhaps we could have a list of Resources, and we wanted to get its own sensitivity or its container's sensitivity, if it exists.
`Embedded` gives us the single `container` property we can query.
```edgeql
select Resource {
  sensitivity := (
    [is Project::ContextAware].sensitivity ??
    [is Mixin::Embedded].container[is Project::ContextAware].sensitivity
  )
}
```
Kinda clunky, but it allows us to do it without enumerating every concrete type that could be embedded (i.e. Posts, Comments, Files, Media).

## Example Pic
Screenshot 2024-01-12 at 8 58 08 AM

Here we see all different kinds of resources, all getting sensitivity in a unique way.
Now including Posts, which have no explicit schema references to project contexts.

## No Recursion

One caveat, is that this is only one level deep. We can't recurse yet, though this is on the roadmap for EdgeDB. So a Post can't be embedded into another Post.

### DIFF
--- a/dbschema/language.esdl
+++ b/dbschema/language.esdl
@@ -1,5 +1,5 @@
 module default {
-  type Language extending Resource, Project::ContextAware, Mixin::Named, Mixin::Pinnable, Mixin::Taggable {
+  type Language extending Mixin::Postable, Resource, Project::ContextAware, Mixin::Named, Mixin::Pinnable, Mixin::Taggable {
     required displayName: str {
       default := .name;
     }
--- a/dbschema/migrations/00038.edgeql
+++ b/dbschema/migrations/00038.edgeql
@@ -1,0 +1,87 @@
+CREATE MIGRATION m1wdw2srbtq3ondtyuhcobwgcbdawourtp4vcapedti3kca2wh46dq
+    ONTO m1xzy4deu3w7kduilieoiydxysopymo7skj4sftggjen4vi3og6kaq
+{
+  CREATE MODULE Post IF NOT EXISTS;
+  CREATE ABSTRACT TYPE Mixin::Embedded {
+      CREATE REQUIRED SINGLE LINK container: default::Resource;
+  };
+  CREATE SCALAR TYPE Post::Shareability EXTENDING enum<Membership, Internal, AskToShareExternally, External>;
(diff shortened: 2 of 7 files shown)


bryanjnelson (approved):
More excellent foundation.  Thanks @CarsonF!

PR #3047, opened 2024-01-16
0703 add project.marketing country

willdch:
https://seed-company-squad.monday.com/boards/3451697530/pulses/5186745678
https://seed-company-squad.monday.com/boards/3451697530/pulses/5814888900
https://seed-company-squad.monday.com/boards/5989610236/pulses/6080997031
https://seed-company-squad.monday.com/boards/5989610236/pulses/6080995756

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5869158536) by [Unito](https://www.unito.io)

API PR: https://github.com/SeedCompany/cord-api-v3/pull/2965

PR #3048, opened 2024-01-16
Added marketingRegion

willdch:
https://seed-company-squad.monday.com/boards/3451697530/pulses/5186746393

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5870648616) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/project/dto/project.dto.ts
+++ b/src/components/project/dto/project.dto.ts
@@ -131,6 +131,7 @@ class Project extends Interfaces {
 
   readonly marketingRegionOverride: Secured<IdOf<Location> | null>;
   readonly fieldRegion: Secured<ID | null>;
+  readonly marketingRegion: Secured<IdOf<Location> | null>;
 
   readonly owningOrganization: Secured<ID | null>;
 
--- a/src/components/project/dto/create-project.dto.ts
+++ b/src/components/project/dto/create-project.dto.ts
@@ -60,6 +60,12 @@ export abstract class CreateProject {
   })
   readonly fieldRegionId?: ID;
 
+  @IdField({
+    description: 'A marketing region ID',
+    nullable: true,
+  })
+  readonly marketingRegionId?: IdOf<Location> | null;
+
   @DateField({ nullable: true })
   readonly mouStart?: CalendarDate;
 
--- a/src/components/project/dto/update-project.dto.ts
+++ b/src/components/project/dto/update-project.dto.ts
@@ -52,6 +52,12 @@ export abstract class UpdateProject {
   })
   readonly fieldRegionId?: ID | null;
 
+  @IdField({
+    description: 'A marketing region ID',
+    nullable: true,
+  })
+  readonly marketingRegionId?: IdOf<Location> | null;
+
   @DateField({ nullable: true })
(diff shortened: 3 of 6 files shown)


CarsonF (changes requested):
This should be a computed. It's not stored in the database or updated in a mutation. It's just a helper calculation.

```ts
Project.marketingRegion =>
	(.marketingRegionOverride ?? .marketingCountry.marketingRegion)
```

atGit2021:
> This should be a computed. It's not stored in the database or updated in a mutation. It's just a helper calculation.
> 
> ```ts
> Project.marketingRegion =>
> 	(.marketingRegionOverride ?? .marketingCountry.marketingRegion)
> ```

@CarsonF, do you recommend that we compute the value inside of the hydrate function within the project repository?  I assume we keep the marketingRegion property in the update DTO in order to pass down the value that a user may select to override to, but do we need it in the create DTO?  Let us know what you are thinking.

CarsonF:
> do you recommend that we compute the value inside of the hydrate function within the project repository? 

That sounds like the easiest thing to do.

> I assume we keep the marketingRegion property in the update DTO in order to pass down the value that a user may select to override to, but do we need it in the create DTO?

No the create & update DTOs should not reference this property. It is computed/calculated so where would that user input save to? It doesn't make sense.

src/components/project/dto/project.dto.ts
```diff
@@ -131,6 +131,7 @@ class Project extends Interfaces {
 
   readonly marketingRegionOverride: Secured<IdOf<Location> | null>;
   readonly fieldRegion: Secured<ID | null>;
+  readonly marketingRegion: Secured<IdOf<Location> | null>;
```
CarsonF, line 134:
```suggestion
  @Calculated()
  readonly marketingRegion: Secured<IdOf<Location> | null>;
```
This let's the permission system know that edits are not applicable.

Also seen here:
https://github.com/SeedCompany/cord-api-v3/blob/e867d3f50f37fb18c670aef0cd01462b024496c4/src/components/engagement/dto/engagement.dto.ts#L96-L106

atGit2021:
> > do you recommend that we compute the value inside of the hydrate function within the project repository?
> 
> That sounds like the easiest thing to do.
> 
> > I assume we keep the marketingRegion property in the update DTO in order to pass down the value that a user may select to override to, but do we need it in the create DTO?
> 
> No the create & update DTOs should not reference this property. It is computed/calculated so where would that user input save to? It doesn't make sense.

So when a user clicks on the new marketingRegion button in the UI and sets the value to something new, we should use the marketingRegionOverride property to pass that value down to the API and update the override property?

CarsonF:
> So when a user clicks on the new marketingRegion button in the UI and sets the value to something new, we should use the marketingRegionOverride property to pass that value down to the API and update the override property?

Correct.

willdch:
Closing in favor of [PR 3047](https://github.com/SeedCompany/cord-api-v3/pull/3047).

- After creating this branch I realized I needed to have access to the `marketingCountry` property on 3047 in order for the coalesce to work. 
- Added the same changes as a commit on 3047's PR.

PR #3050, opened 2024-01-17
[EdgeDB] RichText Scalar

CarsonF:
This declares a `RichText` scalar as a subtype of `json`.
We convert that to our `RichTextDocument` type & instance in app code (thanks to #3049)

I guess it's worth noting that there are some code paths, that don't account for this specific subtype, but uses the plain `json` logic.
For example, converting these documents to EdgeQL literals still does 
```ts
`to_json(${JSON.stringify(value)})`
```
This is actually fine, since our RichTextDocuments serialize to JSON as expected.
Just felt like it was worth calling out that not all code paths use this specific subtype logic.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5875143343) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/common.esdl
+++ b/dbschema/common.esdl
@@ -15,4 +15,6 @@ module default {
        )
     )
   );
+  
+  scalar type RichText extending json;
 }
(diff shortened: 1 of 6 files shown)


bryanjnelson (approved):
🚀

PR #3053, opened 2024-01-18
[EdgeDB] FieldZone Queries

CarsonF:
Actual changes for FieldZones that should be copy/pasted to other modules:
https://github.com/SeedCompany/cord-api-v3/pull/3053/files/f531e474c19f8c53efc08e9814a7ac567abbd097..47e725a217ed4c5c1f7b6d51752828e022eba56e

### DIFF
--- a/src/components/field-zone/dto/field-zone.dto.ts
+++ b/src/components/field-zone/dto/field-zone.dto.ts
@@ -1,10 +1,9 @@
 import { ObjectType } from '@nestjs/graphql';
 import { keys as keysOf } from 'ts-transformer-keys';
 import { e } from '~/core/edgedb';
-import { RegisterResource } from '~/core/resources';
+import { LinkTo, RegisterResource } from '~/core/resources';
 import {
   DbUnique,
-  ID,
   NameField,
   Resource,
   Secured,
@@ -26,7 +25,7 @@ export class FieldZone extends Resource {
   @DbUnique()
   readonly name: SecuredString;
 
-  readonly director: Secured<ID>;
+  readonly director: Secured<LinkTo<'User'>>;
 }
 
 @ObjectType({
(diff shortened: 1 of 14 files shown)


src/components/field-zone/dto/field-zone.dto.ts
```diff
@@ -26,7 +25,7 @@ export class FieldZone extends Resource {
   @DbUnique()
   readonly name: SecuredString;
 
-  readonly director: Secured<ID>;
+  readonly director: Secured<StaticRef<'User'>>;
```
CarsonF, line 28: This is required to allow EdgeDB to return this shape.

This shape is a subset of a User (`{ id: ID }`), whereas previously, a `User`
could not become an `ID` when selecting.
```edgeql
select FieldZone {
  director := .director.id # error
}
```
Luckily this seems to be a pretty easy change to make, and Neo4j can easily accommodate.

CarsonF: Idk if `StaticRef` is the best name. It makes sense when put next to `DynamicRef`.
But the latter is rarely used, where as the former will be used a lot.
I thought about just calling it `Ref`. It's pretty generic though, and other libraries have their own "ref" concepts which could lead to ambiguity.
Perhaps another word would better describe the type & identifier combo?

CarsonF: Ok I renamed this to `LinkTo`. I think that's probably the best name we came up with on zoom today. My gut says we should go this route to have the field have the same name rather than trying to follow a convention with an `Id` suffix.

bryanjnelson (approved):
Very nice, @CarsonF.  I particularly like the `LinkTo`, `PolymorphicLinkTo`, and `LinkToUnknown` modifications.  More descriptive & intuitive!

PR #3057, opened 2024-01-19
[EdgeDB] Automatically handle not found ID errors from user input

CarsonF:
Now user input IDs can be passed straight into EdgeDB queries without validating, and still have a proper error returned when not found. No more pre validation or manual conversion needed.
Screenshot 2024-01-19 at 4 42 23 PM

This does require that the query be composed in a way that gives info enough back in the error for us to do the conversion.
Object casts work great for this.
```edgeql
<User><uuid>"..."
```
https://github.com/SeedCompany/cord-api-v3/blob/3cb6a335e0fa27ee21a50d844763b67b7dfba75a/src/components/field-zone/field-zone.edgedb.repository.ts#L20-L20
A manual assertion won't give enough info, so that will still be returned as a server error, which it could rightly be.
```edgeql
assert_exists((select User filter .id = <uuid>"..."))
```
This gives a generic assertion failed error so it won't be properly converted.
Much more verbose, so just use object casts lol.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5895484751) by [Unito](https://www.unito.io)

### DIFF
--- a/src/common/exceptions/duplicate.exception.ts
+++ b/src/common/exceptions/duplicate.exception.ts
@@ -1,8 +1,4 @@
 import { ArgumentsHost } from '@nestjs/common';
-import {
-  GqlContextType as ContextKey,
-  GqlExecutionContext,
-} from '@nestjs/graphql';
 import { lowerCase, upperFirst } from 'lodash';
 import type { ExclusivityViolationError } from '~/core/edgedb';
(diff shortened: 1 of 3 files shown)


bryanjnelson (approved):
🚀

PR #3063, opened 2024-01-24
[EdgeDB] Create `Product` schema

bryanjnelson:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5925717215) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/product.esdl
+++ b/dbschema/product.esdl
@@ -1,6 +1,7 @@
 module default {
   abstract type Product extending Engagement::Child {
-    overloaded engagement: LanguageEngagement;
+    # https://github.com/edgedb/edgedb/issues/6766
+    # overloaded engagement: LanguageEngagement;
     
     scripture: Scripture::Collection {
       on source delete delete target if orphan;
@@ -39,6 +40,14 @@ module default {
   
   type DerivativeScriptureProduct extending Product {
     required produces: default::Producible;
+    
+    overloaded scripture {
+      rewrite insert, update using (
+        # coalescing `??` bugged with links
+        # https://github.com/edgedb/edgedb/issues/6767
+        if exists .scriptureOverride then .scriptureOverride else .produces.scripture
+      );
+    };
     
     scriptureOverride: Scripture::Collection {
       on source delete delete target if orphan;
--- a/dbschema/product.esdl
+++ b/dbschema/product.esdl
@@ -22,7 +22,7 @@ module default {
     # Enforce no empty collections for this type's use-case. Use null/empty-set instead.
     trigger denyEmptyScriptureCollection after insert, update for each do (
       assert(
-        exists __new__.scripture.verses,
+        not exists __new__.scripture or exists __new__.scripture.verses,
         message := "`Product.scripture` should have a `Scripture::Collection` with verses or be null/empty-set"
       )
     );
--- a/dbschema/product.esdl
+++ b/dbschema/product.esdl
@@ -33,8 +33,8 @@ module default {
       on source delete delete target if orphan;
     };
     
-    #TODO - totalVerses: := ??? - add this computed field here after migration
-    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration
+    totalVerses: int16;
+    totalVerseEquivalents: int32;
   }
   
   type DerivativeScriptureProduct extending Product {
@@ -46,8 +46,8 @@ module default {
     
     required composite: bool { default := false };
     
-    #TODO - totalVerses: := ??? - add this computed field here after migration
-    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration  
+    totalVerses: int16;
+    totalVerseEquivalents: int32;
   }
   
   type OtherProduct extending Product {
--- a/dbschema/product.esdl
+++ b/dbschema/product.esdl
@@ -1,0 +1,119 @@
+module default {
+  abstract type Product extending Engagement::Child {
+    overloaded engagement: LanguageEngagement;
+
+    scriptureReferences: Scripture::Collection {
+      on source delete delete target if orphan;
+    };
+
+    multi mediums: Product::Medium;
+    multi purposes: Product::Purpose;
+    multi steps: Product::Step;
+    methodology: Product::Methodology;
+
+    describeCompletion: str;
+    placeholderDescription: str;
+    pnpIndex: int16;
+    progressTarget: int16;
+    progressStepMeasurement: Product::ProgressMeasurement;
+
+    #TODO - category := ??? - add this computed field here after migration
+
+    # Enforce no empty collections for this type's use-case. Use null/empty-set instead.
+    trigger denyEmptyScriptureCollection after insert, update for each do (
+      assert(
+        exists __new__.scriptureReferences.verses,
+        message := "`Product.scriptureReferences` should have a `Scripture::Collection` with verses or be null/empty-set"
+      )
+    );    
+  }
+
+  type DirectScriptureProduct extending Product {
+    unspecifiedScripture: Scripture::UnspecifiedPortion {
+      on source delete delete target if orphan;
+    };
+
+    #TODO - totalVerses: := ??? - add this computed field here after migration
+    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration
+  }
+
+  type DerivativeScriptureProduct extending Product {
+    required produces: default::Producible;
+
+    scriptureReferencesOverride: Scripture::Collection {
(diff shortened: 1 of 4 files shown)


CarsonF (commented):
Looks good so far!

dbschema/product.esdl
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+module Product {
+  type DirectScriptureProduct extending default::Product {
+    unspecifiedScripture: Scripture::Collection;  #TODO - is this needed?  Or of a different type?
+
+    #TODO - also below, should these two be in their own type and inherited?
+    totalVerses: int16;  #TODO - computed?
+    totalVerseEquivalents: int16;  #TODO - computed?
```
CarsonF, line 30: They could be but their calculations are different. 
Direct has the `unspecifiedScripture` to include.
Derivative has the `override` to consider over the `.produces.scripture`.

`totalVerses` could be calculated by DB fairly easily.
```edgeql
range_get_upper() - range_get_lower() # & sum
```

`totalVerseEquivalents` is harder because there's the book factor that the DB doesn't know about.
https://github.com/SeedCompany/cord-api-v3/blob/f8acf9130ada9dfce175787f6bc6a350fcc3217a/src/components/scripture/book-difficulty-factor.ts

If we encoded the difficult enum or factor in our scripture types, this would be possible.

Currently the app calculates this and writes to DB. So possible that we continue down that route for now...

dbschema/product.esdl
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  type DerivativeScriptureProduct extending default::Product {
+    required produces: default::Producible;
+
+    scriptureReferencesOverride: Scripture::Collection;
+    composite: bool;
```
CarsonF, line 37:
```suggestion
    required composite: bool { default := false };
```

dbschema/product.esdl
```diff
@@ -0,0 +1,108 @@
...
+    progressStepMeasurement: Product::ProgressMeasurement;
+
+    #  createdAt: cal::local_date;  #TODO - is this desired here?  Already exists in Mixin::Timestamped?
+
+    #TODO - category := ??? - add this computed field here after migration
+  }
+}
+
+module Product {
+  type DirectScriptureProduct extending default::Product {
```
CarsonF, line 25: I'd keep the products in the default module. Follows suit of the concretes of projects & engagements & producibles.

CarsonF (changes requested):

dbschema/product.esdl
bryanjnelson: Well, let me propose that we just put  the `totalVerseEquivalents` as a `TODO` for now following the "Migrate first" approach.

Which means that we'd only have `totalVerses` in a new subtype, so let's just punt that idea until later as well.

dbschema/product.esdl
```diff
@@ -0,0 +1,108 @@
...
+    unspecifiedScripture: Scripture::Collection;  #TODO - is this needed?  Or of a different type?
```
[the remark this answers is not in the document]
bryanjnelson: So then are you seeing this as being a new type in the schema as well?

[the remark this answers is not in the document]
CarsonF: Yes, in Scripture module

dbschema/product.esdl
CarsonF: Yeah agreed. The app can continue to calculate and save manually to DB. These are one of the first things to be re-worked after migration anyways.

dbschema/scripture.esdl
```diff
@@ -48,4 +48,12 @@ module Scripture {
       constraint expression on (__subject__ >= 0 and __subject__ <= 31101);
     }
   }
+
+  type Portion {
+    required book: str {
+      readonly := true;
+    }
+
+    required totalVerses: int16;
```
CarsonF, line 57: This readonly too?

dbschema/scripture.esdl
```diff
@@ -48,4 +48,12 @@ module Scripture {
       constraint expression on (__subject__ >= 0 and __subject__ <= 31101);
     }
   }
+
+  type Portion {
```
CarsonF, line 52: A portion of scripture could be `Matthew 1-4`. This why I named it `UnspecifiedScripturePortion` in app code.
Maybe `Scripture::UnspecifiedPortion` here?

dbschema/product.esdl
```diff
@@ -0,0 +1,104 @@
...
+    unspecifiedScripture: Scripture::Portion;
+
+    #TODO - totalVerses: := ??? - add this computed field here after migration
+    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration
+  }
+
+  type DerivativeScriptureProduct extending Product {
+    required produces: default::Producible;
+
+    scriptureReferencesOverride: Scripture::Collection;
```
CarsonF, line 31:
```suggestion
    scriptureReferencesOverride: Scripture::Collection {
      on source delete delete target if orphan;
    };
```

dbschema/product.esdl
```diff
@@ -0,0 +1,104 @@
...
+    placeholderDescription: str;
+    pnpIndex: int16;
+    progressTarget: int16;
+    progressStepMeasurement: Product::ProgressMeasurement;
+
+    #TODO - category := ??? - add this computed field here after migration
+  }
+
+  type DirectScriptureProduct extending Product {
+    unspecifiedScripture: Scripture::Portion;
```
CarsonF, line 22:
```suggestion
    unspecifiedScripture: Scripture::Portion {
      on source delete delete target if orphan;
    };
```

dbschema/product.esdl
```diff
@@ -0,0 +1,104 @@
+module default {
+  abstract type Product extending Engagement::Child {
+    overloaded engagement: LanguageEngagement;
+
+    scriptureReferences: Scripture::Collection;
```
CarsonF, line 5:
```suggestion
    scriptureReferences: Scripture::Collection {
      on source delete delete target if orphan;
    };
```

CarsonF (changes requested):
Other than these few things, looks good!

dbschema/scripture.esdl
bryanjnelson: I was wondering what to do there; cool.

dbschema/product.esdl
```diff
@@ -0,0 +1,127 @@
...
+    #TODO - totalVerses: := ??? - add this computed field here after migration
+    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration
+
+    # Enforce no empty collections for this type's use-case. Use null/empty-set instead.
+    trigger denyEmptyScriptureOverrideCollection after insert, update for each do (
+      assert(
+        exists __new__.scriptureReferencesOverride.verses,
+        message := "`Product.scriptureReferencesOverride` should have a `Scripture::Collection` with verses or be null/empty-set"
+      )
+    );
```
CarsonF, line 58: This is actually the one place we do want to allow that
```suggestion
```

dbschema/product.esdl
```diff
@@ -0,0 +1,127 @@
+module default {
+  abstract type Product extending Engagement::Child {
+    overloaded engagement: LanguageEngagement;
+
+    scriptureReferences: Scripture::Collection {
```
CarsonF, line 5: Thinking more, this is actually not here in app code - it's on Producible. We chose here to not extend Producible, and I stand by that.
DirectScriptureProduct - should have this.
DerivativeScriptureProduct - should have this computed from the producible or the override
OtherProduct - it was here via Producible, but really unused.

So really it's just the first two that have it.
I'm pretty sure EdgeQL doesn't like it when one concrete owns the prop and another concrete computes it. So that's something to consider too.

Maybe we leave it here, and then a rewrite to set it on derivative (and trigger for producible update).

dbschema/product.esdl
```diff
@@ -0,0 +1,127 @@
...
+    );    
+  }
+
+  type DirectScriptureProduct extending Product {
+    unspecifiedScripture: Scripture::UnspecifiedPortion {
+      on source delete delete target if orphan;
+    };
+
+    #TODO - totalVerses: := ??? - add this computed field here after migration
+    #TODO - totalVerseEquivalents := ??? - add this computed field here after migration
```
CarsonF, line 37: These do need to be stored now.

CarsonF (commented):
Sorry for the back and forth here. We can sync in the AM.

CarsonF (approved):

PR #3064, opened 2024-01-24
[EdgeDB] Partnership schema

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5926613100) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/partnership.esdl
+++ b/dbschema/partnership.esdl
@@ -1,0 +1,40 @@
+module default {
+
+  type Partnership extending Project::Child {
+    required partner: Partner;
+    organization := .partner.organization;
+    constraint exclusive on ((.project, .partner));
+    agreement: File;
+    required agreementStatus: Partnership::AgreementStatus {
+      default := Partnership::AgreementStatus.NotAttached;
+    };
+    mou: File; 
+    mouStatus: Partnership::AgreementStatus;
+    mouStart := .mouStartOverride ?? .project.mouStart;
+    mouEnd := .mouEndOverride ?? .project.mouEnd;
+    mouStartOverride: cal::local_date;
+    mouEndOverride: cal::local_date;
+    multi types: Partner::Type;
+    financialReportingType: Partnership::FinancialReportingType;
+    required primary: bool {
+      default := false;
+    }
+
+  }
+
+}
+
+module Partnership {
+  scalar type AgreementStatus extending enum<
+    NotAttached,
+    AwaitingSignature,
+    Signed
+  >;
+
+  scalar type FinancialReportingType extending enum<
+    Funded,
+    FieldEngaged,
+    Hybrid
+  >;
+
+}
--- a/dbschema/partnership.esdl
+++ b/dbschema/partnership.esdl
@@ -1,40 +1,41 @@
 module default {
-
   type Partnership extending Project::Child {
     required partner: Partner;
-    organization := .partner.organization;
     constraint exclusive on ((.project, .partner));
-    agreement: File;
-    required agreementStatus: Partnership::AgreementStatus {
-      default := Partnership::AgreementStatus.NotAttached;
+    organization := .partner.organization;
+    
+    required primary: bool {
+      default := false;
     };
-    mou: File; 
-    mouStatus: Partnership::AgreementStatus;
+    multi types: Partner::Type;
+    financialReportingType: Partnership::FinancialReportingType;
+    
     mouStart := .mouStartOverride ?? .project.mouStart;
     mouEnd := .mouEndOverride ?? .project.mouEnd;
     mouStartOverride: cal::local_date;
     mouEndOverride: cal::local_date;
-    multi types: Partner::Type;
-    financialReportingType: Partnership::FinancialReportingType;
-    required primary: bool {
-      default := false;
-    }
-
+    
+    agreement: File;
+    required agreementStatus: Partnership::AgreementStatus {
+      default := Partnership::AgreementStatus.NotAttached;
+    };
+    mou: File;
+    required mouStatus: Partnership::AgreementStatus {
+      default := Partnership::AgreementStatus.NotAttached;
+    };
   }
-
 }
-
+  
 module Partnership {
   scalar type AgreementStatus extending enum<
     NotAttached,
     AwaitingSignature,
     Signed
   >;
-
+  
   scalar type FinancialReportingType extending enum<
     Funded,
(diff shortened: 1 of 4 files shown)


atGit2021:
@bryanjnelson, I left a couple of comments in the file for discussion with you tomorrow.

dbschema/partnership.esdl
```diff
@@ -0,0 +1,38 @@
...
+    AwaitingSignature,
+    Signed
+  >;
+
+  #Confirm this type should be removed from Partner.esdl and when it should be removed
+  scalar type FinancialReportingType extending enum<
+    Funded,
+    FieldEngaged,
+    Hybrid
+  >;
```
bryanjnelson, line 36: You can keep this here and remove it from `Partner` now.  Just make sure you also change any references to it to `Partner::FinancialReportingType`.

bryanjnelson (changes requested):
Wonderful first schema start!

dbschema/partnership.esdl
```diff
@@ -0,0 +1,38 @@
...
+    mouStatus: Partnership::PartnershipAgreementStatus;
+    mouStart := .mouStartOverride ?? .project.mouStart;
+    mouEnd := .mouEndOverride ?? .project.mouEnd;
+    mouStartOverride: cal::local_date;
+    mouEndOverride: cal::local_date;
+    agreement: File;
+    required partner: Partner;
+    required organization: Organization;
+    multi types: Partner::Type;
+    multi financialReportingTypes: Partnership::FinancialReportingType;
```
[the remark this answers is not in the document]
atGit2021: I'm able to select multiple checkboxes for both Partner Types and Financial Reporting Types via the UI.   partner.esdl also contains the following code snippet, but if things have changed, I can update; we can discuss tomorrow. 
` multi types: Partner::Type;
    multi financialReportingTypes: Partner::FinancialReportingType;`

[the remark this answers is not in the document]
bryanjnelson: Well that's true via the UI, but that `multi` relationship comes via the `Partner` object, not the `Partnership` object.  That's why there is already a `multi` on the relationship in the `Partner` schema (which follows the `partner.dto`.  However, the `partnership.dto` shows just a single.

atGit2021:
@bryanjnelson I'm not sure if I should run the edgedb migrate command or some other command to resolve the CI test error.  I remember you saying not to run that command because you can't rollback, so will wait for your guidance before doing anything else.

bryanjnelson (approved):
Great work @atGit2021!  @CarsonF - on to you...I'm sure we missed something!

dbschema/partnership.esdl
```diff
@@ -0,0 +1,34 @@
...
+    Signed
+  >;
+
+  scalar type FinancialReportingType extending enum<
+    Funded,
+    FieldEngaged,
+    Hybrid
+  >;
+
+}
```
CarsonF, line 34: Always end with an empty line, we are not savages 😉

dbschema/partnership.esdl
```diff
@@ -0,0 +1,34 @@
...
+    required partner: Partner;
+    required organization: Organization;
+    multi types: Partner::Type;
+    financialReportingType: Partnership::FinancialReportingType;
+    primary: bool {
+      default := false;
+    }
+  }
+}
+
```
CarsonF, line 20: If you could actually keep the indentation on these empty lines or just add some spacing on these root level ones, it would help me out. My syntax highlighting breaks without them. I know it's dumb. I don't expect it to be this way forever.

dbschema/partnership.esdl
```diff
@@ -0,0 +1,34 @@
+module default {
+  type Partnership extending Project::Child {
+    agreementStatus: Partnership::AgreementStatus;
```
CarsonF, line 3: This status corresponds to the `agreement` file, as does `mouStatus` and `mou`. If you could sort those together that would be great.

The NotAttached is the default value here, so let's use that instead of null.
```suggestion
    required agreementStatus: Partnership::AgreementStatus {
    		default := Partnership::AgreementStatus.NotAttached;
    };
```

dbschema/partnership.esdl
```diff
@@ -0,0 +1,34 @@
...
+    mouStart := .mouStartOverride ?? .project.mouStart;
+    mouEnd := .mouEndOverride ?? .project.mouEnd;
+    mouStartOverride: cal::local_date;
+    mouEndOverride: cal::local_date;
+    agreement: File;
+    required partner: Partner;
+    required organization: Organization;
+    multi types: Partner::Type;
+    financialReportingType: Partnership::FinancialReportingType;
+    primary: bool {
```
CarsonF, line 15:
```suggestion
    required primary: bool {
```
We also do have logic that we should encode here to move this default around, which is currently in app code. That could just be a TODO comment for now.

dbschema/partnership.esdl
```diff
@@ -0,0 +1,34 @@
...
+    agreementStatus: Partnership::AgreementStatus;
+    mou: File; 
+    mouStatus: Partnership::AgreementStatus;
+    mouStart := .mouStartOverride ?? .project.mouStart;
+    mouEnd := .mouEndOverride ?? .project.mouEnd;
+    mouStartOverride: cal::local_date;
+    mouEndOverride: cal::local_date;
+    agreement: File;
+    required partner: Partner;
+    required organization: Organization;
```
CarsonF, line 12: We don't need to have stored forward links for both of these.
```suggestion
    required partner: Partner;
    organization := .partner.organization;
```
I would ask that these move to the top of the block too since this, `partner`, is one of the keys identifying the `Partnership` (`project` is the other).

Along these lines we need this
```esdl
constraint exclusive on ((.project, .partner));
```
So that a partnership can only be created once for each partner and project.

CarsonF (changes requested):

dbschema/partnership.esdl
atGit2021: Is this the logic you are referring to: `const primary = isFirstPartnership ? true : input.primary;`?  I'm not sure how to write the expression on the primary bool property to check if this is the first partnership.

dbschema/partnership.esdl
atGit2021: not sure if the changes I made help, but just added more space.

CarsonF (approved):

PR #3067, opened 2024-01-25
Adds Social Media URL and Website URL to Partners

brentkulwicki:
Backend for ticket [5428606347
](https://seed-company-squad.monday.com/boards/5989610236/pulses/6080411038)


src/common/secured-url.field.ts
CarsonF, line 1: delete

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -49,6 +50,14 @@ export abstract class CreatePartner {
   @NameField({ nullable: true })
   readonly address?: string;
 
+  @Field(() => String, { nullable: true })
```
CarsonF, line 53:
```suggestion
  @UrlField({ nullable: true })
```

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -49,6 +50,14 @@ export abstract class CreatePartner {
...
+  @Transform(({ value: str }) => (str ? externalUrlWithProtocol(str) : null))
+  readonly websiteUrl?: string;
```
CarsonF, line 55: Did we decide against the `urls` array?

src/components/partner/dto/partner.dto.ts
```diff
@@ -79,6 +79,12 @@ export class Partner extends Interfaces {
   @Field()
   readonly address: SecuredString;
 
+  @Field()
+  readonly websiteUrl: SecuredString;
```
CarsonF, line 83:
```suggestion
  readonly websiteUrl: SecuredURL;
```

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -49,6 +50,14 @@ export abstract class UpdatePartner {
   @NameField({ nullable: true })
   readonly address?: string;
 
+  @Field(() => String, { nullable: true })
```
CarsonF, line 53:
```suggestion
  @UrlField({ nullable: true })
```

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -49,6 +50,14 @@ export abstract class UpdatePartner {
...
+  @Transform(({ value: str }) => (str ? externalUrlWithProtocol(str) : null))
```
CarsonF, line 54: Kinda sucks to have this in so many spots. Maybe we move into `UrlField`? Maybe it could be conditional based on the `require_protocol` field?

src/components/partner/dto/update-partner.dto.ts
```diff
@@ -49,6 +50,14 @@ export abstract class UpdatePartner {
...
+  readonly websiteUrl?: string;
```
CarsonF, line 55: For updates, we specifically want to allow `null` type for TS.
`undefined/?` comes from omission from JSON body, which we interrupt (by PATCH convention) to mean "unchanged".
An explicit `null` from input means to clear the current value.
```suggestion
  readonly websiteUrl?: string | null;
```

src/components/partner/partner.repository.ts
```diff
@@ -57,6 +57,8 @@ export class PartnerRepository extends DtoRepository<
       active: input.active,
       startDate: input.startDate ?? CalendarDate.local(),
       address: input.address,
+      websiteUrl: input.websiteUrl ?? '',
+      socialUrl: input.socialUrl ?? '',
```
CarsonF, line 61:
```suggestion
      websiteUrl: input.websiteUrl,
      socialUrl: input.socialUrl,
```
Empty strings are _not_ values :)

We should however add a migration to retroactively add these props to DB.
See [add-type.migration.ts](https://github.com/SeedCompany/cord-api-v3/blob/f9f3d68245076443c620171686c0fe98169a60f0/src/components/organization/migrations/add-type.migration.ts)

CarsonF (changes requested):

src/components/partner/dto/create-partner.dto.ts
brentkulwicki: Yes. I talked with Amy and Sheri. We decided to go with just 1 URL for each item for now and ticket the possibility of doing multiple URLs later.

src/components/partner/partner.repository.ts
brentkulwicki: Believe it or not, I had a migration and then back tracked on that thought. Will re-add

CarsonF: Oof

sethmcknight:

PR #3070, opened 2024-01-29
Validation Injectability

CarsonF:
This allows validators to be injectable. Prepping for future work.

Example:
```ts
@Injectable()
@ValidatorConstraint({ name: 'Foo', async: true })
export class FooConstraint implements ValidatorConstraintInterface {
  constructor(
    // inject anything...
  ) {}

  async validate(value: unknown, args: ValidationArguments) {
    ...
  }
}

const IsFoo = createValidationDecorator(FooConstraint);
// or
const IsFoo = () => ValidateBy(FooConstraint);
```
Obviously the constraint class provided in a module as well.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5954156315) by [Unito](https://www.unito.io)

### DIFF
--- a/src/common/validators/validateBy.ts
+++ b/src/common/validators/validateBy.ts
@@ -1,16 +1,25 @@
-import { FnLike } from '@seedcompany/common';
+import { Type } from '@nestjs/common';
 import {
   registerDecorator,
   ValidationOptions,
   ValidatorConstraintInterface,
 } from 'class-validator';
+import { MergeExclusive } from 'type-fest';
 
-export interface ValidateByOptions {
-  name: string;
-  constraints?: any[];
-  validator: ValidatorConstraintInterface | FnLike;
-  async?: boolean;
-}
+export type ValidateByOptions =
+  | MergeExclusive<
+      {
+        validator: ValidatorConstraintInterface;
+        name: string;
+        async?: boolean;
+        constraints?: any[];
+      },
+      {
+        validator: Type<ValidatorConstraintInterface>;
+        constraints?: any[];
+      }
+    >
+  | Type<ValidatorConstraintInterface>;
 
 export const ValidateBy =
   (
@@ -19,13 +28,11 @@ export const ValidateBy =
   ): PropertyDecorator =>
   (object: Record<string, any>, propertyName: string | symbol) => {
     registerDecorator({
-      name: options.name,
       target: object.constructor,
       propertyName: propertyName as string,
       options: validationOptions,
-      constraints: options.constraints,
-      validator: options.validator,
-      async: options.async,
+      ...options,
+      validator: typeof options === 'function' ? options : options.validator,
     });
   };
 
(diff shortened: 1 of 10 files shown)


src/common/validators/validateBy.ts
```diff
@@ -1,16 +1,25 @@
...
   registerDecorator,
   ValidationOptions,
   ValidatorConstraintInterface,
 } from 'class-validator';
+import { MergeExclusive } from 'type-fest';
 
-export interface ValidateByOptions {
-  name: string;
-  constraints?: any[];
-  validator: ValidatorConstraintInterface | FnLike;
```
CarsonF, line 11: This `FnLike` should actually be a class reference (a constructor), so I converted to that.
Raw functions do not work.
`name` & `async` are given with constraint class decorator, so fixed those types as well.

src/common/validators/validateBy.ts
```diff
@@ -1,16 +1,25 @@
...
+        name: string;
+        async?: boolean;
+        constraints?: any[];
+      },
+      {
+        validator: Type<ValidatorConstraintInterface>;
+        constraints?: any[];
+      }
+    >
+  | Type<ValidatorConstraintInterface>;
```
CarsonF, line 22: This is just a shortcut
```ts
FooConstraint
// same as
{ validator: FooConstraint }
```
Easier, and the only thing unable to be given is the `constraints` list, which we rarely need and poor DX because untyped.

src/core/validation/validation.pipe.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+      return this.moduleRef.get(type);
+    },
+  };
+
+  protected async validate(
+    object: object,
+    validatorOptions?: ValidatorOptions,
+  ) {
+    useContainer(this.containerForLib, { fallback: true });
```
CarsonF, line 32: `class-validator` has the concept of injection with a container, but it's a global.
So here we set/re-set the container to use to our current instance for this `validate()` call.

src/core/validation/validation.pipe.ts
```diff
@@ -0,0 +1,35 @@
...
+      skipMissingProperties: true,
+      exceptionFactory: (es) => new ValidationException(es),
+    });
+  }
+  private readonly containerForLib = {
+    get: (type: any) => {
+      if (type.name === 'CustomConstraint') {
+        // Prototype-less constraints. Null to fall back to default, which just calls constructor once.
+        return null;
+      }
```
CarsonF, line 23: These types are created dynamically created classes, so we don't have a way to reference/provide them in a nestjs module.
We return null here, and have `fallback: true`, so that the library does the default thing of memoized `new type()`.
These are from built in constraints, or ours like this https://github.com/SeedCompany/cord-api-v3/blob/ccd94582658b53c1e8904e7d27e33786683d7dfa/src/common/validators/email.validator.ts#L13-L16

So they don't need to be injected. Actual custom constraint classes, will have a static class ref, that will be provided to a nestjs module, and used here.

I also have `fallbackOnErrors: false` (as default), so we should get errors when asking for a constraint that hasn't be provided/registered.

bryanjnelson (approved):

PR #3071, opened 2024-01-29
[EdgeDB] Aliases for legacy IDs

CarsonF:
This will allow the API/UI to accept our current IDs even after the migration.
My main motivation for this is to not break all references to our URLs that include IDs.

This works with an `Alias` type/table that maps a string to some object.
It assumes all IDs now (after migration) are some form of a UUID.
Therefore, any string that's not a UUID would make a DB call to convert to the target's UUID.
These aliases still have to be globally unique, as our current IDs are.

This currently works transparently for all input fields/args that are of type `ID`, or rather use the [`IdArg`](https://github.com/SeedCompany/cord-api-v3/blob/bfaae760e62d40d4aada19f4b3f8635bacb6f483/src/common/id.arg.ts#L5-L5) or [`IsId`](https://github.com/SeedCompany/cord-api-v3/blob/bfaae760e62d40d4aada19f4b3f8635bacb6f483/src/common/validators/short-id.validator.ts#L13-L13)/[`IdField`](https://github.com/SeedCompany/cord-api-v3/blob/bfaae760e62d40d4aada19f4b3f8635bacb6f483/src/common/id-field.ts#L7-L7) decorators.

Example with an alias called "root":
Screenshot 2024-01-29 at 4 06 37 PM

This doesn't account for using those aliases or legacy IDs in the output results. So any places accepting CORD data, would still have a breaking change.

---

I could see this becoming a user-facing feature in the future hence the generic name `Alias` instead of "legacy ID".
i.e.
```
cordfield.com/projects/adunu-cluster
```

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5954870901) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/alias.esdl
+++ b/dbschema/alias.esdl
@@ -1,0 +1,10 @@
+module default {
+  type Alias {
+    required name: str {
+      constraint exclusive;
+    };
+    required target: Object {
+      on target delete delete source;
+    };
+  }
+}
--- a/src/common/validators/short-id.validator.ts
+++ b/src/common/validators/short-id.validator.ts
@@ -1,28 +1,54 @@
-import { ArgumentMetadata, PipeTransform } from '@nestjs/common';
-import { ValidationOptions } from 'class-validator';
+import { ArgumentMetadata, Injectable, PipeTransform } from '@nestjs/common';
+import {
+  ValidationArguments,
+  ValidationOptions,
+  ValidatorConstraint,
+  ValidatorConstraintInterface,
+} from 'class-validator';
 import { ValidationException } from '~/core/validation';
 import { isValidId } from '../generate-id';
+import { ID } from '../id-field';
 import { ValidateBy } from './validateBy';
 
 export const IsId = (validationOptions?: ValidationOptions) =>
-  ValidateBy(
-    {
-      name: 'IsId',
-      validator: {
-        validate: isValidId,
-        defaultMessage: () =>
-          validationOptions?.each
-            ? 'Each value in $property must be a valid ID'
-            : 'Invalid ID',
-      },
-    },
-    validationOptions,
-  );
+  ValidateBy(ValidIdConstraint, {
+    message: validationOptions?.each
+      ? 'Each value in $property must be a valid ID'
+      : 'Invalid ID',
+    ...validationOptions,
+  });
 
+@Injectable()
+export class IdResolver {
+  async resolve(value: ID): Promise<ID> {
+    return value;
+  }
+}
+
+@Injectable()
+@ValidatorConstraint({ name: 'IsId', async: true })
+export class ValidIdConstraint implements ValidatorConstraintInterface {
+  constructor(private readonly resolver: IdResolver) {}
+
+  async validate(value: unknown, args: ValidationArguments) {
+    if (isValidId(value)) {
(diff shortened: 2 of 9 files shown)


dbschema/alias.esdl
```diff
@@ -0,0 +1,10 @@
+module default {
+  type Alias {
```
CarsonF, line 2: I'm not convinced this should be in the default module., I also don't think we have an existing module that fits this. Thoughts?

src/common/validators/short-id.validator.ts
```diff
@@ -1,28 +1,54 @@
...
+export class IdResolver {
+  async resolve(value: ID): Promise<ID> {
+    return value;
+  }
+}
+
+@Injectable()
+@ValidatorConstraint({ name: 'IsId', async: true })
+export class ValidIdConstraint implements ValidatorConstraintInterface {
+  constructor(private readonly resolver: IdResolver) {}
```
CarsonF, line 31: This is why #3070 is needed

dbschema/alias.esdl
bryanjnelson: How will this be used within other areas of the schema?  Can you present an example?

CarsonF: Well I do think it could be good to keep this as a standalone thing.
I didn't add an `aliases` backlink anywhere in schema.

It could be used to provide a legacy ID for DOMO, but that would take some more work to pass through the API and in every query.
At one point in my dev I thought that this alias could be integrated into the `readMany` queries. I wanted it to be transparent, so that seemed like a not bad way to do it.
But it didn't prove out well, so I iterated to this approach seen in PR.
But the query was something like
```edgeql
with ids := {"carson", "9b84a0ba-bbba-11ee-9de2-c7870f73b88d", "root"},
  items := (
    (select User filter <str>.id in ids)
    union
    (select Alias filter .name in ids).target[is User]
  )
select items {
  id,
  legacyId := (select .<target[is Alias] filter .name in ids limit 1).name ?? <str>.id,
  ...
}
```

----


My point with this comment though was to question whether this was really a top level feature of CORD or rather something of lesser significance and therefore doesn't belong in the `default` module.
Rather `Helper::Alias` or `Util::Alias`. But both of those suck IMO lol.

bryanjnelson: Yeah, that's why I asked.  I was trying to see more how it would be used so I could better think about where it should live.  I was wondering about it being a Mixin, but after seeing this usage that doesn't make sense.  It really is some kind of "util" function.  Whatever we do, please don't use `Helper` for anything.  lol.  It's a pet peeve of mine...helper of what?  I'd rather go `Util::` for sure.  

However, I could almost see it just staying in default.  I'd even consider just putting it in `common.esdl` maybe?

CarsonF: Ok let's just keep it in default for now. Meh I think it's distinct enough to have it's own file. I don't want common to blow up like a util folder.

bryanjnelson (approved):

dbschema/alias.esdl
bryanjnelson: I'm fine with either.  It'll all become more clear as we progress anyway.

PR #3073, opened 2024-02-05
[EdgeDB] `PeriodicReport` schema

andrewmurraydavid:
## Description

In this PR we are adding the EdgeDB schema for the `PeriodicReport`. This work is to be able to move forward with the EdgeDB migration and unblock future work that requires `PeriodicReport` or children types.

In this work we started using the `range` type from EdgeDB for the period a `PeriodicReport` is in and the `start` and `end` are computed fields from the `period` field.

## Checklist
- [x] Add/update tests if needed or N/A


CarsonF:
Just overloaded ProgressReport.engagement as well to be only LanguageEngagements

bryanjnelson:
@andrewmurraydavid - Thanks for your work here!  🚀

PR #3074, opened 2024-02-07
[EdgeDB] Create `Location` queries; refactor/isolate neo4j `Location` implementation details

bryanjnelson:
Addresses [Task 0953](https://seed-company-squad.monday.com/boards/3451697530/pulses/5774989896)


src/components/location/location.service.ts
```diff
@@ -71,30 +62,26 @@ export class LocationService {
...
+  private secure(dto: UnsecuredDto<Location>, session: Session) {
     return this.privileges.for(session, Location).secure(dto);
   }
 
   async update(input: UpdateLocation, session: Session): Promise<Location> {
-    const location = await this.readOne(input.id, session);
+    const location = await this.repo.readOne(input.id);
 
     const changes = this.repo.getActualChanges(location, input);
     this.privileges.for(session, Location, location).verifyChanges(changes);
```
bryanjnelson, line 81: @CarsonF - Do have any insight into what the two errors here are about?

src/components/location/location.service.ts
```diff
@@ -117,7 +104,10 @@ export class LocationService {
...
     const results = await this.repo.list(input, session);
 
-    return await mapListResults(results, (dto) => this.secure(dto, session));
+    return {
+      ...results,
+      items: results.items.map((dto) => this.secure(dto, session)),
+    };
   }
 
   async addLocationToNode(label: string, id: ID, rel: string, locationId: ID) {
```
bryanjnelson, line 113: @CarsonF - Also, we have extra methods here in the service like this one.  If there is going to be some sort of equivalent to these being called here for EdgeDB, then we should leave them.  But if these are also specific to Neo4j, then now could be the time to purge them as well.  The only way I can think to do this would be to remove this method here (and move any salient pieces into it's sister repo method), for example, and have all of it's callers call directly into the repo instead of here?

Thoughts appreciated.

src/components/location/location.service.ts
CarsonF: Fixed here #3076

src/components/location/location.service.ts
CarsonF: We will need these for EdgeDB too. Since actual GQL migrations will call these methods directly for updates, they make sense to keep. (As opposed to it needing to be a piece of another larger query).

Working on a couple type changes to help with this implementation.

CarsonF: #3077

This should get you started:
```ts
const res = this.resources.getByEdgeDB(label);
const node = e.cast(res.db, e.cast(e.uuid, id));
const location = e.cast(e.Location, e.cast(e.uuid, locationId));
const query = e.update(node, () => ({
  set: {
    [rel]: { '+=': location },
  },
}));
```

src/components/location/location.edgedb.repository.ts
```diff
@@ -0,0 +1,70 @@
...
+          type: input.type,
+          isoAlpha3: input.isoAlpha3,
+          fundingAccount:
+            input.fundingAccountId &&
+            e.cast(e.FundingAccount, e.cast(e.uuid, input.fundingAccountId)),
+          defaultFieldRegion:
+            input.defaultFieldRegionId &&
+            e.cast(e.FieldRegion, e.cast(e.uuid, input.defaultFieldRegionId)),
+          mapImage:
+            input.mapImage && e.cast(e.File, e.cast(e.uuid, input.mapImage)),
```
bryanjnelson, line 37: @CarsonF - I'm unsure about how to handle this with the File type?

CarsonF: Yeah TBD. You could leave a TODO until I figure it out.

src/components/location/location.service.ts
bryanjnelson: I'm sorry; I'm not following.  Is that block going to replace what is in `LocationService.addLocationToNode`?

src/components/location/location.edgedb.repository.ts
```diff
@@ -0,0 +1,70 @@
...
+  PaginatedListType,
+  PublicOf,
+  UnsecuredDto,
+} from '~/common';
+import { e, RepoFor } from '~/core/edgedb';
+import { CreateLocation, Location, LocationListInput } from './dto';
+import { LocationRepository } from './location.repository';
+
+@Injectable()
+export class LocationEdgeDBRepository
```
bryanjnelson, line 14: @CarsonF - I think this has to do with me including `mapImage` in the `hydrate` method, but it needs to be there.  How can I handle this?

src/components/location/location.service.ts
CarsonF: That's `addLocationToNode` for the edgedb repository

src/components/location/location.edgedb.repository.ts
CarsonF:
```diff
-       mapImage: false, //TODO - add this to hydrate
+       mapImage: loc.mapImage.id,
```

CarsonF: Ah wait that doesn't work.
We'll need to change away from `DefinedFile` first.

bryanjnelson: Kinda glad it wasn't that easy. lol.  Yeah I was playing around with `DefinedFile` but didn't get anywhere.

src/components/location/location.repository.ts
```diff
@@ -67,7 +67,7 @@ export class LocationRepository extends DtoRepository(Location) {
       throw new ServerException('Failed to create location');
     }
 
-    return { id: result.id, mapImageId };
+    return await this.readOne(result.id);
```
bryanjnelson, line 70: @CarsonF - Not sure what is going on with the error above?

src/components/location/location.service.ts
```diff
@@ -71,30 +62,26 @@ export class LocationService {
...
-    const location = await this.readOne(input.id, session);
+    const location = await this.repo.readOne(input.id);
 
     const changes = this.repo.getActualChanges(location, input);
     this.privileges.for(session, Location, location).verifyChanges(changes);
 
-    await this.repo.update({ id: input.id, ...changes });
-
-    return await this.readOne(input.id, session);
+    const updated = await this.repo.update({ id: input.id, ...changes });
```
bryanjnelson, line 83: @CarsonF - I'm also unsure how to resolve the error above here.

src/components/location/location.service.ts
```diff
@@ -35,28 +33,21 @@ export class LocationService {
...
-        'location.name',
-        'Location with this name already exists.',
-      );
-    }
 
-    const { id, mapImageId } = await this.repo.create(input, session);
+    const dto = await this.repo.create(input, session);
     await this.files.createDefinedFile(
-      mapImageId,
+      dto.mapImage?.id,
```
bryanjnelson, line 39: @CarsonF - I know we discussed this, and you mentioned that this whole thing needs an EdgeDB version or something like that.  Any idea how I can proceed until that happens?

CarsonF: I'm not sure. It needs some more thought in the file layer before referencing here.

The big question is do files exist in DB without any backing versions.
## If Yes
Without any versions we cannot fulfill the GQL API, so we'd need to filter these out / convert to null in the API layer.
But having these would allow DefinedFiles to exist and work more or less the same.
They also wouldn't change IDs if all versions were deleted and then another uploaded, which seems like a good thing...but not really functionally breaking.
Created At would also be reset, even if deleting version and the uploading new version right after.
Still leaving potential leftovers in DB doesn't seem like the best thing.

## If No
Then we'd need to handle that File nullability.
Besides the opposite of the above, this mostly just changes what the abstraction does.
Creating file once vs as needed on new version upload.


I don't have any answers right now.

CarsonF: I think we should closely match user intent. So we shouldn't create an empty file when the parent is created but rather on the first actual upload.

Resetting the ID, created at, and created by also rub me the wrong way.
We should keep these values for a potential future upload.

One solution might be to use a global and an access policy to automatically hide these empty files.
Idk if this would be more useful than an explicit filter.

I guess a third option would be to adapt the GQL API to handle these empty files.
Showing empty files in a directory list doesn't make sense. But saying a file of x id doesn't exist, but then allowing a version upload to that is also weird.
So perhaps exposing this distinction is the best way.

CarsonF: In talking with Bryan, we decided it is hard & unintuitive to keep version-less files. So we'll delete it in this use case. Between that and creating defined files on demand, there will be no version-less files in EdgeDB.

src/core/database/query/create-node.ts
```diff
@@ -24,7 +23,7 @@ export interface CreateNodeOptions<TResourceStatic extends ResourceShape<any>> {
 type InitialPropsOf<T> = {
   [K in keyof T & string]?:
     | Variable
-    | (T[K] extends FileId | LinkTo<'File'> ? ID : T[K]);
+    | (T[K] & {} extends FileId | LinkTo<'File'> ? FileId : T[K]);
```
CarsonF, line 26: This was my change. I don't think it's needed now.

src/components/location/location.repository.ts
```diff
@@ -131,6 +149,7 @@ export class LocationRepository extends DtoRepository(Location) {
         ])
         .return<{ dto: UnsecuredDto<Location> }>(
           merge('props', {
+            mapImage: { id: 'props.mapImage' },
             fundingAccount: 'fundingAccount.id',
             defaultFieldRegion: 'defaultFieldRegion.id',
             defaultMarketingRegion: 'defaultMarketingRegion.id',
```
CarsonF, line 155: Need to adjust these like `mapImage` to fit your DTO changes. That's why tests are failing.

CarsonF (changes requested):
LGTM other than those two things. Also we'll need to wait for #3111 since you have schema changes here.

src/core/database/query/create-node.ts
CarsonF: Ah ok nullability does need to be ignored in this type condition

CarsonF (approved):
Thanks @bryanjnelson!

PR #3075, opened 2024-02-07
[EdgeDB] New decoupled way to associated DB types with our Resources

CarsonF:
The former implementation (#2977) was causing TS problems.
TS would follow/compute these DB types for inheritance and our `Relations` object,
which was extra work that we didn't need.
It was too hard for it, and it gave up intermittently with the "excessively deep.." error.
It also caused problems with inheritance, which we tried to solve with `abstractType()`.

This new way allows us to map from Resources to EdgeDB types, without
TS needing to consider those DB types when evaluating our Resource types.

So far it appears to have a positive impact on performance which is encouraging too.

I did have to give up the default `hydrate` method being the object's splat (`*`), as I couldn't figure out how to get that type to work.
I don't think it's a big loss, because it should be rare that we _don't_ need to customize hydration.


src/components/funding-account/funding-account.edgedb.repository.ts
```diff
@@ -6,5 +6,7 @@ import { FundingAccountRepository } from './funding-account.repository';
 
 @Injectable()
 export class FundingAccountEdgeDBRepository
-  extends RepoFor(FundingAccount).withDefaults()
+  extends RepoFor(FundingAccount, {
+    hydrate: (fa) => fa['*'],
```
andrewmurraydavid, line 10: We can still use splats? I might've misunderstood the PR description! This is nice!

andrewmurraydavid (commented):
This looks really good! I haven't tested with my branch yet. Once that's tested and working you have my ✅

src/components/funding-account/funding-account.edgedb.repository.ts
CarsonF: We can still use them, just not generically (in RepoFor). Which is why it had to be explicitly added here where we have a concrete FundingAccount type.

andrewmurraydavid (approved):
My local git is borked again, so rebasing wasn't that easy. I've temporarily changed the base on my [PeriodicReport schema PR](https://github.com/SeedCompany/cord-api-v3/pull/3073/checks?sha=b7503fe759f9a6cd0ab0a781b0c5d3d99986798e) and ran the tests and everything passes!

Great work as usual, @CarsonF !!

bryanjnelson (commented):
Thanks @CarsonF!

PR #3080, opened 2024-02-12
[EdgeDB] ProgressReport Workflow schema

atGit2021:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6049737413) by [Unito](https://www.unito.io)


dbschema/project.esdl
```diff
@@ -214,6 +214,26 @@ module Project {
...
+    `In Progress -> In Review`,
+    `Translation Done`,
+    `Translation Reject`,
+    `Withdraw Review Request`,
+    `In Review -> Needs Translation`,
+    `Review Reject`,
+    `Review Approve`,
+    Publish,
+  >;
+
```
bryanjnelson, line 236: I think these make the most sense inside the Workflow Transition type.  They would then be part of module, so could be renamed and reference like `WorkflowTransition.name` & `WorkflowTransition.type`.

dbschema/progress-report-workflow-transition.esdl
```diff
@@ -0,0 +1,28 @@
+module default {
```
bryanjnelson, line 1: I think this is fine as a separate `esdl`, but I would rename to just `workflow-transition.esdl` and have it inside of the `ProgressReport` module.

bryanjnelson (changes requested):

dbschema/workflow-transition.esdl
```diff
@@ -0,0 +1,49 @@
...
+      readonly := true;
+    };
+  }
+
+  type InternalTransition extending WorkflowTransition {
+    required name: WorkflowTransition::Name;
+    from: array<ProgressReport::Status> {
+      readonly := true;
+    };
+    `notify`: MembersWithRoles;
```
bryanjnelson, line 19: @CarsonF - We're a little unsure if we've represented this correctly from the application code.  Thoughts?

bryanjnelson (commented):
@CarsonF - We're sending this over your way.  I think @atGit2021 has done a great job thus far, but it is possible both of us are off on something here.  Thoughts appreciated.

atGit2021:
@CarsonF I'm not sure how to resolve the error that's causing the CI tests to fail.

dbschema/workflow-transition.esdl
CarsonF, line 1: I'd like to keep the workflow loose & flexible, which is why it's currently in app code. That also allows us to make changes that affect all envs (without data migration).

The only data point that's currently stored in the DB is the transition ID in the workflow event.
While that's not ideal for viewing event log without API, I think we should keep it this way at least until after migration.

dbschema/progress-report.esdl
```diff
@@ -4,3 +4,14 @@ module default {
     overloaded engagement: LanguageEngagement;
```
CarsonF, line 4: Need a `workflowEvents` multi link here.

dbschema/workflow-event.esdl
```diff
@@ -0,0 +1,20 @@
...
+    required at: datetime {
+      readonly := true;
+    };
+    transition: ProgressReport::InternalTransition {
+      readonly := true;
+    };
+    required status: ProgressReport::Status {
+      readonly := true;
+    };
+    required notes: RichText {
```
CarsonF, line 15:
```suggestion
    notes: RichText {
```

dbschema/workflow-event.esdl
```diff
@@ -0,0 +1,20 @@
+module default {
+  type ProgressReportWorkflowEvent {
+    required who: User {
+      readonly := true;
```
CarsonF, line 4:
```suggestion
      readonly := true;
      default := currentUser;
```

CarsonF (changes requested):

dbschema/workflow-event.esdl
```diff
@@ -0,0 +1,20 @@
...
+    };
+    required at: datetime {
+      readonly := true;
```
CarsonF, line 7:
```suggestion
      readonly := true;
      default := datetime_of_statement();
```

CarsonF (approved):

PR #3085, opened 2024-02-14
[EdgeDB]: PromptVariantResponse schema

brentkulwicki:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6066857331) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/prompt-variant-response.esdl
+++ b/dbschema/prompt-variant-response.esdl
@@ -1,6 +1,7 @@
 module default {
   type PromptVariantResponse extending Resource, Mixin::Embedded, Mixin::Owned {
     required promptId: str;
+    # TODO - make the items in the list unique
     multi link responses: VariantResponse {
       on source delete delete target;
     }
--- a/dbschema/prompt-variant-response.esdl
+++ b/dbschema/prompt-variant-response.esdl
@@ -5,9 +5,7 @@ module default {
   }
 
   abstract type VariantResponse extending Resource {
-    variant: VariantResponse::Variant {
-      default := VariantResponse::Variant.draft;
-    }
+    variant: VariantResponse::Variant;
     response: RichText;
     creator: User; # Need to figure out if there is a way to reuse the above creator here
   }
--- a/dbschema/prompt-variant-response.esdl
+++ b/dbschema/prompt-variant-response.esdl
@@ -1,34 +1,13 @@
 module default {
-  type PromptVariantResponse extending Resource {
-    creator: User;
-    multi link responses := VariantResponse;
+  type PromptVariantResponse extending Resource, Mixin::Embedded, Mixin::Owned {
+    promptId: str;
+    multi link responses: VariantResponse {
+      on source delete delete target;
+    }
   }
 
-  abstract type VariantResponse extending Resource {
-    variant: VariantResponse::Variant;
-    response: RichText;
-    creator: User;
-  }
-  
-  type PromptResponse extending Resource {
-    parent: str;
-    prompt: Prompt;
+  type VariantResponse extending Resource, Mixin::Owned {
+    required variant: str;
     response: RichText;
   }
-  
-  type Prompt extending Resource {
-    text: RichText;
-    shortLabel: str;
-    min: int16;
-    max: int16; 
-  }
-}
-
-module VariantResponse {
-  scalar type Variant extending enum<
-    draft,
-    translated,
-    fpm,
-    published,
-  >;
 }
--- a/dbschema/prompt-variant-response.esdl
+++ b/dbschema/prompt-variant-response.esdl
@@ -1,14 +1,13 @@
 module default {
-  type PromptVariantResponse extending Resource, Mixin::Embedded, Mixin::Owned {
-    required promptId: str;
-    # TODO - make the items in the list unique
-    multi link responses: VariantResponse {
-      on source delete delete target;
-    }
+  type PromptVariantResponse {
+    responses := .<uniqueResponse[is VariantResponse];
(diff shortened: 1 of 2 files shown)


dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,23 @@
...
+    variant: VariantResponse::Variant {
+      default := VariantResponse::Variant.draft;
+    }
+    response: RichText;
+    creator: User; # Need to figure out if there is a way to reuse the above creator here
+  }
+}
+
+module VariantResponse {
+  scalar type Variant extending enum<
```
brentkulwicki, line 17: This doesn't feel right here as this comes from the highlights dto 🤔

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,34 @@
...
+    prompt: Prompt;
+    response: RichText;
+  }
+  
+  type Prompt extending Resource {
+    text: RichText;
+    shortLabel: str;
+    min: int16;
+    max: int16; 
+  }
```
CarsonF, line 24: Currently Prompts only exist in app code. This is one of a few like this.
We should keep it this way at least until after migration.
So delete this.

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,34 @@
...
+    multi link responses := VariantResponse;
+  }
+
+  abstract type VariantResponse extending Resource {
+    variant: VariantResponse::Variant;
+    response: RichText;
+    creator: User;
+  }
+  
+  type PromptResponse extending Resource {
```
CarsonF, line 13: This exists in app code, but has yet to have a use case. I'd just remove for now.
This is an alternate abstraction that has a single response instead of a response for each variant.

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,34 @@
+module default {
+  type PromptVariantResponse extending Resource {
+    creator: User;
+    multi link responses := VariantResponse;
+  }
+
+  abstract type VariantResponse extending Resource {
+    variant: VariantResponse::Variant;
```
CarsonF, line 8:
```suggestion
    required variant: str;
```

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,34 @@
...
+    response: RichText;
+    creator: User;
```
CarsonF, line 10:
```diff
- creator: User;
+ extending Mixin::Owned
```

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,34 @@
+module default {
+  type PromptVariantResponse extending Resource {
+    creator: User;
+    multi link responses := VariantResponse;
+  }
```
CarsonF, line 5: Need prompt ID here too.

```esdl
promptId: str;
```

CarsonF (changes requested):

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,13 @@
+module default {
+  type PromptVariantResponse extending Resource, Mixin::Embedded, Mixin::Owned {
+    promptId: str;
+    multi link responses: VariantResponse {
```
CarsonF, line 4: It would be nice to include the constraint that the variant can only be used once.
Idk how ATM. Maybe at least a todo here for now?

dbschema/prompt-variant-response.esdl
```diff
@@ -0,0 +1,13 @@
+module default {
+  type PromptVariantResponse extending Resource, Mixin::Embedded, Mixin::Owned {
+    promptId: str;
```
CarsonF, line 3:
```suggestion
    required promptId: str;
```

CarsonF (changes requested):

dbschema/prompt-variant-response.esdl
brentkulwicki: I did a little bit of research and it seems like there is no way to do this in the schema, but rather people will do it at the application level or even do it before inserting to check the list for duplicates

CarsonF: Yeah it could be done if the `VariantResponse` owned the relationship to the `PromptVariantResponse`. Practically it's the same, it would just change inserts a little.

```esdl
type PromptVariantResponse {
  responses := .<x[is VariantResponse];
}

type VariantResponse {
  required x: PromptVariantResponse;
  required variant: str;
  response: RichText;

  constraint exclusive on ((.x, .variant));
}
```

brentkulwicki: Do this now or alongside the queries? I added a TODO for now, but not sure which path forward you'd prefer

CarsonF: It's hard to say for certain without getting my head into the queries. Something may force us one way or the other.
But I want to encode everything we can into the schema. The less we have to do in app code, data wise, the better.
So yeah, I'd do this now. Thanks 🙏 
Also rename `x` to something better...I'm too tired to think what it might be. I had a real hard time naming these objects since they are so generic. If anything `pvr`.

CarsonF (approved):
I think I miscommunicated a bit. I restored the accessors and prompt field.
I also tweaked the field/module names.

I also changed my mind. This _should_ be abstract to match app code. We'll create concrete types for report Highlight/Story.
So I adjusted for that; sorry for waffling on you.

PR #3087, opened 2024-02-20
[EdgeDB] ProgressReport.Child & VarianceExplanation schema

atGit2021:
[Monday Task](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5799622935url)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6100996859) by [Unito](https://www.unito.io)


atGit2021:
@bryanjnelson , I notated a couple of places with "TODO" where I'll need your help.

bryanjnelson (approved):
LGTM, @atGit2021!

On to you @CarsonF...seems too easy.  We probably missed something.

dbschema/variance-explanation.esdl
```diff
@@ -0,0 +1,16 @@
...
+    required report: default::ProgressReport {
+      readonly := true;
+    };
+    
+    reasons: array<str> {
+      readonly := true;
+    };
+
+    comments: default::RichText {
+      readonly := true;
```
CarsonF, line 12:
```suggestion
```
Editable

dbschema/variance-explanation.esdl
```diff
@@ -0,0 +1,16 @@
+module ProgressReport {
+  type VarianceExplanation {
+    required report: default::ProgressReport {
+      readonly := true;
+    };
+    
+    reasons: array<str> {
+      readonly := true;
+    };
```
CarsonF, line 9:
```suggestion
    multi reasons: str;
```
Editable. And I think we should prefer `multi` over `array`.

dbschema/variance-explanation.esdl
```diff
@@ -0,0 +1,16 @@
+module ProgressReport {
+  type VarianceExplanation {
+    required report: default::ProgressReport {
+      readonly := true;
+    };
```
CarsonF, line 5: This needs to be an `Engagement::Child`.
It probably makes sense to have a `ProgressReport::Child` as well following suit since we'll have several of these.

CarsonF (changes requested):

dbschema/variance-explanation.esdl
CarsonF, line 1: I think this file should include "progress report". 
Maybe `progress-report.variance-explanation.esdl`?

dbschema/variance-explanation.esdl
bryanjnelson: The reason we thought these were readonly is because they are marked as such in the dto.

dbschema/variance-explanation.esdl
bryanjnelson: Ah, I didn't know `multi` applied to things other than links.  Gtk.

dbschema/variance-explanation.esdl
CarsonF: That readonly just means the API cannot modify these objects. If it wants to change it, it has to do so with a completely new object identity.
i.e.
```ts
const changed = { ...explanation, reasons: [] };
// disallows
explanation.reasons = [];
```

This is completely independent of whether a value can be changed in DB. For that you should look at the _update_ DTO.

dbschema/variance-explanation.esdl
bryanjnelson: Actually, what about just having it inside of the `progress-report.esdl`?  It could still be in the `default` module in there...or could even be in a `ProgressReport` module in there?

CarsonF: Sure, this will probably be small enough for that to make sense. Esp with the new PR::Child.

dbschema/variance-explanation.esdl
atGit2021: @bryanjnelson or @CarsonF, I'll need some help on this one.  I extended the VarianceExplanation type as an Engagement::Child, but not sure if that is correct, or how to implement your feedback.

bryanjnelson (approved):
@CarsonF - I think this is what you were going for...if it is wrong blame me. 😛 

The only thing I'm not sure of is if `VarianceExplanation` is extending correctly.

dbschema/progress-report.esdl
```diff
@@ -4,3 +4,24 @@ module default {
     overloaded engagement: LanguageEngagement;
   }
 }
+
+module ProgressReport {
+   abstract type Child extending Engagement::Child, Project::ContextAware {
```
CarsonF, line 9: Engagement::Child is already context aware
```suggestion
   abstract type Child extending Engagement::Child {
```

dbschema/progress-report.esdl
```diff
@@ -4,3 +4,24 @@ module default {
...
+    annotation description := "\
+      A type that is a child of a progress report. \
+
+      It will always have a reference to a single progress report and engagement that it is under.";
+    
+    required progressReport: default::ProgressReport {
```
CarsonF, line 15: Seems redundant, can we just call it report?
```suggestion
    required report: default::ProgressReport {
```
Matches 
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/progress-report/variance-explanation/variance-explanation.dto.ts#L27-L27

dbschema/progress-report.variance-explanation.esdl
```diff
@@ -0,0 +1,12 @@
+module ProgressReport {
+  type VarianceExplanation extending ProgressReport::Child, Engagement::Child {
```
CarsonF, line 2:
```suggestion
  type VarianceExplanation extending ProgressReport::Child {
```

I'd be ok moving this to the progress report file too as @bryanjnelson pointed out.

dbschema/progress-report.variance-explanation.esdl
```diff
@@ -0,0 +1,12 @@
...
+    required report: default::ProgressReport {
+      readonly := true;
+    };
```
CarsonF, line 5: Already declared by ancestor, but we need to an exclusive constraint.
```suggestion
    overloaded report {
      constraint exclusive;
    };
```
Since there's a 1-1 relationship between ProgressReports and VarianceExplanation.

dbschema/progress-report.esdl
```diff
@@ -4,3 +4,24 @@ module default {
     overloaded engagement: LanguageEngagement;
```
CarsonF, line 4: I'd add a shortcut here for DX
```esdl
single varianceExplanation := .<report[is ProgressReport::VarianceExplanation];
```

CarsonF (changes requested):

dbschema/progress-report.esdl
bryanjnelson: So this is just a backlink, right?

atGit2021: what is "DX"?

CarsonF: Correct, `:=`

bryanjnelson (approved):

CarsonF (approved):
Thank you guys, looks great!
I'll get the schema migration added and merged.

PR #3089, opened 2024-02-23
[Gel] Create `Product.CompletionDescription` schema

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5799854538)

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6128809801) by [Unito](https://www.unito.io)

https://seed-company-squad.monday.com/boards/5989610236/pulses/5799854716


bryanjnelson (changes requested):

bryanjnelson (commented):
@CarsonF - I'm unsure how/if the possible matches provided by the query search play in here?

dbschema/product.esdl
```diff
@@ -66,6 +66,11 @@ module default {
 }
   
 module Product {
+  type CompletionDescription {
+    query: str;
```
CarsonF, line 70: `query` is the input name.
I'd call this `value` or `description`, as it is currently in the DB.
https://github.com/SeedCompany/cord-api-v3/blob/63a27a21040c8ff7c9b50efd7e6e865e35cb7a61/src/components/product/product.repository.ts#L531-L531
It's also required.

dbschema/product.esdl
```diff
@@ -66,6 +66,11 @@ module default {
...
+    methodology: Methodology;
```
CarsonF, line 71:
```suggestion
    required methodology: Methodology;
```

dbschema/product.esdl
```diff
@@ -66,6 +66,11 @@ module default {
...
+  }
```
CarsonF, line 72: Also need `createdAt` & `lastUsedAt`.
https://github.com/SeedCompany/cord-api-v3/blob/63a27a21040c8ff7c9b50efd7e6e865e35cb7a61/src/components/product/product.repository.ts#L535-L541

dbschema/product.esdl
```diff
@@ -66,6 +66,11 @@ module default {
 }
   
 module Product {
+  type CompletionDescription {
```
CarsonF, line 69: Deviating here, but would prefixing with "Known" be a better name?
```suggestion
  type KnownCompletionDescription {
```

CarsonF (changes requested):

CarsonF:
> @CarsonF - I'm unsure how/if the possible matches provided by the query search play in here?

Yeah this uses full text search. We'll need FTS indexes here eventually. Something like this https://github.com/SeedCompany/cord-api-v3/blob/63a27a21040c8ff7c9b50efd7e6e865e35cb7a61/dbschema/z.named.esdl#L9-L14

The EdgeQL query will be something like
```edgeql
select fts::search(Product::CompletionDescription, '<query>', language := 'eng');
```

dbschema/product.esdl
bryanjnelson: I'm confused as to how these 2 fields are required.  On the UI you can leave them blank.

CarsonF: What UI? The DB nodes always have these [two values](https://github.com/SeedCompany/cord-api-v3/blob/63a27a21040c8ff7c9b50efd7e6e865e35cb7a61/src/components/product/product.repository.ts#L524-L532).

bryanjnelson: When you fill out a goal on the UI you can save without selecting a Methodology (or Completion Description).  And if you do choose one then you can save without having a Completion Description.  Maybe this is not the fields backing that UI on a Product?

CarsonF: Yeah this is just to help suggest existing values for this field
https://github.com/SeedCompany/cord-api-v3/blob/63a27a21040c8ff7c9b50efd7e6e865e35cb7a61/dbschema/product.esdl#L15-L15

CarsonF: I actually don't even know if we need this. Maybe we could just add an FTS index to that Product field.

CarsonF (approved):
I still wonder if we actually need this object or product holds enough info to use.
Still need the FTS index either way. We can revisit with the queries.

rdonigian:
Moving away from Edge/Gel

PR #3090, opened 2024-02-25
Typed IDs v2

CarsonF:
`IdOf` seems to be working well, and helps with DX. This standardizes it and improves the functionality.

- Type generic can now be given to `ID`
- Deprecated `IdOf` for `ID`
- Types/Kinds can be:
  - Resource names: `ID<'User'>`. These are easy to use without needing imports.
  - EdgeDB FQNs: `ID<'default::User'>`. These will allow DB QB types to be compatible.
  - Class "instances": `ID<User>`. These are a little nicer than strings, if already imported. And needed because that's what `IdOf` used.
  - Static class references: `ID<typeof User>`. Here for completeness, but idk the use case.
- These are normalized, so they are compatible with each other
  ```ts
  ID<'User'> == ID<'default::User'> == ID<User> == ID<typeof User>
  ```
- Concrete types are compatible with interface types 😎 
  ```ts
  const existing: ID<LanguageEngagement>;
  const id: ID<Engagement> = existing;
  ```

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6131937445) by [Unito](https://www.unito.io)

### DIFF
--- a/src/common/id-field.test.ts
+++ b/src/common/id-field.test.ts
@@ -1,0 +1,37 @@
+/* eslint-disable @seedcompany/no-unused-vars */
+import { test } from '@jest/globals';
+import type { User } from '../components/user';
+import { ID } from './id-field';
+
+test('only types here', () => undefined);
+
+const NameFromSelf: ID<'User'> = '' as ID<'User'>;
+const NameFromDB: ID<'User'> = '' as ID<'default::User'>;
+const NameFromInstance: ID<'User'> = '' as ID<User>;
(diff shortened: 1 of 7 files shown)


brentkulwicki (approved):
Woah. This is a cool improvement. ✅ 
mind-blown

bryanjnelson (approved):
> `ID<'User'> == ID<'default::User'> == ID<User> == ID<typeof User>`

🤯

PR #3091, opened 2024-02-26
EdgeDb - product progress schema

brentkulwicki:
https://seed-company-squad.monday.com/boards/5989610236/pulses/5792959117

### DIFF
--- a/dbschema/product-progress.esdl
+++ b/dbschema/product-progress.esdl
@@ -1,24 +1,16 @@
-module default {
-  type StepProgress extending Mixin::Timestamped {
-    required productProgress: ProductProgress;
+module ProgressReport::ProductProgress {
+  type Step extending Mixin::Timestamped {
+    required report: default::ProgressReport;
+    required product: default::Product;
+    required variant: Variant;
     required step: Product::Step;
-    completed: float32;
-    
-    constraint exclusive on ((.productProgress, .step));
-  }
-  type ProductProgress extending Mixin::Timestamped {
-    required product: Product;
-    required report: ProgressReport;
-    required variant: ProductProgress::Variant;
-    steps := .<productProgress[is StepProgress];
-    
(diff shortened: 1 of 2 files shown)


dbschema/product-progress.esdl
```diff
@@ -0,0 +1,19 @@
+module default {
+  type ProductProgress extending Mixin::Timestamped {
+    product: Product;
+    report: PeriodicReport;
+    variant: VariantProgress::Variant;
+    multi steps: StepProgress;
+  }
+  type StepProgress extending Mixin::Timestamped {
+    step: Product::Step;
```
CarsonF, line 9: This is required and should be exclusive to each `ProductProgress`. The best way is to flip the relation and do the same thing as above.
```suggestion
			 required productProgress: ProductProgress;
    required step: Product::Step;
    
    constraint exclusive on ((.productProgress, .step));
```

Then the `ProductProgress` references the steps with a backlink.
```diff
- multi steps: StepProgress;
+ steps := .productProgress[is StepProgress];
```

dbschema/product-progress.esdl
```diff
@@ -0,0 +1,19 @@
...
+    completed: float32;
+  }
+}
+
+module VariantProgress {
+  scalar type Variant extending enum<
```
CarsonF, line 15: Typically I've kept variants loose strings, so that they can be changed easily. But since this isn't polymorphic, like PromptVariantResponse, and there's only two, and I doubt this will change I'm fine keeping it.

CarsonF (changes requested):

CarsonF (approved):
Thanks @brentkulwicki!
I talked with @bryanjnelson and @andrewmurraydavid and we decided it would be better to merge these two together.
`ProductProgress` was/is just a container for multiple Steps, so we can do that in a group query instead.

I also tweaked the namespace too.

PR #3092, opened 2024-02-27
Only allow FPMs to edit Project dates if in development

CarsonF:
This is a business constraint that we've always wanted to implement. Now easy with the policy system.
The only logical changes of this PR are in the [policy](https://github.com/SeedCompany/cord-api-v3/pull/3092/files#diff-2ce6dc91076b6679b82573512d72c4b924876b2bde1e363c03f9353d4eb1e782).

There's some further refinement as well to only allow editing in the first ~half of development steps. But this also has workflow implications as well, so I'm leaving it stubbed out for now.


brentkulwicki (approved):
Makes sense and LGTM

bryanjnelson (approved):

PR #2976, opened 2023-11-30
Migrate to new delete permission check

CarsonF:
┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5601024713) by [Unito](https://www.unito.io)

PR #3094, opened 2024-02-27
UX Regression - Budget modification from Project date changes

CarsonF:
Users will change project date ranges frequently while the project is in development.
This can cause budget records to be removed.
This process regressed #2976, which started enforcing that users had permission to delete budget records.
This is weird because they do not create/delete these directly, but rather through the changes of project dates or funding partners.

As records are system-managed objects, I've removed the create/delete perm checks for them. And marked them as calculated, so the perms dumper knows those actions are not applicable.

# Before
```
┌───────────────────────────────────────────────────────────┐
│              Permissions for Project Manager              │
│ Resource                │ Read │ Edit   │ Create │ Delete │
│ Budget                  │ Yes  │ Member │ No     │ No     │
│  .universalTemplateFile │ Yes  │ Member │        │        │
│ Budget Record           │ Yes  │ Member │ No     │ No     │
│  .organization          │ Yes  │ Member │        │        │
│  .fiscalYear            │ Yes  │ Member │        │        │
│  .amount                │ Yes  │ Member │        │        │
└─────────────────────────┴──────┴────────┴────────┴────────┘
```

# After
```
┌───────────────────────────────────────────────────────────┐
│              Permissions for Project Manager              │
│ Resource                │ Read │ Edit   │ Create │ Delete │
│ Budget                  │ Yes  │        │        │        │
│  .universalTemplateFile │ Yes  │ Member │        │        │
│ Budget Record           │ Yes  │        │        │        │
│  .organization          │ Yes  │        │        │        │
│  .fiscalYear            │ Yes  │        │        │        │
│  .amount                │ Yes  │ Member │        │        │
└─────────────────────────┴──────┴────────┴────────┴────────┘
```


brentkulwicki (approved):

src/components/budget/dto/budget-record.dto.ts
```diff
@@ -30,8 +32,10 @@ export class BudgetRecord extends IntersectionType(ChangesetAware, Resource) {
   @Field(() => Budget)
   declare readonly parent: BaseNode;
 
+  @Calculated()
```
bryanjnelson, line 35: @CarsonF - Besides notifying the perm dumper that these are not applicable, what else does adding these `@Calculated` annotations do?

bryanjnelson (approved):

src/components/budget/dto/budget-record.dto.ts
CarsonF: Technically, edit/create/delete checks are forced to false, instead of checking policies.
https://github.com/SeedCompany/cord-api-v3/blob/5b1b85b3922230b3937509b98fde74eede1da735/src/components/authorization/policy/executor/policy-executor.ts#L36-L44

In all of these cases here, we are not checking the perms, or the properties we do check are not updatable.

PR #3095, opened 2024-02-27
[EdgeDB] nanoid scalar to generate types as ID

CarsonF:
This correctly generates types like
```ts
class PromptVariantResponse {
  promptId: ID;
  ...
}
```
This is more accurate than `string`.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6148790225) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/common.esdl
+++ b/dbschema/common.esdl
@@ -17,6 +17,8 @@ module default {
     )
   );
   
+  scalar type nanoid extending str;
+  
   scalar type RichText extending json;
   
   # A fake function to produce valid EdgeQL syntax.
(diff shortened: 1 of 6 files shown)


CarsonF:
I'd really rather these be `uuid` as that's the standard ID type that EdgeDB uses.
But this requires some more thought with migration, so putting it off as a future problem.

Here's how a UUID could be generated based on seed string input.
```ts
import uuid from 'uuid';

const namespace = uuid.v4(); // once & statically store in app code.
const id = uuid.v5('...', namespace);
```
```ts
// example
const processReportWorkflowTransitionsNamespace = '112db88b-be48-45c5-b7db-d62f03237261';
const transitionId = uuid.v5('In Review -> Needs Translation', processReportWorkflowTransitionsNamespace);
```

So this could be what we do in future. But that still leaves data migration outstanding.

PR #3097, opened 2024-02-28
[EdgeDB] Default Repo create/update handle link IDs

CarsonF:
The queries built will now automatically cast IDs for links to their object refs.
The methods will also accept props with an `Id` suffix and strip it off.

This should help make a lot of our queries involving links easier, as demonstrated by FieldZone & FieldRegion repos.

```diff
  create({
    name: input.name,
-   director: e.cast(e.User, e.cast(e.uuid, input.directorId)),
+   directorId: input.directorId,
  })
```
This is only a shortcut: EdgeQL expressions can still be given, and IDs can be given to `director` (no `Id` suffix).

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6155607311) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/edgedb/query-util/easy-insert.ts
+++ b/src/core/edgedb/query-util/easy-insert.ts
@@ -1,0 +1,103 @@
+import { Cardinality } from 'edgedb/dist/reflection';
+import {
+  ConditionalPick,
+  MergeExclusive,
+  RequireExactlyOne,
+  StringKeyOf,
+} from 'type-fest';
+import { ID } from '~/common';
+import { AllResourceDBNames } from '~/core';
+import {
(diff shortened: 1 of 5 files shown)


src/core/edgedb/query-util/easy-insert.ts
CarsonF, line 1: FWIW these types were copied and adapted from edgedb.
Looking at 
RawInsertShape from ~/core/edgedb/generated-client/insert.ts
UpdateShape from ~/core/edgedb/generated-client/update.ts
there's a lot of similarities.

bryanjnelson (approved):
Huge DX win!  🥳

PR #3099, opened 2024-02-29
[EdgeDB] DirectScriptureProduct seeds

willdch:
https://seed-company-squad.monday.com/boards/5989610236/pulses/5793006654

While this is being reviewed I will continue working on the seed data for `DerivativeScriptureProduct` and `OtherProduct`.

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/6160636235) by [Unito](https://www.unito.io)

### DIFF
--- a/dbschema/seeds/011.product.edgeql
+++ b/dbschema/seeds/011.product.edgeql
@@ -1,0 +1,1080 @@
+with
+    productsJson := to_json('[
+     {
+      "project": "Misty Mountains",
+      "scripture":
+        {
+          "label": "Joshua",
+          "verses": [
+            {
+              "label": "Joshua",
+              "end": {
+                "book": "Joshua",
+                "chapter": 24,
+                "verse": 33,
+                "verseId": 2
+              },
+              "start": {
+                "book": "Joshua",
+                "chapter": 1,
+                "verse": 1,
+                "verseId": 1
+              }
+            }
+          ]
+        },
+      "mediums" : ["Print", "Web", "App"],
+      "purposes" : ["Discipleship", "ChurchLife", "SocialIssues" ],
+      "steps": [
+          "ExegesisAndFirstDraft",
+          "TeamCheck",
+          "CommunityTesting",
+          "BackTranslation",
+          "ConsultantCheck",
+          "Completed"
+          ],
+      "methodology": "Paratext",
+      "describeCompletion": "All pre-publication checks complete and published on app and trial print editions",
+      "pnpIndex": 1,
+      "progressTarget": 100,
+      "progressStepMeasurement":"Percent",
+      "totalVerses": 658,
+      "totalVerseEquivalents": 658
+     },
+     {
+      "project": "Misty Mountains",
+      "scripture":
+        {
+          "label": "Proverbs",
+          "verses": [
+            {
+              "label": "Proverbs",
+              "end": {
+                "book": "Proverbs",
+                "chapter": 31,
+                "verse": 31,
+                "verseId": 3
+              },
+              "start": {
+                "book": "Proverbs",
+                "chapter": 1,
+                "verse": 1,
+                "verseId": 4
+              }
+            }
+          ]
+        },
+      "mediums" : ["Print", "Audio"],
+      "purposes" : ["Discipleship"],
+      "steps": [
+          "CommunityTesting",
+          "BackTranslation",
+          "ConsultantCheck"
+          ],
+      "methodology": "Paratext",
+      "pnpIndex": 1,
+      "progressTarget": 100,
+      "progressStepMeasurement":"Percent",
+      "totalVerses": 915,
+      "totalVerseEquivalents": 1143
+     },
+     {
+      "project": "Misty Mountains",
+      "scripture":
+        {
+          "label": "Psalms",
+          "verses": [
+            {
+              "label": "Psalms",
+              "end": {
+                "book": "Psalms",
+                "chapter": 150,
+                "verse": 6,
+                "verseId": 5
+                },
+              "start": {
+                "book": "Psalms",
+                "chapter": 1,
+                "verse": 1,
+                "verseId": 6
+              }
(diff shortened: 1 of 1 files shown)


bryanjnelson (approved):
Looks good to me so far, @willdch!

dbschema/seeds/011.product.edgeql
```diff
@@ -0,0 +1,1080 @@
...
+    }
+    ]'),
+     products := (
+        for directScriptureProduct in json_array_unpack(productsJson)
+        union (
+          (select DirectScriptureProduct filter .project.name = <str>directScriptureProduct['project']
+           and (.scripture.label ?= <str>json_get(directScriptureProduct, 'scripture', 'label')
+                or .unspecifiedScripture.book ?= <str>json_get(directScriptureProduct, 'unspecifiedScripture', 'book'))) ??
+          (with
+            scripture := if (select exists json_get(directScriptureProduct,'scripture')) then (
```
CarsonF, line 1020: I'd rather just inline this. It's only assigned in the insert statement.

dbschema/seeds/011.product.edgeql
```diff
@@ -0,0 +1,1080 @@
...
+                                      verse := <int16>verseRange['end']['verse'],
+                                      verseId := <int16>verseRange['end']['verseId']
+                                  }
+                              )
+                          }
+                      )
+                  )
+              })
+              else {},
+            engagement := assert_single((select Engagement filter .project.name = <str>directScriptureProduct['project']))
```
CarsonF, line 1049: I'd move this to a with statement above the select/insert. This is a datapoint we are expecting to pre-exist as a dependency to this script.
Modeled here: https://github.com/SeedCompany/cord-api-v3/blob/develop/dbschema/seeds/009.language-engagements.edgeql#L37-L42

With that scoped you can change your `select` check to
```edgeql
select DirectScriptureProduct filter .project = engagement.project
```

This will make it easier to adapt this to support multiple engagements within a project. Something we should do soon.

dbschema/seeds/011.product.edgeql
```diff
@@ -0,0 +1,1080 @@
...
+          progressStepMeasurement := <str>directScriptureProduct['progressStepMeasurement'],
+          totalVerses := <int16>json_get(directScriptureProduct, 'totalVerses'),
+          totalVerseEquivalents := <int32>json_get(directScriptureProduct, 'totalVerseEquivalents')
+        }
+      )
+    )),
+      newProducts := (select products filter .createdAt = datetime_of_statement())
+ select { `Added Scripture Products` := "Project: " ++ newProducts.project.name ++ ', Scripture Range: '
+         ++ newProducts.scripture.label ++ ' ' ++ <str>newProducts.scripture.verses.`start`.chapter ++ ' - ' ++ <str>newProducts.scripture.verses.`end`.chapter, `Total Scripture` := count(newProducts.scripture), `Added UnspecifiedScripture Products` := "Project: "
+         ++ newProducts.project.name ++ ', Unspecified Scripture Book: ' ++ newProducts.unspecifiedScripture.book, `Total UnspecifiedScripture` := count(newProducts.unspecifiedScripture)}
```
CarsonF, line 1079: I won't nitpick on formatting here, as we don't have a way yet to enforce it.
But can you put each of these 4 on their own line.

It may just be me, but I'd simplify this format.
```diff
- Project: Misty Mountains, Scripture Range: Proverbs 1 - 31
+ Misty Mountains: Proverbs 1 - 31
```
```diff
- Project: Misty Mountains, Unspecified Scripture Book: Matthew
+ Misty Mountains: Matthew X verses
```
We are already scoped with a
```
"Added Scripture Products": [
  ...
]
```
which gives enough info. So emitting types just adds clutter.
Thoughts?

CarsonF (changes requested):
Wow, this is huge @willdch! 🤯 
This will be so nice to have for development :heart:

dbschema/seeds/011.product.edgeql
CarsonF: Actually the select should be
```edgeql
select DirectScriptureProduct filter .engagement = engagement
```

dbschema/seeds/011.product.edgeql
willdch: I toggled back and forth on this and I feel fine with simplifying it.  I just wanted to make sure I had the necessary info output clearly, which I think is still accomplished with your suggestions.

dbschema/seeds/011.product.edgeql
```diff
@@ -0,0 +1,1080 @@
...
+                "book": "Song of Solomon",
+                "chapter": 8,
+                "verse": 14,
+                "verseId": 41
+              },
+              "start": {
+                "book": "Song of Solomon",
+                "chapter": 1,
+                "verse": 1,
+                "verseId": 42
```
CarsonF, line 719: This start verse ID needs be less than the end ID. Would be worth double checking other places.
You check by running
```edgeql
select Scripture::Collection {*}
```
and seeing if it gives you an error or not.

You can also
```edgeql
delete Product;
```
to be able to re-run the seed script for changes you make.

willdch: @CarsonF , good catch here.  I was getting an error and couldn't figure out why but this makes total sense now.

CarsonF: Ironically Bryan suggested that I write a constraint for this. But I declined because it would be a non trivial trigger assertion and I said that these inserts would be constructed in an app code abstraction. But here we are. So Bryan gets a point 😁

CarsonF (changes requested):
Here are the actual verse IDs. Will you apply?
[verse-ids.json](https://github.com/SeedCompany/cord-api-v3/files/14465651/verse-ids.json)

CarsonF (approved):
Sweet!

PR #3112, opened 2024-03-08
[EdgeDB] Rename seeding files to account for dependencies

bryanjnelson:


CarsonF (approved):
I also wonder if it would be better for these numbers to represent stages. Multiple seeds can exist in the same stage as long as there aren't dependencies.

bryanjnelson:
> I also wonder if it would be better for these numbers to represent stages. Multiple seeds can exist in the same stage as long as there aren't dependencies.

I could see that.  That type of grouping might become more clear as the seeding increases as well.

PR #3113, opened 2024-03-11
[EdgeDB] Comment Schema

wes-cutting:
https://seed-company-squad.monday.com/boards/5989610236/pulses/5792954463

### DIFF
--- a/dbschema/comments.esdl
+++ b/dbschema/comments.esdl
@@ -1,0 +1,21 @@
+module Comments {
+  abstract type Aware extending default::Resource {
+    commentThreads := .<container[is default::Thread];
+  }
+
+  type Thread extending Resource, Mixin::Embedded, Mixin::Owned {
+    overloaded required single link container: Aware {
+      on target delete delete source;
+    };
+    comments := .<thread[is Comment];
+    firstComment := (select .comments order by .createdAt asc limit 1);
+    latestComment := (select .comments order by .createdAt desc limit 1);
+  }
+
+  type Comment extending Resource, Mixin::Owned {
+    required body: default::RichText;
+    required thread: Thread {
+      on target delete delete source;
+    };
+  }
+}
--- a/dbschema/comments.esdl
+++ b/dbschema/comments.esdl
@@ -1,9 +1,9 @@
 module Comments {
   abstract type Aware extending default::Resource {
-    commentThreads := .<container[is default::Thread];
+    commentThreads := .<container[is Thread];
   }
 
-  type Thread extending Resource, Mixin::Embedded, Mixin::Owned {
+  type Thread extending default::Resource, Mixin::Embedded, Mixin::Owned {
     overloaded required single link container: Aware {
       on target delete delete source;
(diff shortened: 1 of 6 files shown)


dbschema/comment-thread.esdl
```diff
@@ -0,0 +1,29 @@
...
+      select Comment filter __source__ = .thread
+    );
+    firstComment := (
+      select Comment filter __source__ = .thread
+      order by .createdAt limit 1
+    );
+    latestComment := (
+      select Comment filter __source__ = .thread
+      order by .createdAt desc limit 1
+    );
```
CarsonF, line 21: These can be shortened to:
```suggestion
    comments := .<thread[is Comment];
    firstComment := (select .comments order by .createdAt asc limit 1);
    latestComment := (select .comments order by .createdAt desc limit 1);
```

dbschema/comment-thread.esdl
```diff
@@ -0,0 +1,29 @@
+module default {
+  type Comment extending Resource, Mixin::Owned {
+    required body: default::RichText;
+    required thread: CommentThread;
```
CarsonF, line 4: Let's allow threads to be deleted with their comments
```suggestion
    required thread: CommentThread {
      on target delete delete source;
    };
```

dbschema/comment-thread.esdl
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  type CommentThread extending Resource, Mixin::Embedded, Mixin::Owned {
```
CarsonF, line 7: Following suit of other things, like BudgetRecord, this could be `Comment::Thread`.
aka `module Comment { type Thread }`. It does feel a little weird since Thread is the entry point. But the name is nice. Thoughts?

CarsonF (changes requested):
Thanks! Just a couple things..

dbschema/comment-thread.esdl
andrewmurraydavid: @CarsonF what about `Comment`? Would that become `Comment::Comment`?

CarsonF: I guess that stays `default::Comment`

wes-cutting: Andrew and I had a discussion about this last Thursday on the first few passes of this schema. Where we are now is if it makes sense to switch up the module and naming convention for how this is implemented then we're fine with the change. It is a matter of opinion and either way makes sense to me

andrewmurraydavid: Feels like naming inconsistency, having `Thread` (which is the main of the comment) in the `Comment` namespace but `Comment` in the default.
There are just thoughts, I'm not leaning either way. Ok with both options.

CarsonF: Yeah I agree Andrew. And Commentable is in a third module :/

CarsonF (approved):
Thanks @wes-cutting! Just added a couple more things & migration

PR #3117, opened 2024-03-13
[EdgeDB] OtherProduct seeds

willdch:
https://seed-company-squad.monday.com/boards/5989610236/pulses/5793006654


bryanjnelson (changes requested):
Looks great @willdch!  Much better formatting as well!  😉

dbschema/seeds/014.product.edgeql
```diff
@@ -1083,3 +1083,121 @@ with
...
+          methodology := <Product::Methodology>json_get(otherProduct,'methodology'),
+          describeCompletion := <str>json_get(otherProduct,'describeCompletion'),
+          progressTarget := <int16>otherProduct['progressTarget'],
+          progressStepMeasurement := <str>otherProduct['progressStepMeasurement']
+        })
+      )
+    )
+  ),
+  newOtherProducts := (select otherProducts filter .createdAt = datetime_of_statement())
+  select { `Added Other Products` := newOtherProducts.project.name ++ ': ' ++ newOtherProducts.title }
```
bryanjnelson, line 1203: Remove the indention on the last line here.  It should be flush on the left.  Also, we always want one new line at the end of every file.  👊🏼

dbschema/seeds/014.product.edgeql
```diff
@@ -1090,20 +1090,15 @@ with
...
     "progressStepMeasurement": "Percent"
   },
   {
     "project": "Misty Mountains",
     "title":  "7 new songs written and recorded",
-    "mediums": [
-      "Audio",
-      "Web"
-      ],
+    "mediums": ["Audio","Web"],
```
bryanjnelson, line 1101: We want spaces in between our array entries, so like this here...and then on any others.

```suggestion
    "mediums": ["Audio", "Web"],
```

bryanjnelson (changes requested):

bryanjnelson (approved):
Thanks @willdch!  LGTM!

@CarsonF?

CarsonF:
Let's put these in their own files. I don't want to try to find the actual logic here in the middle of thousands of lines.
Not to mention your two PRs conflict with each other since they are both modifying the same spot.

CarsonF (approved):

PR #3122, opened 2024-03-19
[EdgeDB] Create Partnership queries

willdch:
https://seed-company-squad.monday.com/boards/5989610236/pulses/5792970379

@CarsonF Still work to be done but would love to get your eyes on these changes while working on the rest.

### DIFF
--- a/src/components/partnership/dto/partnership.dto.ts
+++ b/src/components/partnership/dto/partnership.dto.ts
@@ -5,7 +5,6 @@ import { e } from '~/core/edgedb';
 import { LinkTo, RegisterResource } from '~/core/resources';
 import {
   Calculated,
-  ID,
   IntersectionType,
   Resource,
   ResourceRelationsShape,
@@ -17,9 +16,7 @@ import {
   Sensitivity,
   SensitivityField,
 } from '../../../common';
-import { ScopedRole } from '../../authorization';
 import { ChangesetAware } from '../../changeset/dto';
-import { DefinedFile } from '../../file/dto';
 import { Organization } from '../../organization/dto';
 import { SecuredPartnerTypes } from '../../partner/dto';
 import { IProject } from '../../project/dto';
@@ -54,7 +51,7 @@ export class Partnership extends IntersectionType(ChangesetAware, Resource) {
   } satisfies ResourceRelationsShape;
   static readonly Parent = import('../../project/dto').then((m) => m.IProject);
 
-  readonly project: ID;
+  readonly project: LinkTo<'Project'>;
 
   @Field(() => IProject)
   declare readonly parent: BaseNode;
@@ -62,7 +59,7 @@ export class Partnership extends IntersectionType(ChangesetAware, Resource) {
   @Field()
   readonly agreementStatus: SecuredPartnershipAgreementStatus;
 
-  readonly mou: Secured<LinkTo<'File'>>;
+  readonly mou: Secured<LinkTo<'File'> | null>;
 
   @Field()
   readonly mouStatus: SecuredPartnershipAgreementStatus;
@@ -81,10 +78,10 @@ export class Partnership extends IntersectionType(ChangesetAware, Resource) {
   @Field()
   readonly mouEndOverride: SecuredDateNullable;
 
-  readonly agreement: DefinedFile;
+  readonly agreement: Secured<LinkTo<'File'> | null>;
 
-  readonly partner: Secured<ID>;
-  readonly organization: ID;
+  readonly partner: Secured<LinkTo<'Partner'>>;
+  readonly organization: LinkTo<'Organization'>;
 
   @Field()
   readonly types: SecuredPartnerTypes;
@@ -99,10 +96,6 @@ export class Partnership extends IntersectionType(ChangesetAware, Resource) {
     description: "Based on the project's sensitivity",
   })
   readonly sensitivity: Sensitivity;
-
-  // A list of non-global roles the requesting user has available for this object.
-  // This is just a cache, to prevent extra db lookups within the same request.
-  declare readonly scope: ScopedRole[];
 }
 
 declare module '~/core/resources/map' {
--- a/src/components/partnership/partnership.repository.ts
+++ b/src/components/partnership/partnership.repository.ts
@@ -2,15 +2,17 @@ import { Injectable } from '@nestjs/common';
 import { Node, node, Query, relation } from 'cypher-query-builder';
 import { pickBy } from 'lodash';
 import { DateTime } from 'luxon';
-import { ChangesOf } from '~/core/database/changes';
 import {
+  DuplicateException,
   generateId,
   ID,
   labelForView,
+  NotFoundException,
   ObjectView,
   ServerException,
   Session,
   UnsecuredDto,
+  viewOfChangeset,
 } from '../../common';
 import { DtoRepository } from '../../core';
 import {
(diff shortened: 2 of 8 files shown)


src/components/partnership/dto/partnership.dto.ts
```diff
@@ -54,15 +52,15 @@ export class Partnership extends IntersectionType(ChangesetAware, Resource) {
...
+  readonly project: LinkTo<'Project'>;
 
   @Field(() => IProject)
   declare readonly parent: BaseNode;
 
   @Field()
   readonly agreementStatus: SecuredPartnershipAgreementStatus;
 
-  readonly mou: Secured<LinkTo<'File'>>;
+  readonly mou: Secured<LinkTo<'File'>> | null;
```
CarsonF, line 63: `Secured`'s are never nullable, but they can wrap nullable things, like this file.
```suggestion
  readonly mou: Secured<LinkTo<'File'> | null>;
```

CarsonF (commented):
Looks like you are on the right track!

src/components/partnership/partnership.service.ts
```diff
@@ -143,40 +114,26 @@ export class PartnershipService {
...
-  ): Promise<Partnership> {
+  secure(dto: UnsecuredDto<Partnership>, session: Session) {
     return this.privileges.for(session, Partnership).secure(dto);
   }
 
   async update(input: UpdatePartnership, session: Session, view?: ObjectView) {
     const existing = await this.repo.readOne(input.id, session, view);
     const partner = await this.partnerService.readOne(
-      existing.partner,
+      existing.partner as unknown as ID, // TODO: clean this up
```
bryanjnelson, line 133: @CarsonF - Recommendations here?

src/components/partnership/partnership.service.ts
```diff
@@ -81,10 +76,10 @@ export class PartnershipService {
       input.financialReportingType,
       input.types ?? [],
       partner,
-    );
+    ); // is this a check to be moved to the repository?
```
bryanjnelson, line 79: I think from the comment below on `verifyFinancialReportingType` it was decided that this is okay here for now, but will be moved to the repo after migration.  So this comment can probably either be deleted or changed to a TODO similar to the one below.

bryanjnelson (changes requested):
Looking good!

src/components/partnership/partnership.service.ts
CarsonF:
```suggestion
      existing.partner.id,
```

src/components/partnership/partnership.repository.ts
```diff
@@ -195,25 +222,24 @@ export class PartnershipRepository extends DtoRepository<
...
-
       .match(requestingUser(session))
       .apply(
         this.privileges.forUser(session).filterToReadable({
           wrapContext: oncePerProject,
         }),
       )
       .apply(sorting(Partnership, input))
-      .apply(paginate(input))
+      .apply(paginate(input, this.hydrate(session)))
```
CarsonF, line 232: `hydrate` has not been overridden here to provide the correct shape.
Notice how `readMany`, or anything, doesn't call it either.

Tracing this was simple:
```
TypeError: Cannot read properties of undefined (reading 'id')

    at SyncBudgetRecordsToFundingPartners.syncRecords (src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts:152:53)at SyncBudgetRecordsToFundingPartners.handle (src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts:94:18)
    at SyncEventBus.publish (src/core/events/event-bus.service.ts:48:7)
    at ProjectService.update (src/components/project/project.service.ts:316:5)
    at Connection.withTransaction [as runInTransaction] (src/core/database/transaction.ts:96:12)
    at ProjectService.descriptor.value (src/core/database/transactional.decorator.ts:44:14)
    at ProjectResolver.updateProject (src/components/project/project.resolver.ts:374:21)
    at Object.<anonymous> (test/project-changeset-aware.e2e-spec.ts:211:5)
```
1. `syncRecords` is called from the list result of [determinePartnerships](https://github.com/SeedCompany/cord-api-v3/blob/0937-edgedb-partnership-queries/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts#L91-L91)
2. `determinePartnerships` has several condition branches. We know it's the [`ProjectUpdateEvent`](https://github.com/SeedCompany/cord-api-v3/blob/0937-edgedb-partnership-queries/src/components/budget/handlers/sync-budget-records-to-funding-partners.handler.ts#L137-L138), because `ProjectService.update` is what is in the stacktrace.
3. So I put a breakpoint on the [result of the partnerships list](https://github.com/SeedCompany/cord-api-v3/blob/0937-edgedb-partnership-queries/src/components/partnership/partnership.repository.ts#L234-L234)
4. Ran the test that failed, `project-changeset-aware.e2e-spec.ts:211`, given by stacktrace.
5. Saw the was missing all of the links/relationships. Why? Because the [base hydrate](https://github.com/SeedCompany/cord-api-v3/blob/0937-edgedb-partnership-queries/src/core/database/dto.repository.ts#L153-L154) only grabs the direct props.

A go-to on `hydrate` would've revealed this as well.
Or a `.logIt()` on the query showing that none of the relationships were matched/merged.

bryanjnelson: I walked through this with Will and we track with you the whole way down to 5).  At that point I'm confused though.  From your red/green comment above it looks like you want him to add in the `this.hydrate(session)`, but that already exists on that line, so that doesn't make sense.

From your explanation, on the other hand, I think you're saying that we need to add a `hydrate` method specifically inside the Neo4j repo that actually returns the relationships as well, instead of just falling back on the base `DtoRepository` hydrate method.

Can you clarify for us here?

CarsonF: > From your explanation, on the other hand, I think you're saying that we need to add a `hydrate` method specifically inside the Neo4j repo that actually returns the relationships as well, instead of just falling back on the base `DtoRepository` hydrate method.

That is correct!



> From your red/green comment above it looks like you want him to add in the `this.hydrate(session)` portion of that line, but that already exists, so that doesn't make sense.

You guys added the first call to `hydrate` here, not me. And the default implementation was insufficient to complete the shape we require. But we have no way of confirming that, it's just typecasts with cypher/neo4j.

src/components/partnership/partnership.service.ts
```diff
@@ -59,8 +56,6 @@ export class PartnershipService {
   ): Promise<Partnership> {
     const { projectId, partnerId } = input;
 
-    await this.verifyRelationshipEligibility(projectId, partnerId, changeset);
```
CarsonF, line 62: This is what changed the error message

src/components/partnership/partnership.repository.ts
```diff
@@ -124,50 +152,58 @@ export class PartnershipRepository extends DtoRepository<
...
+              'changedProps.mouEndOverride',
+              'props.mouEndOverride',
+              'projectChangedProps.mouEnd',
+              'projectProps.mouEnd',
+            ),
+            parent: 'project',
+            project: 'project { .id }',
+            partner: 'partner { .id }',
+            organization: 'org { .id }',
+            changeset: 'changeset { .id }',
```
CarsonF, line 201: We're still expecting the changeset to be an ID, not a object/link.
```suggestion
            changeset: 'changeset.id',
```

willdch: Thank you @CarsonF!

CarsonF (approved):
FYI @willdch checkout 5ce7afe073c296a9f10cd106bb8886004830f804

PR #3123, opened 2024-03-20
Bump edgedb lib for object cast on expression fix

CarsonF:
Also a couple query tweaks using object casts on multiple IDs


CarsonF:
@andrewmurraydavid this fixes what we ran into today with the `e.cast` expression.

brentkulwicki (approved):

bryanjnelson (approved):

PR #3127, opened 2024-03-21
[EdgeDB] ProgressReport Workflow Queries

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/5989610236/views/131962501/pulses/5799644779)

### DIFF
--- a/src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
+++ b/src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
@@ -131,7 +131,7 @@ export class ProgressReportWorkflowNotificationHandler
     );
     const report = await this.reportService.readOne(reportId, recipientSession);
     const changedBy = await this.userService.readOne(
-      unsecuredEvent.who,
+      unsecuredEvent.who.id,
       recipientSession,
     );
     const workflowEvent = this.workflowService.secure(
--- a/src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
+++ b/src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
@@ -1,0 +1,82 @@
+import { Injectable } from '@nestjs/common';
+import { SetRequired } from 'type-fest';
+import { ID, PublicOf } from '~/common';
+import { e, edgeql, RepoFor } from '~/core/edgedb';
+import { ProgressReportStatus as Status } from '../dto';
+import { ExecuteProgressReportTransitionInput } from './dto/execute-progress-report-transition.input';
+import { ProgressReportWorkflowEvent } from './dto/workflow-event.dto';
+import { ProgressReportWorkflowRepository } from './progress-report-workflow.repository';
+
+@Injectable()
+export class ProgressReportWorkflowEdgeDBRepository
+  extends RepoFor(ProgressReportWorkflowEvent, {
+    hydrate: (event) => ({
+      ...event['*'],
+      report: true,
+      who: true,
+      transition: event.transitionId,
+    }),
+  }).customize((cls, { defaults }) => {
+    return class extends cls {
+      static omit = [defaults.create, defaults.update, defaults.delete];
+      async list(reportId: ID) {
+        const progressReport = e.cast(e.ProgressReport, e.uuid(reportId));
+        const query = e.select(progressReport.workflowEvents, this.hydrate);
+        return await this.db.run(query);
+      }
+    };
+  })
+  implements PublicOf<ProgressReportWorkflowRepository>
+{
+  async recordEvent({
+    report,
+    ...props
+  }: SetRequired<ExecuteProgressReportTransitionInput, 'status'>) {
+    const query = e.select(
+      e.insert(e.ProgressReport.WorkflowEvent, {
+        report: e.cast(e.ProgressReport, e.uuid(report)),
+        transitionId: props.transition,
+        status: props.status,
+        notes: props.notes,
+      }),
+      this.hydrate,
+    );
+    return await this.db.run(query);
+  }
+
+  async currentStatus(reportId: ID): Promise<Status> {
+    const query = e.cast(e.ProgressReport, e.uuid(reportId)).status;
+    return await this.db.run(query);
+  }
+
+  async changeStatus(_report: ID, _status: Status) {
+    return;
+  }
+
+  async getProjectMemberInfoByReportId(reportId: ID) {
+    const query = edgeql(`
+      with report := <ProgressReport><uuid>$reportId,
+      users := (select report.project.members.user filter exists .email)
+      select users { id, email := assert_exists(.email), roles }
+    `);
+    return await this.db.run(query, { reportId });
+  }
+
+  async getUserIdByEmails(emails: readonly string[]) {
+    const query = edgeql(`
+      with emails := array_unpack(<array<str>>$emails),
+      users := (select User filter .email in emails)
+      select users { id, email := assert_exists(.email) }
+    `);
+    return await this.db.run(query, { emails });
+  }
+
+  async getProjectInfoByReportId(reportId: ID) {
+    const progressReport = e.cast(e.ProgressReport, e.uuid(reportId));
+    const query = e.select({
+      projectId: progressReport.project.id,
+      languageId: progressReport.engagement.language.id,
+    });
+    return await this.db.run(query);
+  }
+}
--- a/src/components/progress-report/workflow/progress-report-workflow.repository.ts
+++ b/src/components/progress-report/workflow/progress-report-workflow.repository.ts
@@ -1,10 +1,10 @@
 import { Injectable } from '@nestjs/common';
 import { inArray, node, Query, relation } from 'cypher-query-builder';
+import { SetRequired } from 'type-fest';
 import {
   ID,
   NotFoundException,
   Order,
-  RichTextDocument,
   Role,
   ServerException,
   Session,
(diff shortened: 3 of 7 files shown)


atGit2021:
> @atGit2021 this worked for me
> 
> ```ts
>   readonly transition?: InternalTransition & SetUnsecuredType<ID | null>;
> ```

Thank you

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
```diff
@@ -0,0 +1,193 @@
...
+    hydrate: (event) => ({
+      ...event['*'],
+      report: true,
+      who: true,
+      transition: event.transitionId,
+    }),
+  }).customize((cls, { defaults }) => {
+    return class extends cls {
+      static omit = [defaults.create, defaults.update, defaults.delete];
+      async list(
```
bryanjnelson, line 28: @CarsonF - Andre's reasoning for customizing this is because the signature in the common EdgeDB repo uses Pagination, but the one in the Neo4j repo does not.  Is customizing the best approach here?  Should we consider altering the Neo4j list instead?

CarsonF: The list isn't paginated in GQL, so we couldn't just change the neo4j query.
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/progress-report/workflow/resolvers/progress-report-workflow-events.resolver.ts#L11-L17

Customizing the method here is the right approach.

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
```diff
@@ -0,0 +1,193 @@
...
+            id: event.who.id,
+            email: event.who.email,
+          },
+        }));
+      }
+    };
+  })
+  implements PublicOf<ProgressReportWorkflowRepository>
+{
+  async recordTransition(
```
bryanjnelson, line 72: @CarsonF - Would it make sense to move both `recordTransition` and `recordBypass` up to the service so that they don't need to be duplicated here?  Neither of them are actually doing anything database specific.

CarsonF: Sure

bryanjnelson (changes requested):
Excellent work, @atGit2021!

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
```diff
@@ -0,0 +1,100 @@
...
+    const users = await this.db.run(query);
+    return users.map((user) => ({
+      id: user.id,
+      email: user.email,
+      roles: user.roles.map((role) => role),
+    }));
+  }
+
+  async getUserIdByEmails(emails: readonly string[]) {
+    const query = e.params({ emails: e.array(e.str) }, ({ emails }) =>
```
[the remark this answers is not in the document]
CarsonF: Params is better because it doesn't change the edgeql query with every different input. This allows the server to cache it.

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
```diff
@@ -0,0 +1,100 @@
...
+    const query = e.select(
+      e.select(progressReport).project.members.user,
+      () => ({
+        id: true,
+        email: true,
+        roles: true,
+      }),
+    );
+    const users = await this.db.run(query);
+    return users.map((user) => ({
```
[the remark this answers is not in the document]
atGit2021: I get the following error, `Types of property 'roles' are incompatible.` when I don't map through the roles, e.g. ` roles: user.roles.map((role) => role),`

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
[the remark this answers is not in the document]
bryanjnelson: Ah.  Well, that makes sense.  It's unfortunate, but it makes sense.

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
```diff
@@ -76,7 +76,7 @@ export class ProgressReportWorkflowNotificationHandler
         this.prepareNotificationObject(
           reportId,
           previousStatus,
-          userId ? { userId } : { email },
+          userId ? { userId } : email ? { email } : { email: '' },
```
CarsonF, line 79: Empty strings to hide type errors are a big ick for me.
Instead, in `getProjectNotifyees`, member's without email addresses can be `flatMap`'ped out.

CarsonF (changes requested):
Thanks @atGit2021! Just a couple things.

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -81,33 +79,7 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
...
-  ) {
-    return await this.recordEvent(report, { status, notes }, session);
-  }
-
-  private async recordEvent(
-    report: ID,
-    props: Record<string, any>,
-    session: Session,
-  ) {
+  async recordEvent(report: ID, props: Record<string, any>, session: Session) {
```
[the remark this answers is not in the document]
atGit2021: @CarsonF using the `ExecuteProgressReportTransitionInput ` type allows for 'undefined' to be passed in to props.status which is not in the ProgressReportStatus Enum.  Should I coalesce the value to "Not Started" if undefined is received or should I code a guard to not update that property if undefined is received?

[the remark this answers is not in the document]
CarsonF: Ah. You can just adjust it to be required.
```ts
SetRequired<ExecuteProgressReportTransitionInput, 'status'>
```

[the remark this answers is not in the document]
atGit2021: @CarsonF I modified recordEvent to use the new type but I'm getting an error, see below.  I'm not sure how to make the types on both the edge and neo4j `recordEvent` method match.

Property 'recordEvent' in type 'ProgressReportWorkflowEdgeDBRepository' is not assignable to the same property in base type 'PublicOf<ProgressReportWorkflowRepository>'.
  Type '({ report, ...props }: { readonly report: ID; readonly transition?: ID | undefined; readonly notes?: RichTextDocument | undefined; readonly status: "NotStarted" | "InProgress" | "PendingTranslation" | "InReview" | "Approved" | "Published"; }, _session: Session) => Promise<...>' is not assignable to type '(report: ID, props: Record<string, any>, session: Session) => Promise<UnsecuredDto<ProgressReportWorkflowEvent>>'.
    Types of parameters '__0' and 'report' are incompatible.

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
atGit2021: @CarsonF , I'm having trouble getting rid of the null email values. I pushed up a version using flatMap in the `getProjectNotifyees` method, so you can see the code.

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -216,7 +173,7 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
       .where({ 'email.value': inArray(emails) })
       .return<{
         id: ID;
-        email: string;
+        email: string | null;
```
CarsonF, line 176: You can revert this change. We know the emails exist, that's our input/starting point.

src/components/progress-report/workflow/progress-report-workflow.repository.ts
```diff
@@ -180,8 +152,8 @@ export class ProgressReportWorkflowRepository extends DtoRepository(
       ])
       .return<{
         id: ID;
-        email: string;
-        roles: Role[];
+        email: string | null;
```
CarsonF, line 155: Let's revert this type change, and change the optional match on the email address to a required one.
Let the DB do the filtering for us.

src/components/progress-report/workflow/handlers/progress-report-workflow-notification.handler.ts
CarsonF: I couldn't get the flatMap to work with types either, not exactly sure why.
In the end, it's better to just to align the DB queries closer to our need.
See my other comments in the repo.

src/components/progress-report/workflow/progress-report-workflow.repository.ts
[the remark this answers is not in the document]
CarsonF: You updated the signature in the edgdb repository but not the neo4j repository. They need to match.

CarsonF (changes requested):

src/components/progress-report/workflow/progress-report-workflow.edgedb.repository.ts
```diff
@@ -57,37 +57,39 @@ export class ProgressReportWorkflowEdgeDBRepository
...
+      with emails := array_unpack(<array<str>>$emails),
+      users := (select User filter .email in emails)
+      select users {
+        id,
+        email := assert_exists(.email),
+      } 
+    `);
+    return await this.db.run(query, { emails });
   }
+  // async getUserIdByEmails(emails: readonly string[]) {
```
bryanjnelson, line 79: Probably just forgot to delete this?

atGit2021: oops, forgot to save the file after deleting it.  will push another commit now

CarsonF (approved):
Looks good! Thanks @atGit2021

PR #3128, opened 2024-03-21
clipanion ← commander / nestjs-console

CarsonF:
This replaces the `nestjs-console` library, which is a wrapper for `commander`, with `clipanion` library.
The former was not type safe, and functionality was limited to whatever the wrapper library exposed.

`clipanion` is type safe. The big change here, is each command is its own class. Since class fields are used to configure the command in a type safe way.

Command discovery and DI were trivial, so I just did them myself.

This library is better at exposing command proxies as well which I'll use in #3121.


bryanjnelson (approved):
Onward!!

PR #3129, opened 2024-03-22
[EdgeDB] Repo.customize allow omitting default methods

CarsonF:
This provides a way to omit default methods when customizing without providing a replacement method.

```ts
.customize((cls, { defaults }) => {
  return class extends cls {

    static omit = [defaults.create, defaults.update, defaults.delete];

  };
})
```


bryanjnelson (approved):
Slick.

PR #3130, opened 2024-03-22
[EdgeDB] Seed script execution v2

CarsonF:
Scripts still work the same as before.
Driving this change is the need to disable access policies while seeding 4985f3959b3ca97029cdab4f6970443b78a2e9a3.
Seeds do not have a user configured so #3126 fails to apply seeds as they are now assumed to executed as an "anonymous user".

# TS seed files
While I was here I added support for seeds to be TS files.

These TS files should export a default function which will be executed in the appropriate order.

They can `return` or `yield` one or more query strings or query builder queries.
They will be executed in order, and the results printed, just like the edgeql files.
```ts
// dbschema/seeds/777.something.ts
import { SeedFn } from '~/core/edgedb/seeds.run';

export default (async function* ({ e }) {

  yield 'select 1';

  yield e.select(2);

} satisfies SeedFn);
```

Queries can also be executed explicitly for result manipulation.
```ts
export default (async function ({ e, db, print }) {

  const root = await e
    .assert_exists(e.select(e.RootUser).assert_single())
    .run(db);

  print(`Root User: ${root.id}`);

} satisfies SeedFn);
```


.eslintrc.cjs
```diff
@@ -191,6 +191,13 @@ const config = {
         'no-console': 'off',
       },
     },
+    {
+      files: './dbschema/seeds/*.ts',
+      rules: {
+        'import/no-default-export': 'off',
+        // 'import/no-named-export': 'error',
```
andrewmurraydavid, line 198: Is the comment left intentionally?

package.json
```diff
@@ -22,7 +22,7 @@
     "repl": "yarn start --entryFile repl",
     "gen-id": "ts-node -Te \"let g = require('./src/common/generate-id').generateId; for (let i = 0, l = process.argv[1] || 1; i < l; i++) { g().then(console.log) }\"",
     "edgedb:gen": "node --loader ts-node/esm/transpile-only --experimental-loader=extensionless src/core/edgedb/generator/run.ts",
-    "edgedb:seed": "cat dbschema/seeds/*.edgeql | edgedb",
+    "edgedb:seed": "node --loader ts-node/esm/transpile-only --experimental-loader=extensionless src/core/edgedb/seeds.run.ts",
```
andrewmurraydavid, line 25: Not looking for a change, but UGH, this is where I _**love**_ bun...

src/core/edgedb/seeds.run.ts
```diff
@@ -0,0 +1,98 @@
...
+    for (const file of files) {
+      if (file.endsWith('.edgeql')) {
+        const query = await fs.readFile(file, 'utf-8');
+        await runAndPrint(query);
+      } else {
+        const script = await import('../../../' + file);
+        const queries = await (script.default as SeedFn)(params);
+        if (!queries) {
+          continue;
+        }
```
andrewmurraydavid, line 75: It's a nice approach, but 72 just looks weird, casting and doing a call on what was casted. I'm sure this'll be a low maintenance part of our code, so I'm ok leaving it as is for now, but that like gives me the "icks".

andrewmurraydavid (approved):
Overall looks good. Nice work, @CarsonF !
I just left some thought, but nothing that needs addressing.

.eslintrc.cjs
CarsonF: Yeah I thought about restricting seeds to not export anything but their own default execution function.
But then I thought that some seed helpers might emerge from sister files, so I didn't want to primitively block that.

package.json
CarsonF: Yeah I want to get there too. We need to refactor the TS keyof transformer first. Currently we need TS to compile our code to get that transformation done. Once that's done, we can compile/transpile with another faster tool.

src/core/edgedb/seeds.run.ts
CarsonF: I think I didn't like my variable name for this fn & typecast so my immediate next thought was just to inline it lol

bryanjnelson (approved):
🚢

PR #3132, opened 2024-03-26
Fix `delete Object` usage

CarsonF:
- Doing `delete Object` would fail on the two trigger assertions.
  - These were to prevent mistakes, but these actions should
only happen in one spot in the schema or app code.
    So I'm fine dropping them to make deletes more ergonomic.
  - FWIW both of these should work from my perspective
with `delete Object` so maybe they are bugs.
- Deleting `Partner` or `Organization` deletes the other.
- Deleting `Language`/`Project`/`Org` deletes their `ProjectContext`
- Deleting `Scripture::VerseRanges` are allowed if their `Scripture::Collection` is being deleted too
- Marking migration files as generated so GitHub doesn't render them by default (or count them in diff stats)

### DIFF
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -97,14 +97,6 @@ module default {
         # projects := {__subject__},
       });
     }
-    # Setting the new project as its own context should be the immediate next thing that happens
-    # So enforce that that happens (as best we can), and assert that the context is ever only itself.
-    trigger enforceContext after update for each do (
-      assert(
-        __new__ in __new__.projectContext.projects and count(__new__.projectContext.projects) = 1,
-        message := "A Project's own context should be itself (no more or less)"
-      )
-    );
     
     trigger createBudgetOnInsert after insert for each do (
       insert default::Budget {
--- a/dbschema/project.esdl
+++ b/dbschema/project.esdl
@@ -96,6 +96,7 @@ module default {
         # https://github.com/edgedb/edgedb/issues/3960
         # projects := {__subject__},
       });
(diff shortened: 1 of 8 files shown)


dbschema/scripture.esdl
```diff
@@ -7,6 +7,7 @@ module Scripture {
     multi verses: VerseRange {
       readonly := true;
       on source delete delete target;
+      on target delete deferred restrict;
```
bryanjnelson, line 10: Similar question here...why exactly is this line needed?

bryanjnelson (approved):
🏅

dbschema/project.esdl
```diff
@@ -96,6 +96,7 @@ module default {
         # https://github.com/edgedb/edgedb/issues/3960
         # projects := {__subject__},
       });
+      on source delete delete target;
```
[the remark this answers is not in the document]
CarsonF: It deletes `target` == `Project::Context`

dbschema/scripture.esdl
CarsonF: I believe when we say `delete Object` it's compiled to some SQL like
```sql
delete from Users;
delete from Scripture::VerseRange;
delete from Scripture::Collection;
```
Since abstract queries are compiled to a union of their concretes.
And in this case `VerseRange` is enumerated before `Collection`, so an error is thrown.
The `deferred restrict` means the VerseRange can be deleted, as long as the `Collection` is also deleted within the same transaction.

PR #3133, opened 2024-03-26
add pull request template

andrewmurraydavid:
## Reason for this PR
We discussed about trying out pull request templates in our repos.

## Description
This PR adds the pull request file.

- [N/A] Change the task url above to the actual Monday task

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,31 @@
+<!--- replace the text in parentheses with Monday link for your task -->
+## [Monday task](https://seed-company-squad.monday.com/boards/.../pulses/...)
+
+<!--- or if you don't have a Monday task, you can remove the line above and give the reason here -->
+<!--- 
+## Reason for this PR
+...
+
+-->
+
+## Description
+<!--- Describe the changes in this pull request and include screenshots if needed -->
+
+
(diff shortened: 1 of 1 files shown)


bryanjnelson:
In regards to the "Reason for this PR" section...do we _only_ use this if we don't have a Monday task?

andrewmurraydavid:
> In regards to the "Reason for this PR" section...do we _only_ use this if we don't have a Monday task?

Not necessarily. I can see this being used in other use cases as well. Maybe if the task and/or story does not express any details as to what needs to be worked.

bryanjnelson:
So "Reason for this PR" / Monday Task is more of the **why**...and the "Description" would be more of the **what**?

andrewmurraydavid:
> So "Reason for this PR" / Monday Task is more of the **why**...and the "Description" would be more of the **what**?

Correct. At least that's what I was envisioning. Happy to adapt it to something that would make more sense.

bryanjnelson (approved):
I'd say this is a fine start for us.  We can always add/remove items later as desired.

Thanks @andrewmurraydavid!

willdch (approved):
This looks great to me!
- [x] 🤙🏼

CarsonF (approved):
All I see is boilerplate I need to delete or avoid while reading the description text.
But if this solves some problem for people, it's fine with me.

PR #3134, opened 2024-03-26
[CI] Inject access policies before checking migration status

CarsonF:


andrewmurraydavid (approved):
Makes sense why my PR was failing! Thanks for the fix!

PR #3138, opened 2024-04-02
Stop using project level scope

andrewdavid48:
## Reason for this PR
A change was made a while ago that moved the project level scope to context-aware. This is tech debt to clean this up.

## Description
Move away from project level scope and use session (context-aware) scope.

- [N/A] Add/update tests if needed (Not applicable, all tests should pass)

### DIFF
--- a/src/components/budget/budget.service.ts
+++ b/src/components/budget/budget.service.ts
@@ -226,13 +226,12 @@ export class BudgetService {
       session,
       viewOfChangeset(changeset),
     );
-    const budget = await this.readOne(br.parent.properties.id, session);
 
     const changes = this.budgetRecordsRepo.getActualChanges(br, input);
     this.privileges.for(session, BudgetRecord, br).verifyChanges(changes);
-    this.privileges
-      .for(session, Budget, budget)
-      .verifyCan('edit', 'universalTemplateFile');
+    // TODO: add this back after the Budget "edit" if fixed
+    // const budget = await this.readOne(br.parent.properties.id, session);
+    // this.privileges.for(session, Budget, budget).verifyCan('edit');
 
     try {
       const result = await this.budgetRecordsRepo.update(
--- a/src/components/budget/budget.service.ts
+++ b/src/components/budget/budget.service.ts
@@ -1,5 +1,4 @@
 import { Injectable } from '@nestjs/common';
-import { intersection } from 'lodash';
 import {
   DuplicateException,
   generateId,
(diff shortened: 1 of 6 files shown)


src/components/budget/budget.service.ts
```diff
@@ -237,7 +236,9 @@ export class BudgetService {
 
     const changes = this.budgetRecordsRepo.getActualChanges(br, input);
     this.privileges.for(session, BudgetRecord, br).verifyChanges(changes);
-    this.privileges.for(session, Budget, budget).verifyCan('edit');
+    this.privileges
+      .for(session, Budget, budget)
+      .verifyCan('edit', 'universalTemplateFile');
```
CarsonF, line 235: Why make this change?

andrewmurraydavid: It breaks with a 403 without it.

andrewmurraydavid: After removing the `this.privileges.for(session, Budget, budget).verifyCan('edit');`, the update goes through. It seems that the edit policy privilege might be messing things up ([talking about this here](https://github.com/SeedCompany/cord-api-v3/pull/3138/files#diff-2ce6dc91076b6679b82573512d72c4b924876b2bde1e363c03f9353d4eb1e782R34)).

CarsonF: > In more detail, the `verifyCan` was expecting a `prop` after the action.

No. It can be either the object level action or an action on a specific property.
The specific property was incorrect, that's not what we were trying to confirm edit access too.



> Is the 403 happening now because of the changes that were made in the `ProjectManagerPolicy` or `ProjectService.currentBudget`?

Idk, both? They are correlated. The idea is to move the check into the policies.
The only tricky part is we need to check this on BudgetRecord, but it doesn't have a status field to check against. Hence the added Budget check here. But obv something wasn't right.

bryanjnelson: Okay.  Andrew just put that property there to get it working; he knew it wasn't a final solution.

I get that we're moving the checks into the policies.  Question though, why was the `canEditFinalized` call removed?

CarsonF: It was supposed be replaced by `verifyCan('edit')` and the policies would do it for us.

CarsonF: Trying here #3139 
Seeing a failure with engagement create flag, but not budgets. Does the test suite not cover this?

andrewmurraydavid:
This will be closed by #3141

PR #3140, opened 2024-04-03
[EdgeDB] `currentUserId` global refactor

CarsonF:
This follows up on #2941

# Interceptor fail

[Previously](https://github.com/SeedCompany/cord-api-v3/commit/30238eef49e609fa1b68d97bf122d9d745916115), the `currentUserId` was put in the EdgeDB globals context via a NestJS Interceptor.

But _interceptors_ only cover the main query/mutation method (like a _controller_).
GQL field resolvers are walked/invoked after this lifecycle as needed.
So those field resolvers would not have the currentUser set ❌ 

Aside: I might have been thrown off here, because the transaction context _is_ applied with an [interceptor](https://github.com/SeedCompany/cord-api-v3/blob/develop/src/core/database/abstract-transactional-mutations.interceptor.ts#). But this is fine because DB changes only happen within the main mutation method. Field resolvers after that are just reading back the new DB state.

# Session updates missed

Additionally, queries/mutations that deal with authentication did not work either. For `session`, that interceptor was [skipped](https://github.com/SeedCompany/cord-api-v3/blob/30238eef49e609fa1b68d97bf122d9d745916115/src/components/authentication/session.interceptor.ts#L77), because the session was being created within that resolver (after interceptors).
For `login/register`, the session was created as anonymous, and the later [updated](https://github.com/SeedCompany/cord-api-v3/blob/bb15f216984349bb091976afc7b5c329e6ee854a/src/components/authentication/login.resolver.ts#L36-L36) to be for that user after the login succeeded. So this change wasn't picked up and applied to the DB global.

# Interceptor -> Middleware

This is fixed by using a _middleware_ for the options ALS context.
Middleware's _do_ compass the field resolvers, and the entire request.

Aside: this could be multiple GQL operations (queries/mutations), but this is fine because the current user is given via an HTTP header, so all operations within the request are the same.

# BehaviorSubjects - to set new option later

However, because middleware's cover the entire request, all we have at that point _is_ the request.
We'd have to process it to get the given session token, and then do a DB lookup to get the associated user ID.
We need a way to tell our EdgeDB code to use "these" options, but be able to update those options later, when we actually have the value.

In c866579cd1facb79754477b7a67df350d6415ad4, I change the options ALS to hold rxjs `BehaviorSubjects`. These have a current value, but are also `Subjects` which can have new values pushed to them. They are also `Observables`, so the value changes can be subscribed to as well.
The cleverest thing here 😅 is that the ALS holds a `BehaviorSubject` of the current _function_ that reduces the previous options and a `BehaviorSubject` of the parent options.
This allows any node in the context tree to change its options, and of its descendants will refresh appropriately.

This sounds complicated, but it practice it makes much more sense.
```ts
const currentUser$ = new BehaviorSubject(
  options => options // no change for now
);
EdgeDB.usingOptions(currentUser$, () => {
  // globals now: { }

  EdgeDB.usingOptions(options => options.withGlobals({ color: 'blue' }), () => {
    // globals now { color: 'blue' }

    // now we have the current user
    currentUser$.next(options => options.withGlobals({ currentUserId: '...' })
    // globals in the outer closure: { currentUserId: '...' }
    // globals in this closure: { currentUserId: '...', color: 'blue' }
  });

  // Again, globals here: { currentUserId: '...' }
});
```

[Here's](https://github.com/SeedCompany/cord-api-v3/blob/0bb6a4d27ea33f448a25df6bd6801f6e26229ecb/src/components/authentication/current-user.provider.ts#L33-L42) the middleware that sets creates the ALS context scope for the request, with a `BehaviorSubject` (initialized as a noop).

# BehaviorSubjects - to receive new Session value

Now we have a way to update that global later when we get it, but how do we do that?
First step: f02b332 - convert the `session` held in the _GQL context_ to a `BehaviorSubject`.
This allows others to subscribe to session changes as they happen.
This fixes the problem with `login/register` where the `session` is updated later, but our "current user logic" is not informed of this.
Then we use an Interceptor again to acquire this GQL context and its `session` `BehaviorSubject`.
Then we can [connect the two BehaviorSubjects together](https://github.com/SeedCompany/cord-api-v3/blob/0bb6a4d27ea33f448a25df6bd6801f6e26229ecb/src/components/authentication/current-user.provider.ts#L54-L57).

# Wrap up

Essentially all of this boils down to:
```
new session value -> set db global: { currentUserId: session.userId }
```

It was a lot of work to get here, but I'm pretty happy with the result.
Setting this global is hard, because:
- we have to do logic first to determine it.
- we have to know where the global should apply (multiple requests at same time use different values).
- EdgeDB options are immutable, so mutating it requires an entire new identity. But there can be several spots applying options, that need to work in harmony.

The end result leaves this one class, [`EdgeDBCurrentUserProvider`](https://github.com/SeedCompany/cord-api-v3/blob/0bb6a4d27ea33f448a25df6bd6801f6e26229ecb/src/components/authentication/current-user.provider.ts), responsible for this connection. And it is loosely coupled to how the current `session` is determined.

https://seed-company-squad.monday.com/boards/5989610236/pulses/6374337425

### DIFF
--- a/src/common/context.type.ts
+++ b/src/common/context.type.ts
@@ -1,5 +1,6 @@
 import { Request, Response } from 'express';
 import { OperationDefinitionNode } from 'graphql';
+import { BehaviorSubject } from 'rxjs';
 import { Session } from './session';
 
 /**
@@ -9,5 +10,5 @@ export interface GqlContextType {
   operation: OperationDefinitionNode;
   request?: Request;
   response?: Response;
-  session?: Session;
+  readonly session$: BehaviorSubject<Session | undefined>;
 }
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -41,11 +41,12 @@ export function loggedInSession(session: Session): Session {
   return session;
 }
 
-const sessionFromContext = (context: GqlContextType) => {
-  if (!context.session) {
+export const sessionFromContext = (context: GqlContextType) => {
+  const session = context.session$.value;
+  if (!session) {
     throw new NoSessionException();
   }
-  return context.session;
+  return session;
 };
 
 export const LoggedInSession = () =>
--- a/src/components/admin/admin.edgedb.repository.ts
+++ b/src/components/admin/admin.edgedb.repository.ts
@@ -24,7 +24,7 @@ export class AdminEdgeDBRepository extends AdminRepository {
       hash: u['<user[is Auth::Identity]'].passwordHash,
     }));
     const query = e.assert_exists(e.assert_single(rootUser));
-    const user = await this.edgedb.withOptions(noAPs, () =>
+    const user = await this.edgedb.usingOptions(noAPs, () =>
       this.edgedb.run(query),
     );
     return {
@@ -48,7 +48,7 @@ export class AdminEdgeDBRepository extends AdminRepository {
           set: { passwordHash },
         })),
       }));
-    await this.edgedb.withOptions(noAPs, () => this.edgedb.run(query));
+    await this.edgedb.usingOptions(noAPs, () => this.edgedb.run(query));
   }
 
   async checkDefaultOrg() {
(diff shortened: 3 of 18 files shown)


bryanjnelson (approved):
Excellent work, @CarsonF!

PR #3141, opened 2024-04-03
Move budget record non-finalized check to policies

CarsonF:
This fixes the `budget.record.canEdit` flag to correctly emit `false` if the budget is finalized & you/FPM cannot edit finalized budgets.
It's also much cleaner.

### DIFF
--- a/src/components/authorization/policies/by-role/project-manager.policy.ts
+++ b/src/components/authorization/policies/by-role/project-manager.policy.ts
@@ -32,7 +32,7 @@ const stepsUntilFinancialEndorsement = takeWhile(
     ]),
 
     r.Budget.read.when(member).edit,
-    r.BudgetRecord.read.when(member).edit,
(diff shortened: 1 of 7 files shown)


src/components/authorization/policies/by-role/project-manager.policy.ts
```diff
@@ -32,7 +32,7 @@ const stepsUntilFinancialEndorsement = takeWhile(
     ]),
 
     r.Budget.read.when(member).edit,
-    r.BudgetRecord.read.when(member).edit,
+    r.BudgetRecord.read.whenAll(member, field('status', 'Pending')).edit,
```
andrewmurraydavid, line 35: This is nice approach to this whole debacle!

andrewmurraydavid (approved):
Awesome work!! 👏

bryanjnelson (commented):
Nice!!

PR #3142, opened 2024-04-03
Change departmentId to str / Fix Project seeds

CarsonF:
I previously assumed it could be cast in the query, but this only works if the field name is changed.
So for app code to function the same, this has to be the right type in schema.

This also allows leading zeros, which could be coming with multiplication projects.

The constraint is changed to just expect a positive 5-digit integer.

Also running the rewrite on insert to enforce constraints on projects created as active.

---

This also exposes the error in project seeds, that projects didn't have a primary location that should have. I added those.

I also discovered that multiple projects which would have a departmentId generated cannot be inserted at the same time, because the logic doesn't account for newly inserted IDs. So it would give the same ID multiple times, leading to constraint errors.

To overcome I migrated the seed file to TS, and run each project insert one at a time.


andrewmurraydavid (approved):
Great work as always!!
I love how "simple" this ended up being 😆

dbschema/seeds/008.projects.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (newProjects.length === 0) {
+    return;
+  }
+
+  for (const { type, ...project } of newProjects) {
+    const insertQ = e.insert(e[`${type}Project`], {
+      ...project,
+      ...mapValues.fromList(
+        ['mouStart', 'mouEnd', 'estimatedSubmission'],
+        (key, { SKIP }) =>
```
bryanjnelson, line 119: @CarsonF - Can you explain what is happening with this `SKIP` business?

CarsonF: If the `project` doesn't specify that key (or rather it's falsy), the pair is omitted from the resulting object.
It would be similar to using `undefined` instead, and expecting code receiving that object to ignore that pair. Most things do, as does the QB code, so it's tomateo tomatto.

PR #3143, opened 2024-04-04
Move engagement "create if in dev" to policies

CarsonF:


src/components/project/project.service.ts
```diff
@@ -369,10 +369,7 @@ export class ProjectService {
     return {
       ...result,
       canRead: perms.can('read', 'engagement'),
-      canCreate:
-        perms.can('create', 'engagement') &&
-        (project.status === ProjectStatus.InDevelopment ||
-          session.roles.includes('global:Administrator')),
+      canCreate: perms.can('create', 'engagement'),
```
andrewmurraydavid, line 372: So much nicer!

src/components/engagement/engagement.service.ts
```diff
@@ -211,7 +208,10 @@ export class EngagementService {
 
   private async verifyCreateEngagement(projectId: ID, session: Session) {
     const project = await this.resources.load(IProject, projectId);
-    const projectPrivileges = this.privileges.for(session, IProject, project);
+    const projectPrivileges = this.privileges.for(session, IProject, {
+      ...project,
+      project,
+    } as any);
```
andrewmurraydavid, line 214: I'm curious why the awkward spreading here.

src/components/authorization/policies/by-feature/engagements-create-delete.policy.ts
```diff
@@ -3,7 +3,7 @@ import { field, Hierarchies, member, Policy } from '../util';
 @Policy([...Hierarchies.Field, ...Hierarchies.Finance], (r) => [
   r.Project,
   r.Engagement.read,
-  r.Engagement.when(member).create,
+  r.Engagement.whenAll(member, field('project.status', 'InDevelopment')).create,
```
andrewmurraydavid, line 6: Path traversal is so cool! It simplifies a lot the writing and reading of these policies!

src/components/engagement/engagement.service.ts
CarsonF: The project object is used as the context object when checking "can create engagement".
It works because project has the same attributes as engagement: sensitivity, isMember, etc.
But the condition I just added is looking for `context.project`, so we also need project at a sub key here.

What would be better altogether is to check the create permission after creating it.
At that point we'd have the object like normal. And then we can just rollback the transaction if permissions fail.
Also EdgeDB will be checking this too with APs...so these statements could also just go away after migration.

andrewmurraydavid: That makes sense!

I'm ok with this, especially if we can move away from it. Normally I'd want to provide what specific fields are needed or trim down a copy of the `project`, but it's ok for now.

CarsonF: It's hard to know what's needed is the problem. The policies are very flexible. I believe it could be possible to add a build step to generate the context types for each resource, based on what's used in policies.

CarsonF:
Missed a second spot that needed the `project` added to the privileges context. Tests should pass now.

PR #3144, opened 2024-04-04
Fix omitting `customize` `omit` acting like omitting all default methods

CarsonF:
@andrewmurraydavid @bryanjnelson both of you guys hit this with project & language repos.
With no omit array given, TS inferred the `OmitKeys` type to be all of the enum members.
Giving it a default value of `never` suppresses that, so it will infer if the array is given or fallback to this default value which removes nothing.
Honestly TS is amazing.


bryanjnelson (approved):
Thanks @CarsonF!

andrewmurraydavid:
Let's go!! Thanks!

PR #3146, opened 2024-04-04
Adjust imports/exports

CarsonF:
- Use `~/common` & `~/core`
- Import/export DTOs always from their folder instead of rollup to component folder
- Stop rolling up database (neo4j) to core
- Shorten where possible


bryanjnelson (approved):
Yuge!  🎉

PR #3149, opened 2024-04-04
Fix TranslationProject sensitivity sync

CarsonF:
Mostly reverts 5829d35e588f7e20eee71e29c8569602df2a7c26
I'm not sure what the error was back then, but it works fine now in 4.7

The actual problem was that we were comparing the engagement to language list, so of course there was no overlap.
```diff
- in .languages
+ in .engagements
```

But the former expression of just selecting the linked project works and should be more performant. So going with that.

### DIFF
--- a/dbschema/engagement.esdl
+++ b/dbschema/engagement.esdl
@@ -84,20 +84,18 @@ module default {
     
     trigger recalculateProjectSensOnInsert after insert for each do (
       update (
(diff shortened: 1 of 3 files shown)


andrewmurraydavid (approved):
Nice!

bryanjnelson (approved):
:shipit:

PR #3151, opened 2024-04-05
[EdgeDB] User Education Queries

atGit2021:
[Monday Task](https://seed-company-squad.monday.com/boards/5989610236/views/131962501/pulses/5792981500)

## Description

Create EdgeDB UserEducation repository

### DIFF
--- a/src/components/user/education/education.edgedb.repository.ts
+++ b/src/components/user/education/education.edgedb.repository.ts
@@ -1,0 +1,23 @@
+import { Injectable } from '@nestjs/common';
+import { ID, PublicOf } from '~/common';
+import { e, RepoFor } from '~/core/edgedb';
+import { Education } from './dto';
+import { EducationRepository } from './education.repository';
+
+@Injectable()
+export class EducationEdgeDBRepository
+  extends RepoFor(Education, {
+    hydrate: (education) => education['*'],
+  }).withDefaults()
+  implements PublicOf<EducationRepository>
+{
+  async getUserIdByEducation(id: ID) {
+    const edu = e.cast(e.User.Education, e.uuid(id));
+    const query = e.assert_exists(
+      e.select(e.User, (user) => ({
+        filter_single: e.op(edu, 'in', user.education),
+      })),
+    );
+    return await this.db.run(query);
+  }
+}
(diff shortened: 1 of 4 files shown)


src/components/user/education/education.edgedb.repository.ts
```diff
@@ -0,0 +1,29 @@
...
+  implements PublicOf<EducationRepository>
+{
+  async getUserIdByEducation(id: ID) {
+    const edu = e.cast(e.User.Education, e.uuid(id));
+    const query = e.select(e.User, (user) => ({
+      id: true,
+      filter_single: e.op(edu, 'in', user.education),
+    }));
+    const result = await this.db.run(query);
+    if (!result) {
```
bryanjnelson, line 21: @CarsonF - I'm not sure if we want to actually throw here?  It is conceivable that not all users will have education data, so I would think that an empty set would be reasonable to return?

src/components/user/education/education.edgedb.repository.ts
```diff
@@ -0,0 +1,29 @@
...
+export class EducationEdgeDBRepository
+  extends RepoFor(Education, {
+    hydrate: (education) => education['*'],
+  }).withDefaults()
+  implements PublicOf<EducationRepository>
+{
+  async getUserIdByEducation(id: ID) {
+    const edu = e.cast(e.User.Education, e.uuid(id));
+    const query = e.select(e.User, (user) => ({
+      id: true,
```
bryanjnelson, line 17: @CarsonF - Is this `id` necessary?  Wouldn't that just be the default?

CarsonF: Not sure, try removing lol

src/components/user/education/education.edgedb.repository.ts
CarsonF: This shouldn't happen since the education cannot exist without the user.
I would just wrap in an `e.assert_exists` instead of throwing.

bryanjnelson (approved):
LGTM @atGit2021!  Thanks!

On to you @CarsonF.

CarsonF (approved):
LGTM thanks @atGit2021!

PR #3137, opened 2024-04-02
[EdgeDB] Project queries

andrewmurraydavid:
## Description
Added Project queries for EdgeDB

- [x] Change the task url above to the actual Monday task
- [N/A] Add/update tests if needed
- [x] Add reviewers to this PR

PR #3152, opened 2024-04-05
[EdgeDB] castToEnum QB helper

CarsonF:
This is an unsafe helper to assert that the given str expression
is actually a string literal union of the given enum.

```ts
const expr = e.str('red') // edgeql str literal whose TS value is `string`

castToEnum(expr, ['red', 'blue']) // expression TS value is now `'red' | 'blue'`
```


andrewmurraydavid (approved):
Yes!! Thank you! I need that in my #3137 PR!

PR #3153, opened 2024-04-07
[EdgeDB] Generate APs for update statements

CarsonF:
Following up on #3126 

Updates policies were previously skipped because we want the app to handle them.
But we actually have to declare them, otherwise EdgeDB defaults to "no access".

- Set `update read` to match `selects`.
  This just means that anything that can be read with a `select` query can be read with an `update` query.

- Set `update write` to always allow.
  The API will handle edit permissions.
  But we need to declare these so the DB knows to ignore.

Schema diff here: https://github.com/SeedCompany/cord-api-v3/commit/3b556c8fa04dc303a01be624a3a9e5638ccf4edd


bryanjnelson (approved):
Makes sense!

PR #3156, opened 2024-04-08
[EdgeDB] Partnership seeds

brentkulwicki:
## Description
Seeding for Partnerships in EdgeDB

### DIFF
--- a/dbschema/seeds/017.partnerships.edgeql
+++ b/dbschema/seeds/017.partnerships.edgeql
@@ -1,49 +1,39 @@
 with
 partnershipsJson := to_json($$[
   {
-      "project": "South Downs",
-      "partner": "Rohan Linguistics",
-      "types": ["Funding"],
-      "agreementStatus": "NotAttached",
-      "primary": true
+    "project": "South Downs",
+    "partner": "Rohan Linguistics",
+    "types": ["Funding"]
   },
   {
-      "project": "Emyn Muil",
-      "partner": "Ered Luin Translation Syndicate",
-      "types": ["Funding"],
-      "agreementStatus": "NotAttached",
-      "primary": true
+    "project": "Emyn Muil",
+    "partner": "Rohan Linguistics",
+    "types": ["Funding", "Managing"],
+    "financialReportingType": "FieldEngaged"
   },
   {
-        "project": "Emyn Muil",
-        "partner": "Rohan Linguistics",
-        "types": ["Funding", "Managing"],
-        "financialReportingType": "FieldEngaged",
-        "agreementStatus": "NotAttached",
-        "primary": true
+    "project": "Emyn Muil",
+    "partner": "Ered Luin Translation Syndicate",
+    "types": ["Funding"],
+    "primary": false
   },
   {
-      "project": "Emyn Muil",
-      "partner": "Rhun For Zero",
-      "types": ["Impact"],
-      "agreementStatus": "NotAttached",
-      "primary": false
+    "project": "Emyn Muil",
+    "partner": "Rhun For Zero",
+    "types": ["Impact"],
+    "primary": false
   },
   {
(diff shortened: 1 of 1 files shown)


willdch (approved):
This looks great to me!

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,65 @@
+with
+partnershipsJson := to_json($$[
```
willdch, line 2: Just a questions as I have not seen this before.  What is the purpose of the `$$` at the beginning and end of the array?

CarsonF (commented):
Looks good! It would be good to have a few projects with multiple partnerships though. And at least one case with multiple _funding_ partners.

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,65 @@
+with
+partnershipsJson := to_json('[
```
CarsonF, line 2: FWIW I prefer the `$$` for these huge json strings. Easier to not mess up escapes.

brentkulwicki: Will change back. I just updated to be similar to what everyone else had, but I like $$ too

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,83 @@
+with
+partnershipsJson := to_json($$[
+  {
+      "partner": "Rohan Linguistics",
+      "project": "South Downs",
+      "mouStatus": "NotAttached",
```
CarsonF, line 6: NotAttached is default for mouStatus and agreementStatus, and really I don't think it's used much. Let's just remove from these seeds.

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,83 @@
...
+      "financialReportingType": "Funded",
```
CarsonF, line 7: Only `Managing` type can set this field.
https://github.com/SeedCompany/cord-api-v3/blob/edgedb%2Fbump-gen-lib/src/components/partner/partner.service.ts#L126-L126

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,83 @@
...
+    select (
+      (select Partnership filter .partner = partner and .project = project) ??
+      (insert Partnership {
+        project := project,
+        partner := partner,
+        projectContext := project.projectContext,
+        mouStatus := <str>json_get(entry, 'mouStatus'),
+        financialReportingType := <str>json_get(entry, 'financialReportingType'),
+        agreementStatus := <str>json_get(entry, 'agreementStatus'),
+        primary := <bool>json_get(entry, 'primary'),
```
CarsonF, line 74: Maybe we just assume primary true here?
```suggestion
        primary := <bool>json_get(entry, 'primary') ?? true,
```
And then only specify in json if false

dbschema/seeds/017.partnerships.edgeql
```diff
@@ -0,0 +1,83 @@
...
+      "project": "South Downs",
+      "mouStatus": "NotAttached",
+      "financialReportingType": "Funded",
+      "agreementStatus": "NotAttached",
+      "primary": true,
+      "types": ["Funding"]
+  },
+  {
+      "partner": "Ered Luin Translation Syndicate",
+      "project": "Emyn Muil",
```
CarsonF, line 14: Partnerships are typically thought about by `project` first. Would you mind putting those first here?
Then `partner`, then `types`.

CarsonF (commented):
I think the data is close enough. Hard to see some of it without a UI to correlate, but we can iterate later if we need more code paths.

CarsonF (approved):
Thanks @brentkulwicki!

PR #3159, opened 2024-04-08
[EdgeDB] Fix delete cascade

CarsonF:
Following up on #3132

That PR added a on Project delete, delete ProjectContext.
All project children are also configured to delete with Project as well. ([src](https://github.com/SeedCompany/cord-api-v3/blob/95af5f9041cd91bda38e3c9401571271e94a832e/dbschema/project.esdl#L136))
But this sibling deletion chain was not respecting that sibling deletion chain.

So setting them both to chain deletes works.

In practice, I could see `delete Project::Context` deleting all directly associated objects (like this PR does), but I could also see it restricting because it still has things referenced.
I guess our hand is forced though. These contexts should be rarely referenced anyways, and would/should never be deleted directly.


bryanjnelson:
Thanks @CarsonF!

PR #3161, opened 2024-04-09
[Neo4j] Don't create empty nodes for optional, now null relationships

CarsonF:
Big big facepalm here. Can't believe this has lived for so long, we must've not been using it much. And I think we've made a recent push to fix the "update to nulls" in API/UI.

Creating these empty nodes does affect cardinality, which is why the DOMO sync was throwing a duplicate error.

I'll just drop these empty nodes in prod adhoc, I don't want to create a script.


bryanjnelson (approved):
🩹 :shipit:

PR #3163, opened 2024-04-11
Refactor root user creation

CarsonF:
Now an ID & email can be given via environment.
The default fallback will be a static ID for dev,
and a hash of the email/password for prod
(though we'll configure a static one there too).

For EdgeDB, we now create the RootUser in both seeds and on app boot. This allows seeds to work after wiping data.
This also removes the need to have the root user in the migrations/ folder.


bryanjnelson (approved):
Great work @CarsonF!

PR #3166, opened 2024-04-12
[EdgeDB] User Unavailability Queries

atGit2021:
## Description
Create EdgeDB User Unavailability Queries

### DIFF
--- a/src/components/user/unavailability/unavailability.edgedb.repository.ts
+++ b/src/components/user/unavailability/unavailability.edgedb.repository.ts
@@ -1,0 +1,70 @@
+import { Injectable } from '@nestjs/common';
+import { Range } from 'edgedb';
+import { ID, PublicOf } from '~/common';
+import { e, RepoFor } from '~/core/edgedb';
+import {
+  CreateUnavailability,
+  Unavailability,
+  UpdateUnavailability,
+} from './dto';
+import { UnavailabilityRepository } from './unavailability.repository';
+
+@Injectable()
+export class UnavailabilityEdgeDBRepository
+  extends RepoFor(Unavailability, {
+    hydrate: (unavailability) => ({
+      ...unavailability['*'],
+      start: e.assert_exists(e.range_get_lower(unavailability.dates)),
+      end: e.assert_exists(e.range_get_upper(unavailability.dates)),
+    }),
+  }).customize((cls) => {
+    return class extends cls {
+      async create(input: CreateUnavailability) {
+        const user = e.cast(e.User, e.uuid(input.userId));
+        const inserted = e.insert(e.User.Unavailability, {
+          description: input.description,
+          dates: new Range(input.start, input.end),
+        });
+        const updatedUser = e.update(user, () => ({
+          set: {
+            unavailabilities: { '+=': inserted },
+          },
+        }));
+        const query = e.select(inserted, (u) => ({
+          ...this.hydrate(u),
+          updatedUser, // Attach to query, so it is executed.
+        }));
+        return await this.db.run(query);
+      }
+
+      async update(input: UpdateUnavailability) {
+        const unavailability = e.cast(e.User.Unavailability, e.uuid(input.id));
+        const updated = e.update(unavailability, () => ({
+          set: {
+            description: input.description,
+            dates: e.cast(
+              e.range(e.datetime),
+              e.range(
+                input.start ?? e.range_get_lower(unavailability.dates),
+                input.end ?? e.range_get_upper(unavailability.dates),
+              ),
+            ),
+          },
+        }));
+        const query = e.select(updated, this.hydrate);
+        return await this.db.run(query);
+      }
+    };
+  })
+  implements PublicOf<UnavailabilityRepository>
+{
+  async getUserIdByUnavailability(id: ID) {
+    const unavailability = e.cast(e.User.Unavailability, e.uuid(id));
+    const query = e.assert_exists(
+      e.select(e.User, (user) => ({
+        filter_single: e.op(unavailability, 'in', user.unavailabilities),
+      })),
+    );
+    return await this.db.run(query);
+  }
+}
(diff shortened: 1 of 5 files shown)


atGit2021:
@bryanjnelson @CarsonF  The neo4j dto for unavailability contains a `start` and `end` property, but the edgeql schema uses one property called `dates`  which is a  range<cal::local_date>.   Should I customize a separate create in the edge repo or should I refactor the neo4j repo?  I'm not sure what the correct solution should be.

CarsonF:
> Should I customize a separate create in the edge repo or should I refactor the neo4j repo? I'm not sure what the correct solution should be.

Handle it within the edgedb repo

src/components/user/unavailability/unavailability.edgedb.repository.ts
```diff
@@ -0,0 +1,78 @@
...
+        await this.db.run(query);
+        const results = e.assert_exists(
+          e.select(e.User.Unavailability, (u) => ({
+            filter_single: { id: unavailability.id },
+            ...u,
+            start: e.assert_exists(e.range_get_lower(u.dates)),
+            end: e.assert_exists(e.range_get_upper(u.dates)),
+          })),
+        );
+        return await this.db.run(results);
```
bryanjnelson, line 42: @CarsonF - Would this be possible now that the start/end are up in the hydrate?

```suggestion
          e.select(e.User.Unavailability, (u) => ({
            filter_single: { id: unavailability.id },
          })),
        );
        return await this.db.run(results, this.hydrate);
```

src/components/user/unavailability/unavailability.repository.ts
```diff
@@ -1,7 +1,11 @@
 import { Injectable } from '@nestjs/common';
 import { node, relation } from 'cypher-query-builder';
-import { ChangesOf } from '~/core/database/changes';
-import { ID, ServerException, Session } from '../../../common';
+import {
+  ID,
+  NotFoundException,
+  ServerException,
+  Session,
+} from '../../../common';
```
bryanjnelson, line 8:
```suggestion
} from '~/common';
```

Same for the one below while you're in there.

src/components/user/unavailability/unavailability.service.ts
```diff
@@ -1,14 +1,6 @@
 import { Injectable } from '@nestjs/common';
-import {
-  ID,
-  NotFoundException,
-  ObjectView,
-  ServerException,
-  Session,
-  UnsecuredDto,
-} from '../../../common';
+import { ID, ObjectView, Session, UnsecuredDto } from '../../../common';
```
bryanjnelson, line 2:
```suggestion
import { ID, ObjectView, Session, UnsecuredDto } from '~/common';
```

src/components/user/unavailability/unavailability.edgedb.repository.ts
CarsonF: Yes it would. But you also shouldn't be running two queries.
Try
```ts
const updatedUser = e.update(user, () => ({
  set: {
    unavailabilities: { '+=': unavailability },
  },
}));

// Might need to wrap in assert single/exists
const created = e.select(updatedUser.unavailabilities, (u) => ({
  filter: e.op(u.id, '=', unavailability.id)
}));

const query = e.select(created, this.hydrate);
```

atGit2021: @CarsonF, I tried to get it working but for some reason it seems like the hydrate is not picking up/adding the additional properties.  I pushed the code so it's easier for you to review.

src/components/user/unavailability/unavailability.edgedb.repository.ts
```diff
@@ -0,0 +1,70 @@
...
+import { e, RepoFor } from '~/core/edgedb';
+import {
+  CreateUnavailability,
+  Unavailability,
+  UpdateUnavailability,
+} from './dto';
+import { UnavailabilityRepository } from './unavailability.repository';
+
+@Injectable()
+export class UnavailabilityEdgeDBRepository
```
bryanjnelson, line 13: @CarsonF - I'm not sure why this is showing that the `create` method is only returning an ID.  I would expect it to return the entire option since the `this.hydrate` is in there?

CarsonF: Replied in other thread.

src/components/user/unavailability/unavailability.edgedb.repository.ts
atGit2021: ok, thank you!

CarsonF:
Looks like the edgedb schema has `dates` as a range of `CalendarDates` instead of `DateTimes` like the app expects. I'll fix that in my other PR which also makes schema changes.

src/components/user/unavailability/unavailability.edgedb.repository.ts
```diff
@@ -15,44 +15,44 @@ export class UnavailabilityEdgeDBRepository
...
+            unavailabilities: { '+=': inserted },
           },
         }));
-        const created = e.assert_exists(
-          e.select(updatedUser.unavailabilities, (u) => ({
-            filter: e.op(u.id, '=', unavailability.id),
-          })),
-        );
-        return await this.db.run(created, this.hydrate);
+        const query = e.select(inserted, (u) => ({
```
bryanjnelson, line 33: @CarsonF - So this select statement actually executes the update statement inside of it?

bryanjnelson (approved):

src/components/user/unavailability/unavailability.edgedb.repository.ts
atGit2021: I didn't realize you could "daisy chain" statements/queries either; cool.

CarsonF: Correct.
```edgeql
with inserted := (insert ...)
select inserted {
  *,
  updatedUser := (update User ...)
}
```
No different than
```edgeql
with updated := (update User ...)
select updated {*}
```

CarsonF (approved):

src/components/user/unavailability/unavailability.edgedb.repository.ts
bryanjnelson: 🤯

PR #3171, opened 2024-04-17
Fix permissions for `SessionOutput.impersonator`

CarsonF:
We were using the _impersonatee's_ perms to load the _impersonator_.

So when the impersonatee couldn't view the impersonator, the UI would not be able to **stop impersonating** because it didn't know the impersonator's roles (to conditionally show the toggle).

I think EdgeDB would've thrown a not found error here too.


bryanjnelson (approved):
🧠

PR #3174, opened 2024-04-17
Expose scripture parsing to GQL query as an adhoc util

CarsonF:
Screenshot 2024-04-17 at 2 15 31 PM


atGit2021 (approved):
Tested a couple of strings in Apollo, LGTM!

PR #3182, opened 2024-04-19
Strip superfluous Session coupling

CarsonF:


brentkulwicki (approved):
🤯 ✅

bryanjnelson (approved):
Talk about droppin' the dead weight; onward!

PR #3183, opened 2024-04-22
[EdgeDB] - Ceremony queries

brentkulwicki:
## Description
Ceremony Queries for EdgeDB

## Ready for review checklist

### DIFF
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -2,7 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { NotImplementedException, PublicOf } from '~/common';
 import { castToEnum, RepoFor } from '~/core/edgedb';
 import { CeremonyRepository } from './ceremony.repository';
-import { Ceremony, CeremonyType, CreateCeremony } from './dto';
+import { Ceremony, CeremonyType } from './dto';
 
 @Injectable()
 export class CeremonyEdgeDBRepository
@@ -12,10 +12,14 @@ export class CeremonyEdgeDBRepository
       engagement: true,
       type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
     }),
-  }).withDefaults()
+  }).customize((cls, { defaults }) => {
+    return class extends cls {
+      static omit = [defaults.create];
+    };
+  })
   implements PublicOf<CeremonyRepository>
 {
-  override create(input: CreateCeremony): never {
+  async create(): Promise<never> {
     throw new NotImplementedException();
   }
 }
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -1,8 +1,10 @@
 import { Injectable } from '@nestjs/common';
 import { PublicOf } from '~/common';
+import { DbTypeOf } from '~/core';
+import { ChangesOf } from '~/core/database/changes';
 import { RepoFor } from '~/core/edgedb';
 import { CeremonyRepository } from './ceremony.repository';
-import { Ceremony, CreateCeremony } from './dto';
+import { Ceremony, CreateCeremony, UpdateCeremony } from './dto';
 
 @Injectable()
 export class CeremonyEdgeDBRepository
@@ -11,6 +13,12 @@ export class CeremonyEdgeDBRepository
   }).customize((cls) => {
     return class extends cls {
       async create(input: CreateCeremony): Promise<any> {
+        return;
+      }
+      async update(
+        existing: DbTypeOf<Ceremony>,
+        changes: ChangesOf<Ceremony, UpdateCeremony>,
+      ) {
         return;
       }
     };
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -1,16 +1,8 @@
 import { Injectable } from '@nestjs/common';
-import { PaginatedListType, PublicOf, UnsecuredDto } from '~/common';
-import { DbTypeOf } from '~/core';
-import { ChangesOf } from '~/core/database/changes';
+import { PublicOf } from '~/common';
 import { castToEnum, RepoFor } from '~/core/edgedb';
 import { CeremonyRepository } from './ceremony.repository';
-import {
-  Ceremony,
-  CeremonyListInput,
-  CeremonyType,
-  CreateCeremony,
-  UpdateCeremony,
-} from './dto';
+import { Ceremony, CeremonyType } from './dto';
 
 @Injectable()
 export class CeremonyEdgeDBRepository
@@ -20,27 +12,10 @@ export class CeremonyEdgeDBRepository
       engagement: true,
       type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
     }),
-  }).customize((cls, { defaults }) => {
-    return class extends cls {
-      static omit = [defaults.readOne, defaults.readMany];
-      async create(input: CreateCeremony): Promise<any> {
-        // EdgeDB creates these ceremonies in the process of creating the parent
-        return;
-      }
-      async update(
-        existing: DbTypeOf<Ceremony>,
-        changes: ChangesOf<Ceremony, UpdateCeremony>,
-      ): Promise<DbTypeOf<Ceremony>> {
-        return await this.defaults.update({ existing, changes });
-      }
-      async list({
-        filter,
-        ...input
-      }: CeremonyListInput): Promise<
-        PaginatedListType<UnsecuredDto<Ceremony>>
-      > {
-        return await this.defaults.list({ filter, ...input });
-      }
-    };
-  })
-  implements PublicOf<CeremonyRepository> {}
+  }).withDefaults()
+  implements PublicOf<CeremonyRepository>
+{
+  override create() {
+    // nothing needed here
+  }
+}
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -1,24 +1,23 @@
 import { Injectable } from '@nestjs/common';
 import { PublicOf } from '~/common';
-import { DbTypeOf } from '~/core';
-import { ChangesOf } from '~/core/database/changes';
 import { RepoFor } from '~/core/edgedb';
 import { CeremonyRepository } from './ceremony.repository';
-import { Ceremony, CreateCeremony, UpdateCeremony } from './dto';
+import { Ceremony, CreateCeremony } from './dto';
 
 @Injectable()
 export class CeremonyEdgeDBRepository
   extends RepoFor(Ceremony, {
     hydrate: (ceremony) => ({ ...ceremony['*'], engagement: true }),
-  }).customize((cls) => {
+  }).customize((cls, { defaults }) => {
     return class extends cls {
+      static omit = [
+        defaults.update,
+        defaults.list,
+        defaults.readMany,
+        defaults.readOne,
+      ];
       async create(input: CreateCeremony): Promise<any> {
-        return;
-      }
-      async update(
-        existing: DbTypeOf<Ceremony>,
-        changes: ChangesOf<Ceremony, UpdateCeremony>,
-      ) {
+        // EdgeDB creates these ceremonies in the process of creating the parent
         return;
       }
     };
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -1,24 +1,45 @@
 import { Injectable } from '@nestjs/common';
-import { PublicOf } from '~/common';
-import { RepoFor } from '~/core/edgedb';
+import { PaginatedListType, PublicOf, UnsecuredDto } from '~/common';
+import { DbTypeOf } from '~/core';
+import { ChangesOf } from '~/core/database/changes';
+import { castToEnum, RepoFor } from '~/core/edgedb';
 import { CeremonyRepository } from './ceremony.repository';
-import { Ceremony, CreateCeremony } from './dto';
+import {
+  Ceremony,
+  CeremonyListInput,
+  CeremonyType,
+  CreateCeremony,
+  UpdateCeremony,
+} from './dto';
 
 @Injectable()
 export class CeremonyEdgeDBRepository
   extends RepoFor(Ceremony, {
-    hydrate: (ceremony) => ({ ...ceremony['*'], engagement: true }),
+    hydrate: (ceremony) => ({
+      ...ceremony['*'],
+      engagement: true,
+      type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
+    }),
   }).customize((cls, { defaults }) => {
     return class extends cls {
-      static omit = [
-        defaults.update,
-        defaults.list,
-        defaults.readMany,
-        defaults.readOne,
-      ];
+      static omit = [defaults.readOne, defaults.readMany];
       async create(input: CreateCeremony): Promise<any> {
         // EdgeDB creates these ceremonies in the process of creating the parent
         return;
+      }
+      async update(
+        existing: DbTypeOf<Ceremony>,
+        changes: ChangesOf<Ceremony, UpdateCeremony>,
+      ): Promise<DbTypeOf<Ceremony>> {
+        return await this.defaults.update({ existing, changes });
+      }
+      async list({
+        filter,
+        ...input
+      }: CeremonyListInput): Promise<
+        PaginatedListType<UnsecuredDto<Ceremony>>
+      > {
+        return await this.defaults.list({ filter, ...input });
       }
     };
   })
--- a/src/components/ceremony/ceremony.edgedb.repository.ts
+++ b/src/components/ceremony/ceremony.edgedb.repository.ts
@@ -1,0 +1,18 @@
+import { Injectable } from '@nestjs/common';
+import { PublicOf } from '~/common';
+import { RepoFor } from '~/core/edgedb';
+import { CeremonyRepository } from './ceremony.repository';
+import { Ceremony, CreateCeremony } from './dto';
+
+@Injectable()
+export class CeremonyEdgeDBRepository
+  extends RepoFor(Ceremony, {
+    hydrate: (ceremony) => ({ ...ceremony['*'], engagement: true }),
+  }).customize((cls) => {
+    return class extends cls {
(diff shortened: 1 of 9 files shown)


src/components/ceremony/ceremony.edgedb.repository.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from '@nestjs/common';
+import { PublicOf } from '~/common';
+import { RepoFor } from '~/core/edgedb';
+import { CeremonyRepository } from './ceremony.repository';
+import { Ceremony, CreateCeremony } from './dto';
+
+@Injectable()
+export class CeremonyEdgeDBRepository
```
brentkulwicki, line 8: Hey @CarsonF - Need help here when you have a moment. I thought that the `static omit = [*defaults*]` section below would fix a lot of the TS issues I am dealing with, but I am still seeing the below error on this CeremonyEdgeDBRepository that I am trying to make sense of: 
```
TS2420: Class CeremonyEdgeDBRepository incorrectly implements interface PublicOf<CeremonyRepository>
Type CeremonyEdgeDBRepository is missing the following properties from type PublicOf<CeremonyRepository>: update, list, readOne, readMany
```

CarsonF: The current neo4j repository still has those methods, so we'll need to keep them here until we fully cut over.

src/components/ceremony/ceremony.repository.ts
```diff
@@ -32,15 +32,20 @@ export class CeremonyRepository extends DtoRepository<
       actualDate: input.actualDate,
       canDelete: true,
     };
-    return await this.db
+    const result = await this.db
       .query()
       .apply(await createNode(Ceremony, { initialProps }))
-      .return<{ id: ID }>('node.id as id')
+      .return<{ node: UnsecuredDto<Ceremony> }>('node')
```
CarsonF, line 38: That's not what this does. I'd revert this. The object with single ID is sufficient.

src/components/ceremony/ceremony.repository.ts
```diff
@@ -32,15 +32,20 @@ export class CeremonyRepository extends DtoRepository<
...
+
+    if (!result) {
+      throw new ServerException('failed to create a ceremony');
+    }
+    return result;
   }
 
   async update(
-    existing: Ceremony,
+    existing: DbTypeOf<Ceremony>,
```
CarsonF, line 48: Follow this pattern instead for updates.
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/funding-account/funding-account.service.ts#L103-L104
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/funding-account/funding-account.repository.ts#L38-L38

src/components/ceremony/handlers/create-engagement-default-ceremony.handler.ts
```diff
@@ -22,6 +22,7 @@ export class CreateEngagementDefaultCeremonyHandler
           ? CeremonyType.Dedication
           : CeremonyType.Certification,
     };
+    // if DB is Edge - do nothing
```
CarsonF, line 25: https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/project/handlers/set-department-id.handler.ts#L26-L28

src/components/ceremony/ceremony.service.ts
```diff
@@ -29,13 +28,9 @@ export class CeremonyService {
 
   async create(input: CreateCeremony): Promise<ID> {
     try {
-      const result = await this.ceremonyRepo.create(input);
+      const { node } = await this.ceremonyRepo.create(input);
```
CarsonF, line 31: Rename for consistency
```suggestion
      const { node } = await this.repo.create(input);
```

src/components/ceremony/ceremony.service.ts
```diff
@@ -29,13 +28,9 @@ export class CeremonyService {
...
 
-      if (!result) {
-        throw new ServerException('failed to create a ceremony');
-      }
-
-      return result.id;
+      return node.id;
     } catch (exception) {
       this.logger.warning('Failed to create ceremony', {
```
CarsonF, line 35: Log message is redundant.

src/components/ceremony/ceremony.edgedb.repository.ts
brentkulwicki: I guess I misunderstood what this did. I was thinking it omitted them from the customization, not all together. 

So then is there a way to customize the `create` but use the defaults for the rest? It's making me re-create them all to work around TS errors if I customize this one, but I don't feel like I should have to do that.

CarsonF: Yes `omit` removes them all together. If you are customizing it you don't need to omit it.
I think there's a bug/annoyance in the TS error. It gives you that, but the real problem is probably the return type of the `create`.
If you do temporarily add all those methods it says are missing and just forward them to the defaults, TS should give you the actual error.

CarsonF: It's the `type` that the DTO wants, but EdgeDB schema doesn't have. We can compute it on read, in the `hydrate` method.
```ts
  // Engagement::DedicationCeremony -> Dedication, etc.
  type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
```
Just like we do for projects
https://github.com/SeedCompany/cord-api-v3/blob/develop/src/components/project/project.edgedb.repository.ts#L21-L22

brentkulwicki: Added on but still seeing some TS errors. here I now see: 
```
TS2420: Class CeremonyEdgeDBRepository incorrectly implements interface PublicOf<CeremonyRepository>
Type CeremonyEdgeDBRepository is missing the following properties from type PublicOf<CeremonyRepository>: readOne, readMany
```

brentkulwicki: I have literally never struggled with TS errors and how to resolve this much in my life but I can't make sense of these EdgeDB/neo4j mashup TS errors. 😞

src/components/ceremony/ceremony.edgedb.repository.ts
```diff
@@ -0,0 +1,46 @@
...
+      static omit = [defaults.readOne, defaults.readMany];
+      async create(input: CreateCeremony): Promise<any> {
+        // EdgeDB creates these ceremonies in the process of creating the parent
+        return;
+      }
+      async update(
+        existing: DbTypeOf<Ceremony>,
+        changes: ChangesOf<Ceremony, UpdateCeremony>,
+      ): Promise<DbTypeOf<Ceremony>> {
+        return await this.defaults.update({ existing, changes });
```
brentkulwicki, line 34: TS errors I am struggling to make sense of here and at the same place in `list` too. @CarsonF

src/components/ceremony/ceremony.repository.ts
```diff
@@ -32,18 +31,24 @@ export class CeremonyRepository extends DtoRepository<
...
-    existing: Ceremony,
-    changes: ChangesOf<Ceremony, UpdateCeremony>,
-  ) {
-    return await this.updateProperties(existing, changes);
+  async update(changes: UpdateCeremony) {
+    const { id, ...simpleChanges } = changes;
+    await this.updateProperties({ id }, simpleChanges);
+    // I don't understand why something like funding-account does not require args but this one does?
+    // I also don't feel like I know what args or where they should come from so feel stuck
+    return await this.readOne(id);
```
brentkulwicki, line 51: Help here would be great @CarsonF

src/components/ceremony/ceremony.repository.ts
brentkulwicki: Right but aren't we trying to return the entire entity now vs just an ID? I feel like I am going in circles on everything and don't have a clear idea on what the expectations are for query implementations

CarsonF: Well generally we are. Because this object isn't created directly by users though, the returned object isn't really used. Only the ID. Keeping the return as `{ id: ID }` is best as that allows the edgedb implementation to do that plus more without it being a breaking type.

src/components/ceremony/ceremony.repository.ts
CarsonF: Ok so `readOne/Many` all call `hydrate` to populate the shape of the object that the app needs. Sometimes `hydrate` needs some more info though, like the current user to know if you are a project member or not.
That is the case here, we can see that `hydrate` arg below. 
You can see that `readOne/Many`
https://github.com/SeedCompany/cord-api-v3/blob/1e2dfedf2cc5273df3550df9bc47c1dcaf7349aa/src/core/database/dto.repository.ts#L87-L87
reference the `HydrateArgs` generic, which is defined here
https://github.com/SeedCompany/cord-api-v3/blob/1e2dfedf2cc5273df3550df9bc47c1dcaf7349aa/src/components/ceremony/ceremony.repository.ts#L24-L24

So in this case we should also pass `session` in to this `update` method, so `readOne` can use it.

`FundingAccount` may have been too simple of an example for this case.

I know there's a lot of inconsistencies between the module repositories.
But we aim to solve that with this migration.

src/components/ceremony/ceremony.edgedb.repository.ts
CarsonF: Ok I totally agree with you that the error it's giving you isn't helpful.
I think it's an intersection between the "`implements` neo4j repo" and the `customize()` signature.
The former is short lived, and I want to rework `customize()` as well. Because I agree it's bad right now.

So I removed the `static omit` you have. And then added
```ts
readOne(id: ID) {
  return this.defaults.readOne(id);
}
readMany(id: readonly ID[]) {
  return this.defaults.readMany(id);
}
```
(which I totally agree you shouldn't need, but this gives the actual type error).
Oh I also changed update to
```ts
async update(changes: UpdateCeremony)
```
to match your neo4j version. (The `session` arg is needed there, but it can be omitted here since it's the last arg and we don't need it here.)

_Finally_ TS tells us the error is

>Types of property `estimatedDate` are incompatible.
>  Type `CalendarDate | null` is not assignable to type `CalendarDate`

This is another long-standing inconsistency we've had.
We declare
https://github.com/SeedCompany/cord-api-v3/blob/1e2dfedf2cc5273df3550df9bc47c1dcaf7349aa/src/components/ceremony/dto/ceremony.dto.ts#L36-L36
which is just a secure wrapped value of `CalendarDate`.

But the EdgeDB schema we've recently declared is `optional` (implicit)
https://github.com/SeedCompany/cord-api-v3/blob/1e2dfedf2cc5273df3550df9bc47c1dcaf7349aa/dbschema/ceremony.esdl#L6-L6
aka nullable.

The neo4j queries don't have strict types nor another schema to compare against, which is why this has lived so long. Now it actually matters as EdgeDB is forcing us to be accurate here.

We need to fix the fields in the DTO to be `SecuredDateNullable` which translates to a secure wrapped value of `CalendarDate | null`.

Hopefully all this makes sense.

src/components/ceremony/ceremony.edgedb.repository.ts
CarsonF: As I mentioned above, this should just be
```ts
async update(changes: UpdateCeremony)
```
to match the new neo4j signature you have.
Which does in fact match the default signature here.
So we don't need to declare.

----

On that note, you are not actually changing the signature of `create` either.
So you don't actually need to `customize()` you can just `override` in the final class.
```ts
class CerRepo extends RepoFor().withDefaults() {
  override create() {
    // noop
  }
}
```

brentkulwicki: So what you're saying here is this should look like:
```
@Injectable()
export class CeremonyEdgeDBRepository
  extends RepoFor(Ceremony, {
    hydrate: (ceremony) => ({
      ...ceremony['*'],
      engagement: true,
      type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
    }),
  }).withDefaults()
  implements PublicOf<CeremonyRepository> {
    override create() {}
}
```

brentkulwicki: > As I mentioned above, this should just be
> 
> ```ts
> async update(changes: UpdateCeremony)
> ```
> 
> to match the new neo4j signature you have. Which does in fact match the default signature here. So we don't need to declare.
> 
> Same for `list`.
> 
> On that note, you are not actually changing the signature of `create` either. So you don't actually need to `customize()` you can just `override` in the final class.
> 
> ```ts
> class CerRepo extends RepoFor().withDefaults() {
>   override create() {
>     // noop
>   }
> }
> ```

I made this change and still see issues with the return signature. Essentially it seems like one of the `update` is returning `Promise<readonly<entity>>` while the other is returning `readonly<entity>`

brentkulwicki: ☝️ Biggest struggle with all of these is knowing which `update` needs to be updated since they are all named the same thing. It's killing me.

brentkulwicki: Ahh... I fixed the DTO as you mentioned above and I think the update is working properly afterwards. That makes sense now too.

src/components/ceremony/ceremony.edgedb.repository.ts
```diff
@@ -0,0 +1,21 @@
...
+  extends RepoFor(Ceremony, {
+    hydrate: (ceremony) => ({
+      ...ceremony['*'],
+      engagement: true,
+      type: castToEnum(ceremony.__type__.name.slice(12, -8), CeremonyType),
+    }),
+  }).withDefaults()
+  implements PublicOf<CeremonyRepository>
+{
+  override create() {
```
brentkulwicki, line 18: I think I have this mostly ready to go now with your other comments, but I do still have some errors here with the return types not matching. It feels like this is supposed to be the case as these are actually being created as part of the engagement schema, so not sure if this should just be ignored or if there is a true way to resolve the issue? @CarsonF 

TS2416: Property `create` in type `CeremonyEdgeDBRepository` is not assignable to the same property in base type `PublicOf<CeremonyRepository>`
Type () => void is not assignable to type
```
(input: CreateCeremony) => Promise<{   id: ID; }>
```
Type `void` is not assignable to type `Promise<{ id: ID; }>`
TS2416: Property create in type `CeremonyEdgeDBRepository `is not assignable to the same property in base type
```
RepoFor<typeof Ceremony, $expr_PathNode<TypeSet<$Ceremony, Cardinality.Many>, null>, { engagement: true; type: { [x: number]: $expr_Operator<ScalarType<"std::str", string, string, string>, Cardinality.One>; [x: `${number}:${number}`]: $expr_Operator<...>; [x: `${number}:`]: $expr_Operator<...>; [x: `:${number}`]: $e...
```
Type `() => void` is not assignable to type
```
(input: { id?: ID | TypeSet<$uuid, Cardinality.AtMostOne | Cardinality.One | Cardinality.Empty> | undefined; createdAt?: DateTime | TypeSet<...> | undefined;  ... 6 more ...; project: TypeSet<...> | ... 1 more ... | TypeSet<...>; } | ... 4 more ... | { ...; }) => Promise<...>
```
Type `void` is not assignable to type
```
Promise<Readonly<{   id: ID;   createdAt: DateTime;   sensitivity: 'Low' | 'Medium' | 'High';   type: 'Dedication' | 'Certification';   isMember: boolean;   engagement: Readonly<{     id: ID;   }>;   ... 4 more ...;   planned: boolean; }>>
```

CarsonF: Yeah this is a tricky situation. In neo4j, the ceremony is created and then attached to the engagement. In EdgeDB, the engagement is given as a required link of the ceremony. This isn't the first time we've hit this.
For this case though, a schema trigger creates it for us, so we don't have to bother.
I'd just
```ts
create(input: CreateCeremony): never {
  throw new NotImplementedException();
}
```
That will make the types compatible, and we already the condition to not call any of this create logic for EdgeDB.

src/components/ceremony/ceremony.service.ts
```diff
@@ -56,36 +51,44 @@ export class CeremonyService {
...
     this.privileges.for(session, Ceremony, object).verifyChanges(changes);
-    return await this.ceremonyRepo.update(object, changes);
+    const updated = await this.repo.update(
+      {
+        id: input.id,
+        ...changes,
+      },
+      session,
+    );
+    return updated;
```
CarsonF, line 81: The return type of this update method should still be `Promise<Ceremony>`.
You just need to wrap it in a secure()
```suggestion
    return this.secure(updated, session);
```

CarsonF (commented):
Almost there!

src/components/ceremony/ceremony.edgedb.repository.ts
brentkulwicki: Is this not done as an override? Still seeing TS having problems after pushing the most recent commit with this: 
```
  override create(input: CreateCeremony): never {
    throw new NotImplementedException();
  }
  ```
  
 `create` method throws this: 
 TS2416: Property `create` in type `CeremonyEdgeDBRepository` is not assignable to the same property in base type
```
RepoFor<typeof Ceremony, $expr_PathNode<TypeSet<$Ceremony, Cardinality.Many>, null>, { engagement: true; type: { [x: number]: $expr_Operator<ScalarType<"std::str", string, string, string>, Cardinality.One>; [x: `${number}:${number}`]: $expr_Operator<...>; [x: `${number}:`]: $expr_Operator<...>; [x: `:${number}`]: $e...
```
Type `(input: CreateCeremony) => never` is not assignable to type
```
(input: { modifiedAt?: DateTime | TypeSet<$datetime, Cardinality.AtMostOne | Cardinality.One | Cardinality.Empty> | undefined;  ... 7 more ...; engagement: TypeSet<...> | ... 1 more ... | TypeSet<...>; } | ... 4 more ... | { ...; }) => Promise<...>
```

Types of parameters `input` and `input` are incompatible.
Type
```
{ modifiedAt?: DateTime | TypeSet<$datetime, Cardinality.AtMostOne | Cardinality.One | Cardinality.Empty> | undefined;  ... 7 more ...; engagement: TypeSet<...> | ... 1 more ... | TypeSet<...>; } | ... 4 more ... | { ...; }
```
is not assignable to type `CreateCeremony`
Property `type` is missing in type
```
{ modifiedAt?: DateTime | TypeSet<$datetime, Cardinality.AtMostOne | Cardinality.One | Cardinality.Empty> | undefined;  ... 7 more ...; engagement: TypeSet<...> | ... 1 more ... | TypeSet<...>; }
```
but required in type `CreateCeremony`

CarsonF: Ah that's unfortunate. So we will need to have this `create` in a `customize()`. Sorry for that whiplask.

CarsonF (approved):
Thanks @brentkulwicki!

PR #3184, opened 2024-04-22
Update Project Workflow for Multiplication Projects

atGit2021:
[Monday task](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5934478741)

## Description
Create simplified project workflow for Multiplication projects

### DIFF
--- a/src/components/project/handlers/send-step-change-notification-on-change-request-approved.ts
+++ b/src/components/project/handlers/send-step-change-notification-on-change-request-approved.ts
@@ -1,6 +1,6 @@
 import { EmailService } from '@seedcompany/nestjs-email';
 import { node, relation } from 'cypher-query-builder';
-import { ID } from '../../../common';
+import { UnsecuredDto } from '../../../common';
 import {
   ConfigService,
   DatabaseService,
@@ -12,7 +12,7 @@ import {
 import { ACTIVE, INACTIVE } from '../../../core/database/query';
 import { ProjectStepChanged } from '../../../core/email/templates';
 import { ProjectChangeRequestApprovedEvent } from '../../project-change-request/events';
-import { ProjectStep } from '../dto';
+import { Project, ProjectStep } from '../dto';
 import { ProjectRules } from '../project.rules';
 
 type SubscribedEvent = ProjectChangeRequestApprovedEvent;
@@ -56,11 +56,11 @@ export class SendStepChangeNotificationsOnChangeRequestApproved
       .with('node, currentStep, previousStep')
       .orderBy('previousStep.createdAt', 'DESC')
       .return<{
-        projectId: ID;
+        project: UnsecuredDto<Project>;
         currentStep: ProjectStep;
         previousStep: ProjectStep;
       }>(
-        'node.id as projectId, currentStep.value as currentStep, collect(previousStep.value)[0] as previousStep',
+        'node as project, currentStep.value as currentStep, collect(previousStep.value)[0] as previousStep',
       )
       .first();
 
@@ -73,15 +73,14 @@ export class SendStepChangeNotificationsOnChangeRequestApproved
     }
 
     const recipients = await this.projectRules.getNotifications(
-      result.projectId,
-      result.currentStep,
+      result.project,
       event.session.userId,
       result.previousStep,
     );
 
     this.logger.info('Notifying', {
       emails: recipients.map((r) => r.recipient.email.value),
-      projectId: result.projectId,
+      projectId: result.project.id,
       step: result.currentStep,
       previousStep: result.previousStep,
     });
--- a/src/components/project/handlers/send-step-change-notification-on-change-request-approved.ts
+++ b/src/components/project/handlers/send-step-change-notification-on-change-request-approved.ts
@@ -1,6 +1,5 @@
 import { EmailService } from '@seedcompany/nestjs-email';
 import { node, relation } from 'cypher-query-builder';
-import { UnsecuredDto } from '../../../common';
 import {
   ConfigService,
   DatabaseService,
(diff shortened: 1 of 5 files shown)


src/components/project/project-step.resolver.ts
```diff
@@ -3,24 +3,37 @@ import { stripIndent } from 'common-tags';
...
+    step: SecuredProjectStep & {
+      parentId: ID;
+      changeset?: ID;
+    },
     @AnonSession() session: Session,
   ): Promise<ProjectStepTransition[]> {
     if (!step.canRead || !step.canEdit || !step.value) {
       return [];
     }
+    const project = await this.projectService.readOneUnsecured(
```
bryanjnelson, line 29: @CarsonF - I know that you mentioned using a loader here.  I'm not entirely sure what would be used for the key for the loader or how that would work.  We looked at some other examples, but couldn't quite get there.

src/components/project/project.rules.ts
```diff
@@ -281,26 +301,46 @@ export class ProjectRules {
...
-              to: ProjectStep.FinalizingProposal,
-              type: TransitionType.Reject,
-              label: 'Send Back for Corrections',
-            },
-            {
-              to: ProjectStep.Rejected,
-              type: TransitionType.Reject,
-              label: 'Reject',
-            },
+            ...(isMultiplication
```
bryanjnelson, line 304: @CarsonF - Now that this is coded, it makes me wonder if it would be better to have two separate implementations, one for the momentum/intern flow and one for the multiplication flow, instead of splitting logic like this.  I know that wasn't part of what Andre was supposed to do; I was just thinking about it.

bryanjnelson (approved):
@CarsonF - I don't have a lot to add here since I'm not terribly familiar with the code.  I just added a few comments.  On to you.

src/components/project/project-step.resolver.ts
CarsonF: https://github.com/SeedCompany/cord-api-v3/blob/dfb31cc2a5922d776695c11582e5d447d772ad22/src/components/product/product.resolver.ts#L84-L84
```ts
const project = await projects.load(step.parentId)
```

atGit2021: I tried this, but step.parentId is of type ID and the load method says it receives a type of 'key'.  I wasn't sure how to transform the ID argument into a type of 'key'

src/components/project/handlers/send-step-change-notification-on-change-request-approved.ts
```diff
@@ -56,11 +56,11 @@ export class SendStepChangeNotificationsOnChangeRequestApproved
       .with('node, currentStep, previousStep')
       .orderBy('previousStep.createdAt', 'DESC')
       .return<{
-        projectId: ID;
+        project: UnsecuredDto<Project>;
         currentStep: ProjectStep;
         previousStep: ProjectStep;
       }>(
-        'node.id as projectId, currentStep.value as currentStep, collect(previousStep.value)[0] as previousStep',
+        'node as project, currentStep.value as currentStep, collect(previousStep.value)[0] as previousStep',
```
CarsonF, line 63: This is not right.
You could do
```cql
node { .id, .type } as project
```
```ts
project: Pick<Project, 'id' | 'type'>
```

src/components/project/project.rules.ts
CarsonF: File's a mess anyways. It's on the chopping block.

In this case theres no need for the outer array, it could be removed.

src/components/project/project.rules.ts
```diff
@@ -281,26 +301,46 @@ export class ProjectRules {
...
+                  {
+                    to: ProjectStep.PendingFinanceConfirmation,
+                    type: TransitionType.Approve,
+                    label: 'Submit for Finance Confirmation',
+                  },
+                  {
+                    to: ProjectStep.DidNotDevelop,
+                    type: TransitionType.Reject,
+                    label: 'End Development',
+                  },
```
CarsonF, line 320: Why `DidNotDevelop` here instead of `Rejected`?

src/components/project/project.rules.ts
```diff
@@ -330,31 +370,56 @@ export class ProjectRules {
...
-            },
-            {
-              to: ProjectStep.Rejected,
-              type: TransitionType.Reject,
-              label: 'Reject',
-            },
+            ...(isMultiplication
+              ? [
+                  {
+                    to: ProjectStep.Active,
```
CarsonF, line 376: Same on both sides of condition, extract.

CarsonF (changes requested):
Great first stab @atGit2021

src/components/project/project-step.resolver.ts
CarsonF: Ah right sorry
```ts
load({ id: step.parentId, view: { active: true } })
```

atGit2021: ok, thank you sir

bryanjnelson: It's the `view` part we didn't really get I think.

CarsonF: It's apart of the changesets work. "active" is the normal view, not inside a changeset. It's only half baked, and being dropped in edgedb migration.

src/components/project/project.rules.ts
atGit2021: I reached out to Sheri to confirm the [requirement](https://seed-company-squad.monday.com/boards/3451697530/pulses/5934478741/posts/3100606414)

atGit2021: updating to include both steps per product teams feedback

CarsonF (approved):
LGTM! Thanks @atGit2021

PR #3185, opened 2024-04-23
[EdgeDB] Film queries

willdch:
## Description

Implementing the EdgeDB Query Updates for the Film data

- [n/a] Add/update tests if needed

### DIFF
--- a/src/components/film/film.service.ts
+++ b/src/components/film/film.service.ts
@@ -1,18 +1,15 @@
 import { Injectable } from '@nestjs/common';
 import {
-  DuplicateException,
   ID,
   ObjectView,
-  SecuredList,
   ServerException,
   Session,
-} from '../../common';
-import { DbTypeOf, HandleIdLookup, ILogger, Logger } from '../../core';
-import { ifDiff } from '../../core/database/changes';
-import { mapListResults } from '../../core/database/results';
+  UnsecuredDto,
+} from '~/common';
+import { HandleIdLookup } from '~/core';
+import { ifDiff } from '~/core/database/changes';
 import { Privileges } from '../authorization';
 import { isScriptureEqual } from '../scripture';
-import { ScriptureReferenceService } from '../scripture/scripture-reference.service';
 import {
   CreateFilm,
   Film,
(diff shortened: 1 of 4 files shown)


src/components/film/film.service.ts
```diff
@@ -91,7 +63,7 @@ export class FilmService {
   }
 
   async update(input: UpdateFilm, session: Session): Promise<Film> {
-    const film = await this.readOne(input.id, session);
+    const film = await this.repo.readOne(input.id);
```
bryanjnelson, line 66: @CarsonF - Any advice on how to proceed with the error below.  I think the issue is the start/end vs range issue?

CarsonF: Yeah it's because `secure()` also parses the scripture range IDs back into objects.
https://github.com/SeedCompany/cord-api-v3/pull/3185/files#diff-5578576cbb43b3c5ac0e42edf81d777514503b411c133ae9575f36a2906f2d82R59-R61

src/components/scripture/edgedb.utils.ts
```diff
@@ -66,13 +66,13 @@ export const insert = (param: $expr_Param<string, typeof type>) => {
...
           verse: e.cast(e.int16, start.verse),
-          verseId: e.cast(e.int32, start.verseId),
+          verseId: e.cast(e.int16, start.verseId),
         }),
         end: e.insert(e.Scripture.Verse, {
           book: e.cast(e.str, end.book),
           chapter: e.cast(e.int16, end.chapter),
           verse: e.cast(e.int16, end.verse),
-          verseId: e.cast(e.int32, end.verseId),
+          verseId: e.cast(e.int16, end.verseId),
```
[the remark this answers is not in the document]
willdch: While I was testing things out I got a type error: `InvalidPropertyTargetError: invalid target for property 'verseId' of object type 'Scripture::Verse': 'std::int32' (expecting 'std::int16'))`.  After sending this to Carson he said he missed that and showed me where to fix it.

CarsonF (changes requested):
Diff cypher repository with story repo. The variations are why the tests are failing.

src/components/film/film.service.ts
```diff
@@ -25,111 +22,63 @@ import { FilmRepository } from './film.repository';
...
       await this.repo.deleteNode(film);
     } catch (exception) {
-      this.logger.error('Failed to delete', { id, exception });
       throw new ServerException('Failed to delete', exception);
     }
-
-    this.logger.debug(`deleted film with id`, { id });
   }
 
   async list(input: FilmListInput, session: Session): Promise<FilmListOutput> {
```
willdch, line 77: @CarsonF Just saw that a missed fixing the signature here.  Will get that fixed and pushed

CarsonF (approved):
Right on. I squashed some commits down, but didn't change anything.

PR #3186, opened 2024-04-23
feat: project member queries

adam-soltech:
## Reason for this PR
Migrate Project Member Queries to edgedb

## Description

### DIFF
--- a/src/components/project/project-member/project-member.repository.ts
+++ b/src/components/project/project-member/project-member.repository.ts
@@ -42,7 +42,6 @@ export class ProjectMemberRepository extends DtoRepository<
   constructor(
     private readonly users: UserRepository,
     @Logger('project:member:repository') private readonly logger: ILogger,
-    private readonly service: ProjectMemberService,
   ) {
     super();
   }
@@ -134,43 +133,9 @@ export class ProjectMemberRepository extends DtoRepository<
       .first();
   }
 
-  async update(input: UpdateProjectMember, session: Session) {
-    const existing = await this.readOne(input.id, session);
-    await this.assertValidRoles(input.roles, () => {
-      const user = existing.user;
-      if (!user) {
-        throw new UnauthorizedException(
-          'Cannot read user to verify roles available',
-        );
-      }
-      return user;
-    });
-    const changes = this.getActualChanges(existing, input);
-    this.service.privileges
-      .for(session, ProjectMember, existing)
-      .verifyChanges(changes);
-
-    await this.updateProperties(existing, changes);
-    return await this.readOne(existing.id, session);
-  }
-
-  async assertValidRoles(
-    roles: readonly Role[] | undefined,
-    forUser: () => UnsecuredDto<User>,
-  ) {
-    if (!roles || roles.length === 0) {
-      return;
-    }
-    const user = forUser();
-    const availableRoles = user.roles ?? [];
-    const forbiddenRoles = difference(roles, availableRoles);
-    if (forbiddenRoles.length) {
-      const forbiddenRolesStr = forbiddenRoles.join(', ');
-      throw new InputException(
-        `Role(s) ${forbiddenRolesStr} cannot be assigned to this project member`,
-        'input.roles',
-      );
-    }
+  async update(input: UpdateProjectMember, changes: any, session: Session) {
+    await this.updateProperties(input, changes);
+    return await this.readOne(input.id, session);
   }
 
   protected hydrate(session: Session) {
--- a/src/components/project/project-member/project-member.repository.ts
+++ b/src/components/project/project-member/project-member.repository.ts
@@ -31,7 +31,7 @@ export class ProjectMemberRepository extends DtoRepository<
     super();
   }
 
-  async verifyRelationshipEligibility(projectId: ID, userId: ID) {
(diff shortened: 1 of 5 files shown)


src/components/project/project-member/project-member.service.ts
```diff
@@ -104,7 +59,8 @@ export class ProjectMemberService {
 
     const id = await generateId();
     const createdAt = DateTime.local();
-    await this.repo.verifyRelationshipEligibility(projectId, userId);
+    this.repo.verifyRelationshipEligibility &&
+      (await this.repo.verifyRelationshipEligibility(projectId, userId));
```
atGit2021, line 63: the following code should work, `await this.repo.verifyRelationshipEligibility(projectId, userId);` if you remove the optional operator, `?`,  in the repo.

src/components/project/project-member/project-member.repository.ts
```diff
@@ -31,7 +31,7 @@ export class ProjectMemberRepository extends DtoRepository<
     super();
   }
 
-  async verifyRelationshipEligibility(projectId: ID, userId: ID) {
+  async verifyRelationshipEligibility?(projectId: ID, userId: ID) {
```
atGit2021, line 34: I think you can remove the optional operator `?` and adjust the service file per my comment below.

atGit2021 (commented):
@adam-soltech , Do you have time to sync tomorrow? not sure if you saw this wiki 
[page](https://github.com/SeedCompany/cord-docs/wiki/Migration--%5C---Refactoring-Work), that Bryan created for us to follow when creating queries.  We can discuss a few things.

src/components/project/project-member/project-member.service.ts
```diff
@@ -44,49 +33,15 @@ import { ProjectMemberRepository } from './project-member.repository';
...
     @Logger('project:member:service') private readonly logger: ILogger,
     private readonly privileges: Privileges,
     private readonly repo: ProjectMemberRepository,
   ) {}
 
-  protected async verifyRelationshipEligibility(
-    projectId: ID,
-    userId: ID,
-  ): Promise<void> {
-    const result = await this.repo.verifyRelationshipEligibility(
```
atGit2021, line 63: if you move the exception handling to the repo, then this can be modified to: 
_await this.repo.verifyRelationshipEligibility..._

src/components/project/project-member/project-member.service.ts
```diff
@@ -44,49 +33,15 @@ import { ProjectMemberRepository } from './project-member.repository';
...
-        'projectMember.userId',
-      );
-    }
-
-    if (result.member) {
-      throw new DuplicateException(
-        'projectMember.userId',
-        'Person is already a member of this project',
-      );
-    }
```
atGit2021, line 87: may need to move these checks to the verifyRelationshipEligibility function in the project-member.repository.ts file

src/components/project/project-member/project-member.service.ts
atGit2021, line 1: In the create method, you will want to remove the readOne, and modify the repo to do the readOne and send back the ProjectMember.  This will enable you to do one repo call from the create method.  You will also need to move the error handling checks to the repo too.

In the update method, you will want to change the readOne to point to the repo, and then also remove the second readOne after the update is returned.  Also need to modify the repo to do the readOne and return the ProjectMember instead of just the id.  Also will need to migrate the error handling to the repo too.

In the delete method, the readOne should be pointed to the repo and the error handling should also migrate to the repo.

src/components/project/project-member/project-member.service.ts
atGit2021: @adam-soltech , just remembered you mentioned that this method in the service file is not being called, so prob can ignore this piece, but may want to migrate the exception handling to the repo, see comment below.

bryanjnelson: Well, it turns out that in '22 I made the mistake of removing the call to the `verifyRelationshipEligibility` service method.  🤦🏼 

It _should_ be used right now.  So that means that yes, you should move those three checks down into the `verifyRelationshipEligibility` repo method as @atGit2021 suggested!

atGit2021 (commented):
@adam-soltech I believe you can leave the following code, `   this.service.privileges
      .for(session, ProjectMember, existing)
      .verifyChanges(changes);
` back in the service file.  @bryanjnelson  mentioned that the code related to privileges checking don't have to move to the repo at this time.  This will allow you to also remove the service file from the constructor and import list which should then resolve the test files from breaking too.  Let me know if you want to discuss.

bryanjnelson:
> @adam-soltech I believe you can leave the following code, ` this.service.privileges .for(session, ProjectMember, existing) .verifyChanges(changes);` back in the service file. @bryanjnelson mentioned that the code related to privileges checking don't have to move to the repo at this time. This will allow you to also remove the service file from the constructor and import list which should then resolve the test files from breaking too. Let me know if you want to discuss.

Yes, you'll need to leave that `privileges.for` call in there.  Also, it looks like that entire method was gutted.  It should still contain one read and one update (along with the `getActualChanges` and `verifyChanges` calls.  I am unsure about the `assertValidRoles` call.  I _think_ that should stay, but I'll let @CarsonF weigh in on that for certainty.

CarsonF:
Closing in favor of #3204

PR #3188, opened 2024-04-24
Fix up Unavailability queries for EdgeDB

CarsonF:
Few things missed in #3166


atGit2021 (approved):
I'm not familiar with some of the things you added but it LGTM!  So by adding a string literal, e.g. start and end, it acts like other properties such as, `dates`?  I notice you could remove those props from the hydrate so curious to know when to use one way vs the other.

PR #3192, opened 2024-04-24
Owner -> Audited / Actor / SystemAgent / Self

CarsonF:
This kinda ballooned into several related refactors.

Our authorization system is rule based. i.e. we don't have an ACL lists in our data. We felt like `owners` violated this principle. It served as the object's creator, but also was mutable to allow an owner change. The waters mudded more with Users, and how they were "owned".

So `owner` (schema & policy conditions) is replaced by `createdBy/creator`.
This has clearer split between raw data & authorization rules based on real data.

Related, we've also wanted to track who creates & modifies most things.
So to solve this we have `Audited` that holds `createdBy` & `modifiedBy`.
`Resource` was updated to extend `Audited` so most things can have this tracked now in a standard way.

We wanted these creators/modifiers to be required, as it's always someone that's making these changes. But there are edge cases.
So now we have `Actor` type that's barebones and have `User` & `SystemAgent` extend from that.
Hopefully a few pre-defined `SystemAgents` help fulfill these edge cases.
- _Ghost_ - we don't know who did it. Legacy migrated data.
- _Anonymous_ - It was actually an anonymous actor that did the action. Currently the only anonymous mutation we have is _register_. So this will allow us to know the user was created by someone else already logged in, or they registered themselves, or they were migrated in. Idk how helpful this will be.
- _External Mailing Group_ - For email notifications that reference mailing groups. I want to migrate to this reference instead of the _Root User_. It's possible that RootUser goes away soon after this change, as the ties to this user get clipped.

Finally I replaced the `currentUser` `alias` with a few `globals`.
https://github.com/SeedCompany/cord-api-v3/blob/0c18f9b7501d34c07b95291000a449531014ffec/dbschema/common.esdl#L3-L6
Globals (single, not multi) are evaluated once per query, so this should help with performance. And be less ambiguous in use.

### DIFF
--- a/dbschema/engagement.esdl
+++ b/dbschema/engagement.esdl
@@ -76,8 +76,8 @@ module default {
       insert Engagement::DedicationCeremony {
         createdAt := datetime_of_statement(),
         modifiedAt := datetime_of_statement(),
-        createdBy := assert_exists(currentUser),
-        modifiedBy := assert_exists(currentUser),
+        createdBy := assert_exists(global currentActor),
+        modifiedBy := assert_exists(global currentActor),
         engagement := __new__,
         project := __new__.project,
         projectContext := __new__.projectContext,
@@ -130,8 +130,8 @@ module default {
       insert Engagement::CertificationCeremony {
         createdAt := datetime_of_statement(),
         modifiedAt := datetime_of_statement(),
-        createdBy := assert_exists(currentUser),
-        modifiedBy := assert_exists(currentUser),
+        createdBy := assert_exists(global currentActor),
+        modifiedBy := assert_exists(global currentActor),
         engagement := __new__,
         project := __new__.project,
         projectContext := __new__.projectContext,
--- a/dbschema/engagement.esdl
+++ b/dbschema/engagement.esdl
@@ -76,6 +76,8 @@ module default {
       insert Engagement::DedicationCeremony {
         createdAt := datetime_of_statement(),
         modifiedAt := datetime_of_statement(),
+        createdBy := assert_exists(currentUser),
+        modifiedBy := assert_exists(currentUser),
         engagement := __new__,
         project := __new__.project,
         projectContext := __new__.projectContext,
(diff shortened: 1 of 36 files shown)


bryanjnelson:
So you did end up combining `UserStamped` and `TimeStamped` into `Audited`?

CarsonF:
> So you did end up combining `UserStamped` and `TimeStamped` into `Audited`?

Yeah. I thought UserStamped made less sense with the change to Actor. Looking around Audited was suggested. And with that it made sense to include timestamps in that name. Plus it doesn't really make sense to have users/actors without timestamps.

bryanjnelson:
Yeah, I like it.  Seems simplified...but yet at the same time more robust.

bryanjnelson (approved):
Thanks @CarsonF!

dbschema/engagement.esdl
```diff
@@ -130,8 +130,8 @@ module default {
       insert Engagement::CertificationCeremony {
         createdAt := datetime_of_statement(),
         modifiedAt := datetime_of_statement(),
-        createdBy := assert_exists(currentUser),
-        modifiedBy := assert_exists(currentUser),
+        createdBy := assert_exists(global currentActor),
```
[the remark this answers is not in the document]
CarsonF: The current user is the current actor, if it is a user, otherwise null. current actor can be a system agent. created/modifiedBy is an actor type so we use that here to be the most accepting. current user should be used sparingly probably, when a user type is required.

src/components/user/actor.repository.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Injectable()
+export abstract class ActorRepository {
+  @CachedByArg()
+  async getAnonymous() {
+    return await this.upsertAgent('Anonymous');
+  }
+
+  @CachedByArg()
+  async getGhost() {
```
brentkulwicki, line 13: What is this concept of ghost, anon, etc? Not sure I follow the application here

brentkulwicki (approved):
One question for my clarity, but this looks good!

src/components/user/actor.repository.ts
CarsonF: Does the PR description describe this with enough clarity or are you looking for more?

PR #3194, opened 2024-04-30
Simplify EdgeDB Repo customize

CarsonF:
Instead of putting customized methods in an intermediary class, they can now just go in the final class.
If the signature is compatible with the default one, then nothing more is needed, which is the same as before.
If the signature is _not_ compatible with the default one, then it can just be marked in the `omit` array to ignore the default signature.

Overall I think this simpler for customization cases.

For default cases, it doesn't appear that we will ever not want to start with default methods.
If we did we'd just extend from CommonRepo instead. Maybe future work here if needed.


bryanjnelson (approved):
Absolutely love it!

PR #3195, opened 2024-05-01
add financial approver

andrewdavid48:
## Description

- [N/A] Add reviewers to this PR


atGit2021:
@CarsonF , I still have work to do on the edgedb repo side of things, but I have a question so pushing up what I have so far.  First question is whether I should commit and push the following file, dbschema/migrations/00006-m1fbgja.edgeql, which was generated from the new migration create that we did today.  
Second question, is regarding the neo4j repo side, bot the set and list are now finding the financial approver and loading the result dto, but both methods give me an error in graphql when returning, which says, `message": "Cannot return null for non-nullable field ProjectTypeFinancialApprover.user.",`.  I'm not sure why it's saying null when I see a dto object being returned (at least from the repo).

atGit2021:
@CarsonF, I messed up my edgedb branch environment and lost the new edge branch called financial approver that we created yesterday so all those errors are back.  Since I really don't understand or remember all the steps you walked me through yesterday to get the new FinancialApprover type working,  would you be able to fix and rebase that into this PR so I can pull it down?

CarsonF:
We need to get this moving. Closing in favor of #3206

PR #3199, opened 2024-05-03
fix: fixing ghost user bug

brentkulwicki:
## Reason for this PR
Error in local dev environment because of a modifying user not being set

- [N/A] Change the task url above to the actual Monday task

### DIFF
--- a/src/components/admin/admin.edgedb.repository.ts
+++ b/src/components/admin/admin.edgedb.repository.ts
@@ -69,8 +69,6 @@ export class AdminEdgeDBRepository {
     await this.db
       .withOptions((o) =>
         o
-          // eslint-disable-next-line @typescript-eslint/naming-convention
-          .withConfig({ allow_user_specified_id: true })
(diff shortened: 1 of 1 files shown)


src/components/admin/admin.edgedb.repository.ts
```diff
@@ -63,8 +63,16 @@ export class AdminEdgeDBRepository {
...
   async updateEmail(id: ID, email: string) {
+    const ghost = await this.actors.getGhost();
     const u = e.cast(e.User, e.uuid(id));
     const query = e.update(u, () => ({ set: { email } }));
-    await this.db.run(query);
+    await this.db
+      .withOptions((o) =>
+        o
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          .withConfig({ allow_user_specified_id: true })
```
CarsonF, line 73: No user specified ID being inserted here.
```suggestion
```

CarsonF (approved):

PR #3200, opened 2024-05-03
Fix current report picking final prematurely

CarsonF:
Sort periodic reports by `start` date instead of `end` date.

This fixes incorrectly matching the current report to the final one.
Given reports:
Last regular:  1/1/2024 - 3/31/2024
Final:        3/31/2024 - 3/31/2024

So if using the end date, one of those two is arbitrarily picked, and has so far just been a coincidence to have been picked correctly.

### DIFF
--- a/src/components/periodic-report/periodic-report.repository.ts
+++ b/src/components/periodic-report/periodic-report.repository.ts
@@ -205,15 +205,22 @@ export class PeriodicReportRepository extends DtoRepository<
     return (query: Query) =>
       query.comment`matchCurrentDue()`
         .match([
-          node('baseNode', 'BaseNode', { id: parentId }),
-          relation('out', '', 'report', ACTIVE),
-          node('node', `${reportType}Report`),
-          relation('out', '', 'end', ACTIVE),
-          node('end', 'Property'),
+          [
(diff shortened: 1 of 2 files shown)


src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -205,15 +205,22 @@ export class PeriodicReportRepository extends DtoRepository<
     return (query: Query) =>
       query.comment`matchCurrentDue()`
         .match([
-          node('baseNode', 'BaseNode', { id: parentId }),
-          relation('out', '', 'report', ACTIVE),
-          node('node', `${reportType}Report`),
-          relation('out', '', 'end', ACTIVE),
-          node('end', 'Property'),
```
bryanjnelson, line 212: So still _matching_ on the `endDate`, but now _sorting_ on the `startDate`?

bryanjnelson (approved):
LGTM!

src/components/periodic-report/periodic-report.repository.ts
CarsonF: Yes still filtering on what's applicable according to the end date and now, but of those candidates now picking (via sort) the by the earliest start date.

PR #3205, opened 2024-05-08
[EdgeDB] EthnoArt queries

willdch:
## Description

Implementing the EdgeDB Query Updates for the EthnoArt data

- [na] Add/update tests if needed


CarsonF (approved):
Nice. Just squashed down and stripped out a stale comment in list().

PR #3207, opened 2024-05-09
Project Workflow: All Controllers -> FinancialApprovers

atGit2021:
[Monday Task](https://seed-company-squad.monday.com/boards/3451697530/pulses/6074211485)

### DIFF
--- a/src/components/project/project.rules.ts
+++ b/src/components/project/project.rules.ts
@@ -26,6 +26,7 @@ import {
   ProjectStepTransition,
   TransitionType,
 } from './dto';
+import { FinancialApproverRepository } from './financial-approver';
 import { ProjectService } from './project.service';
 
 type EmailAddress = string;
@@ -73,6 +74,7 @@ export class ProjectRules {
     @Inject(forwardRef(() => AuthenticationService))
     private readonly auth: AuthenticationService & {},
     private readonly configService: ConfigService,
+    private readonly financialApproverRepo: FinancialApproverRepository,
     // eslint-disable-next-line @seedcompany/no-unused-vars
     @Logger('project:rules') private readonly logger: ILogger,
   ) {}
@@ -85,6 +87,10 @@ export class ProjectRules {
   ): Promise<StepRule> {
     const mostRecentPreviousStep = (steps: ProjectStep[]) =>
       this.getMostRecentPreviousStep(id, steps, changeset);
+    const financialApprovers = async () =>
+      (await this.financialApproverRepo.read(projectType)).map(
+        ({ user }) => user.id,
+      );
     const isMultiplication =
       projectType === ProjectType.MultiplicationTranslation;
 
@@ -405,7 +411,7 @@ export class ProjectRules {
               type: TransitionType.Approve,
               label: 'Confirm Project 🎉',
               notifiers: async () => [
-                ...(await this.getRoleEmails(Role.Controller)),
+                ...(await financialApprovers()),
                 'project_approval@tsco.org',
                 'projects@tsco.org',
               ],
@@ -413,7 +419,7 @@ export class ProjectRules {
           ],
           getNotifiers: async () => [
             ...(await this.getProjectTeamUserIds(id)),
-            ...(await this.getRoleEmails(Role.Controller)),
+            ...(await financialApprovers()),
           ],
         };
       case ProjectStep.OnHoldFinanceConfirmation:
@@ -425,7 +431,7 @@ export class ProjectRules {
               type: TransitionType.Approve,
               label: 'Confirm Project 🎉',
               notifiers: async () => [
-                ...(await this.getRoleEmails(Role.Controller)),
+                ...(await financialApprovers()),
                 'project_approval@tsco.org',
                 'projects@tsco.org',
               ],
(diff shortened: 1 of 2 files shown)


src/components/project/project.rules.ts
```diff
@@ -405,15 +407,19 @@ export class ProjectRules {
               type: TransitionType.Approve,
               label: 'Confirm Project 🎉',
               notifiers: async () => [
-                ...(await this.getRoleEmails(Role.Controller)),
+                ...(await this.financialApproverRepo.read([projectType])).map(
+                  (fa) => fa.user.email!,
+                ),
```
CarsonF, line 412: We should use the `user.id` here instead. Otherwise these emails are treated as distribution lists and could be filtered out erroneously.
https://github.com/SeedCompany/cord-api-v3/blob/5e49de34c52a891c5d39ccce2bac9496ad38e756/src/components/project/project.rules.ts#L1012-L1016
Also these would be faked to the root user id instead of using the actual user.
https://github.com/SeedCompany/cord-api-v3/blob/5e49de34c52a891c5d39ccce2bac9496ad38e756/src/components/project/project.rules.ts#L1114-L1133

----

Also you could follow suit of this
https://github.com/SeedCompany/cord-api-v3/blob/5e49de34c52a891c5d39ccce2bac9496ad38e756/src/components/project/project.rules.ts#L86-L87
And provide a helper function to call in each use rather than duplicating all of it.
```ts
const financialApprovers = () => this.financialApproverRepo
  .read(projectType)
  .map(({ user }) => user.id);

**Review by CarsonF — CHANGES_REQUESTED** [review]

File: src/components/project/project.rules.ts (hunk 1 above)
Comment (CarsonF) [reply src/components/project/project.rules.ts:412]: I guess you could check for email and filter those out without. But that wouldn't save anything in practice probably. These users wouldn't be approvers and not have email addresses. Also we still filter them out later if that did happen

**Comment by atGit2021** [comment]
@CarsonF  The only thing missing per the requirements is the desire to have the partner name added to the email notification template.  I pushed up the latest commit since that will require a little bit more work as Partner is not currently passed into the template.  I also assume that when Sheri says "Partner Name" added to the template, we are actually talking about the Organization name?  Please verify

File: src/components/project/project.rules.ts [hunk 2]
```diff
@@ -85,6 +87,11 @@ export class ProjectRules {
   ): Promise<StepRule> {
     const mostRecentPreviousStep = (steps: ProjectStep[]) =>
       this.getMostRecentPreviousStep(id, steps, changeset);
+    const financialApprovers = async () => [
+      ...(await this.financialApproverRepo.read(projectType)).map(
+        ({ user }) => user.id,
+      ),
+    ];
```
CarsonF, line 94: There's no need to spread the array into another array.
```suggestion
    const financialApprovers = async () =>
      (await this.financialApproverRepo.read(projectType)).map(
        ({ user }) => user.id,
      );
```

CarsonF (approved):
This looks good to me so far. If you want to merge and deploy this and circle back to more email info that sounds like a good idea too.

src/components/project/project.rules.ts
atGit2021: If I don't spread it, then I get errors elsewhere, 
```
[{
	"resource": "/Users/andreturner/Dev/Repos/cord-api-v3/src/components/project/project.rules.ts",
	"owner": "typescript",
	"code": "2322",
	"severity": 8,
	"message": "Type '() => Promise<(string | ID[])[]>' is not assignable to type 'Notifiers | undefined'.\n  Type '() => Promise<(string | ID[])[]>' is not assignable to type '() => Promisable<Many<Notifier>>'.\n    Type 'Promise<(string | ID[])[]>' is not assignable to type 'Promisable<Many<Notifier>>'.",
	"source": "ts",
	"startLineNumber": 508,
	"startColumn": 11,
	"endLineNumber": 508,
	"endColumn": 23,
	"relatedInformation": [
		{
			"startLineNumber": 49,
			"startColumn": 3,
			"endLineNumber": 49,
			"endColumn": 15,
			"message": "The expected type comes from property 'getNotifiers' which is declared here on type 'StepRule'",
			"resource": "/Users/andreturner/Dev/Repos/cord-api-v3/src/components/project/project.rules.ts"
		}
	]
}]
```

CarsonF: That must be VSCode lying to you

atGit2021: okie dokie, I can push it without, but not sure how the 6 errors are showing up for me but not for you.

CarsonF: Remove the array brackets too

atGit2021: ok, thank you.

atGit2021:
> This looks good to me so far. If you want to merge and deploy this and circle back to more email info that sounds like a good idea too.

Ok, once @bryanjnelson approves, I'll squash/fixup and merge this PR and then start working a new PR for the Partner name info for the email notification

CarsonF:
> Ok, once @bryanjnelson approves, I'll squash/fixup and merge this PR and then start working a new PR for the Partner name info for the email notification

I got it. I'd like there to be two distinct commits here since there are two distinct changes here. I don't like "fix up previous & add new thing" in one commit.

PR #3208, opened 2024-05-10
[EdgeDB] Pin queries

adam-soltech:
## [Monday task](https://seed-company-squad.monday.com/boards/.../pulses/...)
https://seed-company-squad.monday.com/boards/5989610236/pulses/5792971662

## Description


src/components/pin/pin.edgedb.repository.ts
```diff
@@ -0,0 +1,83 @@
...
+import { DateTime } from 'luxon';
+import { ID, PublicOf, Session } from '~/common';
+import { Client } from '~/core/edgedb';
+import { PinRepository } from './pin.repository';
+
+@Injectable()
+export class PinEdgeDBRepository implements PublicOf<PinRepository> {
+  constructor(@Inject(Client) private readonly client: Client) {}
+
+  async isPinned(id: ID, session: Session): Promise<boolean> {
```
bryanjnelson, line 11: @CarsonF - Since the schema itself defines a `pinned` property that is automatically inherited by any type that implements the `Pinnable` abstract type, doesn't that make this `isPinned` method unneeded in the EdgeDB world?

CarsonF: The `pinned` property is a computed shortcut. 
`User.pins` holds the links for the `Pinnable` objects that user has pinned.
So this could query cast the ID to the `Pinnable` instance and then check the `pinned` property on it.
That uses the current user global which has already been connected in the app code in another layer.
```edgeql
select (<Pinnable><uuid>$id).pinned
```
Or
```edgeql
with obj := <Pinnable><uuid>$id
select obj.pinned
```

src/components/pin/dto/pinnable.dto.ts
CarsonF, line 1: This isn't needed since `Pinnable` isn't a `Resource`.

src/components/pin/pin.edgedb.repository.ts
```diff
@@ -0,0 +1,83 @@
+import { Inject, Injectable } from '@nestjs/common';
+import { DateTime } from 'luxon';
+import { ID, PublicOf, Session } from '~/common';
+import { Client } from '~/core/edgedb';
+import { PinRepository } from './pin.repository';
+
+@Injectable()
+export class PinEdgeDBRepository implements PublicOf<PinRepository> {
+  constructor(@Inject(Client) private readonly client: Client) {}
```
CarsonF, line 9:
```suggestion
export class PinEdgeDBRepository 
  extends CommonRepository
  implements PublicOf<PinRepository> {
```

CarsonF (changes requested):
Was this a paste from AI? `BaseNode`, `PinnedRelation`, `pinned_by` don't exist within our schema.

Pins can be adjusted by `+=` or `-=` from the `User.pins` link.
```edgeql
update global currentUser
set {
  pins += <Mixin::Pinnable><uuid>$id
}
```
Though QB is preferred. Look at the other edgedb repos in the codebase for examples.

PR #3210, opened 2024-05-13
Add primary partner name to project workflow notification

atGit2021:
[Monday task](https://seed-company-squad.monday.com/boards/3451697530/pulses/6074211485)
https://seed-company-squad.monday.com/boards/5989610236/pulses/6509819818

### DIFF
--- a/src/components/project/project.rules.ts
+++ b/src/components/project/project.rules.ts
@@ -16,11 +16,13 @@ import {
 } from '~/common';
 import { ConfigService, ILogger, Logger } from '~/core';
 import { DatabaseService } from '~/core/database';
-import { ACTIVE, INACTIVE } from '~/core/database/query';
+import { ACTIVE, INACTIVE, merge } from '~/core/database/query';
 import { AuthenticationService } from '../authentication';
 import { withoutScope } from '../authorization/dto';
 import { EngagementService } from '../engagement';
 import { EngagementStatus } from '../engagement/dto';
+import { OrganizationService } from '../organization';
+import { Organization } from '../organization/dto';
 import { UserService } from '../user';
 import { User } from '../user/dto';
 import {
@@ -61,6 +63,7 @@ export interface EmailNotification {
   changedBy: Pick<User, 'id' | 'displayFirstName' | 'displayLastName'>;
   project: Pick<Project, 'id' | 'modifiedAt' | 'name' | 'step' | 'type'>;
   previousStep?: ProjectStep;
+  primaryPartnerName?: string | undefined;
 }
 
 const rolesThatCanBypassWorkflow: Role[] = [Role.Administrator];
@@ -75,6 +78,7 @@ export class ProjectRules {
     private readonly projectService: ProjectService & {},
     @Inject(forwardRef(() => EngagementService))
     private readonly engagements: EngagementService & {},
+    private readonly organizations: OrganizationService,
     @Inject(forwardRef(() => AuthenticationService))
     private readonly auth: AuthenticationService & {},
     private readonly configService: ConfigService,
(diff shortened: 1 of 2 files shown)


src/components/project/project.rules.ts
```diff
@@ -57,6 +58,7 @@ export interface EmailNotification {
   changedBy: Pick<User, 'id' | 'displayFirstName' | 'displayLastName'>;
   project: Pick<Project, 'id' | 'modifiedAt' | 'name' | 'step' | 'type'>;
   previousStep?: ProjectStep;
+  partnerName?: Pick<Organization, 'name'>;
```
bryanjnelson, line 61: Since we're only using one field from here, perhaps this type should just be a `string`?  I see why you did what you did though.  I'll let @CarsonF weigh in on that.

bryanjnelson (approved):
LGTM, @atGit2021!

@CarsonF - I'm leaving the Cypher for your review.

src/components/project/project.rules.ts
CarsonF: I agree there's a mismatch between the property type & name. I would go one way or the other to make them match.

atGit2021: Technically, the property needed is the organization name, so I can just change the property name to match the actual type/property that is being retrieved.  I think I named it "partner" because the UI and requirements view it as a "partner" name.

CarsonF: It's more of the fact that it's called "name" but it's an _object_ with a name.

atGit2021: ah, gotcha, so just leave pass and receive an "organization" similar to how we pass a project and then pull the props out when it arrives at the template?

atGit2021: I'm not sure how to change the cypher query to return the object vs the specific property in order to send the object.

atGit2021: If not that way, then I could change the type from the object to a string `organizationName: string` then I don't have to mess with the cypher query. let me know which you prefer

src/components/project/project-rules/project.rules.ts
```diff
@@ -57,6 +59,7 @@ export interface EmailNotification {
   changedBy: Pick<User, 'id' | 'displayFirstName' | 'displayLastName'>;
   project: Pick<Project, 'id' | 'modifiedAt' | 'name' | 'step' | 'type'>;
   previousStep?: ProjectStep;
+  organizationName?: string | undefined;
```
CarsonF, line 62:
```suggestion
  primaryPartnerName?: string;
```
Would be better.
"Which partner?" - the primary one.
We'll try to stop using organizations probably, so favor that wording.

src/components/project/project-rules/project.rules.ts
```diff
@@ -1131,11 +1136,22 @@ export class ProjectRules {
       recipientSession,
     );
 
+    let organization;
+    const perms = this.privileges.for(recipientSession, Organization);
+
+    if (perms.can('read')) {
```
CarsonF, line 1142: This is checking if the recipient can read _all_ organizations, which isn't what we want.
```ts
const orgId = db.query(...);
const org = this.resourcesLoader.load(Organization, orgId);
const name = this.privileges
    .for(recipientSession, Organization, org)
    .can('read', 'name')
  ? org.name : undefined
```

CarsonF (changes requested):
I wouldn't do all this refactoring in this same PR with the business logic change.
Additionally, it needs to be rewritten all together to match progress report workflow.
This is the only way EdgeDB will be able to use "previous step".

atGit2021:
> I wouldn't do all this refactoring in this same PR with the business logic change.
> Additionally, it needs to be rewritten all together to match progress report workflow.
> This is the only way EdgeDB will be able to use "previous step".

@CarsonF are you referring to creating the new module and splitting out the two repo files in order to port to edgedb?  If so, should I revert all the edge related work and only test to make sure it works on the neo4j side?

CarsonF:
> @CarsonF are you referring to creating the new module and splitting out the two repo files in order to port to edgedb? If so, should I revert all the edge related work and only test to make sure it works on the neo4j side?

Yes and yes

atGit2021:
@CarsonF I tried to get a rid of a conflict that was only showing on the remote side, but not sure if I did it correctly and why I don't see these errors on my local.

CarsonF (approved):
LGTM!

atGit2021:
@CarsonF tests look good!

PR #3213, opened 2024-05-16
[Hotfix] Change `Promise.all` to `for await` in project status handler to catch UnauthorizedException

andrewmurraydavid:
## Description
This PR changes how the engagements are updated in the project handler.
There seems to be an issue, either with the the way `Promise.all` handles the UnauthorizedException for the edgecase where a project has 3 engagements, one (or more) of them being in a step previous (for example 2 in `Active change plan` and 1 in `Discussing change plan`), the engagement update would fail with the UnauthorizedException. 

It is assumed that because Promis.all not handling the exception properly, would cause a neo4j transaction creation to retry in an infinite loop.

This infinite loop caused the API service to crash completely being OOM (out of memory).

## Steps to reproduce the bug
In the description

## Expected behavior
Not crash, lol

- [x] THIS IS A HOTFIX

### DIFF
--- a/src/components/engagement/engagement.rules.ts
+++ b/src/components/engagement/engagement.rules.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-case-declarations */
 import { Injectable } from '@nestjs/common';
 import { node, relation } from 'cypher-query-builder';
-import { first, intersection } from 'lodash';
+import { first, intersection, startCase } from 'lodash';
 import {
   ID,
   ServerException,
@@ -383,7 +383,9 @@ export class EngagementRules {
     );
     if (!validNextStatus) {
       throw new UnauthorizedException(
-        'This status is not in an authorized sequence',
+        `One or more engagements cannot be changed to ${startCase(
+          nextStatus,
+        )}. Please check engagement statuses.`,
         'engagement.status',
       );
     }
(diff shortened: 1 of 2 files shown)


src/components/engagement/engagement.rules.ts
```diff
@@ -383,7 +383,9 @@ export class EngagementRules {
     );
     if (!validNextStatus) {
       throw new UnauthorizedException(
-        'This status is not in an authorized sequence',
+        `One or more engagements cannot be changed to ${startCase(
```
bryanjnelson, line 386: Is this called from anywhere else?  Meaning, will changing this message inadvertently show it somewhere else where it doesn't make sense?

andrewmurraydavid: It's "complicated" a bit. But very straight forward. Anytime there's a mutation to update the engagement or and handler call to update the engagement.

It's called only from the engagement service:
image

Which is called by the handler and resolver:
image
image

brentkulwicki (approved):

bryanjnelson (approved):

CarsonF:
UnauthorizedExceptions should not be treated as retryable. But maybe it was the error from the other queries?
I don't think this is the root solution, other queries in parallel could trigger this symptom too.
But I'm pretty sure EdgeDB doesn't even let you run queries in parallel while in a transaction so there's that too. Which warrants this change.

Good job @andrewmurraydavid! Race conditions are hard to track down :+1:

PR #3218, opened 2024-05-29
Bump neo4j driver / Improve connection handling in CLI

CarsonF:
- Bump neo4j driver
- Patch neo4j driver to stop preventing the process from closing when waiting to connect.
  This is felt in CLI (console/REPL), where a command could finish before DB connection is created or needed. This is also exacerbated by the fact that sometimes the socket connection doesn't emit data (from server, though I'm not convinced it's an actual server problem) in these short windows. So the driver is left waiting for a successful response to a `logon` command, but it never shows up. 
- Allow the process to close while waiting for `waitForRootUserId()`
  Like the above, if the retry loop is a timed out waiting state, don't let that block the process from closing.
- Patch neo4j driver to fix infinite loop with errors in transactions.
  This will be released in 5.21 tomorrow(?) but applying now since we are sensitive to this.
- Don't run neo4j migrations in REPL
- Share neo4j queries asking for server info.
  Now multiple calls asking for this (or using this to "wait for successful connection"), will share a single underlying promise. If the promise is successful, it will be cached for 3 minutes.


bryanjnelson (approved):
Nice!

PR #3220, opened 2024-05-29
[Neo4j] `DtoRepo.delete`

CarsonF:
- Add `DtoRepo.delete` to replace `deleteNode`
  This query is specific to the resource type of the class, which is safer (assumes less) than any type.
  It also matches the EdgeDB version so migration should be seamless.
- Mask generic Neo4j errors just like with EdgeDB
  There shouldn't be any need to:
  ```ts
  } catch (e) {
    ...
    throw new ServerException('Failed to do X');
  }
  ```
  Now you can `throw e` or skip `catch` all together if there's no other logic.

### DIFF
--- a/src/core/database/dto.repository.ts
+++ b/src/core/database/dto.repository.ts
@@ -16,7 +16,7 @@ import { DbChanges, getChanges } from './changes';
 import { CommonRepository } from './common.repository';
 import { DbTypeOf } from './db-type';
 import { OnIndex } from './indexer';
-import { matchProps } from './query';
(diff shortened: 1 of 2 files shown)


bryanjnelson (approved):
❤️

CarsonF:
Ah this doesn't work because it forces all subclasses to conform to this `delete` signature.
We still want subclasses to be able to change that as this is just a builder/helper class.
I'm not going to go through the effort of implementing the `omit` logic here, which is how we got around this on the EdgeDB side.

PR #3225, opened 2024-05-31
Fix MadeEnum type to provide value & label from entries implicitly

CarsonF:
## Fix type to provide value & label from entries implicitly

Currently:
```ts
const Status = makeEnum(...);
Status.entry(...) // => { value: Status, label: string }
```
BUT if referencing the type generically:
```ts
const statusEnum: MadeEnum<Status>;
```
The entries didn't have a shape:
```ts
statusEnum.entry(...) // => empty object
```
Now they have the implicit shape:
```ts
statusEnum.entry(...) // => { value: Status, label: string }
```

## Flip `Extra` positional generic

This allows declaring extra entry declarations easier, by making it the second arg instead of third.
```ts
const statusEnum: MadeEnum<Status, { color: string }>;
statusEnum.entry(...).color // => string
```

This `Extra` generic really has nothing to do with the enum functionality, it's just extra "statics".
The type could be written as:
```ts
MadeEnum<Status> & { <Extras> }
```
Which we can still do now when needing to reference a generic enum with extra "statics".

I decided to keep the generic though as that helps TS infer how to emit the type.
```ts
const X = makeEnum(...); // X is represented as "MadeEnum<...>"
```
As opposed to some obtuse object literal shape.

### DIFF
--- a/src/common/make-enum.ts
+++ b/src/common/make-enum.ts
@@ -2,7 +2,6 @@ import { registerEnumType } from '@nestjs/graphql';
 import { cleanJoin, mapKeys, nonEnumerable } from '@seedcompany/common';
 import { lowerCase } from 'lodash';
 import { titleCase } from 'title-case';
-import { SetRequired } from 'type-fest';
 import { inspect, InspectOptionsStylized } from 'util';
 
 export type EnumType<Enum> = Enum extends MadeEnum<infer Values, any, any>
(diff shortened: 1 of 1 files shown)


src/common/make-enum.ts
```diff
@@ -191,21 +200,19 @@ type ValuesOfDeclarations<ValueDeclaration extends EnumValueDeclarationShape> =
  * properties as optional.
  */
 type NormalizedValueDeclaration<Declaration extends EnumValueDeclarationShape> =
-  SetRequired<Pick<EnumValueDeclarationObjectShape<any>, 'label'>, 'label'> &
```
CarsonF, line 194: Essentially just moved this from here, which is only referenced in output of `makeEnum` to [new line 19](https://github.com/SeedCompany/cord-api-v3/pull/3225/files#diff-5d4630304550842a4c2a9d27a4019353b81c45c87a67a02eceeba07a0682cff2R19) within `MadeEnum`.

This allows others to use the `MadeEnum` type and get the implicits that were previously only available from the return type of `makeEnum`

bryanjnelson (approved):
You're a TypeMaster.  💪🏼

PR #3227, opened 2024-06-04
Fix `MadeEnum<'a'>` to be compatible with `MadeEnum<string>`

CarsonF:
```ts
// Given
let Color: MadeEnum<'red' | 'blue'>;
let AnEnum: MadeEnum<string>;

// Now
AnEnum = Color;
```
Two changes to make this compatible.

First, disallow direct access to member values when referencing generically.

Without this change `Color` can't be assigned to `AnEnum`, because the TS defines the member values as an index: `{ [x: string]: string }`. So since one has an index accessor and one doesn't, making them incompatible.

Secondly, The `entry()` arg type change also allows them to be compatible. Without it TS thinks something like `entry()` fn accepting any `string` is not compatible with `entry()` fn only accepting `'red' | 'blue'`. Using another generic at the function level somehow works around this, while still maintaining all the strictness.

### DIFF
--- a/src/common/make-enum.ts
+++ b/src/common/make-enum.ts
@@ -12,13 +12,18 @@ export type MadeEnum<
   Values extends string,
   ValueDeclaration = EnumValueDeclarationShape,
   Extra = unknown,
-> = {
-  readonly [Value in Values & string]: Value;
-} & EnumHelpers<
+> = EnumHelpers<
   Values,
   ValueDeclaration & ImplicitValueDeclarationShape<Values>
 > &
-  Readonly<Extra>;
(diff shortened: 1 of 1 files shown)


brentkulwicki (approved):
This all makes sense, but my curiosity wants to know a real life example of where this will be handy in our codebase? Or are you optimizing for the future here?

CarsonF:
> This all makes sense, but my curiosity wants to know a real life example of where this will be handy in our codebase? Or are you optimizing for the future here?

PR incoming for **Workflow**, which will have a _state_ enum, typed generically.

PR #3229, opened 2024-06-05
Project workflow

atGit2021:
Creating Draft PR in order to tie this to the front end PR


CarsonF:
I'd rather have the PR come from myself for this for historical records and since it's so large.

CarsonF:
> Sorry, I created a branch off of your branch for the 0883 task and thought it would be a separate PR, but after I opened it I realized it just gets added to your project-workflow PR.

np that one just needs the base branch changed to this branch. I'll do that. I'll still use this branch to make a PR.


> I also created a cordfield repo PR for the front end work, but I needed a backend PR to tie it too so the schema would not throw errors during the CI tests

Yeah that's valid. I probably would've made a draft PR for you for that. That one is cool hanging out for a bit though. This work doesn't have breaking changes.

PR #3230, opened 2024-06-06
[EdgeDB] Workaround QB typing overloads from ancestors

CarsonF:
Currently the QB doesn't map types correctly for overloaded properties/links that are declared above the direct parent.
This works around that by manually fixing the shapes for our schema.


CarsonF:
@bryanjnelson This will fix your problem with `Engagement.project` type.

bryanjnelson (approved):
Sick. 🚢

PR #3231, opened 2024-06-06
Project Workflow - PMs should not be able to bypass Consultant checks for Internships

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/5641346058)


src/components/project/workflow/transitions/conditions.ts
```diff
@@ -23,6 +24,45 @@ export const IsMultiplication: Condition = {
   },
 };
 
+export const hasValidRoleForProjectType: Condition = {
```
CarsonF, line 27: This is an authorization concern, so the policies should handle it. We need to just figure out how to get the "internship type condition" there. Let me think about it.
I get that's it's kinda weird we have two ways and both have conditions.

src/components/project/workflow/transitions/conditions.ts
```diff
@@ -23,6 +24,45 @@ export const IsMultiplication: Condition = {
...
+        };
+      default:
+        return {
+          status: 'OMIT',
+        };
+    }
+  },
+};
+
+export const hasEngagement: Condition = {
```
CarsonF, line 56:
```suggestion
export const HasEngagement: Condition = {
```

src/components/project/workflow/transitions/conditions.ts
```diff
@@ -23,6 +24,45 @@ export const IsMultiplication: Condition = {
...
+  description: 'Has an engagement',
+  resolve({ project }) {
+    return {
+      status: project.engagementTotal > 0 ? 'ENABLED' : 'DISABLED',
+      disabledReason: `The project requires an engagement`,
```
CarsonF, line 61:
```suggestion
      disabledReason: `Create an engagement first`,
```
Is this better?

CarsonF:
This was actually implemented in 55777cd168e8a69eb1c42b4189a5306cf5f5f9f7

PR #3234, opened 2024-06-07
New way to merge interfaces that's variadic and DRY for TS & GQL

CarsonF:
I held off on this because TS doesn't natively support variadic generic parameters, so the conventional workaround is to add overloads for 1-10ish parameters before giving up.

But what we do right now is so clunky and we need up enumerating each interface 3 times.
So I bit the bullet. The type code is long, but cleans up the business logic nicely.

I also moved the remaining secured enum wrappers to their respective files. Most were already there, but there were some stragglers.


bryanjnelson (approved):
🤯 🚀

PR #3239, opened 2024-06-13
Bugfix for Update Language 

atGit2021:
[Monday task](https://seed-company-squad.monday.com/boards/3451697530/pulses/6587752956)


atGit2021:
I think I'll need to add a check for an empty object for simpleChanges too, but will wait for your initial feedback

CarsonF (commented):
This doesn't appear change anything logically.
What error are you trying to fix?
How does this change that error?

CarsonF:
Real fix: https://github.com/SeedCompany/cord-field/pull/1541

PR #3240, opened 2024-06-14
[Gel] Add `FinancialReport` and `NarrativeReport` seeding

bryanjnelson:
## Description

Add seeding for `FinancialReport`s and `NarrativeReport`s.

- [X] Change the task url above to the actual Monday task
- [X] Add reviewers to this PR

### DIFF
--- a/dbschema/seeds/020.financial-narrative-reports.edgeql
+++ b/dbschema/seeds/020.financial-narrative-reports.edgeql
@@ -1,0 +1,109 @@
+with
+  periodicReportsJson := to_json('[
+    {
+      "type": "Narrative",
+      "project": "Misty Mountains",
+      "start": "2022-01-01",
+      "end": "2022-03-31",
+      "receivedDate": "2022-04-01",
+      "skippedReason": "No funding"
+    },
+    {
+      "type": "Narrative",
+      "project": "Arnor Lake",
+      "start": "2023-01-01",
+      "end": "2023-03-31",
+      "skippedReason": "Change in translation team"
+    },
+    {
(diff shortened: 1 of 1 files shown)


bryanjnelson:
@CarsonF - The open question here is if we would also like to include seeding for `ProgressReport`s in this same file.  Since it is considerably different, I lean toward making that an entirely different seeding file.  Thoughts?

CarsonF (commented):
Seems fine. I don't really like the denormalized dates, since these are based on the project date range.

CarsonF:
Oh can we pick a project to be monthly financial reporting and add those here.

bryanjnelson:
> Oh can we pick a project to be monthly financial reporting and add those here.

```
    {
      "type": "Financial",
      "project": "Glorfindel - Exegetical Facilitator",
      "start": "2025-01-01",
      "end": "2025-03-31",
      "receivedDate": "2025-04-01"
    },
```

That one is in there and the project looks like this already:

```
  {
    name: 'Glorfindel - Exegetical Facilitator',
    step: 'DiscussingChangeToPlan',
    mouStart: '2021-07-01',
    mouEnd: '2024-12-30',
    primaryLocation: 'Sweden',
    estimatedSubmission: '2024-10-01',
    financialReportPeriod: 'Monthly',
  },
  ```
  
  Is that what you were going after?

CarsonF:
> Is that what you were going after?

Oh yeah. So that start/end date is wrong. The reports should be monthly not quarterly.

rdonigian:
Moving away from Edge/Gel

PR #3251, opened 2024-06-20
Notify Financial Approvers only when they need to take action

CarsonF:
Notify Financial Approvers when they need to take action, not when they just have taken action


atGit2021 (commented):
LGTM, but I'll leave Sheri/Renato to confirm/approve the notification business rules

CarsonF:
Going to merge to let them review in dev UI

renatorossiter (approved):
It looks good to me. Sheri confirmed yesterday, and I agree with her.

PR #3264, opened 2024-07-08
More commentable resources

andrewmurraydavid:
## Description
Made more resources commentable


bryanjnelson (approved):
What's here looks good to me, although we should have @CarsonF weigh in.  I'm not sure what else an addition like this could entail.

CarsonF:
Thanks for getting this started @andrewmurraydavid.
I'm finishing it off in #3287.
I adjusted the `__typename` in the queries & updated EdgeDB schema to match.

PR #3265, opened 2024-07-09
Update Partners to include growthPartnersClient field

rdonigian:
https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/6787472941

### DIFF
--- a/dbschema/migrations/00012-m1pv2oj.edgeql
+++ b/dbschema/migrations/00012-m1pv2oj.edgeql
@@ -1,0 +1,9 @@
+CREATE MIGRATION m1pv2ojzfsv7yjfeac7r4pb23nogec235dasqgmosh2utwzome3bcq
+    ONTO m17u4aufxga7wgmcebhiaapulc7xrqg34hcbmucddbvm4tw5c63kyq
+{
+  ALTER TYPE default::Partner {
+      ALTER PROPERTY globalInnovationsClient {
+          RENAME TO growthPartnersClient;
+      };
+  };
+};
--- a/src/components/partner/dto/create-partner.dto.ts
+++ b/src/components/partner/dto/create-partner.dto.ts
@@ -36,8 +36,14 @@ export abstract class CreatePartner {
   })
   readonly pmcEntityCode?: string;
 
-  @Field({ nullable: true })
+  @Field({
+    nullable: true,
+    deprecationReason: 'Use `growthPartnersClient` instead',
+  })
   readonly globalInnovationsClient?: boolean;
+
+  @Field({ nullable: true })
+  readonly growthPartnersClient?: boolean;
 
   @Field({ nullable: true })
   readonly active?: boolean;
--- a/src/components/partner/dto/partner.dto.ts
+++ b/src/components/partner/dto/partner.dto.ts
@@ -51,8 +51,13 @@ export class Partner extends Interfaces {
   @Field()
   readonly pmcEntityCode: SecuredStringNullable;
 
+  @Field({
+    deprecationReason: 'Use `growthPartnersClient` instead',
+  })
+  readonly globalInnovationsClient?: SecuredBoolean;
+
   @Field()
-  readonly globalInnovationsClient: SecuredBoolean;
+  readonly growthPartnersClient: SecuredBoolean;
 
   @Field()
   readonly active: SecuredBoolean;
--- a/src/components/partner/partner.resolver.ts
+++ b/src/components/partner/partner.resolver.ts
@@ -15,6 +15,7 @@ import {
   loadSecuredIds,
   LoggedInSession,
   mapSecuredValue,
+  SecuredBoolean,
   Session,
 } from '~/common';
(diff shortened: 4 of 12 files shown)


brentkulwicki (commented):
👍 Let's let @bryanjnelson check this out as well, but I have no issues with it

brentkulwicki (approved):

src/components/partner/dto/partner.dto.ts
```diff
@@ -51,8 +51,13 @@ export class Partner extends Interfaces {
...
   readonly pmcEntityCode: SecuredStringNullable;
 
+  @Field({
+    deprecationReason: 'Use `growthPartnersClient` instead',
+  })
+  readonly globalInnovationsClient?: SecuredBoolean;
+
   @Field()
-  readonly globalInnovationsClient: SecuredBoolean;
+  readonly growthPartnersClient: SecuredBoolean;
```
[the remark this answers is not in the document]
brentkulwicki: Ignore this. Figured it out as I went on.

dbschema/migrations/00012-m1pv2oj.edgeql
bryanjnelson, line 1: @CarsonF - Are us developers attaching these migration files to PRs now, or is this something that you are still doing in a separate PR?

src/components/partner/partner.resolver.ts
```diff
@@ -86,6 +87,11 @@ export class PartnerResolver {
     );
   }
 
+  @ResolveField()
+  growthPartnersClient(@Parent() partner: Partner): SecuredBoolean {
+    return partner.growthPartnersClient ?? partner.globalInnovationsClient;
```
bryanjnelson, line 92: I see what you're trying to accomplish here.  I think this works, however, I've never seen the same field both here as a `ResolveField` _and_ on the dto itself as well.  I think you should probably remove it from the dto and keep it here instead during the transition.  Could you make that change and test it that way?

bryanjnelson: Hmmm....I guess I was wrong here.  Ignore this.  🙃

CarsonF (changes requested):
Let's rename this. I now understand that this is just a flag to denote that the partner is for our "Growth Partner" strategy/initiative.

"client" is redundant.

Maybe `Partner.isGrowthPartner`. The "Partner" suffix seems redundant as well, but I'm not sure if "isGrowth" could be associated with our "growth partners" initiative.
Maybe
```
growthPartnerInitiative
growthInitiativePartner
isGrowthInitiative
```

---

My other thought is converting to an enum. `Partner.type = "Growth" or "X"`.

We already have `Partner.types: PartnerType[]`, but I see now that this is misnamed.
We should change it to 
```
Partner.approvedPartnershipRoles: PartnershipRole[]
Partnership.roles: PartnershipRole[]
```

src/components/partner/dto/partner.dto.ts
```diff
@@ -51,8 +51,13 @@ export class Partner extends Interfaces {
   @Field()
   readonly pmcEntityCode: SecuredStringNullable;
 
+  @Field({
+    deprecationReason: 'Use `growthPartnersClient` instead',
+  })
+  readonly globalInnovationsClient?: SecuredBoolean;
```
CarsonF, line 57: This is not needed, because we have the field resolver that will map the new property name/value to the old one.
We don't want to declare that our internal shapes have this property - they have the new one instead.

src/components/partner/dto/create-partner.dto.ts
```diff
@@ -36,9 +36,15 @@ export abstract class CreatePartner {
   })
   readonly pmcEntityCode?: string;
 
-  @Field({ nullable: true })
+  @Field({
+    nullable: true,
+    deprecationReason: 'Use `growthPartnersClient` instead',
+  })
```
CarsonF, line 42: Maintaining this _input_ field (as deprecated) is appropriate for a graceful migration. But they need to be used. They should be mapped to the new input properties in the service/repo.

src/components/partner/partner.resolver.ts
CarsonF: Yes, you're correct. The `deprecatedReason` should be declared here too.

dbschema/migrations/00012-m1pv2oj.edgeql
CarsonF: Either way. It's more fine now since we have edgedb branches. But it is a handful of steps for the author, and more if there's a need to change or rebase it.

CarsonF (changes requested):
Hold off until we finalize the new name, but here's what needs to change other than that.

Thanks!

CarsonF:
Alright after talking with @sethmcknight, he's informed us that really this needs to turn into `PartnerProject` & `PartnerEngagement`. And then this could be something like "does this Partner have an active project?"

With that in mind I say we leave this boolean as is for now so we can replace it all together with the more robust functionality.

PR #3267, opened 2024-07-10
Remove legacy APIs & transitional code for project workflow

CarsonF:
UI is migrated to new paths & doesn't reference these type names.

### DIFF
--- a/src/components/project/dto/create-project.dto.ts
+++ b/src/components/project/dto/create-project.dto.ts
@@ -17,7 +17,6 @@ import {
 } from '~/common';
 import { type Location } from '../../location/dto';
 import { ReportPeriod } from '../../periodic-report/dto';
-import { ProjectStep } from './project-step.enum';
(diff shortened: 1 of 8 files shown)


CarsonF:
@bryanjnelson thoughts here? We aren't using this in UI, DOMO, or multiplication sync? :shipit: ?

bryanjnelson:
> @bryanjnelson thoughts here? We aren't using this in UI, DOMO, or multiplication sync? :shipit: ?

Originally we said we thought we were still going to be using this path from the multiplication sync.  I guess we landed on not doing that?  🤔

CarsonF:
> Originally we said we thought we were still going to be using this path from the multiplication sync. I guess we landed on not doing that? 🤔

Right we aren't doing that. We decided to the proper thing there and look at the available transitions.

bryanjnelson (approved):
Out with the old! :shipit:

PR #3271, opened 2024-07-31
Add Project.partnership(partner: id)

atGit2021:
[Monday task](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/7046940396)

### DIFF
--- a/src/components/project/project.resolver.ts
+++ b/src/components/project/project.resolver.ts
@@ -42,10 +42,12 @@ import { OrganizationLoader } from '../organization';
 import { SecuredOrganization } from '../organization/dto';
 import { PartnershipLoader } from '../partnership';
 import {
+  Partnership,
   PartnershipListInput,
   SecuredPartnership,
   SecuredPartnershipList,
 } from '../partnership/dto';
+import { PartnershipByProjectAndPartnerLoader } from '../partnership/partnership-by-project-and-partner.loader';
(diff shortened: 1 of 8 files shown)


CarsonF (approved):

src/components/project/project.resolver.ts
```diff
@@ -254,6 +258,20 @@ export class ProjectResolver {
...
+    @Parent() project: Project,
+    @IdArg({ name: 'partner' }) partnerId: ID,
+    @Loader(PartnershipByProjectAndPartnerLoader)
+    loader: LoaderOf<PartnershipByProjectAndPartnerLoader>,
+  ): Promise<Partnership> {
+    const result = await loader.load({
+      project: project.id,
+      partner: partnerId,
+    });
+    return result.partnership;
```
atGit2021, line 272: @CarsonF we see the additional `id` property coming back which contains the project id and the partner id, but we only send back results.partnership.  Should we revise so that we are sending both the `id` and `partnership` objects back to the front end?

CarsonF: I don't understand? are you getting a schema error?

atGit2021: No error, we see the partnership object coming through on the front end, but wasn't sure in your design plans whether you wanted that id object also passed back up to the front end.

atGit2021: ok

PR #3279, opened 2024-08-16
[Gel] - Add triggers to create/update periodic reports upon project creation

### LINKED ISSUE #3478: Gel - PeriodicReport - Triggers

bryanjnelson:
Closes #3478.

## Description
This adds `createPeriodicReports` and `addRemovePeriodicReports` triggers to create/update the appropriate number of periodic reports when a project is created, based upon project `mouStart`, `mouEnd`, and `financialReportPeriod`.

Open items:

- [ ] The biggest issue now is that only single points of change are accounted for in these triggers.  Meaning that an mou project start date change will be handled, and an mou project end date change will be handled, but if both change at the same time the update trigger might not handle that appropriately.  All single field updates are tested and accounted for however.
- [ ] Some functions could potentially be combined to take advantage of polymorphism, however, in practice this has caused issues
- [ ] Naming?
- [ ] Function refactoring for more concise code?

### DIFF
--- a/dbschema/project.gel
+++ b/dbschema/project.gel
@@ -60,6 +60,7 @@ module default {
     status := Project::statusFromStep(.step);
     latestWorkflowEvent := (select .workflowEvents order by .at desc limit 1);
     workflowEvents := .<project[is Project::WorkflowEvent];
+
     trigger assertMatchingLatestWorkflowEvent after insert, update for each do (
       assert(
         __new__.latestWorkflowEvent.to ?= __new__.step
@@ -130,7 +131,13 @@ module default {
       }
     );
 
-    trigger createPeriodicReports after insert for each do (
+    trigger createPeriodicReports after insert for each 
+      when (
+        exists __new__.financialReportPeriod
+        and exists __new__.mouStart
+        and exists __new__.mouEnd
+      )  
+      do (
       with
         interval := (select 
           if __new__.financialReportPeriod = default::ReportPeriod.Monthly then '1' else '3'),
@@ -138,30 +145,9 @@ module default {
           __new__.mouStart,
           __new__.mouEnd,
           interval
-        )
-      for reportRange in reportRanges 
-      union (
-        (insert default::FinancialReport {
-          createdAt := datetime_of_statement(),
-          modifiedAt := datetime_of_statement(),
-          createdBy := assert_exists(global currentActor),
-          modifiedBy := assert_exists(global currentActor),
-          project := __new__,
-          projectContext := __new__.projectContext,
-          container := __new__,
-          period := reportRange
-        }),
-        (insert default::NarrativeReport {
-          createdAt := datetime_of_statement(),
-          modifiedAt := datetime_of_statement(),
-          createdBy := assert_exists(global currentActor),
-          modifiedBy := assert_exists(global currentActor),
-          project := __new__,
-          projectContext := __new__.projectContext,
-          container := __new__,
-          period := reportRange
-        })
-      )
+        ),
+        insertedReports := Project::insert_reports(array_agg(reportRanges), __new__)
+      select insertedReports
     );
 
     trigger addRemovePeriodicReports after update for each 
@@ -192,86 +178,44 @@ module default {
             select existingReports
             filter not exists .reportFile
         )  
-      select if not exists __new__.mouStart or not exists __new__.mouEnd then (     
+      select
+      # start date, end date, or financial report period were deleted
+      if not exists __new__.mouStart
+        or not exists __new__.mouEnd
+        or not exists __new__.financialReportPeriod then (
         for report in reportsForDeletion 
         union (
           delete report 
         )
       )
+      # financial report period changes, start date is moved forward, or end date is moved backward
       else if __old__.financialReportPeriod ?!= __new__.financialReportPeriod
         or (__new__.mouStart > __old__.mouStart) ?? false 
         or (__new__.mouEnd < __old__.mouEnd) ?? false then (
         with
-          requestedReportPeriodsForInsertion := (
-            select requestedReportPeriods
-            filter requestedReportPeriods not in existingReports.period
+          periodsForInsertion := Project::determine_requested_report_periods(
+            array_agg(requestedReportPeriods), 
+            array_agg(existingReports)
           ),
-          financialReports := (for reportPeriod in requestedReportPeriodsForInsertion
+          insertedReports := Project::insert_reports(array_agg(periodsForInsertion), __new__),
+          deletedReports := (for report in reportsForDeletion 
           union (
-            insert default::FinancialReport {
-              createdAt := datetime_of_statement(),
-              modifiedAt := datetime_of_statement(),
-              createdBy := assert_exists(global currentActor),
-              modifiedBy := assert_exists(global currentActor),
-              project := __new__,
-              projectContext := __new__.projectContext,
-              container := __new__,
-              period := reportPeriod,
-            }
-          )),
-          narrativeReports := (for reportPeriod in requestedReportPeriodsForInsertion
-          union (
-            insert default::NarrativeReport {
-              createdAt := datetime_of_statement(),
-              modifiedAt := datetime_of_statement(),
-              createdBy := assert_exists(global currentActor),
-              modifiedBy := assert_exists(global currentActor),
-              project := __new__,
-              projectContext := __new__.projectContext,
-              container := __new__,
-              period := reportPeriod,
-            }
+            delete report
+            filter report.period not in requestedReportPeriods
           ))
-        for report in reportsForDeletion 
-        union (
-          delete report
-          filter report.period not in requestedReportPeriods
-        )  
-      ) 
+        select insertedReports
+      )
+      # start or end dates otherwise change
       else if newMouStart ?!= oldMouStart 
         or newMouEnd ?!= oldMouEnd then (
         with
-          requestedReportPeriodsForInsertion := (
-            select requestedReportPeriods
-            filter requestedReportPeriods not in existingReports.period
-          ),
-          financialReports := (for reportPeriod in requestedReportPeriodsForInsertion
-          union (
-            insert default::FinancialReport {
-              createdAt := datetime_of_statement(),
-              modifiedAt := datetime_of_statement(),
-              createdBy := assert_exists(global currentActor),
-              modifiedBy := assert_exists(global currentActor),
-              project := __new__,
-              projectContext := __new__.projectContext,
-              container := __new__,
-              period := reportPeriod,
-            }
-          )),
-          narrativeReports := (for reportPeriod in requestedReportPeriodsForInsertion
-          union (
-            insert default::NarrativeReport {
-              createdAt := datetime_of_statement(),
-              modifiedAt := datetime_of_statement(),
-              createdBy := assert_exists(global currentActor),
-              modifiedBy := assert_exists(global currentActor),
-              project := __new__,
-              projectContext := __new__.projectContext,
-              container := __new__,
-              period := reportPeriod,
-            }
-          ))
-        select financialReports union narrativeReports
+          periodsForInsertion := Project::determine_requested_report_periods(
+            array_agg(requestedReportPeriods), 
+            array_agg(existingReports)
+          ),        
+          insertedReports := Project::insert_reports(array_agg(periodsForInsertion), __new__)
+        select insertedReports
+      # nothing changes
       ) else (
         select <PeriodicReport>{}
       )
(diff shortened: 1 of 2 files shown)


dbschema/common.esdl
```diff
@@ -27,8 +27,4 @@ module default {
   # This will be reflected to dynamically inject portable shapes in our EdgeQL queries.
   function hydrate(typeName: str, scopedValue: json) -> str
     using (typeName);
-  
-  # Get the inclusive upper bound of the given date range.
-  function date_range_get_upper(period: range<cal::local_date>) -> cal::local_date
-    using (<cal::local_date><str>assert_exists(range_get_upper(period)) - <cal::date_duration>"1 day");
```
CarsonF, line 33: This is still needed. The typecasting workaround was fixed though
```edgeql
  function date_range_get_upper(period: range<cal::local_date>) -> cal::local_date
    using (assert_exists(range_get_upper(period)) - <cal::date_duration>"1 day");
```

Ranges are half closed `[start, end)` meaning the endpoint is not considered to be included.
This matters when we start doing math, like does x date fall within this range?
For example this work week. We see it as
```
[2024-09-02, 2024-09-06]
```
but SQL/EdgeQL stores that as
```
[2024-09-02, 2024-09-06)
```
meaning the 6th is NOT included in that range.

That's why we add one day
```
[2024-09-02, 2024-09-07) == [2024-09-02, 2024-09-06]
```

And that's why this function exists, because when we display the "end date" we want it to be the inclusive endpoint, not the half closed one.

dbschema/project.esdl
```diff
@@ -119,6 +119,41 @@ module default {
...
+        (insert default::FinancialReport {
+          createdAt := datetime_of_statement(),
+          modifiedAt := datetime_of_statement(),
+          createdBy := assert_exists(global currentActor),
+          modifiedBy := assert_exists(global currentActor),
+          project := __new__,
+          projectContext := __new__.projectContext,
+          container := __new__,
+          period := reportRange
+          #receivedDate := __new__.financialReportReceivedAt, //TODO - what is this at the project level?
```
CarsonF, line 143: That's what the user defines when the file is uploaded. "Here is the word doc from the field partner which I'm uploading on the 7th, but I actually received the email from them on the 4th"

dbschema/project.esdl
```diff
@@ -206,4 +241,22 @@ module Project {
       on target delete allow;
     };
   }
+
+  # creates the ranges for the given start and end dates based upon the given month interval
+  function create_periodic_report_ranges(startDate: cal::local_date, endDate: cal::local_date,
```
CarsonF, line 246: This is wild. Just noting that I have not vetted this for accuracy.

CarsonF (changes requested):
Wow love the initiative!

For this to be fully out of app code, we also need to handle:
- Date range expanding (on update), which would add new periods.
  Jan-March -> Jan-July
- Date range shrinking (on update), which would delete _empty_ periods
  Jan-May -> Jan-March
  But we only want to delete reports that [don't have a file uploaded or a progress report started](https://github.com/SeedCompany/cord-api-v3/blob/8919871e4651016fa5a7610b150d88a85de4700f/src/components/periodic-report/periodic-report.repository.ts#L371-L377)
- On financial reporting frequency change create new and delete old empty periods.
  Monthly (Jan, Feb, March) -> Quarterly (Jan-March)

Honestly "on insert" will never be hit by users since they don't define the date ranges when they create the project.
May be useful for migration though?

> Should the create_periodic_report_ranges function be made more generic to handle other use cases? (Probably not.)

Probably not

> Is the `create_periodic_report_ranges` function located in the best place? It could be moved to the `common.esdl`, but if this is the only place it will be used then no. It could also potentially be moved to the `periodic-report.esdl`?

Probably their own files for each concrete report. They all have distinct use-cases / logic.

> It's currently problematic that the `mouStart`, `mouEnd`, and `financialReportPeriod` fields are not required on a `Project`; this will obviously prevent any reports from being generated correctly. We'll need to figure out the best way to handle that. Make them required? Update the trigger to also run on update when all 3 are finally populated?

That's what we do in app code. If all 3 are given, synchronize.

dbschema/common.esdl
bryanjnelson: I _think_ I understand what you're saying conceptually, but something still isn't adding up for me in practice.

When I run the `create_periodic_report_ranges` in the context of the trigger I created I receive the expected results.  For example, if I call `create_periodic_report_ranges` alone in the editor like this:  

```
create_periodic_report_ranges('2023-01-01', '2023-12-30', 3)
```

I receive this output:

```
[
  {"lower": "2023-01-01T00:00:00", "upper": "2023-03-31T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-04-01T00:00:00", "upper": "2023-06-30T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-07-01T00:00:00", "upper": "2023-09-30T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-10-01T00:00:00", "upper": "2023-12-31T00:00:00", "inc_lower": true, "inc_upper": false},
  ...
]
```

which seems accurate and what we'd want.

Now if I actually run the trigger and then look at the data on the `start` and `end` fields on the `Financial Report` in the data explorer I see this:

```
[
  {"lower": "2023-01-01T00:00:00", "upper": "2023-03-30T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-04-01T00:00:00", "upper": "2023-06-29T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-07-01T00:00:00", "upper": "2023-09-29T00:00:00", "inc_lower": true, "inc_upper": false},
  {"lower": "2023-10-01T00:00:00", "upper": "2023-12-30T00:00:00", "inc_lower": true, "inc_upper": false},
  ...
]
```

The  end dates seem to be one less than what I would expect to see.  I understand that they are calculated values applying the custom `date_range_get_upper` function and thus subtracting a day, but is that what we want?

I do see that the `period` on the `Financial Report` is correct, however.

...

I guess the only thing I can think of is that I'm incorrect in my assessment that we really want a financial reporting period to look like this:  `2023-01-01 -- 2023-03-31`.  Perhaps conceptually we want to think of the _period_ as `2023-01-01 -- 2023-04-01`?  If that's the case then I'm just creating the periods in my function with an end date 1 less than we desire?

dbschema/project.esdl
```diff
@@ -153,7 +152,107 @@ module default {
...
+        ).period,
+        interval := (
+          select (if __new__.financialReportPeriod = default::ReportPeriod.Monthly then '1' else '3')
+          ),
+        requestedReportPeriods := Project::create_periodic_report_ranges(
+          __new__.mouStart,
+          __new__.mouEnd,
+          interval
+        )
+      if __old__.financialReportPeriod != __new__.financialReportPeriod (
```
bryanjnelson, line 177: @CarsonF  - The code within these conditional blocks works; I've tested it just like I have the other trigger above.  However, I'm having some issue with the syntax on this `if` statement here.  I'm getting a compilation error on this line that says:

> error: Missing '['

Just before the parenthesis.  GPT and the documentation have been most unhelpful in explaining why.

dbschema/project.esdl
```diff
@@ -259,4 +358,44 @@ module Project {
           ))
       select reportPeriodRanges
   )
+
+  # TODO - Toying with the idea of abstracting some of this logic in some capacity...
```
bryanjnelson, line 362: @CarsonF - After the trigger code settles down I think there's some opportunity for some common methods like I started here.

CarsonF: I don't think functions can mutate data right now. I don't really understand the limitation.

dbschema/project.esdl
CarsonF: Missing `then` between condition and truth expression

dbschema/project.esdl
```diff
@@ -119,6 +119,140 @@ module default {
...
+          projectContext := __new__.projectContext,
+          container := __new__,
+          period := reportRange
+        })
+      )
+    );
+
+    trigger addRemovePeriodicReports after update for each 
+      when (
+        __old__.mouStart != __new__.mouStart
```
CarsonF, line 159: I believe you want `?!=` on these 3 compassions. Since they are all optional fields. We want null on either side to count as a valid value to compare against, instead of skipping the comparison all together when null is encountered. It's a semi well known footgun that's complained about. Docs have notes on these ?-prefixed operators.

dbschema/project.esdl
bryanjnelson: Well, no.  That's how I got here.  With that in place we get this:

> error: Unexpected keyword 'THEN'

Apparently within triggers only you don't use the keyword `then`.  But...that also isn't working as you saw above.

bryanjnelson: Named-Only_Arguments_-_User-Defined_Functions___Tutorial___EdgeDB_Docs

bryanjnelson: I can't get either way to work, tbh.

CarsonF: The AI is lying to you. It does that a lot with EdgeDB. I never trust its answers either way in this context unfortunately.

CarsonF: Oh you have a `with` expression preceding this. You need to wrap the `if` in a `select`

bryanjnelson: Ah!  The `select` wrap was it!  I thought I tried that, but apparently not.  🤷🏼 

And yes, I have had some less than stellar answers with the AI as well.  But since I couldn't get it working either way I didn't know who was right.

dbschema/project.esdl
```diff
@@ -174,13 +174,13 @@ module default {
           __new__.mouEnd,
           interval
         )
-      if __old__.financialReportPeriod != __new__.financialReportPeriod (
+      select (if __old__.financialReportPeriod ?!= __new__.financialReportPeriod then (
```
bryanjnelson, line 177: @CarsonF  - I'm getting this error here.

Window_and_project_esdl_—_cord-api-v3

I'm trying to avoid `ProgressReport`s here since that's a different flow.  Any ideas on the best way to handle this?

dbschema/project.esdl
```diff
@@ -119,6 +119,142 @@ module default {
...
+          projectContext := __new__.projectContext,
+          container := __new__,
+          period := reportRange
+        })
+      )
+    );
+
+    trigger addRemovePeriodicReports after update for each 
+      when (
+        __old__.mouStart ?!= __new__.mouStart
```
bryanjnelson, line 159: @CarsonF - I'm wondering if this is the place where we can also handle the case in which all 3 of these values exist for the first time?  We could check if they've each changed, but also that none of them are null?

On a related note, can any of these values go back to null once set?

dbschema/project.gel
```diff
@@ -121,6 +121,149 @@ module default {
...
+            select existingReports
+            filter not exists .reportFile
+          ),
+          deletedReportPeriods := (
+            for reportPeriod in reportPeriodsWithoutFiles 
+            union (
+              delete reportPeriod
+            )
+          ),
+          financialReports := (for reportPeriod in requestedReportPeriods
```
bryanjnelson, line 191: @CarsonF - I don't like that we're looping through the financials and the narratives separately here.  If there's a way to combine them both into one loop please let me know.  I wanted to do that (if it is possible) before I abstract this functionality out into it's own function.

CarsonF: I do not think it's possible to unify the insert clauses, or dynamically reference different types for inserts.

I _think_ the loops could be combined though if that's what you mean.
```edgeql
new := (
  for reportPeriod in requestedReportPeriods union ({
    insert Fin {},
    insert Narr {},
  })
)
```

bryanjnelson: Yeah, that's the structure I originally had, but was running into this kind of error

> operator 'std::IF' cannot be applied to operands of type 'tuple<default::FinancialReport, default::NarrativeReport>', 'std::bool', 'default::PeriodicReport'

on account of Gel not being able to reconcile that the else statement could only return in reality FR & NR.

The solution was to split them out and manually merge them like this

> select financialReports union narrativeReports

CarsonF: Gotcha. Then that's probably necessary.

dbschema/project.gel
```diff
@@ -208,4 +303,65 @@ module Project {
...
+      with
+        requestedReportPeriodsForInsertion := (
+          select array_unpack(requestedReportPeriods)
+          filter array_unpack(requestedReportPeriods) not in (
+            for report in array_unpack(existingReports)
+            union (
+              select report.period
+            )
+          )
+        ),
```
bryanjnelson, line 338: @CarsonF - I think this whole thing is pretty close, but on this statement here I'm receiving this warning:

`possibly more than one element returned by an expression in a FILTER clause`

on line 335.  Any ideas on how to remedy that?

bryanjnelson: I tried using an `assert_single`, but I'm not even sure if that's safe.

bryanjnelson: This seems to take care of the the warning:

```edgeql
        distinctRequestedReportPeriods := distinct array_unpack(requestedReportPeriods),
        requestedReportPeriodsForInsertion := (
          select distinctRequestedReportPeriods
          filter distinctRequestedReportPeriods not in (
            for report in array_unpack(existingReports)
            select report.period
          )
        ),
```

bryanjnelson:
@rdonigian - Although we'll close this out (obviously) in light of Gel fading into the sunset, there is a lot of good logic that we're going to want to remember in this PR, as it took the initiative of pulling out periodic report creation from app code into stored procedures.  We can use this logic for our PostGres migration.

PR #3219, opened 2024-05-29
[EdgeDB] Add `Engagement` queries | refactor service/repo layers appropriately

bryanjnelson:
## Description

- Refactor `Engagement` functionality to prepare for EdgeDB migration
- Create EdgeDB `Engagement` repository 

- [X] Change the task url above to the actual Monday task
- `[N/A]` Add/update tests if needed
- [X] Add reviewers to this PR

PR #3284, opened 2024-09-05
[EdgeDB] Polymorphic queries & Engagement Repo

CarsonF:
- Allow our `__typename` fields to be EdgeDB FQN names
- Switch to temp unreleased `@edgedb/generate` branch
  edgedb/edgedb-js#1090
- Use `Engagement.resolve` function instead of `.__typename === '...'` to decouple the strings
- Created EdgeDB version of the Engagements repo with polymorphic hydration.
  What's here is working which is the main thing, but it could be better.
  The QB types could still be better to allow merging shapes & automatically handling EdgeQL type intersections & TS discrimated unions. One day.
- Pulled the rest of Bryan's queries & work from #3219

[Monday Task](https://seed-company-squad.monday.com/boards/5989610236/pulses/5792966980)

### DIFF
--- a/dbschema/engagement.esdl
+++ b/dbschema/engagement.esdl
@@ -51,7 +51,20 @@ module default {
     property firstScripture := (
       exists .language.firstScriptureEngagement
     );
-    
+
+    trigger denyDuplicateFirstScriptureBasedOnExternal after insert, update for each do ( 
+      assert(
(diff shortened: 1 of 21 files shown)


bryanjnelson (approved):
Yuge step forward, @CarsonF!  🚀

PR #3288, opened 2024-09-20
Expose progressReports GQL query

CarsonF:
https://seed-company-squad.monday.com/boards/5989610236/pulses/7125526862

### DIFF
--- a/src/components/progress-report/dto/progress-report-list.dto.ts
+++ b/src/components/progress-report/dto/progress-report-list.dto.ts
@@ -1,4 +1,4 @@
-import { InputType, ObjectType } from '@nestjs/graphql';
+import { Field, InputType, ObjectType } from '@nestjs/graphql';
 import {
   FilterField,
   OmitType,
@@ -8,6 +8,7 @@ import {
 } from '~/common';
 import { EngagementFilters } from '../../engagement/dto';
 import { PeriodicReportListInput } from '../../periodic-report/dto';
+import { ProgressReportStatus } from './progress-report-status.enum';
 import { ProgressReport } from './progress-report.entity';
 
 @InputType()
@@ -15,6 +16,11 @@ export abstract class ProgressReportFilters extends PickType(
   PeriodicReportListInput,
   ['start', 'end', 'parent'],
 ) {
+  @Field(() => [ProgressReportStatus], {
+    nullable: true,
+  })
+  readonly status?: readonly ProgressReportStatus[];
+
   @FilterField(() => EngagementFilters)
   readonly engagement?: EngagementFilters & {};
 }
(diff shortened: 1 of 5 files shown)


andrewmurraydavid (approved):
LGTM! Thanks, @CarsonF

src/components/progress-report/progress-report.repository.ts
```diff
@@ -1,6 +1,95 @@
...
+    return result!; // result from paginate() will always have 1 row.
+  }
+
+  protected hydrate(session: Session) {
+    return (query: Query) =>
+      query
+        .match([
+          node('node'),
+          relation('in', '', 'report', ACTIVE),
+          node('parent', 'Engagement'),
```
rdonigian, line 62: This should target 'LanguageEngagement' specifically due to Internship's generating progress reports unknowingly.

CarsonF: Yep will do that for now and address that bug separately 👍

src/components/progress-report/dto/progress-report-list.dto.ts
```diff
@@ -1,7 +1,41 @@
...
+  PeriodicReportListInput,
+  ['start', 'end', 'parent'],
+) {
+  @Field(() => [ProgressReportStatus], {
+    nullable: true,
+  })
+  readonly status?: readonly ProgressReportStatus[];
+
+  @FilterField(() => EngagementFilters)
+  readonly engagement?: EngagementFilters & {};
```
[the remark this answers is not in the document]
CarsonF: It's a weird TS compiler thing. 
When TS is emitting metadata it looks at these property values and adds runtime references in decoration metadata.
The same thing happens in constructor parameters
```ts
@Injectable()
class Foo {
  constructor(private bar: BarService) {}
}
```
This compile time emission is how NestJS is able to know that `BarService` should be injected for `bar`.

In some cases though we have circular references. Node in CJS mode, would just silently have these references as undefined. But now in ESM mode, Node explicitly throws saying "X cannot be referenced yet because it has yet to be defined" (paraphrasing).
This is better DX in most ways, because it calls it out directly instead of silently failing and then later in execution we get a "cannot access foo on undefined".

So that error can happen in models as well with circular references. In these cases, we don't need the runtime value, we are just defining the TS type. The runtime value is encapsulated in the `() => EngagementFilters`. It is wrapped in a function exactly for the reason above - Node does not try to reference the class until the function is executed, which happens later in NestJS init (after Node loads all the files/scripts).

TS will only emit the metadata decoration if it is a direct reference to a class.
So the workaround to have TS not emit the value, is to make it a "complex" type.
`& {}` means "and not null/undefined".
So type-wise `EngagementFilters & {}` is the same thing as `EngagementFilters`, but it avoids the direct ref to the class.

Anywhere you see a `forwardRef(() => X)`, this is to explicitly tell NestJS what to inject in a "lazy" way, you'll also see the type being `x: X & {}` to avoid referencing the class directly.
Example:
https://github.com/SeedCompany/cord-api-v3/blob/be2f3dd9cd7f8b6eb71a2bc71ba0f9824a59f3a4/src/components/file/file-url.controller.ts#L24-L25

Hopefully this is clear 😵

PR #3290, opened 2024-09-20
Migrate external cache to our library

CarsonF:
And bump our lru-cache to v7 -> v11, which caused a little type churn.


bryanjnelson (approved):
Thanks @CarsonF!

PR #3291, opened 2024-09-23
BUG: Don't create ProgressReports for InternshipEngagements

CarsonF:
Remarkably we never checked the project/engagement type before creating/syncing progress reports.
These should only be for TranslationProjects/LanguageEngagements.

Add migration to drop these rogue entries that are never accessed.


willdch (approved):
Looks A ok to me.

bryanjnelson (approved):
Incredulous that this was even a thing this far along in the game!

LGTM!

PR #3292, opened 2024-09-24
Express -> Fastify / Compression

CarsonF:
This is an easy migration to a newer http framework.

# Express 👎 

Express is so so old. v4.0.0 was released 10 years ago. Although they did finally cut v5 a couple weeks ago, our ecosystem is yet to adopt.

I don't like its ecosystem; all of its supporting packages/plugins are obtusely named with no relation to express: `connect`, `serve-static`, `accepts`, `body-parser`, `cookie-parser`, `mime`, `send`, `content-disposition`. No idea without looking that each of these are for the express/connect framework.
No idea why each header needs its handling package either 🙄 
None have first party types either so 3rd party types `@types/` are needed, doubling the packages on our side.

One of the main pain points experienced here, is express' types. They have a global-ish `e` namespace. At least for me, Webstorm always gets confused on whether I want the `e` global type or to import express or to import our EdgeDB query builder. We never want to reference express, but we had no way to block that suggestion.

# Fastify 👍 

Fastify has better types and doesn't suffer from the ambiguity headache with `e`.
All of its packages are under its org, `@fastify/*`, so they are easy to identify.
They also have first party support for TS, so each package doesn't need the additional `@types/x` package.

Admittedly these are the main reasons for this change, which could be considered relatively minor. Which is probably why this hasn't happened for years.

## Speed
Fastify is reportedly faster as well, but from my benchmarks, there is no difference.
Meaning that the http framework is not the bottleneck, but rather our services/repositories.
This is to be expected since the http layer is so thin for us.

<details>
<summary>Benchmark Results</summary>
This is running the UI's `PartnerDetailEngagements` operation 1k times with `autocannon`.

### Express

```
┌─────────┬────────┬────────┬────────┬────────┬───────────┬──────────┬────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 228 ms │ 274 ms │ 379 ms │ 398 ms │ 279.31 ms │ 35.09 ms │ 537 ms │
└─────────┴────────┴────────┴────────┴────────┴───────────┴──────────┴────────┘
┌───────────┬───────┬───────┬────────┬────────┬────────┬────────┬───────┐
│ Stat      │ 1%    │ 2.5%  │ 50%    │ 97.5%  │ Avg    │ Stdev  │ Min   │
├───────────┼───────┼───────┼────────┼────────┼────────┼────────┼───────┤
│ Req/Sec   │ 3     │ 3     │ 36     │ 40     │ 34.49  │ 7.26   │ 3     │
├───────────┼───────┼───────┼────────┼────────┼────────┼────────┼───────┤
│ Bytes/Sec │ 42 kB │ 42 kB │ 504 kB │ 560 kB │ 483 kB │ 102 kB │ 42 kB │
└───────────┴───────┴───────┴────────┴────────┴────────┴────────┴───────┘

Req/Bytes counts sampled once per second.
# of samples: 29

1k requests in 29.03s, 14 MB read
```

### Fastify

```
┌─────────┬────────┬────────┬────────┬────────┬───────────┬──────────┬────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 216 ms │ 272 ms │ 401 ms │ 452 ms │ 279.77 ms │ 44.54 ms │ 523 ms │
└─────────┴────────┴────────┴────────┴────────┴───────────┴──────────┴────────┘
┌───────────┬─────────┬─────────┬────────┬────────┬────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼────────┼────────┼────────┼─────────┼─────────┤
│ Req/Sec   │ 6       │ 6       │ 36     │ 40     │ 34.49  │ 6.87    │ 6       │
├───────────┼─────────┼─────────┼────────┼────────┼────────┼─────────┼─────────┤
│ Bytes/Sec │ 83.6 kB │ 83.6 kB │ 502 kB │ 558 kB │ 480 kB │ 95.6 kB │ 83.6 kB │
└───────────┴─────────┴─────────┴────────┴────────┴────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.
# of samples: 29

1k requests in 29.04s, 13.9 MB read
```

</details>

## Other Options

I'm aware there other other hyped frameworks right now (h3, hono, etc).
Fastify is an easy migration path with NestJS & Apollo having well supported integrations already built.
Nothing is forever though 😄 

# Compression ⚡ 

I've configured compressing the HTTP responses (brotli, gzip).
This could've been done in Express for FWIW.

This is a bit of big miss 😳 
My tests show that this **reduces** data over the wire by **75%** 🤯

<details>
<summary>Results</summary>

Operation | Resource Size | Compressed Size | Reduced By
-- | -- | -- | --
EngagementList | 13.5 kB | 1.4 kB | 90%
ProjectList | 7.4 kB | 1.1 kB | 85%
ProjectOverview | 6.1 kB | 1.7 kB | 72%
ProjectEngagementListOverview | 2.5 kB | 887 B | 65%
ProjectPostList | 1.3 kB | 818 B | 37%
ProjectDirectory | 5 kB | 1.4 kB | 72%
Engagement | 3.7 kB | 1.4 kB | 62%
ProductDetail | 3.7 kB | 1.4 kB | 62%
Users | 1.1 kB | 693 B | 37%
Languages | 1.5 kB | 715 B | 52%
Partners | 1.6 kB | 765 B | 52%
Partner | 1.7 kB | 879 B | 48%
PartnerProjects | 8.4 kB | 1.3 kB | 85%
PartnerDetailEngagements | 13.6 kB | 1.4 kB | 90%
**Totals** | **49.9 kB** | **13.2 kB** | 74%

</details>

The larger the data returned the greater affect this had. These are results from just local data which is probably smaller in response size than production.

So this is a huge huge win


package.json
```diff
@@ -152,9 +153,13 @@
     "neo4j-driver-bolt-connection@npm:5.20.0": "patch:neo4j-driver-bolt-connection@npm%3A5.20.0#~/.yarn/patches/neo4j-driver-bolt-connection-npm-5.20.0-1f7809f435.patch",
     "neo4j-driver-core@npm:5.20.0": "patch:neo4j-driver-core@npm%3A5.20.0#~/.yarn/patches/neo4j-driver-core-npm-5.20.0-99216f6938.patch",
     "@apollo/server-plugin-landing-page-graphql-playground": "npm:empty-npm-package@*",
+    "@apollo/server/express": "npm:empty-npm-package@*",
```
andrewmurraydavid, line 156: Trying to understand why

andrewmurraydavid (commented):
I'm surprised how plug & play this change has been.

CarsonF:
> I'm surprised how plug & play this change has been.

Well this is following #3296 #3298

But yeah we are fairly decoupled here which is great.

package.json
CarsonF: `@apollo/server` requires/depends on `express`. This is for BC.
I don't want express being pulled in, so this swaps it to an empty package.

andrewmurraydavid:
Ah! Well that makes a lot more sense! But still, great work!

PR #3293, opened 2024-09-24
Fix current report due prioritizing the final report

CarsonF:
This was supposed to be fixed with #3200

Now the _current report due_ is the most recently ending report (of the reports that have already ended) and of the reports with the same end date, the one that started first.

### DIFF
--- a/src/components/periodic-report/periodic-report.repository.ts
+++ b/src/components/periodic-report/periodic-report.repository.ts
@@ -454,7 +454,10 @@ export const matchCurrentDue =
       ])
       .raw(`WHERE end.value < date()`)
       .with('node, start')
-      .orderBy('start.value', 'desc')
+      .orderBy([
+        ['end.value', 'desc'],
+        ['start.value', 'asc'],
+      ])
       .limit(1);
 
 export const periodicReportSorters = defineSorters(IPeriodicReport, {});


src/components/periodic-report/periodic-report.repository.ts
```diff
@@ -454,7 +454,7 @@ export const matchCurrentDue =
       ])
       .raw(`WHERE end.value < date()`)
       .with('node, start')
-      .orderBy('start.value', 'desc')
+      .orderBy('start.value', 'asc')
```
atGit2021, line 457: I believe this is going to always bring back the earliest/oldest quarter in the range, not the most recent. I just tried testing in my local, but I could be testing it incorrectly.  It always brings the Published report back as the current report due. 
Screenshot 2024-09-24 at 11 50 04 AM

CarsonF: Bleh 🤦

CarsonF: Thank for testing. Can you try this? Now we use the most recently ending report and of the reports with the same end date, the one that starts first.

atGit2021 (approved):
LGTM!

CarsonF:
I'm realizing that we never have a cut off for this last report.
If the project ended in 2020, now we will still show the last real quarter report as the currently due report.
This doesn't seem right, but maybe should be addressed separately.
This can focus on the bug of showing the final report instead of the real last quarter report.

PR #3294, opened 2024-09-25
Infer/Decouple Platform types (HTTP/GraphQL)

CarsonF:
I've created `~/core/http` to hold all http platform specifics. The codebase can reference this now and we don't have implementation details littered all over the codebase.

---

Also `.switchToHttp()` & `GqlExecutionContext` have strict return types (via declaration merging).
So the default return values of `any` is avoided and there's no need to give explicit generic at every usage.


bryanjnelson (approved):
Progress!

PR #3300, opened 2024-10-01
Notifications GQL schema

CarsonF:
https://seed-company-squad.monday.com/boards/5989610236/pulses/7542832281

### DIFF
--- a/src/components/notifications/dto/notification.dto.ts
+++ b/src/components/notifications/dto/notification.dto.ts
@@ -1,0 +1,35 @@
+import { Field, InterfaceType, ObjectType } from '@nestjs/graphql';
+import { keys as keysOf } from 'ts-transformer-keys';
+import { Resource, SecuredProps } from '~/common';
+import { LinkTo, RegisterResource } from '~/core/resources';
+
+@RegisterResource()
+@InterfaceType({
+  implements: [Resource],
+})
+export class Notification extends Resource {
+  static readonly Props = keysOf<Notification>();
+  static readonly SecuredProps = keysOf<SecuredProps<Notification>>();
+
+  readonly owner: LinkTo<'User'>;
+
+  @Field(() => Boolean)
+  readonly unread: boolean;
+}
+
+@ObjectType({
+  implements: [Notification],
+})
+export class SimpleTextNotification extends Notification {
+  @Field(() => String)
+  readonly content: string;
+}
+
+declare module '~/core/resources/map' {
+  interface ResourceMap {
+    Notification: typeof Notification;
+  }
+  // interface ResourceDBMap {
+  //   Notification: typeof e.Notification;
+  // }
+}
--- a/src/components/notifications/notification.resolver.ts
+++ b/src/components/notifications/notification.resolver.ts
@@ -23,7 +23,7 @@ export class NotificationResolver {
   }
 
   @Mutation(() => Notification)
-  async markNotificationRead(
+  async readNotification(
     @LoggedInSession() session: Session,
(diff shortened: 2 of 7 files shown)


src/components/notifications/dto/notification.dto.ts
```diff
@@ -0,0 +1,35 @@
...
+  implements: [Resource],
+})
+export class Notification extends Resource {
+  static readonly Props = keysOf<Notification>();
+  static readonly SecuredProps = keysOf<SecuredProps<Notification>>();
+
+  readonly owner: LinkTo<'User'>;
+
+  @Field(() => Boolean)
+  readonly unread: boolean;
```
CarsonF, line 17: Or would the flipped `read` be better?

src/components/notifications/notification.resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Query(() => NotificationList)
+  async notifications(
+    @LoggedInSession() session: Session,
+    @ListArg(NotificationListInput) input: NotificationListInput,
+  ): Promise<NotificationList> {
+    throw new NotImplementedException().with(input, session);
+  }
+
+  @Mutation(() => Notification)
+  async markNotificationRead(
```
CarsonF, line 26: Actually I think 
```suggestion
  async readNotification(
```
would be better.

src/components/notifications/dto/notification.dto.ts
bryanjnelson: That's a tough one; I was thinking about that before I saw your comment.  Typically I would say to flip it; positive approach over double-negative and all that...but this case is so well known and little thought of as "Oh, these are my 'read' notifications."  Of course, that's from a UI/visual perspective, so maybe that shouldn't play in to the model.  Kinda torn here...just typing it out so far.

src/components/notifications/notification.resolver.ts
bryanjnelson: Eh, the problem with `readNotification` is it's ambiguous.   Is "read" a past tense verb, a present tense verb, or a noun here without context?

CarsonF: True. But this is in the context of a mutation action?

bryanjnelson: Point.  I can get on board with it.

src/components/notifications/dto/notification.dto.ts
rdonigian: Put me down as another one who's torn the name, but I would keep it as is.

src/components/notifications/dto/notification.dto.ts
```diff
@@ -0,0 +1,35 @@
...
+export class SimpleTextNotification extends Notification {
+  @Field(() => String)
+  readonly content: string;
+}
+
+declare module '~/core/resources/map' {
+  interface ResourceMap {
+    Notification: typeof Notification;
+  }
+  // interface ResourceDBMap {
```
rdonigian, line 32: Is there a need for the comment here?

CarsonF: Just stubbed. It will be uncommented when implemented & we have edgedb schema types.

rdonigian (approved):

src/components/notifications/dto/notification.dto.ts
bryanjnelson: I'm fine with keeping it as `unread` as well.

bryanjnelson (approved):

PR #3303, opened 2024-10-07
Apollo -> Yoga Server

CarsonF:
The goal here is to switch to the underdog (Yoga/Hive) that's been grinding while the champ (Apollo/Studio/GraphOS) has been sleeping. But still avoid being locked into an inferior solution.
There are big implications here with how we continue forward with a Supergraph SaaS, so I've outlined some of that below.

Notable features/aspects and their change:
- Explorer UI: Unchanged.
  - We still render Apollo Sandbox / Studio Explorer ourselves now. We could remove this and switch to the default GraphiQL if we decide we like it just as much.
- Schema Reporting: Effectively unchanged.
  - Before however, we reported to Apollo's Studio/GraphOS on every app boot.
  - Now we publish as a CodePipeline deploy step.
  - We also publish to GraphQL Hive as well (if configured)
- Usage Reporting: Changed.
  - We no longer publish usage data to Apollo. Seen under "Insights" in Studio.
  - We do publish to Hive, when configured.
  - With Apollo's free plan we only had 1 day of retention anyways.
  - Hive's free plan is 1 week. And their _very_ cheap paid plan is 3 months.
- Subscriptions: Changed.
  - Apollo Server restricted the flexibility of how these were managed. They did provide WS support out of the box.
  - Yoga offers WS & SSE protocols out of the box.
  - Yoga also has flexibility within the lifecycle. For example, we can reset data loaders between subscription events. Which was seemingly impossible with Apollo.
- Supergraph: Unchanged.
  - Yoga supports Apollo-style federated subgraphs.
  - These could be plugged into Hive or GraphOS.
- Supergraph subscriptions: Choice limited.
  - Hive supports this with an SSE protocol to the subgraph.
  - Apollo uses a "HTTP Callback" protocol which Yoga does not _yet_ support.

Ultimately, Hive seems like the better option to buy into. They are dirt cheap for effectively the same product and all of their stuff is open source and built to be compatible with all.
Still, even now we could almost use Apollo's GraphOS/Router with this new setup. Minus usage reporting and subscriptions. Plus a big bill.

<details>
<summary>WIP checklist</summary>

- [x] Initial Driver change
- [x] File Uploads
- [x] Yoga/NestJS `@Plugin()` discovery/registration
- [x] Connect operation context to our `GqlContextHost` (ALS)
- [x] Convert logging/tracing/etc plugins from Apollo to Yoga
- [x] Error formatting/handling
- [x] Continue with Apollo Sandbox / Studio redirect or try GraphiQL or both?
      Lots of code here. Not sure how much abstraction is actually needed
      https://github.com/apollographql/apollo-server/tree/main/packages/server/src/plugin/landingPage
- [x] Apollo schema reporting & Hive schema reporting
      Could probably port their smallish plugin 
      https://github.com/apollographql/apollo-server/blob/main/packages/server/src/plugin/schemaReporting
  - [x] #3317
  - [x] https://github.com/SeedCompany/infra/commit/a6565744f069a562d7c43b5c5db13a4b32fefada
- [x] Usage reporting
  - ❌ Apollo. Too hard and not that important.
      Lots of code here. Def not ported. Not even sure if integratabtle with Yoga.
      https://github.com/apollographql/apollo-server/blob/main/packages/server/src/plugin/usageReporting/plugin.ts
  - ✅ Hive. Plugin added to do this. We can check there now if needed.
- [ ] Federated Subscriptions / Subscription Callback Protocol?
      Hive (Apollo's GraphOS/Router competitor) supports subscriptions over WS, SSE, & HTTP Callbacks.
      So this may block us out of Apollo. But the functionality overall can still be achieved with Hive. At least that appears to be the current state of things.
      https://the-guild.dev/graphql/hive/docs/use-cases/apollo-studio
- [x] Drop Apollo libraries

</details>

https://seed-company-squad.monday.com/boards/3451697530/pulses/7493638980


bryanjnelson (approved):
Exciting times!  Thanks @CarsonF!

PR #3305, opened 2024-10-08
Expose canCreate flags for Comments

CarsonF:
Currently the comments canCreate just points to threads canCreate. I think this is fine, until it's not.
It's probable that we'd want to simplify and remove threads from authorization, since they just group comments.
But first we'd need to do the policy refactor for embedded resources
```diff
- r.X.children(c => c.comments...)
+ r.Comment.for('X')...
```

https://seed-company-squad.monday.com/boards/3451697530/pulses/7124883698


rdonigian (approved):
Had a question on the removing all of the awaits, but now understand that this is part of lingering tech debt.  Otherwise, LGTM!

PR #3307, opened 2024-10-08
Shuffle `~/core/graphql` code around

CarsonF:
No logical changes here, just code shuffling.

- Don't export graphql from `core/index.ts` following suit of other core modules.
  IMO that DX context is key/identifiable and shouldn't be flattened into `~/core`
- Rename `graphql.config.ts` so it isn't mistaken for IDE config.
  My IDE (prob vscode too) tries to read/parse it as like a gql client configuration file.
- Move graphql error handling logic to a separate class.
  Just a lot of code that could be segregated off.


bryanjnelson (approved):
🧹

PR #3324, opened 2024-11-01
PnP: Mismatched Reporting Quarter / Missing Summary % -> 0

atGit2021:
Consider all "missing" summary percents to be 0

We don't really know if an empty cell is a valid 0 or coming from other cells not being updated.

As long as the reporting quarter is valid, we'll count these zeros as valid.
That appeared to be the root of the problem that we were trying to solve before,
plus just trying to guard against bad xlsx navigation which has been vetted at this point.

Also see this for "missing" details: https://github.com/SeedCompany/cord-api-v3/pull/3324#issuecomment-2455038346

[Monday](https://seed-company-squad.monday.com/boards/3451697530/pulses/7763761968)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new getter method for calculating the reporting quarter, enhancing data accuracy.
	- Added new functions for improved validation of fiscal year and quarter data.
- **Bug Fixes**
	- Enhanced error handling for reporting quarters, providing clearer error messages when invalid data is encountered.
	- Improved handling of planned and actual values to ensure accurate data reporting.

### DIFF
--- a/src/common/fiscal-year.ts
+++ b/src/common/fiscal-year.ts
@@ -2,6 +2,9 @@ import { range } from 'lodash';
 import { DateTime } from 'luxon';
 import { CalendarDate, DateInterval } from './temporal';
 
+export const fiscalQuarterLabel = (date: CalendarDate) =>
+  `Q${fiscalQuarter(date)} FY${fiscalYear(date)}`;
+
 export const fiscalYear = (dt: DateTime) => dt.year + (dt.month >= 10 ? 1 : 0);
 
 export const fiscalYears = (start?: DateTime, end?: DateTime) =>
@@ -44,3 +47,12 @@ export const fullFiscalQuarter = (
     fiscalQuarterStartDate.endOf('quarter'),
   );
 };
+
+export const isReasonableYear = (year: unknown) =>
+  isInt(year) && year >= 1970 && year <= 3000;
+
+export const isQuarterNumber = (quarter: unknown) =>
+  isInt(quarter) && quarter >= 1 && quarter <= 4;
+
+// Not true for falsy case, thus not built-in, but fine for our cases here.
+const isInt = (x: unknown): x is number => Number.isInteger(x);
--- a/src/components/pnp/progress-sheet.ts
+++ b/src/components/pnp/progress-sheet.ts
@@ -1,4 +1,5 @@
 import { LazyGetter as Once } from 'lazy-get-decorator';
+import { fullFiscalQuarter, isQuarterNumber, isReasonableYear } from '~/common';
 import { Column, Range, Row, Sheet, WorkBook } from '~/common/xlsx.util';
 
 export abstract class ProgressSheet extends Sheet {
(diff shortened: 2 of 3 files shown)


CarsonF (commented):
I thought we couldn't distinguish between `0` and null?

atGit2021:
> I thought we couldn't distinguish between `0` and null?

I can't recall historical bugs or bus reasons, but David is saying that it should be 0 progress and not classified as missing.

CarsonF:
> I can't recall historical bugs or bus reasons, but David is saying that it should be 0 progress and not classified as missing.

Ok yeah. The cells that exist always have a value. Cells that are formulas, like these summary cells, always exist.
So while the book shows that it is "empty", we received `0`.

```
> pnp.progress.cell('AN22').asNumber
0
``` 
Screenshot 2024-11-01 at 1 25 40 PM

I agree that this would be nice. But it's a limitation of the current process.

atGit2021:
> > I can't recall historical bugs or bus reasons, but David is saying that it should be 0 progress and not classified as missing.
> 
> Ok yeah. The cells that exist always have a value. Cells that are formulas, like these summary cells, always exist. So while excel shows that it is "empty", we received `0`.
> 
> ```
> > pnp.progress.cell('AN22').asNumber
> 0
> ```
> 
> Screenshot 2024-11-01 at 1 25 40 PM
> 
> I agree that this would be nice. But it's a limitation of the current process.

I'm not sure if I understand what you are inferring, but if returning zero values impacts other code, then how about leaving the checks for the validation error using  "==" vs "===" so that the zero is not captured as an error, but revert the return statement so that it uses the original check "===" so that undefined, null and zero are checked on line 113?

CarsonF:
> I'm not sure if I understand what you are inferring, but if returning zero values impacts other code, then how about leaving the checks for the validation error using "==" vs "===" so that the zero is not captured as an error, but revert the return statement so that it uses the original check "===" so that undefined, null and zero are checked on line 113?

What I'm saying is `planned` & `actual` will never be `null` or `undefined`.

CarsonF:
> Unless I'm testing incorrectly, when I step through the code and test I see `undefined` come in on empty cells, and `0` come in on the cells that say zero in the planned and actual fields.

Are you looking at the summary cells? From what I was seeing these summary cells have formulas, and thus were never empty.

atGit2021:
> > Unless I'm testing incorrectly, when I step through the code and test I see `undefined` come in on empty cells, and `0` come in on the cells that say zero in the planned and actual fields.
> 
> Are you looking at the summary cells? From what I was seeing these summary cells have formulas, and thus were never empty.

I see undefined for actual and planned, screen shot attached below. 
Screenshot 2024-11-04 at 8 57 29 AM

CarsonF:
Ok I am seeing that some "empty" cells are parsed as string cells, instead of number cells.
So those string cells would correctly return undefined for `asNumber` since we guard the cell type when returning the value.

It looks like the _4 quarter columns_ & the _yearly actual_ don't have any empty numbers - the empty cells are of type string.
The _year planned_ column has number cells up to the last row reported, even if those cells are empty/0.
All the rows of the _3 cumulative columns_ are of type number.

Processing what this means for us...

CarsonF (approved):

PR #3333, opened 2024-11-13
GQL Subscriptions over WebSockets

CarsonF:
This PR is pushed the need for #3330 

Yoga already has built-in support for Subscriptions over SSE (which I forgot until today).
These work fine when deployed to AWS which uses HTTP/2.
Locally, over HTTP/1, however there is the a 6 connection limit.

Apollo Sandbox/Explorer/Studio also does not support SSE, they have a similar "apollo http multipart" protocol.

Websockets are supported, so this is for them, and just to have flexibility in subscription protocol.
UI client could use either going forward.

# How it works

This ties `fastify`, `yoga`, and `graphql-ws` together.
Execution layers in order:
1. fastify route (http path matching)
2. fastify's websocket plugin (http upgrade request & websocket open/close)
    This allows our fastify hooks to be executed.
    And provides a consistent Fastify `Request` type,
    instead of a raw `IncomingMessage`.
 3. `graphql-ws`'s fastify handler (adapts # 2 to graphql-ws)
 4. `graphql-ws` (handles specific gql protocol over websockets)
 5. `graphql-yoga` is unwrapped to `envelop`.
    Yoga just wraps `envelop` and handles more of the http layer.
    We really just reference `envelop` hooks with our "Yoga Plugins".
    So this allows our "yoga" plugins to be executed.

`@graphql-yoga/nestjs` goes a completely different direction with its subscription logic, which is why I abandoned it. It adapts from the `ApolloDriver`, doing a bunch of unnecessary low level work. It couldn't assume fastify, because express is also supported.
This approach here fully embraces fastify & yoga, and both of their lifecycle hooks. 

@coderabbitai ignore

## Release Notes

- **New Features**
	- Added support for WebSocket connections and GraphQL subscriptions, enhancing real-time capabilities.
	- Introduced new methods to handle subscription operations and errors within the GraphQL context.

- **Bug Fixes**
	- Improved error handling for subscription-related issues.

- **Chores**
	- Updated project dependencies to include `@fastify/websocket` and `graphql-ws`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@fastify/compress": "^8.0.1",
     "@fastify/cookie": "^11.0.1",
     "@fastify/cors": "^11.0.1",
+    "@fastify/websocket": "^11.0.2",
     "@ffprobe-installer/ffprobe": "^2.1.2",
     "@golevelup/nestjs-discovery": "^4.0.0",
     "@graphql-hive/yoga": ">=0.41.0",
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "graphql": "^16.9.0",
     "graphql-parse-resolve-info": "^4.14.0",
     "graphql-scalars": "^1.22.4",
+    "graphql-ws": "^6.0.4",
     "graphql-yoga": "^5.13.4",
     "human-format": "^1.2.0",
     "image-size": "^2.0.2",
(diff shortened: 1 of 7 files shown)


rdonigian (approved):

CarsonF:
I rebased this and dropped the patches that handled the mismatch between graphql-ws & fastify v5 / newer websocket plugin. Everything is aligned now.

PR #3336, opened 2024-12-05
Revised PnP error message

atGit2021:
[Monday ](https://seed-company-squad.monday.com/boards/5989610236/pulses/7994409983)

### DIFF
--- a/src/components/pnp/extract-scripture.ts
+++ b/src/components/pnp/extract-scripture.ts
@@ -71,7 +71,7 @@ export const extractScripture = (
       }\` is **${totalVersesFromNoteCol}** verses, but the goal declares **${totalVerses}** verses to translate \`${
         totalVersesCell.ref
       }\` For more information see the PnP Troubleshooting 
-            <a href="https://github.com/SeedCompany/cord-docs/wiki/PnP-Extraction-Validation:-Errors-and-Troubleshooting-Steps#1-mismatch-between-the-planned-scripture-in-my-notes-column-and-the-number-of-verses-to-translate" target="_blank">Guide</a>`,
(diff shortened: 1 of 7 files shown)


CarsonF (commented):
I'd really rather not store this link in the db for every error.
Can we just a generic link to the UI instead?

atGit2021:
> I'd really rather not store this link in the db for every error. Can we just a generic link to the UI instead?

when you say "generic link", do you mean just use this [link](https://github.com/SeedCompany/cord-docs/wiki/PnP-Extraction-Validation:-Errors-and-Troubleshooting-Steps) which just points to the overall page vs using the specific links that point to the relative error within the page?  And if so, do you want me to pull it from the local.eve config file?

CarsonF:
> when you say "generic link", do you mean just use this [link](https://github.com/SeedCompany/cord-docs/wiki/PnP-Extraction-Validation:-Errors-and-Troubleshooting-Steps) which just points to the overall page vs using the specific links that point to the relative error within the page? And if so, do you want me to pull it from the local.env config file?

Yeah and no it could be hard coded. If that's not acceptable I think we need to do something else.

atGit2021:
> > when you say "generic link", do you mean just use this [link](https://github.com/SeedCompany/cord-docs/wiki/PnP-Extraction-Validation:-Errors-and-Troubleshooting-Steps) which just points to the overall page vs using the specific links that point to the relative error within the page? And if so, do you want me to pull it from the local.env config file?
> 
> Yeah and no it could be hard coded. If that's not acceptable I think we need to do something else.

I think it will be ok, but will ping product team to confirm

CarsonF:
Pulled into #3342

PR #3337, opened 2024-12-06
Fix PnP progress step comparison with project date range

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/7995665803)

### DIFF
--- a/src/components/pnp/isProgressCompletedOutsideProject.ts
+++ b/src/components/pnp/isProgressCompletedOutsideProject.ts
@@ -15,6 +15,7 @@ export const isProgressCompletedOutsideProject = (
   if (!completeDate) {
     return false;
   }
+  // Steps completion dates smallest unit is quarters, so expand project range to that.
   const projectTimeframe =
     pnp.planning.projectDateRange.expandToFull('quarter');
   if (projectTimeframe.contains(completeDate)) {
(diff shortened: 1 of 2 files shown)


atGit2021:
@bryanjnelson @CarsonF, I'm not sure if this solution will have ripple effects in other areas of the application, but throwing it out to see what you guys think.

CarsonF:
Yeah I'd rather keep this how it is. This should technically be able to be any date, and we want to match user intent as close as possible. Business logic / date checks based on these values should be adjusted.

I think I'd rather change the logic in 
https://github.com/SeedCompany/cord-api-v3/blob/283cf07a2aad9e25358c3a1fc35107c61fbdc272/src/components/pnp/isProgressCompletedOutsideProject.ts#L8-L8
to
```ts
// Steps completion dates smallest unit is quarters, so expand project range to that.
const projectTimeframe = pnp.planning.projectDateRange.expandToFull('quarter')

...
projectTimeframe.contains(completeDate)
projectTimeframe.isAfter(completeDate)
```

src/components/pnp/isProgressCompletedOutsideProject.ts
```diff
@@ -15,7 +15,9 @@ export const isProgressCompletedOutsideProject = (
   if (!completeDate) {
     return false;
   }
-  if (pnp.planning.projectDateRange.contains(completeDate)) {
+  const projectTimeframe =
+    pnp.planning.projectDateRange.expandToFull('quarter');
```
CarsonF, line 19:
```suggestion
  // Steps completion dates smallest unit is quarters, so expand project range to that.
  const projectTimeframe =
    pnp.planning.projectDateRange.expandToFull('quarter');
```

CarsonF (approved):

PR #3338, opened 2024-12-06
Added Engagement.milestoneReached

atGit2021:
[Monday](https://seed-company-squad.monday.com/boards/3451697530/views/78801638/pulses/7966558836)


atGit2021:
Posting my questions/comments sent to Carson so Rob is in the loop too.

This PR contains the new milestoneReached property for the Engagement table.  The value is being queried/mutated ok, but when it comes back from the resolver it gets "lost", meaning apollo continues to show the value as null   The screen shot below shows the query returning the correct value, FullBible, on the return, but Apollo shows null   Based on testing with Rob/Bryan, we believe its related to the value not being secured correctly, but we haven't had much success figuring it out.  In the screen shot you will notice that lukePartnership is secured and an object with value, canRead, canEdit is returned, but for some reason i'm only getting a string value back for milestoneReached.

rdonigian (commented):
We discussed the enum values on slack, this was the agreed on config.  There was discussion with product that stakeholders may want to add more values in the future (no examples), this seems easily done with this implementation.  

The field name `milestoneReached` could be perceived as a boolean.  How about flipping the term to `reachedMilestone` or `completedMilestone`?

CarsonF:
> The field name `milestoneReached` could be perceived as a boolean. How about flipping the term to `reachedMilestone` or `completedMilestone`?

Same difference IMO

rdonigian (approved):

CarsonF (approved):

PR #3339, opened 2024-12-12
Add `getUserByEmail` query

bryanjnelson:
[Monday Story #0322](https://seed-company-squad.monday.com/boards/3451697530/pulses/8002040129)

### DIFF
--- a/src/components/user/user.edgedb.repository.ts
+++ b/src/components/user/user.edgedb.repository.ts
@@ -46,6 +46,15 @@ export class UserEdgeDBRepository
     return !!result;
   }
 
+  getUserByEmailAddress(email: string) {
+    const query = e.select(e.User, (user) => ({
+      ...this.hydrate(user),
+      filter_single: { email },
+    }));
+
+    return this.db.run(query);
+  }
+
   protected listFilters(user: ScopeOf<typeof e.User>, input: UserListInput) {
     if (!input.filter) return [];
     return [
--- a/src/components/user/user.edgedb.repository.ts
+++ b/src/components/user/user.edgedb.repository.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { ID, NotImplementedException, PublicOf } from '~/common';
-import { e, RepoFor, ScopeOf } from '~/core/edgedb';
+import { disableAccessPolicies, e, RepoFor, ScopeOf } from '~/core/edgedb';
 import {
   AssignOrganizationToUser,
   RemoveOrganizationFromUser,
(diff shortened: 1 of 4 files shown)


src/components/user/user.edgedb.repository.ts
```diff
@@ -46,6 +52,20 @@ export class UserEdgeDBRepository
     return !!result;
   }
 
+  getUserByEmailAddress(
```
bryanjnelson, line 55: @CarsonF - There's an opportunity here to remove `doesEmailAddressExist` and replace with this new query since it could be used as a dual purpose.  I believe if this happened though, I'd have to remove the `e.assert_exists` and handle the `null` back in the service layer with an exception.  Would that work?  Is that even desirable?  Or is there another way?  I suppose we could also leave both methods as is...

CarsonF: I would yes. I would make the query nullable. There's not an expectation that the email will exist IMO. Contrasted to IDs which we expect to exist.

src/components/user/user.repository.ts
```diff
@@ -243,6 +244,30 @@ export class UserRepository extends DtoRepository<typeof User, [Session | ID]>(
...
+      .matchNode('node', 'User')
+      .where(
+        path([
+          node('node'),
+          relation('out', '', 'email', ACTIVE),
+          node({
+            value: email,
+          }),
+        ]),
+      )
```
CarsonF, line 259: I think we also need to `filterToReadable` here. We don't want anyone to be able to probe users by public email addresses

src/components/user/user.edgedb.repository.ts
rdonigian: I would agree on the removal, it doesn't seem as if there's anything `doesEmailAddressExist` is accomplishing that a `null` from your new method couldn't handle.

CarsonF: Yeah I'm saying the not finding a user with this email is expected behavior. The caller can do whatever they want with the null.

bryanjnelson:
@CarsonF - Can you throw a 🦴 on that type error in the EdgeDB repo?

CarsonF:
> @CarsonF - Can you throw a 🦴 on that type error in the EdgeDB repo?

You're returning `{ dto: User }` from the neo4j repo. Unwrap it.

CarsonF:
Ok I think I made a miss here. The checkEmail is anonymous, no access policies. The new method is with access policies, so they cannot be combined.

bryanjnelson:
So it doesn't work to just call it with that anonymous session?

CarsonF:
> So it doesn't work to just call it with that anonymous session?

It could be fine for Neo4j, but not EdgeDB.

PR #3341, opened 2024-12-13
Normalize  PnP Problems

CarsonF:


src/components/product-progress/handlers/extract-pnp-progress.handler.ts
```diff
@@ -125,25 +126,36 @@ export class ExtractPnpProgressHandler {
...
+              step: error.step,
             });
           }
         }
       }),
     );
   }
 }
+
+const StepNotPlanned = PnpProblemType.register({
```
rdonigian, line 142: Why does StepNotPlanned fall under a handler and not in it's own file like the other errors in the pnp folder?

rdonigian (approved):

PR #3350, opened 2025-01-08
Fix Engagement Date Filtering

rdonigian:
The `WITH` clause here acts to carry forward the node along with date override(`startDateOverride`, `endDateOverride`) and project date(`mouStart`, `mouEnd`) into the next part of the query.

### DIFF
--- a/src/components/engagement/engagement.repository.ts
+++ b/src/components/engagement/engagement.repository.ts
@@ -733,7 +733,7 @@ export const engagementFilters = filter.define(() => EngagementFilters, {
     node('', 'Language', { id }),
   ]),
   startDate: filter.dateTime(({ query }) => {
-    query.optionalMatch([
+    query.match([
       [
         node('node'),
         relation('out', '', 'startDateOverride', ACTIVE),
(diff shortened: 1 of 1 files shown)


CarsonF (changes requested):
This doesn't work, the variables carried forward in the `with` are not reflected in the `coalecse()`.

Are you sure swapping the `optional match` to `match` doesn't fix this too?
I checked and we don't actually need the `optional` here, the pattern will always exist.

rdonigian:
> This doesn't work, the variables carried forward in the `with` are not reflected in the `coalecse()`.
> 
> Are you sure swapping the `optional match` to `match` doesn't fix this too? I checked and we don't actually need the `optional` here, the pattern will always exist.

The `with` worked locally, but I checked `match` again and it does work.  Not sure what my problem was with it before.

src/components/engagement/engagement.repository.ts
```diff
@@ -764,6 +765,7 @@ export const engagementFilters = filter.define(() => EngagementFilters, {
         node('mouEnd', 'Property'),
       ],
     ]);
+
```
CarsonF, line 768:
```suggestion
```

src/components/engagement/engagement.repository.ts
```diff
@@ -747,10 +747,11 @@ export const engagementFilters = filter.define(() => EngagementFilters, {
         node('mouStart', 'Property'),
       ],
     ]);
+
```
CarsonF, line 750:
```suggestion
```

PR #3351, opened 2025-01-23
Fix UpdateLocation.mapImage regression

rdonigian:
This provides functionality to include an update for the location map image.


CarsonF (approved):
Thanks @rdonigian!

PR #3353, opened 2025-02-11
[PnP] Fix regression finding latest cumulative summary

CarsonF:
Regressed with #3324

Since that PR changed the `summaryFrom` to always return an object with zeros, this check on line 64 failed.
Updated to check to ignore zeros, or fallback to returning zeros.

Regrettably this would've been caught had this eslint rule been enabled:
https://github.com/SeedCompany/cord-api-v3/blob/79875d5134689bf979bfd42a73d6b7c430a2ff2c/.eslintrc.cjs#L190-L191


bryanjnelson (approved):
:shipit:

PR #3357, opened 2025-03-05
Expose Ethnologue values to Global Search

rdonigian:
- Expose ethnologue value to global search
- Create new pattern within search service for unique searches
- Filter for duplicate id's on returned items within search service


src/components/language/language.edgedb.repository.ts
```diff
@@ -49,4 +49,12 @@ export class LanguageEdgeDBRepository
     const query = e.op('exists', lang.firstScriptureEngagement);
     return await this.db.run(query);
   }
+
+  async languageByEth(ethnologueId: ID) {
+    const ethnologue = e.cast(e.Language.ethnologue, e.uuid(ethnologueId));
+    const language = e.select(e.Language, () => ({
+      filter_single: { id: ethnologue.language.id },
+    }));
+    return await this.db.run(language.id);
```
CarsonF, line 58:
```suggestion
    return await this.db.run(ethnologue.language.id);
```

src/components/search/search.service.ts
```diff
@@ -107,7 +126,13 @@ export class SearchService {
     );
 
     return {
-      items: hydrated.filter(isNotFalsy).slice(0, input.count),
+      items: hydrated
+        .filter(
+          (item, index, self) =>
+            index === self.findIndex((t) => t!.id === item!.id),
+        )
```
CarsonF, line 133: What's this about?

src/components/search/search.service.ts
```diff
@@ -69,23 +75,36 @@ export class SearchService {
...
+          const inputTypeMap = {
+            EthnologueLanguage: inputTypes.includes('Language')
+              ? [
+                  {
+                    id: result.id,
+                    type: 'LanguageByEth' as const,
+                    matchedProps: ['ethnologue'] as const,
+                  },
+                ]
+              : [],
```
CarsonF, line 89: You are missing the case here where EthnologueLanguage is asked for, but it is never returned.
That's what the new line 91 does for Orgs.

Also missing adding the search type implicitly. Like the new line 57 does for Orgs.

src/components/search/dto/search-results.dto.ts
```diff
@@ -73,6 +75,7 @@ const searchableAbstracts = {
   TranslationProject,
   Product,
   PeriodicReport,
+  EthnologueLanguage,
```
CarsonF, line 78: This is not abstract.
```suggestion
```

src/components/search/search.service.ts
rdonigian: This removes duplicates from the results.  When a query hits for both the language name and the ethnologue name, right now you get back 2 results and the ethnologue then gets converted to the language so the final results has 2 of the same languages.

CarsonF: Ok I reverted this. There are no duplicates. What you were seeing is a language result and then an ethnologue language result. And in the UI you were using a language card for both if I remember.
Instead the UI should say it only wants Languages, and then the API will filter out the EthnologueLanguage results.

Going further, I reverted the addition of EthnologueLanguage as a search type. I cannot think of a use case where we'd need this at a top level. We always just want Languages.
So the work here to expose languages via ethnologue matches is sufficient.

CarsonF (approved):
Ok this LGTM now. I wouldn't mind a second set of eyes on what I changed though.

I did refactor the search types to be better. They kinda sucked, and were years old. Now I think they better convey the process of
[SearchType](https://github.com/SeedCompany/cord-api-v3/blob/5827462f84a15b4e268191dfda0d16081dd8d1ff/src/components/search/search.service.ts#L59-L59) -> [ResourceType](https://github.com/SeedCompany/cord-api-v3/blob/5827462f84a15b4e268191dfda0d16081dd8d1ff/src/components/search/search.service.ts#L61-L65) -> [SearchResultType](https://github.com/SeedCompany/cord-api-v3/blob/5827462f84a15b4e268191dfda0d16081dd8d1ff/src/components/search/search.service.ts#L84-L84).

I didn't feel like the object map was much different than the conditional branches, so I stuck with that. This also keeps the engine from executing the entire map only to maybe pick one value from it.

I did also revert exposing EthnologueLanguage as a top level search result type/filter. Which I know is different than Org/Partner, and something new. I just think that both Org/Partner can have value (though that's kinda a different question), where as EthnologueLanguage is truly just a sub-object of Language.

src/components/search/search.service.ts
rdonigian: This is using `develop` branch on FE.  This is what I was encountering.

Screenshot 2025-03-11 at 9 35 12 AM
Screenshot 2025-03-11 at 9 35 23 AM

rdonigian:
I like the refactor, gives the search needed type safety around the inputs and makes this a bit easier to add on to in the future!

src/components/search/search.service.ts
CarsonF: This was with my code changes?

PR #3360, opened 2025-03-13
Refactor & polish search service code

CarsonF:
This mainly hasn't been touched in 3 years and it needed some love.
This has been pulled out of #3357.

This is mainly naming changes & type tweaks.

Functionally though there are a few changes:
- We actually ensure that the resource types returned are within the search types filter
- Because of this we also expand the interfaces listed in the `types` filter to their searchable concretes.
- We ensure that the results are unique. I don't believe this affects anything currently, but with #3357 there could be matches on two different objects that are then hydrated to the same object.

## Future
A big outstanding piece I want to do here is to convert this to use the data loaders.
This way there's only `O(1+R)` requests where `R` is the number of resources.
Currently it's `O(1+M)` where `M` is the number of matches.
So a search result of 3 projects, is 4 DB queries, but it could be reduced to 2 DB queries.


rdonigian:
LGTM

PR #3362, opened 2025-03-14
Enable batching up to 25 GQL operations over http

CarsonF:
The default limit is 10. I'm not really sure if we should use that or 25 is fine or what.

This allows a single HTTP request instead of 25 requests for 25 operations.
So this is an optimization for this protocol layer.
All of these operations are still processed individually/independently.

But it is possible for future work to optimize more. Like the data loaders could be cached by batched request instead of each operation.


bryanjnelson (approved):
Took awhile, but I got through all the changes.  LGTM!

PR #3363, opened 2025-03-14
Fix default last chapter/verse in scripture input

CarsonF:
I'm guessing this regressed with a7187c6.
Fixed by using an actual getter/setter for the `book` in the initial `ScriptureReferenceInput` class. This allows for overriding in `ScriptureReferenceEndInput`. The getter/setter is re-defined to be a normal property on the first set. So this makes it function like it did before at runtime.

This issue was exacerbated by the fact that this failure to produce a valid scripture reference still produced a `Verse` object that happened to be invalid. This invalid `Verse` object caused an infinite loop in `mergeVerseRanges`. This was fixed here:
https://github.com/SeedCompany/libs/commit/b02ae4033c5bc573eb3d392c4e424051b3df45b8 Where runtime checks were added to prevent those invalid objects from being created, thus avoiding the infinite loop.

For some reason Gel doesn't like this getter/setter change. Typecasting to a `Simplify`ed version restores the type compat.

### DIFF
--- a/src/components/scripture/dto/scripture-reference.dto.ts
+++ b/src/components/scripture/dto/scripture-reference.dto.ts
@@ -17,7 +17,17 @@ export abstract class ScriptureReferenceInput {
     description: 'The Bible book',
   })
   @IsValidBook()
-  readonly book: string;
+  get book(): string {
+    return undefined as any;
+  }
+  protected set book(book: string) {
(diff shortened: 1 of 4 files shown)


bryanjnelson (approved):
Excellent find, @CarsonF !

PR #3375, opened 2025-03-26
Limit roles assigned to the project creator's membership to only those applicable

CarsonF:
Roles like "Leadership" or "Administrator" shouldn't be assigned to projects


sethmcknight:
LGTM

PR #3382, opened 2025-04-03
[Neo4j] Switch to the new subquery import syntax

CarsonF:
Now instead of
```
CALL {
  WITH import
  ...
}
```
it is
```
CALL (import) {
  ...
}
```
This is obviously much better.

---

The biggest change here is with `UNIONs`. Since they previously would have to `WITH import` for each `UNION`.
Now all unioned parts(?) share the same single import declaration.
So if you see lines straight up removed in this diff that is why; instead of the existing import only applying to the first part of the subquery it all applies to all unioned parts.


bryanjnelson (approved):
LGTM!

CarsonF:
I reverted this because the new syntax is not able to alias an import like before
```cql
CALL (node) {
  WITH node as partner
  MATCH (partner)-[:organization]->(node:Organization)
  ...
}
```
the `(node:Organization)` is not a new `node` "variables" but rather the same imported `node`.
So this pattern match is looking for `(node)-->(node)` and fails to match what we expect.

This appears to apply only one level deep, so a work around might be another nested subquery. 
```cql
CALL (node) {
  WITH node as partner
  CALL (partner) {
    MATCH (partner)-[:organization]->(node:Organization)
    ...
  }
  ...
}
```

PR #3383, opened 2025-04-04
Allow admins to set departmentId on existing projects

CarsonF:
Following up on #3368 to allow same thing on existing projects


bryanjnelson (approved):
🚀

PR #3392, opened 2025-04-17
Server errors for creation failures

CarsonF:
- `CreationFailed` error class
   This DRYs the strings and allows for identification programmatically & statically
- Ensure that invalid reads immediately following a creation throw a server error: `ReadAfterCreationFailed`
  This is a server problem, not a client one. Having the API should communicate this, allows other processes to know that they could try again later, without needing an input change.


bryanjnelson (approved):
👍🏼

PR #3394, opened 2025-04-17
Update Financial Analyst policies to allow editing of engagement start/end date overrides

bryanjnelson:


CarsonF:
This follows suit with Project dates privileges. This is because Finance changes the project dates with changes to plan, since they need to approve it. This would obviously change in the future if these date revisions could be expressed with dedicated Change to Plan functionality.

CarsonF (commented):
I think we should make the same change here too right?
https://github.com/SeedCompany/cord-api-v3/blob/c2b679a0ecb83a4d01f61341d594b3e1ef4b2b13/src/components/authorization/policies/by-role/financial-analyst.policy.ts#L30-L30

bryanjnelson:
> I think we should make the same change here too right?
> 
> https://github.com/SeedCompany/cord-api-v3/blob/c2b679a0ecb83a4d01f61341d594b3e1ef4b2b13/src/components/authorization/policies/by-role/financial-analyst.policy.ts#L30-L30

That's a good question, although slightly different.  In that case FAs are only allowed to edit those dates if they are members.  I would assume that since FAs can already take that same action on project start/end dates, that it follows they should be able to do the equivalent on the engagement overrides.

So yes, although this use case doesn't come into play in the Monday integration, it exists in Cord already.

CarsonF (approved):

PR #3396, opened 2025-04-22
Validate Project/Engagement date ranges

CarsonF:
- Project date range now has to be valid
- Engagement override date range now has to be valid
- Engagement override dates have to be within the project date range

`Engagement` now also hydrates _labels_ to help identity in various situations.

I still need to verify production data/fix

### DIFF
--- a/src/components/engagement/engagement.gel.repository.ts
+++ b/src/components/engagement/engagement.gel.repository.ts
@@ -63,19 +63,51 @@ const languageHydrate = e.shape(e.LanguageEngagement, (le) => ({
   ...baseHydrate(le),
   __typename: le.__type__.name,
   ...languageExtraHydrate,
+  label: e.select({
+    project: le.project.name,
+    language: le.language.name,
+    intern: e.cast(e.str, e.set()),
+  }),
 }));
 
 const internshipHydrate = e.shape(e.InternshipEngagement, (ie) => ({
   ...baseHydrate(ie),
   __typename: ie.__type__.name,
   ...internshipExtraHydrate,
+  label: e.select({
+    project: ie.project.name,
+    language: e.cast(e.str, e.set()),
+    intern: e.array_join_maybe(
+      e.array([ie.intern.displayFirstName, ie.intern.displayLastName]),
+      ' ',
+    ),
+  }),
 }));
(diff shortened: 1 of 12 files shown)


src/components/engagement/engagement.gel.repository.ts
```diff
@@ -63,19 +63,51 @@ const languageHydrate = e.shape(e.LanguageEngagement, (le) => ({
   ...baseHydrate(le),
   __typename: le.__type__.name,
   ...languageExtraHydrate,
+  label: e.select({
+    project: le.project.name,
+    language: le.language.name,
+    intern: e.cast(e.str, e.set()),
```
bryanjnelson, line 69: Why does this `intern` property have to be up here in the `languageHydrate`?

src/components/project/project.service.ts
```diff
@@ -611,3 +618,23 @@ export class ProjectService {
...
+    if (start && end && start > end) {
+      const field = changes.mouEnd ? 'project.mouEnd' : 'project.mouStart';
+      throw new ProjectDateRangeException({ start, end }, field);
+    }
+  }
+
+  constructor(readonly value: Range<CalendarDate>, readonly field: string) {
+    const message = "Project's MOU start date must be before the MOU end date";
+    super({ message, field });
+  }
```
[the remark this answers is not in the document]
CarsonF: It is exposed to the gql consumer and logged

bryanjnelson (approved):
More than a little work here; thanks @CarsonF!

src/components/engagement/engagement.gel.repository.ts
CarsonF: Because all engagements have this property, so we need to produce the null here for intern engagements.
https://github.com/SeedCompany/cord-api-v3/blob/373939981aad28b1625be54f0266a37a974c0c18/src/components/engagement/dto/engagement.dto.ts#L87

bryanjnelson: You mean produce the null for language engagements?

CarsonF: My b, correct

PR #3398, opened 2025-04-23
Require certain Project fields when not in development

CarsonF:
I don't really think this is a long term solution but it at least is something for now.

### DIFF
--- a/src/common/required-when.ts
+++ b/src/common/required-when.ts
@@ -46,16 +46,14 @@ RequiredWhen.verify = <TResourceStatic extends ResourceShape<any>>(
   obj: UnsecuredDto<TResourceStatic['prototype']>,
 ) => {
   const res = EnhancedResource.of(resource);
-  const props = [...res.props];
-  const missing = props.flatMap((prop: string) => {
+  const missing = [...res.props].flatMap((prop: string) => {
     const condition = RequiredWhenMetadata.get(resource, prop);
-    return condition?.isEnabled(obj) && (condition.isMissing ?? isNil)(obj)
-      ? [
-          {
-            field: condition.field ?? prop,
-            description: condition.description,
-          },
-        ]
+    return condition?.isEnabled(obj) &&
+      (condition.isMissing?.(obj) ?? obj[prop] == null)
+      ? {
+          field: condition.field ?? prop,
+          description: condition.description,
+        }
       : [];
   });
   if (missing.length > 0) {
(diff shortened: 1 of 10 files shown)


src/components/project/dto/project.dto.ts
```diff
@@ -196,10 +197,16 @@ class Project extends Interfaces {
...
    * This doesn't account for changesets or item filters.
    */
-  @RequiredWhenNotInDev({
-    field: 'engagements',
-    isMissing: (project) => project.engagementTotal === 0,
-  })
+  @Disabled(`
+    I'm not convinced this wont have unintended consequences.
+    it is still handled with the workflow condition currently and deletes are
+    restricted, so this is a super edge case effectively.
```
bryanjnelson, line 203: Proof.

    I'm not convinced this won't have unintended consequences.
    It is still handled with the workflow condition currently and deletes are
    restricted, so this is effectively a super edge case.

src/components/engagement/dto/engagement.dto.ts
```diff
@@ -55,6 +56,12 @@ export const resolveEngagementType = (val: Pick<AnyEngagement, '__typename'>) =>
     ? LanguageEngagement
     : InternshipEngagement;
 
+const RequiredWhenNotInDev = RequiredWhen(() => Engagement)({
+  description: 'the engagement is not in development',
+  isEnabled: ({ status }) =>
+    status !== 'InDevelopment' && status !== 'DidNotDevelop',
+});
```
bryanjnelson, line 63: I guess there's no way to DRY this with the one in the Project dto?

bryanjnelson (approved):
💪🏼

src/common/required-when.ts
```diff
@@ -0,0 +1,87 @@
...
+  resource: TResourceStatic,
+  obj: UnsecuredDto<TResourceStatic['prototype']>,
+) => {
+  const res = EnhancedResource.of(resource);
+  const props = [...res.props];
+  const missing = props.flatMap((prop: string) => {
+    const condition = RequiredWhenMetadata.get(resource, prop);
+    return condition?.isEnabled(obj) && (condition.isMissing ?? isNil)(obj)
+      ? [
+          {
```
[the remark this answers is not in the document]
CarsonF: Good bot

src/components/engagement/dto/engagement.dto.ts
CarsonF: The description is different. Beyond that we could but it's not worth it IMO. The fact that the actual check happens to be the same is anecdotal, they are different enums.

PR #3403, opened 2025-04-26
Fix null input handling

CarsonF:
Handle GraphQL nulls from "optional" fields as undefined.

GraphQL's / Nest's decorators have `nullable` option, which means the value can be null or omitted - which naturally translates to null or undefined in JS.

We actually care about this distinction, though, especially on update mutations.
undefined means no change.
null means clear the value.

We carelessly allowed both meaningly that nulls could mistakenly be stored in lots of spots that are unexpected.
Granted, this only would happen in Neo4j, as Gel's schema checks would prevent this.

Going forward we have a new `optional` option as a sibling to `nullable`.
`optional` lets the field be omitted or undefined. GraphQL can still give us nulls, but be will we convert them to undefined.
`nullable` continues to let null & undefined through.

---

I also added `null` types where we actually expect/allow nulls.

I also created `ListField` to help dedupe items, and convert empty lists to undefined (for filters), and to disallow empty arrays where expected.

### DIFF
--- a/src/common/rich-text.scalar.ts
+++ b/src/common/rich-text.scalar.ts
@@ -1,5 +1,5 @@
 import { applyDecorators } from '@nestjs/common';
-import { Field, FieldOptions, ObjectType } from '@nestjs/graphql';
+import { ObjectType } from '@nestjs/graphql';
 import { setToStringTag } from '@seedcompany/common';
 import { markSkipClassTransformation } from '@seedcompany/nest';
 import { IsObject } from 'class-validator';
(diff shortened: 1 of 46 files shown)


src/common/rich-text.scalar.ts
```diff
@@ -76,11 +76,14 @@ export class RichTextDocument {
...
   applyDecorators(
-    Field(() => RichTextScalar, options),
-    IsObject(),
-    Transform(({ value }) => RichTextDocument.fromMaybe(value)),
+    OptionalField(() => RichTextScalar, {
+      optional: false,
+      ...options,
+      transform: RichTextDocument.fromMaybe,
+    }),
+    IsObject(), // TODO validate empty blocks becomes null becomes validation error
```
CarsonF, line 86: Todo

PR #3425, opened 2025-05-09
Fix Gel cardinality warnings

CarsonF:


dbschema/language.gel
```diff
@@ -87,7 +87,7 @@ module default {
       select LanguageEngagement filter __source__ = .language
     );
     projects := (
-      select TranslationProject filter __source__ = .languages
+      select TranslationProject filter __source__ in .languages
```
bryanjnelson, line 90: Did this even work before?

CarsonF: It did, surprisingly.

bryanjnelson (approved):
Thanks man!

PR #3428, opened 2025-05-14
Create AI Guidelines Markdowns

### LINKED ISSUE #3470: CF - create Github Copilot instructions

rdonigian:


CarsonF (changes requested):
This is AI slop :/
Wrong info, duplicated info, irrelevant connections, standards info  explicitly spelled out that is just industry best practice.

All of this info can be obtained by looking at the codebase.
If AI is able to make this doc, then why does AI needs this doc?

PR #3434, opened 2025-05-19
Fix inheritance on ID/ResourceName types

CarsonF:
I think this must've changed when we removed the static `Props` declarations.
Previously comparing static types worked, probably because of the static `Props` list.

Now comparing instance types appears to work, which seems more correct anyway.

However, this uses duck typing / shape matching, not actual inheritance.
So we had a problem with `Pinnable` & `Actor`, both of which only require an `id` prop.
I "fixed", worked around this by declaring the `__typename` string literals for `Actor`/`User`/`SystemAgent`.
This is more feasible than enumerating the concrete types of `Pinnable` & doing the same on them.
It's working now, but not the most stable situation to be in. Will continue pondering other ways to address this.

### DIFF
--- a/src/components/user/system-agent.gel.repository.ts
+++ b/src/components/user/system-agent.gel.repository.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { type Role } from '~/common';
-import { disableAccessPolicies, edgeql, Gel } from '~/core/gel';
+import { disableAccessPolicies, e, Gel } from '~/core/gel';
 import { SystemAgentRepository } from './system-agent.repository';
 
 @Injectable()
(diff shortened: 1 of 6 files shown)


src/components/user/system-agent.gel.repository.ts
```diff
@@ -12,15 +12,15 @@ export class SystemAgentGelRepository extends SystemAgentRepository {
...
-      ) {*}
-    `);
-    return await this.db.run(query, { name, roles });
+    const upserted = e.op(
+      e.select(e.SystemAgent, () => ({ filter_single: { name } })),
+      '??',
+      e.insert(e.SystemAgent, { name, roles }),
+    );
+    const query = e.select(upserted, (agent) => ({
+      __typename: e.str('SystemAgent' as const),
```
CarsonF, line 21: I had to rewrite this query to use the query builder, because EdgeQL doesn't understand string literals, so even though I could write the constant "SystemAgent" it would still be widened to a TS `string`, by the EdgeQL server.

PR #3439, opened 2025-05-20
`AuthLevel` decorator

CarsonF:
Replacing `Anonymous/LoggedIn` decorators with an expanded `AuthLevel`.

This new one allows declaring that a session should not even be tried to be resumed, which is  needed with our HTTP routes.

So the 3 levels:
- `sessionless` - do not attempt a credential gathering from the request
- `anonymous` - gather credential info & validate token is in db & resume session
- `authenticated` - assert that the session is tied to a user aka "logged in"

Now `SessionInterceptor` doesn't have to make assumptions about the codebase.
It doesn't have to know to avoid starting a session for the `session` gql query.
It doesn't have to know to avoid starting a session for http requests unless
it has the session watermark/(session decorator).

This is a smaller PR but distinct with its business logic change, so I wanted to give it its own PR.

### DIFF
--- a/src/components/authentication/anonymous.decorator.ts
+++ b/src/components/authentication/anonymous.decorator.ts
@@ -1,7 +1,0 @@
-import { createMetadataDecorator } from '@seedcompany/nest';
-
-export const LoggedIn = () => Anonymous(false);
-export const Anonymous = createMetadataDecorator({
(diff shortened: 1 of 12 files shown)


bryanjnelson (approved):
👍🏼

PR #3446, opened 2025-05-21
Project.membership.active/roles filters

CarsonF:


src/components/project/dto/list-projects.dto.ts
```diff
@@ -96,6 +97,17 @@ export abstract class ProjectFilters {
   })
   readonly isMember?: boolean;
 
+  @FilterField(() => ProjectMemberFilters, {
+    description:
+      "Only projects with the requesting user's membership that matches these filters",
+  })
+  readonly membership?: ProjectMemberFilters & {};
```
CarsonF, line 104: I picked this key because it matches our Gel schema.
https://github.com/SeedCompany/cord-api-v3/blob/5f322bf882504d67b396ed5431ee14fc188f4078/dbschema/project.gel#L84-L84

But I'm not sure it is the most intuitive for "my membership". Thoughts?
Idk why `myMembership` seems weird to me..

rdonigian: I'm struggling to think of anything better and definitely agree on `myMembership`.  It seems redundant.

bryanjnelson: Maybe I'm not following the context here, but `myMembership` seems okay to me?  Can't membership be in a variety of contexts?  His membership?  Their membership?  Why not My membership?

bryanjnelson (approved):

PR #3461, opened 2025-05-23
Support specific aliases for PnP step labels

rdonigian:
This will remedy the custom step header error for a leadership approved list of alternative headers by replacing custom steps with standard ones.

### DIFF
--- a/src/components/pnp/findStepColumns.ts
+++ b/src/components/pnp/findStepColumns.ts
@@ -1,4 +1,4 @@
-import { sortBy } from '@seedcompany/common';
+import { mapOf, sortBy } from '@seedcompany/common';
 import levenshtein from 'fastest-levenshtein';
 import { startCase } from 'lodash';
 import { type Column } from '~/common/xlsx.util';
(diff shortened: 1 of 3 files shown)


CarsonF (approved):

src/components/pnp/findStepColumns.ts
```diff
@@ -69,7 +69,7 @@ const chooseStep = (
 ): Step | undefined => {
   const alias = ApprovedAliases.get(label.toLowerCase());
   if (alias) {
-    return alias;
+    return available.has(alias) ? alias : undefined;
```
bryanjnelson, line 72: I must not be reading this correctly.  How does this line do what the commit message states?

bryanjnelson (approved):

src/components/pnp/findStepColumns.ts
CarsonF: The loop above that calls this method facilities the problem reporting & remaining steps. Is that what you mean? 
https://github.com/SeedCompany/cord-api-v3/blob/b83b214a2de85f1765c6259d362b609a72af99fc/src/components/pnp/findStepColumns.ts#L55-L61

bryanjnelson: I got it now; for some reason I was kind of inverting it yesterday in my head.

Thanks!

PR #3486, opened 2025-06-06
Assign directors as members when the project's region is assigned

CarsonF:


CarsonF:
Finished implementation here #3504

PR #3515, opened 2025-06-26
Add User Status Filter

rdonigian:
Expose filtering on user status.

### DIFF
--- a/src/components/user/dto/list-users.dto.ts
+++ b/src/components/user/dto/list-users.dto.ts
@@ -9,6 +9,7 @@ import {
   Role,
   SortablePaginationInput,
 } from '~/common';
+import { UserStatus } from './user-status.enum';
 import { User } from './user.dto';
 
 @InputType()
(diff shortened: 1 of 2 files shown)


src/components/user/dto/list-users.dto.ts
```diff
@@ -22,6 +23,9 @@ export abstract class UserFilters {
   @OptionalField()
   readonly title?: string;
 
+  @OptionalField()
```
CarsonF, line 26: Enum types cannot be inferred by TS, so they have to be explicitly declared. Currently it is inferred to be a `String`.
```suggestion
  @OptionalField(() => UserStatus)
```

src/components/user/user.repository.ts
```diff
@@ -407,6 +407,7 @@ export const userFilters = filter.define(() => UserFilters, {
   }),
   title: filter.propPartialVal(),
   roles: filter.stringListProp(),
+  status: filter.propVal(),
```
CarsonF, line 410: Ordering of these filters actually declares the order of the generated cypher, so we want cheaper filters first. i.e don't do the hard work of full-text matching a name for all users, when we are then going to filter that down to the ~20 disabled ones.

I'd say this could go after `pinned`.

CarsonF (changes requested):

CarsonF (approved):

PR #3530, opened 2025-07-18
Expose FieldRegion.projects & FieldZone.projects

### LINKED ISSUE #3500: Expose FieldRegion.projects & ProjectFilter.fieldRegion
Probably Zones too

rdonigian:
BE for https://github.com/SeedCompany/cord-field/issues/1709

### Expose FieldRegion.projects and FieldZone.projects in GraphQL API

Adds a paginated, filterable projects relationship to both FieldRegion and FieldZone so the UI can display project tables on each detail page.

### Changes:

**Schema** — added computed links to both types:

FieldRegion.projects → direct back-reference via .<fieldRegion[is Project]
FieldZone.projects → aggregated through regions via .<fieldZone[is FieldRegion].projects
Project gel repository — added fieldRegion.id and fieldRegion.fieldZone.id filter handling to listFilters (these were already supported in the Neo4j path but missing from the Gel implementation)

**Services** — listProjects() added to both FieldRegionService and FieldZoneService, delegating to ProjectService.list() with the appropriate filter. FieldZone sets canCreate: false since zones don't own projects directly.

**Resolvers** — @ResolveField projects added inline to FieldRegionResolver and FieldZoneResolver, supporting full ProjectListInput pagination and filtering.

**Circular dependency fixes** — introducing ProjectModule into FieldRegionModule and FieldZoneModule created new cycles through LocationModule and OrganizationModule; wrapped those imports with forwardRef.

### DIFF
--- a/dbschema/field-region.gel
+++ b/dbschema/field-region.gel
@@ -6,5 +6,7 @@ module default {
     
     required fieldZone: FieldZone;
     required director: User;
+
+    projects := .<fieldRegion[is Project];
   }
 }
--- a/dbschema/field-zone.gel
+++ b/dbschema/field-zone.gel
@@ -7,5 +7,6 @@ module default {
     required director: User;
     
     fieldRegions := .<fieldZone[is FieldRegion];
+    projects := .<fieldZone[is FieldRegion].projects;
   }
 }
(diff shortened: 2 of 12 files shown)


CarsonF (dismissed):
Looks good, just need to spread in the sub filter objects

brentkulwicki (approved):
Gel stuff I kind of just assume is fine and also doesn't matter longterm anyway. The rest of it looks good.

PR #3542, opened 2025-08-06
Create Tool & EngagementTool

rdonigian:


dbschema/tool.gel
```diff
@@ -0,0 +1,14 @@
+module default {
+  type Tool extending Resource, Mixin::Named, Mixin::Timestamped {
```
CarsonF, line 2: Resource is already timestamped so don't need that. name needs to have its own unique constraint, as all _tool names_ need to be unique.
```suggestion
  type Tool extending Resource, Mixin::Named {
      overloaded name {
        constraint exclusive;
      }
```

dbschema/tool.gel
```diff
@@ -0,0 +1,14 @@
...
+    aiBased: bool {
+      default := false;
+    };
+
+    required appliedBy: default::User {
+      readonly := true;
+      on target delete delete source;
+    };
+
+    applicationDate: datetime;
```
CarsonF, line 12: What are these supposed to be? We have the tool created date/time. Applied to what? this is a "global" object.

dbschema/engagement.gel
```diff
@@ -159,6 +159,17 @@ module default {
       }
     );
   }
+
+
+  type EngagementTool extending Resource, Mixin::Timestamped {
```
CarsonF, line 164: I don't like this name. This is not a certain kind of tool for engagements, it's a usage of a tool in an engagement.

```suggestion
  type ToolUsage extending Resource {
```

```suggestion
  type EngagedTool extending Resource {
```

dbschema/tool.gel
rdonigian: @sethmcknight mentioned us tracking the user who is using the tool or at least the one adding it to the engagement I believe.  Maybe he can confirm.  This can probably be moved to the the other object, open to names.

dbschema/engagement.gel
rdonigian: I went with EngagementTool thinking that there maybe something like ProjectTool in the future so it keeps the naming flexible and open to other interactions.

CarsonF: Actually can you put the usage in a tool module, that follows our convention better.
```esdl
module default {
  type Tool {
    usages := .<tool[is Tool::Usage];
  }
}
module Tool {
  type Usage {}
}
```

CarsonF:
Finishing impl in #3544

PR #3550, opened 2025-08-27
Polls v2

CarsonF:
## Identified voters
Voters must identify themselves via registration. This info is exposed in the result, and prevents behavior where a single voter can vote multiple times to increase the tally of a certain choice. Not that we don't still expect/need good faith actors.

## Strict workflow
Voters must register themselves, only once. Results can only be viewed once a poll is closed. Voting is blocked after poll is closed.
This is to avoid unexpected silent behavior. For example, it prevents a mistake where some voter logic is run multiple times. Or some voter logic accidentally runs after the results are consumed.

## Renames

In general there are just a lot of renames.

And all classes are exported via a `Polls` namespace now, to prevent global pollution/ambiguity in `~/common` and promote discoverability. 

## Move to library

I will move this to our library now, but I wanted a code diff here for reference.

### DIFF
--- a/src/common/poll.ts
+++ b/src/common/poll.ts
@@ -1,63 +1,166 @@
-import { setOf } from '@seedcompany/common';
+import { cmpBy, groupBy, setOf } from '@seedcompany/common';
+import { type Simplify } from 'type-fest';
 
-export class PollResults<T> {
-  constructor(protected readonly data: PollData<T>) {}
+class PollState<Choice, Voter> {
+  closed = false;
+  voters = new Set<Voter>();
+  votes = new Map<Voter, Choice>();
+  vetoers = new Set<Voter>();
+}
 
-  /** Returns true if there were any votes */
-  get anyVotes() {
-    return this.numberOfVotes > 0;
+class VoteTally<Choice, Voter> {
+  protected readonly state: PollState<Choice, Voter>;
+  protected readonly tallies: ReadonlyArray<
+    Readonly<{
+      choice: Choice;
+      count: number;
+      voters: ReadonlySet<Voter>;
+    }>
+  >;
+
+  constructor(state: PollState<Choice, Voter>) {
+    this.state = state;
+    this.tallies = groupBy(state.votes, ([, vote]) => vote)
+      .map((entries) => {
+        const voters = setOf(entries.map(([voter]) => voter));
+        return {
+          choice: entries[0][1],
+          voters,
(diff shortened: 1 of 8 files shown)


src/common/poll.ts
```diff
@@ -1,110 +1,255 @@
...
+  /**
+   * Veto the poll as a certain voter.
+   *
+   * This can only be done once, per voter, even if the voter identity is the same.
+   *
+   * @see Ballot.veto
+   */
+  veto(voter: Voter) {
+    this.registerVoter(voter).veto();
   }
```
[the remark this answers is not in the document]
CarsonF: Nope. These shortcut methods can only be called once. If they want to call vote or veto multiple times then they need to use the `registerVoter` method directly.
This prevents accidentally shadowing another voter.

src/core/authentication/session/session.manager.ts
```diff
@@ -103,15 +103,16 @@ export class SessionManager {
...
       await this.sessionHost.withSession(requesterSession, async () => {
         const event = new CanImpersonateHook(
           requesterSession,
           allowImpersonation,
         );
         await this.hooks.run(event);
       });
-      if (!(allowImpersonation.plurality && !allowImpersonation.vetoed)) {
+      const res = allowImpersonation.close();
+      if (!(res.winner?.choice && !res.vetoed)) {
```
bryanjnelson, line 115: Maybe comment on this condition?  It's a little obtuse.

bryanjnelson (approved):
LGTM.

src/core/authentication/session/session.manager.ts
CarsonF: https://github.com/SeedCompany/cord-api-v3/blob/a1ca1865ea46652b59734ef0d945c83be4617866/src/components/authorization/handler/can-impersonate-via-privileges.handler.ts#L15-L15
This second arg is the `choice`.
https://github.com/SeedCompany/cord-api-v3/blob/a1ca1865ea46652b59734ef0d945c83be4617866/src/common/poll.ts#L164-L164

But I get that the word `choice` isn't used until this point.

bryanjnelson: Ok, no comment needed now that it's tweaked.  A far as the naming goes, in this case `choice` seems to represent the result of the choice made.  So what about `result` or `outcome`? `result` is kind of loaded.

PR #3560, opened 2025-09-17
Add migration to create department id blacklist

### LINKED ISSUE #3559: Mark additional Department ID's as unavailable for assignment
There are more Department ID's that have been assigned outside of Cord, and we're running into collisions with Multiplication Projects.

Use the attached file to add these to unavailable blocks of IDs.

[Department IDs.xlsx](https://github.com/user-attachments/files/22369859/Department.IDs.xlsx)

bryanjnelson:
Closes #3559.

### DIFF
--- a/src/components/project/handlers/set-department-id.handler.ts
+++ b/src/components/project/handlers/set-department-id.handler.ts
@@ -97,24 +97,22 @@ export class SetDepartmentId implements IEventHandler<SubscribedEvent> {
       // Get used IDs
       .subQuery((sub) =>
         sub
-          .match([
-            node('', 'Project'),
-            relation('out', '', 'departmentId', ACTIVE),
-            node('deptIdNode', 'Property'),
-          ])
-          .where({ 'deptIdNode.value': not(isNull()) })
-          .return(collect('deptIdNode.value').as('used')),
+          .subQuery((sub2) =>
+            sub2
+              .match([
+                node('', 'Project'),
+                relation('out', '', 'departmentId', ACTIVE),
+                node('deptIdNode', 'Property'),
+              ])
+              .where({ 'deptIdNode.value': not(isNull()) })
+              .return('deptIdNode.value as id')
+              .union()
+              .match(node('external', 'ExternalDepartmentId'))
+              .return('external.departmentId as id'),
+          )
+          .return(collect('id').as('used')),
       )
-      // Get blacklisted IDs
-      .subQuery((sub) =>
-        sub
-          .match(node('blacklist', 'BlacklistDepartmentId'))
-          .return(collect('blacklist.departmentId').as('blacklisted')),
-      )
-      // Distill to available (excluding both used AND blacklisted)
-      .with(
-        '[id in enumerated where not id in used and not id in blacklisted][0] as next',
-      )
+      .with('[id in enumerated where not id in used][0] as next')
       // collapse cardinality to zero if none available
       .raw('unwind next as nextId')
 
--- a/src/components/project/handlers/set-department-id.handler.ts
+++ b/src/components/project/handlers/set-department-id.handler.ts
@@ -105,8 +105,16 @@ export class SetDepartmentId implements IEventHandler<SubscribedEvent> {
           .where({ 'deptIdNode.value': not(isNull()) })
           .return(collect('deptIdNode.value').as('used')),
       )
-      // Distill to available
-      .with('[id in enumerated where not id in used][0] as next')
+      // Get blacklisted IDs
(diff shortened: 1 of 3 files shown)


CarsonF:
I probably wouldn't bother with a neo4j migration. It's not like each env has its own data that needs its own manipulation. and this is really only needed in prod.

Also we don't want the full list from this file, since it has IDs of our own projects. We'll need to exclude those.

bryanjnelson:
@CarsonF - I ran this locally and received the expected results:

`[database:results:stats] Query +4s { nodesCreated: 564, propertiesSet: 2820, labelsAdded: 564 }`

Could you give this a look before we actually modify Prod?  Also, I'm unclear on what to put as the value of `createdBy`?

(Also, we're not going to check this code in as a migration; this is just allowing easy testing and code review.)

src/components/project/handlers/set-department-id.handler.ts
```diff
@@ -105,8 +105,16 @@ export class SetDepartmentId implements IEventHandler<SubscribedEvent> {
...
-      .with('[id in enumerated where not id in used][0] as next')
+      // Get blacklisted IDs
+      .subQuery((sub) =>
+        sub
+          .match(node('blacklist', 'BlacklistDepartmentId'))
+          .return(collect('blacklist.departmentId').as('blacklisted')),
+      )
+      // Distill to available (excluding both used AND blacklisted)
+      .with(
+        '[id in enumerated where not id in used and not id in blacklisted][0] as next',
```
CarsonF, line 116: I'd prefer to merge our own used and the externally used IDs into one list first, before this logic happens.

```ts
// Get used IDs
.subQuery((sub) => sub
  .subQuery((sub2) => sub2
    .match([
      node('', 'Project'),
      relation('out', '', 'departmentId', ACTIVE),
      node('deptIdNode', 'Property'),
    ])
    .where({ 'deptIdNode.value': not(isNull()) })
    .return('deptIdNode.value as id')
    .union()
    .match(node('external', 'ExternalDepartmentId'))
    .return('external.departmentId as id'))
  .return(collect('id').as('used'))
)

File: src/components/project/migrations/create-used-dept-id-list.migration.ts [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+    );
+
+    await this.db
+      .query()
+      .unwind(prunedIntaactList, 'dept')
+      .create(node('blacklist', 'BlacklistDepartmentId'))
+      .setValues({
+        'blacklist.id': variable(apoc.create.uuid()),
+        'blacklist.departmentId': variable('dept.id'),
+        'blacklist.departmentName': variable('dept.name'),
```
CarsonF, line 48: I would rather just have `name` here. Idk if it's ever referred to as a "Department Name".

src/components/project/migrations/create-used-dept-id-list.migration.ts
```diff
@@ -0,0 +1,55 @@
...
+        'blacklist.createdAt': Date.now(),
```
CarsonF, line 49:
```ts
'blacklist.createdAt': DateTime.now(),
```
Or if running manually just ensure this gets to be a neo4j datetime, not an integer.

src/components/project/migrations/create-used-dept-id-list.migration.ts
```diff
@@ -0,0 +1,55 @@
...
+        'blacklist.createdBy': 'bulk_import',
```
CarsonF, line 50: I don't think this exists or makes sense for this feature. Nor do we have `createdBy` anywhere else in our Neo4j schema.

bryanjnelson:
Thanks @CarsonF.  @rdonigian, I'll let you make the changes in the handler and I'll make the changes in the migration script.

src/components/project/migrations/create-used-dept-id-list.migration.ts
bryanjnelson: I actually meant to use neo4j's "datetime()" here.

CarsonF (commented):
The handler query change looks good to me.

The init import script can just be this query:
```cql
load csv from 'https://url.csv' as row
with row[0] as deptId, row[1] as name
where not (:DepartmentId { value: deptId })
create (:ExternalDepartmentId {
  id: randomUUID(),
  departmentId: deptId,
  name: name,
  createdAt: datetime(),
})
```

bryanjnelson:
Closed in favor of [3561](https://github.com/SeedCompany/cord-api-v3/pull/3561).

PR #3561, opened 2025-09-22
Exclude external department IDs from being available

CarsonF:
Cherry-picks work from #3560 and adds schema parity for Gel.


bryanjnelson (approved):
👍🏼

PR #3607, opened 2025-12-17
Create BudgetRecord.preApprovedAmount and .initialAmount

rdonigian:


src/components/budget/dto/budget-record.dto.ts
```diff
@@ -37,9 +37,22 @@ export class BudgetRecord extends Interfaces {
...
 
+  @Field({
+    description: 'The amount that was pre-approved for this budget record.',
+  })
+  readonly preApprovedAmount: SecuredFloatNullable;
+
+  @Field({
+    description: 'The current amount including any change to plans',
+  })
+  readonly adjustedAmount: SecuredFloatNullable;
```
CarsonF: We have a similar concept with project/engagements end date. But the "canonical" property is the current value, and we store/sync the "initial" value.

https://github.com/SeedCompany/cord-api-v3/blob/6a5215abd72fcc950cc07cc7916ea72e2776ee45/src/components/project/dto/project.dto.ts#L156-L157
https://github.com/SeedCompany/cord-api-v3/blob/6a5215abd72fcc950cc07cc7916ea72e2776ee45/src/components/engagement/dto/engagement.dto.ts#L135-L136

It would probably be good to follow suit here as well.
`amount` & `initialAmount`. That way the UI is always writing to the `amount` field, and we sync the initial on the API side if the project is still in dev.

CarsonF (approved):

PR #3611, opened 2026-01-06
Mass drop our `*Id` suffixes in GQL relation inputs

CarsonF:
I've wanted to change this for a while. Some of the upcoming change-watching subscriptions also need to convert input fields to output types, and I don't want to continue trying to map between the two field names.


rdonigian (approved):
This looks good!

PR #3613, opened 2026-01-07
Project Mutation Subscriptions (CDC)

CarsonF:
This exposes a few ways to listen for project mutations/changes. ([Schema diff here](https://app.graphql-hive.com/seedco/cord/dev/checks/704769b0-c498-48db-a3a6-443e73a640d6?filter_changed=false&filter_failed=false))
There are subscriptions for specific actions: `projectCreated`/`projectUpdated`/`projectDeleted`.
Each of these can listen for "all" projects (implicit) or scope to a specific project by id.

There's also a `projectMutations` subscription, that merges these 3 streams together so all can be listened to at once without needing 3 subscriptions on the consumer side. This subscription also has a hook allowing sub-objects to add in more streams. Like `ProjectMemberCreated`, etc.

There's a lot here that's basically boilerplate. Almost all of this PR would need to be replicated for other resources. A bit concerning, but maybe that doesn't matter in 2026 with AI. I don't want to abstract too early either.
All of this really makes me wish had some Zod-like API for declaring these types so composition/mapping would be easier. Classes kinda suck for this.

There is essentially duplicate functionality here with the `@live` queries that I don't want to forget about. It is probably a different use case though. `@live` just keep the current data one is looking at fresh, when/what/how changes are made are unknown. These subscriptions allow reacting to certain specific changes. Syncing data to other systems seems like it could take advantage of knowing what change was pushed so it knows how to carry that forward.

There's a lot of code here to compose the `ProjectUpdated`/`ProjectMutation` GQL objects.
I'm wondering if this can also play into an "activity history" db table & UI view.
I've resisted in the past creating a "duplicate" view of the project that would be partial just for that case. But it seems important in this CDC case, and it really fits in perfectly for that activity-history representation need.
It also highlights that the partial change view is shallow, its only fields are fields that it owns, where as the canonical `Project` type also represents rich relations. Just to say that it is not really a "duplicate" at all.

# TODOs
- [x] How are changes to 1-Many relations represented? i.e. `Project.otherLocations`
- [x] Object naming (TODOs in code)
- [x] Consider replacing output types of mutations with these new objects? `UpdateProjectOutput` -> `ProjectUpdated`. The only thing that hinders if the mutation itself needs to return something unique to the mutation that wouldn't be represented in a `ProjectUpdated` payload. So far though over the years this hasn't been needed.
  - Yes, will do this is rename in separate PR for whole codebase.

### DIFF
--- a/src/components/project/dto/project-mutations.dto.ts
+++ b/src/components/project/dto/project-mutations.dto.ts
@@ -50,7 +50,7 @@ export class ProjectCreated extends ProjectMutation {
 
 @ObjectType()
 export class ProjectUpdate extends AsUpdateType(UpdateProject, {
-  omit: ['id'],
+  omit: ['id', 'changeset'],
   links: [
     'primaryLocation',
     'marketingLocation',


CarsonF:
Rebased. But added these commits:
46ea2f5990320915f41c9231ec277bf26458c90f
649e717db672a88e2e737f7f4e794b7cecb15e8f
52866355cb9f0d19e9f74a7f2cf95967cc266d3d

The TODOs for naming changes & the 1-Many relations are finished.
Ready for final review/merge.

PR #3624, opened 2026-01-17
Rename mutation outputs to unify with subscription events

CarsonF:
Following up on #3611, #3622, #3613
This is the last big breaking change planned of this run.

All mutations return a type that conveys an action was taken. This "taken action" will also work in subscriptions emitting this action to other users.

Project's have this work done already with #3613, and here we can see how that will carry forward as more fields are added to these "taken action" types.

I also took the opportunity to rework some weird signatures with auth & users.

I also took the opportunity to remove/replace the misleading `{ success: true }` in mutation outputs. It only exists to have a single field on the GQL type, so that the type is valid. But `success` kinda leads consumers to assume the need to check it. Now it's `{ doNotUse: null }` which is much more intuitive to ignore.

### DIFF
--- a/src/common/mutation-placeholder.output.ts
+++ b/src/common/mutation-placeholder.output.ts
@@ -6,13 +6,12 @@ import { stripIndent } from 'common-tags';
  */
 @ObjectType({ isAbstract: true })
 export abstract class MutationPlaceholderOutput {
-  @Field({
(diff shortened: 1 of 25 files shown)


bryanjnelson (approved):
Lots of work here!

PR #3670, opened 2026-03-09
Rev79 Module

rdonigian:


src/components/rev79/rev79.service.ts
```diff
@@ -0,0 +1,245 @@
...
+    }
+    const engagementId = engagementMatches[0]!.id;
+
+    // --- Resolve progress report ---
+    const quarterStart = fullFiscalQuarter(period.quarter, period.year).start;
+    const report = await this.periodicReportService.getReportByDate(
+      engagementId,
+      quarterStart,
+      'Progress',
+    );
```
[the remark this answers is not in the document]
rdonigian: Rev79 reports track with our fiscal calendar so this is correct as-is.

src/components/rev79/rev79.service.ts
```diff
@@ -0,0 +1,245 @@
...
+      }
+    }
+
+    if (item.productProgress?.length) {
+      for (const progress of item.productProgress) {
+        await this.productProgressService.update({
+          ...progress,
+          report: reportId,
+        });
+      }
```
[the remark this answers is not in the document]
rdonigian: The `uploadRev79ProgressReports` mutation is already wrapped in a transaction by the existing `TransactionalMutationsInterceptor`, so any failure in `processReportItem` rolls back all writes automatically. No change needed.

src/components/rev79/dto/quarter-period.input.ts
```diff
@@ -0,0 +1,17 @@
+import { Field, InputType, Int } from '@nestjs/graphql';
+
+@InputType({
+  description: 'A fiscal/calendar quarter within a specific year.',
```
bryanjnelson, line 4: About your comment...aren't fiscal and calendar years two separate things?  Isn't `year` here just a calendar year?

src/components/rev79/dto/upload-progress-reports.input.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+  UpdateStepProgress,
+  VariantProgressArg,
+} from '../../product-progress/dto';
+import { MediaCategory } from '../../progress-report/media/media-category.enum';
+import { type Prompt } from '../../prompts/dto';
+import { QuarterPeriodInput } from './quarter-period.input';
+
+@InputType()
+export class Rev79MediaInput {
```
bryanjnelson, line 19: Concerning the shape of this object, is this a Cord shape or a Rev79 shape?

bryanjnelson (approved):

src/components/rev79/dto/quarter-period.input.ts
rdonigian: It's a fiscal year so this needs to be updated

src/components/rev79/dto/upload-progress-reports.input.ts
rdonigian: It's the shape for Rev79.

bryanjnelson: Maybe I'm not thinking clearly on this, but don't we want to keep the inputs on the Cord side as abstract as possible (meaning Cord shapes) so that this is reusable in the future.  Meaning wouldn't we want to do the mapping of a Rev79 to Cord shape in `seed-api`?

PR #3673, opened 2026-03-20
Makes multiplication pdfs anonymous/accessible in IRP

brentkulwicki:
## Description
We essentially are trying to make multiplication report files anonymous so that they can be opened and spliced into the investor reports in IRP. 

IRP PR: https://github.com/SeedCompany/investor-portal/pull/335
Seed API PR: https://github.com/SeedCompany/seed-api/pull/985
Cord Field PR: https://github.com/SeedCompany/cord-field/pull/1764

- [x] Add/update tests if needed


rdonigian:
Walked through it, checked Neo to make sure the backfill migration was working and it looks good.  Just remove the pdf and I think we're set.

brentkulwicki:
> Walked through it, checked Neo to make sure the backfill migration was working and it looks good. Just remove the pdf and I think we're set.

Remove the pdf? You mean the .md file? @rdonigian

rdonigian (approved):

PR #3685, opened 2026-04-01
Find Communities by Rev79ProjectId

rdonigian:
### Add rev79ProjectCommunities query

seed-api syncs Rev79 quarterly data by calling rev79QuarterlyReportContext once per community ID. When a Rev79 community ID has no corresponding Cord engagement, this produces a Rev79CommunityNotFound error on every sync for every unmatched community — with no way to pre-filter.

This adds a new query:

**rev79ProjectCommunities(rev79ProjectId: String!): [Rev79Community!]!**

Returns all Rev79 community IDs (with language display names) that have mapped language engagements under the Cord project for the given Rev79 project ID. Returns an empty list if no Cord project is found — never throws.

seed-api can now do one preflight call to get the valid community IDs, then only call rev79QuarterlyReportContext for communities known to exist in Cord, eliminating the spurious errors entirely.

### DIFF
--- a/src/components/rev79/dto/index.ts
+++ b/src/components/rev79/dto/index.ts
@@ -1,4 +1,5 @@
 export * from './quarter-period.input';
+export * from './rev79-community';
 export * from './rev79-quarterly-report-context.input';
 export * from './rev79-quarterly-report-context.result';
 export * from './upload-progress-reports.input';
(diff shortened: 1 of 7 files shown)


bryanjnelson (dismissed):
🚀

sethmcknight (approved):

PR #3694, opened 2026-04-09
Give Display name priority for User.fullName

rdonigian:


bryanjnelson (approved):
I feel like this has been an issue for a long time.

PR #3695, opened 2026-04-09
Create Language.publicName

rdonigian:
This is a resolve field which will handle the ternary logic of first trying to use `displayName` if it exists otherwise defaulting to the actual language `name`

### DIFF
--- a/src/components/language/languageName.ts
+++ b/src/components/language/languageName.ts
@@ -1,8 +1,0 @@
-import { type Language } from './dto';
-
-export const languageName = (
-  language: Partial<Pick<Language, 'name' | 'displayName'>>,
-): string | undefined =>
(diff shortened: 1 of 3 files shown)


src/components/language/languageName.ts
bryanjnelson, line 1: I'm questioning whether this is the correct placement for this.  We really don't have individual object files like this.  (If we did, it should be named `language-name.ts`).  But I feel like this is better placed inside the existing `/language/dto/language.dto.ts` file.

bryanjnelson (changes requested):

bryanjnelson (dismissed):
👍🏼

bryanjnelson (approved):

PR #3716, opened 2026-04-27
#5 [pg] Users Migration

### LINKED ISSUE #3729: • User domain — Drizzle repo + SelfCondition + CreatorCondition ports

rdonigian:
## Summary

First PostgreSQL port. Introduces the Drizzle base infrastructure and ports the User, Education, Unavailability, and SystemAgent domains. Auth-adjacent tables (sessions, identities, password reset tokens) ship together so login flows work against PG end-to-end.

Toggled per-engine via `splitDb()`; Neo4j and Gel paths are unchanged.

## Schema (DDL)

Migration `0000_create_auth_tables.sql` — minimal users table + auth surface:

```sql
CREATE TYPE "user_status" AS ENUM ('Active', 'Disabled');

CREATE TABLE "users" (
  "id"         text        PRIMARY KEY NOT NULL,
  "is_root"    boolean     NOT NULL DEFAULT false,
  "status"     user_status NOT NULL,
  "email"      text        NOT NULL UNIQUE,
  "created_at" timestamptz NOT NULL DEFAULT now()
);

CREATE TABLE "user_global_roles" (
  "user_id" text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "role"    text NOT NULL,
  PRIMARY KEY ("user_id", "role")
);

CREATE TABLE "auth_sessions" (
  "token"        text PRIMARY KEY,
  "user_id"      text REFERENCES "users"("id") ON DELETE CASCADE,
  "created_at"   timestamptz NOT NULL DEFAULT now(),
  "logged_in_at" timestamptz,
  "active"       boolean NOT NULL DEFAULT true
);
CREATE INDEX "auth_sessions_active_idx"  ON "auth_sessions" ("token") WHERE active = true;
CREATE INDEX "auth_sessions_user_id_idx" ON "auth_sessions" ("user_id");

CREATE TABLE "auth_identities" (
  "user_id"       text PRIMARY KEY REFERENCES "users"("id") ON DELETE CASCADE,
  "password_hash" text NOT NULL,
  "updated_at"    timestamptz NOT NULL DEFAULT now()
);

CREATE TABLE "auth_password_reset_tokens" (
  "token"      text PRIMARY KEY,
  "user_id"    text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "email"      text NOT NULL,
  "created_on" timestamptz NOT NULL DEFAULT now()
);
CREATE INDEX "auth_password_reset_tokens_user_id_idx" ON "auth_password_reset_tokens" ("user_id");
```

Migration `0001_add_user_profile.sql` — User profile fields + sibling domains:

```sql
CREATE TYPE "gender" AS ENUM ('Male', 'Female');
CREATE TYPE "degree" AS ENUM ('Primary', 'Secondary', 'Associates', 'Bachelors', 'Masters', 'Doctorate');

ALTER TABLE "users" ALTER COLUMN "email" DROP NOT NULL;
ALTER TABLE "users"
  ADD COLUMN "real_first_name"    text NOT NULL DEFAULT '',
  ADD COLUMN "real_last_name"     text NOT NULL DEFAULT '',
  ADD COLUMN "display_first_name" text NOT NULL DEFAULT '',
  ADD COLUMN "display_last_name"  text NOT NULL DEFAULT '',
  ADD COLUMN "phone"      text,
  ADD COLUMN "timezone"   text NOT NULL DEFAULT 'America/Chicago',
  ADD COLUMN "about"      text,
  ADD COLUMN "title"      text,
  ADD COLUMN "gender"     "gender",
  ADD COLUMN "photo_id"   text,
  ADD COLUMN "updated_at" timestamptz NOT NULL DEFAULT now(),
  ADD COLUMN "deleted_at" timestamptz;

CREATE TABLE "educations" (
  "id"          text PRIMARY KEY,
  "user_id"     text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "degree"      "degree" NOT NULL,
  "major"       text NOT NULL,
  "institution" text NOT NULL,
  "created_at"  timestamptz NOT NULL DEFAULT now(),
  "updated_at"  timestamptz NOT NULL DEFAULT now(),
  "deleted_at"  timestamptz
);
CREATE INDEX "educations_user_id_idx" ON "educations" ("user_id");

CREATE TABLE "unavailabilities" (
  "id"          text PRIMARY KEY,
  "user_id"     text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "description" text NOT NULL,
  "start"       timestamptz NOT NULL,
  "end"         timestamptz NOT NULL,
  "created_at"  timestamptz NOT NULL DEFAULT now(),
  "updated_at"  timestamptz NOT NULL DEFAULT now(),
  "deleted_at"  timestamptz,
  CONSTRAINT "unavailabilities_valid_range_chk" CHECK ("end" > "start")
);
CREATE INDEX "unavailabilities_user_id_idx" ON "unavailabilities" ("user_id");

CREATE TABLE "system_agents" (
  "id"         text PRIMARY KEY,
  "name"       text NOT NULL UNIQUE,
  "roles"      text[] NOT NULL DEFAULT '{}',
  "created_at" timestamptz NOT NULL DEFAULT now()
);
```

## Drizzle base infrastructure

- `DrizzleService` — exposes `client` getter that delegates to an `AsyncLocalStorage`-bound transaction when one is open, otherwise the base pool. `inTx()` wraps a function in a Drizzle transaction and binds it via ALS so all repo calls inside flow through the same connection.
- `DrizzleTransactionalMutationsInterceptor` — wraps every GraphQL mutation in `inTx()` so hook handlers commit/rollback atomically.
- `DrizzleDtoRepository<TTable, TDto>` — abstract base with `readOne`, `readMany`, `softDelete`, `updateColumns`, `paginatedSelect`. Honors `deletedAt` (filters reads, supplies soft delete) and `updatedAt` (auto-stamps on partial update) when those columns exist.
- `DrizzleMigrator` — runs `drizzle-orm/migrator` on `OnModuleInit` when `DATABASE=postgres`.
- Helpers: `catchUniqueViolation(constraintMatch, field, message)` (maps Postgres `23505` → `DuplicateException`), `escapeLikePattern(value)` (escapes `%`, `_`, `\\` for safe ILIKE), `resolveOrderBy(input, map, fallback)` (one-stop ASC/DESC + sort-key lookup), `PgErrorCode` constants.

### `pgEnum` for closed sets, `text` array for open ones

- `user_status`, `gender`, `degree` are pgEnum — small fixed sets with known values; the DB enforces validity.
- `roles` is `text[]` — the `Role` set evolves frequently and is owned by application code; pgEnum would force a migration on every addition. Branded with `$type<Role[]>()` so we still get TS narrowing at the column level.

### Column-level type branding via `$type<>()`

- Schema columns are branded directly: `id: text('id').$type<ID<'User'>>().primaryKey()`, `userId: text('user_id').$type<ID<'User'>>()...`, `gender: genderEnum('gender').$type<Gender>()`.
- Result: query rows come back already typed as the domain shape (`row.id` is `ID<'User'>`, `row.gender` is `Gender`). No `as` casts inside repos.

### Explicit FK indexes

- Added on `educations.user_id`, `unavailabilities.user_id`, `auth_sessions.user_id`, `auth_password_reset_tokens.user_id`.
- Postgres does **not** auto-index FK columns (unlike MySQL). Without these, `WHERE user_id = $1` queries (very common for these child tables) seq-scan, and `ON DELETE CASCADE` on `users` triggers a full scan of every child table.

### Partial index on `auth_sessions(token) WHERE active = true`

- Session resume is the hot path and only touches active rows; partial index is materially smaller than a full token index.

### Range check on `unavailabilities`

- `CHECK ("end" > "start")` enforced at the DB level; the GraphQL layer can't be the only validator.

### Soft delete via `deletedAt`

- User, Education, Unavailability use a nullable `deleted_at` column; reads filter with `IS NULL`. `DrizzleDtoRepository.readMany` and `softDelete` do this automatically when the column exists.
- SystemAgent has no soft delete (rows are managed by the upsert-on-startup flow).

### Email becomes nullable

- Required to match the existing GraphQL contract (`email: SecuredEmail` allows null) and the Neo4j data model. The unique constraint stays — Postgres allows multiple NULLs in a unique column.

### `updateColumns` (not `updateProperties`)

- Renamed away from the Neo4j-flavored `updateProperties` since columns is what we update in SQL. Drops `undefined` values, stamps `updatedAt` when present, no-ops if no fields remain.

### `delete(id)` (not `deleteNode(objectOrId)`)

- Drizzle path uses `delete(id: ID)`. The Neo4j-flavored `deleteNode(objectOrId)` is being phased out codebase-wide as a separate cleanup; the Unavailability service-side call site was switched in this branch (Path B per-domain) so the Drizzle repo doesn't need to ship a Neo4j shim.

### `splitDb` Drizzle slot uses `as any`

- Gel repos satisfy `PublicOf<XRepository>` because `RepoFor` extends `CommonRepository`, which mirrors the Neo4j auth/helper surface (`privileges`, `getActualChanges`, `isUnique`, etc.). Our Drizzle base intentionally omits all of that. Forcing `implements PublicOf` would mean bolting Neo4j-flavored stubs onto every Drizzle repo. The cast in module wiring is the line; tracked with `migration-todo`.

### `AsDrizzleParams` distinct from `AsCypherParams`

- Same shape today, different name. Self-condition's Drizzle method now imports `AsDrizzleParams` so a Drizzle file doesn't import a Cypher-named type. `AsCypherParams` stays with a `migration-todo` until Neo4j retires.

## Deferred items

- `assignOrganizationToUser` / `removeOrganizationFromUser` — stubs that throw `NotImplementedException`. Will be implemented when the Organization domain is ported (next branch).
- `pinned` on `User` — returned as `false` from `toDto`. This is per-requesting-user state and lives on the requester, not the user row; will be wired up when the pinning subsystem is ported.
- `SystemAgent` cast in `toAgent` — the DTO is abstract; the row→class shape needs a cast until SystemAgent is restructured.
- Codebase-wide `deleteNode` → `delete` rename — tracked as a separate cleanup PR (User and Unavailability done here; Location and others pending).
- `AsCypherParams` — retained for Cypher-side conditions until Neo4j is retired.


src/components/user/user.drizzle.repository.ts
```diff
@@ -0,0 +1,382 @@
...
+        (r) =>
+          // migration-todo: SystemAgent is abstract; cast bridges plain row → class shape
+          ({
+            ...r,
+            __typename: 'SystemAgent' as const,
+            roles: r.roles as Role[],
+            createdAt: DateTime.fromJSDate(r.createdAt),
+          }) as unknown as UnsecuredDto<User | SystemAgent>,
+      ),
+    ];
```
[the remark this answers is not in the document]
rdonigian: DataLoaderFactory re-orders results by propertyKey ([data-loader.factory.js:71](vscode-webview://1pu95m9uvo8p6fsq84lsflr22g0g5h7tb2tsh2bts0bhih41rpgp/node_modules/@seedcompany/data-loader/dist/data-loader.factory.js#L71)), so loadMany has no input-order contract, and the Neo4j/Gel impls don't preserve order either; no fix needed.

src/components/user/user.drizzle.repository.ts
```diff
@@ -0,0 +1,382 @@
...
+    await this.identity.asUser(id, async () => {
+      await this.files.createDefinedFile(
+        photoId,
+        'Photo',
+        id,
+        'photo',
+        input.photo,
+        true,
+      );
+    });
```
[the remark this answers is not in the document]
rdonigian: The functionality is continuing from the Neo pattern.

src/core/drizzle/dto.repository.ts
```diff
@@ -0,0 +1,119 @@
...
+        .limit(pageSize)
+        .offset(offset),
+    ]);
+    const total = countResult[0]?.total ?? 0;
+    return {
+      rows: rows as Array<TTable['$inferSelect']>,
+      total,
+      hasMore: offset + rows.length < total,
+    };
+  }
```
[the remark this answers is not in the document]
rdonigian: The asymmetry is intentional and documented at [dto.repository.ts:83-84](vscode-webview://1pu95m9uvo8p6fsq84lsflr22g0g5h7tb2tsh2bts0bhih41rpgp/src/core/drizzle/dto.repository.ts#L83-L84) — paginatedSelect is a query helper that takes a caller-built predicate, while readMany is an opinionated by-id lookup; all three current call sites already include isNull(deletedAt) in their conditions, so there's no missing filter to fix. Before flagging consistency-style findings here, please verify both methods' contracts and check actual call sites — auto-filtering would silently double the predicate or require a signature change the finding itself rules out.

src/core/drizzle/dto.repository.ts
```diff
@@ -0,0 +1,119 @@
...
+  async readOne(id: ID): Promise<UnsecuredDto<TDto>> {
+    const rows = await this.readMany([id]);
+    const row = rows.find((r) => r.id === id);
+    if (!row) throw new NotFoundException();
+    return row;
+  }
+
+  async readMany(ids: readonly ID[]): Promise<Array<UnsecuredDto<TDto>>> {
+    if (ids.length === 0) return [];
+    const conds: SQL[] = [inArray(this.table.id, [...ids])];
```
bryanjnelson: nit: spell out `conditions`

src/core/drizzle/dto.repository.ts
```diff
@@ -0,0 +1,119 @@
...
+    const deletedAt = (this.table as Record<string, unknown>).deletedAt as
+      | AnyPgColumn
+      | undefined;
+    if (deletedAt) conds.push(isNull(deletedAt));
+    const rows = (await this.db.db
+      .select()
+      .from(this.table as PgTable)
+      .where(and(...conds))) as Array<TTable['$inferSelect']>;
+    return rows.map((r) => this.toDto(r));
```
bryanjnelson: nit: prefer `row` over `r` here

src/core/drizzle/dto.repository.ts
bryanjnelson, line 1: There're a lot of `as whatever` constructs in here.  My understanding is that that is often not very type safe, and that even using `satisfies` would be safer.  We might want to look at that.

src/components/user/education/education.drizzle.repository.ts
```diff
@@ -0,0 +1,100 @@
...
+  typeof educations,
+  Education
+> {
+  constructor(db: DrizzleService) {
+    super(db, educations);
+  }
+
+  async create(input: CreateEducation): Promise<UnsecuredDto<Education>> {
+    const id = await generateId();
+    await this.db.db.insert(educations).values({
```
bryanjnelson: `this.db.db.` 👉🏼  awkward

src/components/user/education/education.drizzle.repository.ts
```diff
@@ -0,0 +1,100 @@
...
+    return await this.readOne(id);
+  }
+
+  async update(changes: UpdateEducation): Promise<UnsecuredDto<Education>> {
+    const { id, ...fields } = changes;
+    await this.updateProperties(id, fields);
+    return await this.readOne(id);
+  }
+
+  async getUserIdByEducation(id: ID): Promise<{ id: ID }> {
```
bryanjnelson: Our naming generally is in reverse like this:

`getUserByEducationId`

src/components/user/education/education.drizzle.repository.ts
```diff
@@ -0,0 +1,100 @@
...
+      sortColumns[input.sort as keyof typeof sortColumns] ??
+      educations.institution;
+
+    const { rows, total, hasMore } = await this.paginatedSelect({
+      predicate: and(...conditions),
+      orderBy: [dir(orderCol)],
+      page: input.page,
+      count: input.count,
+    });
+    return { items: rows.map((r) => this.toDto(r)), total, hasMore };
```
bryanjnelson: nit: prefer `row` over `r` here

src/components/user/unavailability/unavailability.drizzle.repository.ts
```diff
@@ -0,0 +1,108 @@
...
+      conditions.push(eq(unavailabilities.userId, input.filter.userId));
+
+    const dir = input.order === 'ASC' ? asc : desc;
+    const sortColumns = {
+      description: unavailabilities.description,
+      start: unavailabilities.start,
+      end: unavailabilities.end,
+    } satisfies Partial<Record<keyof Unavailability, unknown>>;
+    const orderCol =
+      sortColumns[input.sort as keyof typeof sortColumns] ??
```
bryanjnelson: I see this pattern throughout this PR a lot.   This is hard to reason on.  I wonder if there's a better way to do this?

src/components/user/user.drizzle.repository.ts
```diff
@@ -0,0 +1,328 @@
...
+  ) {
+    super(db, users);
+  }
+
+  override async readMany(
+    ids: readonly ID[],
+  ): Promise<Array<UnsecuredDto<User>>> {
+    const rows = await this.db.db.query.users.findMany({
+      where: (u, { and: a, inArray: inArr, isNull: n }) =>
+        a(inArr(u.id, [...ids]), n(u.deletedAt)),
```
bryanjnelson: What's the reasoning for aliasing these?

src/components/user/user.drizzle.repository.ts
```diff
@@ -0,0 +1,328 @@
...
+        email: input.email ?? null,
+        realFirstName: input.realFirstName,
+        realLastName: input.realLastName,
+        displayFirstName: input.displayFirstName,
+        displayLastName: input.displayLastName,
+        phone: input.phone ?? null,
+        timezone: input.timezone ?? 'America/Chicago',
+        about: input.about ?? null,
+        title: input.title ?? null,
+        gender: (input.gender ?? null) as (typeof users.$inferInsert)['gender'],
```
bryanjnelson: I see this pattern throughout this PR a lot.  Is there a way we could create a type or something for this `$inferWhatever` business?  Maybe not, I didn't think about it much.  It just seems awkward

src/components/user/user.drizzle.repository.ts
```diff
@@ -0,0 +1,328 @@
...
+        or(
+          ilike(users.realFirstName, term),
+          ilike(users.realLastName, term),
+          ilike(users.displayFirstName, term),
+          ilike(users.displayLastName, term),
+        )!,
+      );
+    }
+    if (input.filter?.title) {
+      const escapedTitle = input.filter.title.replace(/[%_\\]/g, '\\$&');
```
bryanjnelson: Maybe these become more global at some point, if they continue to be used elsewhere...

src/core/drizzle/dto.repository.ts
```diff
@@ -0,0 +1,119 @@
...
+   * after filtering.
+   */
+  protected async updateProperties(
+    id: ID,
+    changes: Partial<TTable['$inferInsert']>,
+  ): Promise<void> {
+    const entries = Object.entries(changes).filter(([, v]) => v !== undefined);
+    if (entries.length === 0) return;
+    const set: Record<string, unknown> = Object.fromEntries(entries);
+    if ((this.table as Record<string, unknown>).updatedAt) {
```
bryanjnelson: I see this pattern throughout this PR a lot.  There's a lot of `unknown`.  I'm wondering if there's a better structure to use here at some point.

src/components/user/user.drizzle.repository.ts
```diff
@@ -243,29 +247,28 @@ export class UserDrizzleRepository extends DrizzleDtoRepository<
...
-    const rolesByUser = groupBy(allRoles, (r) => r.userId);
+    const rolesByUser = groupBy(allRoles, (row) => row.userId);
 
     return {
       total,
-      // migration-todo: items are UnsecuredDto<User>; service layer secures before resolving
-      items: rows.map((r) =>
-        this.toDto({ ...r, globalRoles: rolesByUser[r.id] ?? [] }),
-      ) as unknown as User[],
+      items: rows.map((row) =>
```
bryanjnelson, line 262: Was that migration comment intentionally left out on the refactor?

bryanjnelson (approved):
🚀

PR #3718, opened 2026-04-27
Fix: Project mou date policies

### LINKED ISSUE #3681: Restrict MOU date changes to the Financial Services team
Only Financial Services roles should be permitted to update MOU dates in Cord Field once a project has been Approved. I am not entirely certain, but I believe this was the previous behavior.

Justin and Bety agree that it is not ideal for RDs and FODs to update these dates, as this could cause issues such as dates being altered without an approved CtP, or making it hard to identify the original dates without reviewing the project's documents.

rdonigian:
Prevent non financial service roles from altering active projects mou dates.


bryanjnelson (approved):
🔒

PR #3764, opened 2026-04-28
Add Tool filter to ProjectFilters & EngagementFilters, create ToolContainerType

rdonigian:
Adds `tool: ToolFilters` (with `id` and `name`) to both ProjectFilters and
EngagementFilters so the frontend can drive project/engagement lists with
full server-side sort, filter, and pagination by tool — replacing the current
client-side approach.

ToolContainerType enum is simplified from separate LanguageEngagement /
InternshipEngagement values to a single Engagement value. Both Neo4j (via
CASE WHEN in Cypher) and Gel (via post-processing) normalize the concrete
subtypes before returning.

Also wires up the previously-unregistered ToolContainerSummaryLoader and
adds the filteredUsages / containerSummary resolver fields on Tool that the
frontend was already expecting.

### DIFF
--- a/src/components/tools/tool-usage/tool-usage.service.ts
+++ b/src/components/tools/tool-usage/tool-usage.service.ts
@@ -23,7 +23,15 @@ import { Privileges } from '../../authorization';
 import { Tool } from '../tool/dto';
 import { type ToolKey } from '../tool/dto/tool-key.enum';
 import { ToolRepository } from '../tool/tool.neo4j.repository';
-import { type CreateToolUsage, ToolUsage, type UpdateToolUsage } from './dto';
+import {
+  type CreateToolUsage,
+  type SecuredToolUsageList,
+  type ToolContainerSummary,
+  type ToolContainerType,
+  ToolUsage,
+  type ToolUsageFilters,
+  type UpdateToolUsage,
+} from './dto';
 import { type UsagesByContainer } from './tool-usage-by-container.loader';
 import { type UsagesByTool } from './tool-usage-by-tool.loader';
 import { ToolUsageRepository } from './tool-usage.neo4j.repository';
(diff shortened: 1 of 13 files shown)


src/components/tools/tool-usage/tool-usage.service.ts
```diff
@@ -124,6 +132,65 @@ export class ToolUsageService {
...
+    if (!row) {
+      return {
+        items: [],
+        total: 0,
+        hasMore: false,
+        canRead: true,
+        canCreate: false,
+      };
+    }
+    const usagesRaw = await Promise.all(
```
bryanjnelson:
```suggestion
    const rawUsages = await Promise.all(
```

src/components/tools/tool-usage/tool-usage.service.ts
```diff
@@ -124,6 +132,65 @@ export class ToolUsageService {
...
+  async readContainerSummaryForTools(
+    tools: readonly Tool[],
+  ): Promise<Array<{ tool: Tool; summary: ToolContainerSummary[] }>> {
+    const toolsById = mapKeys.fromList(tools, (t) => t.id).asMap;
+    const rows = await this.repo.containerSummaryForTools(
+      tools.map((t) => t.id),
+    );
+    // Group rows by tool id and filter out types not in the ToolContainerType enum
+    const validTypes = new Set<string>(['Engagement', 'Project']);
+    const byToolId = new Map<ID, ToolContainerSummary[]>();
```
bryanjnelson: This variable name could be more descriptive.  What exactly is by tool id?

src/components/tools/tool-usage/tool-usage.service.ts
```diff
@@ -124,6 +132,65 @@ export class ToolUsageService {
...
+      entries.push({
+        containerType: row.containerType as ToolContainerType,
+        total: Number(row.total),
+      });
+      byToolId.set(row.tool.id, entries);
+    }
+    return tools.map((tool) => ({
+      tool: toolsById.get(tool.id)!,
+      summary: byToolId.get(tool.id) ?? [],
+    }));
```
bryanjnelson: This pattern with the for loop and the continue is something we don't often use.  I wonder if it could be written more cleanly with a `.filter()` or `.map()` or something?

src/components/tools/tool-usage/tool-usages.resolver.ts
```diff
@@ -17,4 +26,27 @@ export class ToolUsagesResolver {
...
+    description:
+      'Distinct container types in use and their counts — for tab display',
+  })
+  async containerSummary(
+    @Parent() tool: Tool,
+    @Loader(() => ToolContainerSummaryLoader)
+    loader: LoaderOf<ToolContainerSummaryLoader>,
+  ): Promise<ToolContainerSummary[]> {
+    const { summary } = await loader.load(tool);
+    return summary;
```
bryanjnelson:
```suggestion
return (await loader.load(tool)).summary;
```

bryanjnelson (changes requested):

src/components/tools/tool-usage/tool-usage.service.ts
```diff
@@ -124,6 +120,81 @@ export class ToolUsageService {
...
+        const labels: readonly string[] = usage.container.value?.labels ?? [];
+        // labels may be Neo4j node labels or Gel FQN type names; both contain
+        // the substring 'Engagement' or 'Project' for the relevant subtypes
+        const containerType: ToolContainerType | null = labels.some((l) =>
+          l.includes('Engagement'),
+        )
+          ? 'Engagement'
+          : labels.some((l) => l.includes('Project'))
+            ? 'Project'
+            : null;
```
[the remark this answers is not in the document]
rdonigian: The labels field comes from two different backends with different shapes:

Neo4j: actual node labels — includes ancestor types, e.g. ['LanguageEngagement', 'Engagement', 'Resource', 'BaseNode']. Exact match on 'Engagement' works here.
Gel: e.array_agg(e.set(res.__type__.name)) — only the concrete FQN, e.g. ['default::LanguageEngagement']. Exact match on 'Engagement' returns false here, breaking Gel entirely.
The includes check is deliberate — it's the only predicate that works correctly for both backends simultaneously. 'default::LanguageEngagement'.includes('Engagement') is true; 'default::LanguageEngagement' === 'Engagement' is false.

The suggested fix (l.trim() === 'Engagement') would silently produce empty summaries for any tool when running against the Gel repo. No change needed.

bryanjnelson (approved):
👍🏼

PR #3765, opened 2026-04-29
#6 [pg] Location Migration

### LINKED ISSUE #3730: Location domain — wire FieldZone/FieldRegion pilot into develop + E2E smoke test

rdonigian:
## Summary

Second PostgreSQL port. Adds the `locations` table and ports `LocationRepository` to Drizzle. Builds on `users-pg` (base infrastructure + User/Education/Unavailability/SystemAgent/Auth) — that PR must merge first.

Engine routing via `splitDb()`; Neo4j and Gel paths unchanged.

## Schema (DDL)

```sql
CREATE TYPE location_type AS ENUM (
  'Country', 'City', 'County', 'Region', 'State', 'CrossBorderArea'
);

CREATE TABLE locations (
  id                          text PRIMARY KEY,
  name                        text NOT NULL UNIQUE,
  type                        location_type NOT NULL,
  iso_alpha3                  text UNIQUE,
  funding_account_id          text,
  default_field_region_id     text,
  default_marketing_region_id text REFERENCES locations(id),
  map_image_id                text,
  created_at                  timestamptz NOT NULL DEFAULT now(),
  updated_at                  timestamptz NOT NULL DEFAULT now(),
  deleted_at                  timestamptz
);

CREATE INDEX "locations_default_marketing_region_id_idx"
  ON "locations" ("default_marketing_region_id");
```

### `location_type` is `pgEnum`, not `text`

Closed set of six values, owned by application code but rarely changes. pgEnum gives DB-side validity and pairs with `$type<LocationType>()` in the schema for end-to-end type narrowing on read.

### FK constraints deferred for cross-domain references

- `funding_account_id` and `default_field_region_id` are plain `text` columns, not `REFERENCES`. The FundingAccount and FieldRegion domains haven't been ported to PG yet — adding the constraint would prevent writes against PG since those parent rows don't exist. Marked with a `migration-todo` to add the constraint when those domains land.
- `default_marketing_region_id` is a self-FK (Location → Location), so the constraint is added now.

### FK index on `default_marketing_region_id`

Postgres doesn't auto-index FK columns. Without this, `WHERE default_marketing_region_id = $1` queries seq-scan, and `ON DELETE` triggers on `locations` would fan out unindexed lookups.

### Soft delete via `deleted_at`

Matches the User domain pattern: nullable `deleted_at`, reads filtered with `IS NULL`. `DrizzleDtoRepository.readMany` and `softDelete` honor it automatically.

### `name` and `iso_alpha3` are `UNIQUE`

`catchUniqueViolation('name', 'name', 'Location with this name already exists.')` maps Postgres `23505` → `DuplicateException` on conflict. ISO alpha-3 codes are world-unique by definition, so the constraint is appropriate even though no business logic enforces it.

### Schema columns branded with `$type<>()`

- `id: text('id').$type<ID<'Location'>>().primaryKey()`
- `type: locationTypeEnum('type').$type<LocationType>().notNull()`
- `fundingAccountId: text('funding_account_id').$type<ID<'FundingAccount'>>()`, etc.

Result: query rows come back already typed as the domain shape; `toDto` has no `as ID` / `as FileId` casts.

### `list()` returns `PaginatedListType<UnsecuredDto<Location>>`

The repo produces unsecured DTOs; the service secures them with `privileges.for(Location).secure(dto)` before returning. Earlier WIP typed `list()` as `LocationListOutput` (secured items) and used an `as unknown as Location[]` cast to satisfy the compiler. That cast is removed and the type now matches reality — same pattern as User/Education/Unavailability list methods.

### Reuses cleanup helpers from `users-pg`

- `escapeLikePattern(value)` for the name ILIKE filter (escapes `%`, `_`, `\\`).
- `resolveOrderBy(input, sortMap, fallback)` + `SortMap<keyof Location>` for sort resolution.
- `EnhancedResource.of(Location)` cached as a `private readonly resource` field instead of constructed twice.

### `splitDb` Drizzle slot uses `as any`

Same pattern as User/Education/Unavailability: Gel repos satisfy `PublicOf<XRepository>` because `RepoFor` extends `CommonRepository` which mirrors the Neo4j surface (`privileges`, `getActualChanges`, etc.); `DrizzleDtoRepository` deliberately omits all that. The cast is the line; tracked with `migration-todo`.

## Deferred items

- `addLocationToNode`, `removeLocationFromNode`, `listLocationsFromNodeNoSecGroups` — stubs that throw `NotImplementedException`. These manage relationships between Locations and other resources (Project, Organization, etc.) and need their parent domains ported before they can be implemented.
- FK constraints for `funding_account_id` and `default_field_region_id` — added when FundingAccount / FieldRegion port to PG.
- FK indexes for the same two columns — deferred with the constraints (no point indexing a column we don't yet query through joins).
- `splitDb` `as any` cast — see decisions above.

### DIFF
--- a/src/components/location/location.drizzle.repository.ts
+++ b/src/components/location/location.drizzle.repository.ts
@@ -1,0 +1,215 @@
+import { Injectable } from '@nestjs/common';
+import { and, eq, ilike, inArray, isNull, type SQL } from 'drizzle-orm';
+import { DateTime } from 'luxon';
+import {
+  EnhancedResource,
+  generateId,
+  type ID,
+  NotImplementedException,
+  type PaginatedListType,
+  ServerException,
+  type UnsecuredDto,
+} from '~/common';
+import {
+  catchUniqueViolation,
+  DrizzleDtoRepository,
+  escapeLikePattern,
+  resolveOrderBy,
+  type SortMap,
+} from '~/core/drizzle';
+import { DrizzleService } from '~/core/drizzle/drizzle.service';
+import { locations } from '~/core/drizzle/schema';
+import { type ResourceNameLike } from '~/core/resources';
+import { PolicyExecutor } from '../authorization/policy/executor/policy-executor';
+import { FileService } from '../file';
+import { type FileId } from '../file/dto';
+import {
+  type CreateLocation,
+  Location,
+  type LocationListInput,
+  type LocationListOutput,
+  type UpdateLocation,
+} from './dto';
+
+const catchNameUnique = catchUniqueViolation(
+  'name',
+  'name',
+  'Location with this name already exists.',
+);
+
+@Injectable()
+export class LocationDrizzleRepository extends DrizzleDtoRepository<
+  typeof locations,
+  Location
+> {
+  private readonly resource = EnhancedResource.of(Location);
+
+  constructor(
+    db: DrizzleService,
+    private readonly files: FileService,
+    private readonly executor: PolicyExecutor,
+  ) {
+    super(db, locations);
+  }
+
+  async create(input: CreateLocation): Promise<UnsecuredDto<Location>> {
+    const id = await generateId();
+    const mapImageId = await generateId<FileId>();
+
+    await this.db
+      .insert(locations)
+      .values({
+        id,
+        name: input.name,
+        type: input.type,
+        isoAlpha3: input.isoAlpha3 ?? null,
+        fundingAccountId: input.fundingAccount ?? null,
+        defaultFieldRegionId: input.defaultFieldRegion ?? null,
+        defaultMarketingRegionId: input.defaultMarketingRegion ?? null,
+        mapImageId,
+      })
+      .catch(catchNameUnique);
+
+    const dto = await this.readOne(id);
+
+    await this.files.createDefinedFile(
+      mapImageId,
+      input.name,
+      id,
+      'mapImage',
+      input.mapImage,
+      true,
+    );
+
+    return dto;
+  }
+
+  async update(changes: UpdateLocation): Promise<UnsecuredDto<Location>> {
+    const { id, mapImage, ...fields } = changes;
+
+    await this.updateColumns(id, {
+      name: fields.name,
+      type: fields.type,
+      isoAlpha3: fields.isoAlpha3,
+      fundingAccountId: fields.fundingAccount,
+      defaultFieldRegionId: fields.defaultFieldRegion,
+      defaultMarketingRegionId: fields.defaultMarketingRegion,
+    }).catch(catchNameUnique);
+
+    if (mapImage !== undefined) {
+      const location = await this.readOne(id);
+      if (!location.mapImage) {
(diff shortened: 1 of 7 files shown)


src/core/drizzle/schema/index.ts
```diff
@@ -236,3 +238,47 @@ export const authPasswordResetTokensRelations = relations(
...
+    mapImageId: text('map_image_id').$type<ID<'File'>>(),
+    createdAt: timestamp('created_at', { withTimezone: true })
+      .notNull()
+      .defaultNow(),
+    updatedAt: timestamp('updated_at', { withTimezone: true })
+      .notNull()
+      .defaultNow(),
+    deletedAt: timestamp('deleted_at', { withTimezone: true }),
+  },
+  (t) => [
```
bryanjnelson, line 277: Prefer something more than `t`.

src/components/location/location.drizzle.repository.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  constructor(
+    db: DrizzleService,
+    private readonly files: FileService,
+    private readonly executor: PolicyExecutor,
+  ) {
+    super(db, locations);
+  }
+
+  async create(input: CreateLocation): Promise<UnsecuredDto<Location>> {
```
bryanjnelson, line 55: I'm not sold on this, but I'd like your thoughts on if it would be worth naming this type and using it in this file for readability:

```tst
type UnsecuredLocation = UnsecuredDto<Location>;
```

It wouldn't by a ton...but is it more readable?

rdonigian: We use UnsecuredDto generic without a new declaration throughout the whole code base.  I think it's fine.

src/core/drizzle/schema/index.ts
rdonigian: This is the standard per Drizzle docs to use t for table in the schema.  I'm normally not a fan of these very shorthand var names but I think within this file it's appropriate.

bryanjnelson: Oh it's for a table?  Hmm...okay.  I will reluctantly concede.  😉

bryanjnelson (approved):
🚀

src/components/location/location.drizzle.repository.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  // migration-todo: implement when location-node relationships are migrated to PG
+  listLocationsFromNodeNoSecGroups(
+    _label: string,
+    _rel: string,
+    _id: ID,
+    _input: LocationListInput,
+  ): Promise<LocationListOutput> {
+    throw new NotImplementedException();
+  }
```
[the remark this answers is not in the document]
rdonigian: splitDb is repo-level so we can't delegate, the PG join tables for Project/Language/User don't exist yet, and production never uses DATABASE=postgres so the stubs can't be hit there — they're intentional gaps tagged migration-todo, not bugs.

PR #3793, opened 2026-05-22
upgrades packages to fix critical dependabot alerts

brentkulwicki:
## Reason for this PR
This is fixing all the critical dependabot alerts in this repo. 

-->

## Description
See above for desc. 

## Ready for review checklist

@rdonigian and maybe @renatorossiter  - The only thing I want to test in QA that I have any concerns about (and they are minimal concerns) is the file storage. Just want to test to make sure when we upload a file it is saving properly in s3 and accessible. And even then, I feel like it's <5% chance there is an issue here.


rdonigian (approved):
LGTM

PR #3805, opened 2026-06-24
[pg] Postgres Test Harness

rdonigian:
### Summary

Adds an ephemeral-Postgres e2e harness, wires `postgres` into CI, and brings the Phase 2 domains up under `DATABASE=postgres`. All changes sit at the repository + test-infra layer, so they ride the existing `splitDb()` routing without altering service/resolver code.

### Test harness

- `ephemeralPg()` (`test/setup/pg-setup.ts`) provisions a uniquely named database per spec file and lets the app apply its Drizzle migrations and bootstrap root objects on boot — so every spec exercises the real migration path end to end.
- Teardown drops the DB inside a try/finally and reaps stale `test_*` databases (>1h) on startup, so the admin pool is always released and abandoned DBs self-clean across parallel workers.
- `createApp()` provisions the ephemeral DB, injects its URL through `ConfigService`, and tears it down on `app.close()`.

### CI

- Adds a `postgres:16-alpine` service container and a `postgres` matrix entry (`POSTGRES_URL`).
- The postgres run covers the 10 domains that have a Drizzle repo on this branch — `authentication`, `user`, `education`, `unavailability`, `location`, `organization`, `region`, `zone`, `funding-account`, `tool`. The matrix comment ties each pattern to a matching `*.drizzle.repository.ts` so the list grows in lockstep with the migration.

### Repository-layer changes

- `AdminDrizzleService`/`-Repository` bootstrap the root user (`users.isRoot`) so a fresh DB comes up usable.
- Drizzle logout re-anonymizes the session (`userId`/`loggedInAt` nulled, token stays active), so the token continues as a valid anonymous session; the `active` flag is reserved for the `deactivate*Sessions` revoke paths.
- Implements `assignOrganizationToUser` / `removeOrganizationFromUser` over `user_organizations`.
- `listLocationsFromNodeNoSecGroups` returns an empty page, which is the accurate result for PG-resident resources whose node edges haven't migrated yet, and keeps reads like `user.locations` working.
- Renames `deleteNode` → `delete` for FieldZone / FieldRegion.
- Gates `photo`/`mapImage` DefinedFile creation behind the File domain migration (tagged `migration-todo`).

### DIFF
--- a/src/components/admin/admin.drizzle.repository.ts
+++ b/src/components/admin/admin.drizzle.repository.ts
@@ -27,7 +27,7 @@ export class AdminDrizzleRepository {
     await callback();
   }
 
-  async doesRootUserExist() {
+  async findRootUser() {
     const [row] = await this.drizzle.client
       .select({
         id: users.id,
@@ -41,13 +41,15 @@ export class AdminDrizzleRepository {
     return row;
   }
 
+  /**
+   * Atomic: a partial failure would leave a row with isRoot=true but no admin
+   * role/password, and findRootUser would then skip re-creation forever.
+   * Bootstrap isn't a GraphQL mutation, so the transactional-mutations
+   * interceptor doesn't cover it — wrap explicitly. (client is ALS-bound, so
+   * the auth write below joins the same tx.)
+   */
   async createRootUser(id: ID, email: string, passwordHash: string) {
     const userId = id as ID<'User'>;
-    // Atomic: a partial failure would leave a row with isRoot=true but no admin
-    // role/password, and doesRootUserExist would then skip re-creation forever.
-    // Bootstrap isn't a GraphQL mutation, so the transactional-mutations
-    // interceptor doesn't cover it — wrap explicitly. (client is ALS-bound, so
-    // the auth write below joins the same tx.)
     await this.drizzle.inTx(async () => {
       await this.drizzle.client.insert(users).values({
         id: userId,
@@ -66,11 +68,13 @@ export class AdminDrizzleRepository {
     });
   }
 
+  /**
+   * Idempotent on re-boot when the row already exists (id conflict → no-op).
+   * Scope the conflict to the id only: if a *different* org already holds
+   * this name, let the name unique-index violation surface rather than
+   * silently no-op and leave config.defaultOrg.id pointing at no row.
+   */
   async mergeDefaultOrg(id: ID, name: string) {
-    // Idempotent on re-boot when the row already exists (id conflict → no-op).
-    // Scope the conflict to the id only: if a *different* org already holds
-    // this name, let the name unique-index violation surface rather than
-    // silently no-op and leave config.defaultOrg.id pointing at no row.
     await this.drizzle.client
       .insert(organizations)
       .values({ id: id as ID<'Organization'>, name })
--- a/src/components/admin/admin.drizzle.repository.ts
+++ b/src/components/admin/admin.drizzle.repository.ts
@@ -1,0 +1,86 @@
+import { Injectable } from '@nestjs/common';
+import { ModuleRef } from '@nestjs/core';
+import { and, eq, isNull } from 'drizzle-orm';
+import { LazyGetter as Once } from 'lazy-get-decorator';
+import { type ID, Role } from '~/common';
+import { AuthenticationRepository } from '~/core/authentication/authentication.repository';
+import { DrizzleService } from '~/core/drizzle/drizzle.service';
+import {
+  authIdentities,
+  organizations,
+  userGlobalRoles,
+  users,
+} from '~/core/drizzle/schema';
+
+@Injectable()
+export class AdminDrizzleRepository {
+  constructor(
+    private readonly drizzle: DrizzleService,
+    private readonly moduleRef: ModuleRef,
+  ) {}
+
+  @Once() get auth() {
+    return this.moduleRef.get(AuthenticationRepository, { strict: false });
+  }
+
+  async finishing(callback: () => Promise<void>) {
+    await callback();
+  }
+
+  async doesRootUserExist() {
+    const [row] = await this.drizzle.client
+      .select({
+        id: users.id,
+        email: users.email,
+        hash: authIdentities.passwordHash,
+      })
+      .from(users)
+      .leftJoin(authIdentities, eq(authIdentities.userId, users.id))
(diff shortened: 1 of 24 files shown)


src/components/admin/admin.drizzle.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    // the background and allow webserver to start.
+    if (this.config.jest) {
+      await finishing;
+    } else {
+      finishing.catch((exception) => {
+        this.logger.error('Failed to setup root objects', {
+          exception,
+        });
+      });
+    }
```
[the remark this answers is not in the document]
rdonigian: It's a faithful port of the existing neo4j/gel AdminService pattern (intentional + documented), and the PG path is dev-only until cutover. Making startup block is a cross-engine architecture decision, not a harness fix.

src/components/user/user.drizzle.repository.ts
```diff
@@ -252,14 +249,56 @@ export class UserDrizzleRepository extends DrizzleDtoRepository<
...
+        .onConflictDoUpdate({
+          target: [userOrganizations.userId, userOrganizations.organizationId],
+          set: { primary: primary ?? false },
+        });
+    } catch (exception) {
+      throw new ServerException(
+        'Failed to assign organization to user',
+        exception,
+      );
+    }
```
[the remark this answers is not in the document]
rdonigian: assignOrganizationToUser atomicity — invalid: DrizzleTransactionalMutationsInterceptor already wraps every GraphQL mutation in inTx() under Postgres, so the primary-reset + upsert are already atomic. The migration notes explicitly say not to add redundant explicit transactions in repo methods.

src/components/admin/admin.drizzle.repository.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  @Once() get auth() {
+    return this.moduleRef.get(AuthenticationRepository, { strict: false });
+  }
+
+  async finishing(callback: () => Promise<void>) {
+    await callback();
+  }
+
+  async doesRootUserExist() {
```
bryanjnelson: Seems a bit oddly named.  I'd expect it to return a boolean with this name.  Change name or change what it returns??

src/components/admin/admin.drizzle.repository.ts
```diff
@@ -0,0 +1,86 @@
...
+      .from(users)
+      .leftJoin(authIdentities, eq(authIdentities.userId, users.id))
+      .where(and(eq(users.isRoot, true), isNull(users.deletedAt)))
+      .limit(1);
+    return row;
+  }
+
+  async createRootUser(id: ID, email: string, passwordHash: string) {
+    const userId = id as ID<'User'>;
+    // Atomic: a partial failure would leave a row with isRoot=true but no admin
```
bryanjnelson: Prefer this comment (and the one on the next method) move up to the method level.

src/core/drizzle/errors.ts
```diff
@@ -2,6 +2,21 @@ import { DatabaseError } from 'pg';
...
 
+/**
+ * Resolve the underlying pg `DatabaseError` from a thrown value. drizzle-orm
+ * rethrows query-execution failures wrapped in a `DrizzleQueryError`
+ * ("Failed query: ...") with the original pg error on `.cause`, so a bare
+ * `instanceof DatabaseError` check no longer matches. Unwrap one level so the
+ * constraint-mapping helpers keep working. Reuse for any future pg-error
+ * matcher (FK violation, check constraint, etc.).
+ */
+const asDatabaseError = (e: unknown): DatabaseError | undefined =>
```
bryanjnelson, line 13: Noice.

test/setup/pg-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export const ephemeralPg = async () => {
+  if (process.env.DATABASE !== 'postgres') {
+    return undefined;
+  }
+
+  const base = new URL(
+    process.env.POSTGRES_URL ??
+      'postgresql://postgres:postgres@localhost:5432/cord',
+  );
```
bryanjnelson: Do we want to default here if the ENV VAR isn't set?  What's the purpose of doing that?  Maybe throw instead?

test/setup/pg-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  return { url: url.toString(), cleanup };
+};
+
+async function dropStale(admin: Pool) {
+  // Escape `_` (a LIKE single-char wildcard) in the prefix so it matches
+  // literally; parameterized to keep it scoped to our prefix only.
+  const likePattern = TEST_DB_PREFIX.replace(/_/g, String.raw`\_`) + '%';
+  const res = await admin.query<{ datname: string }>(
```
bryanjnelson:
```suggestion
  const staleCandidatesResult = await admin.query<{ datname: string }>(
```

test/setup/pg-setup.ts
```diff
@@ -0,0 +1,86 @@
...
+    'select datname from pg_database where datname like $1',
+    [likePattern],
+  );
+
+  const stale = res.rows
+    .map((row) => row.datname)
+    .filter((name) => {
+      const ts = Number(name.slice(TEST_DB_PREFIX.length).split('_')[0]);
```
bryanjnelson:
```suggestion
      const timestamp = Number(name.slice(TEST_DB_PREFIX.length).split('_')[0]);
```

I know that's a little heavy, but I literally had no idea that `ts` was a timestamp just looking at it.

bryanjnelson (changes requested):
@rdonigian - Great work; thanks!

src/components/admin/admin.drizzle.repository.ts
rdonigian: This was just a carryover from the same function in gel.  How does `findRootUser` sound?

test/setup/pg-setup.ts
rdonigian: Erroring sounds appropriate.

src/components/admin/admin.drizzle.repository.ts
bryanjnelson: 👍🏼

test/setup/pg-setup.ts
bryanjnelson: Agreed!

bryanjnelson (approved):
🧪

PR #3810, opened 2026-07-02
[pg] Project + ProjectMember → Postgres (Drizzle) (PR 2 of 3)

### LINKED ISSUE #3736: db-migration: Phase 3 — Project + ProjectMember

### LINKED ISSUE #3739: Project + ProjectMember + Workflow — Drizzle repo
Project + ProjectMember + Workflow — Drizzle repo (must land together — circular filter dep). Includes: status as GENERATED ALWAYS AS, project_workflow_events, MemberCondition + MemberWithRolesCondition ports, SetDepartmentId retry on 23505. Assign most experienced dev.

rdonigian:
### Summary
- PR 2 of 3 of the Project → Postgres recut, stacked on **#3809** (PR 1 = schema + migration `0010`). This is the repository + authorization layer; **no schema/migration changes here** (all in PR 1). Neo4j/Gel paths untouched — everything routes through `splitDb`.

### Repositories
- `ProjectDrizzleRepository` + `ProjectMemberDrizzleRepository` — migrated together because their `*FilterClauses` import each other (the circular Project↔ProjectMember filter dependency).
- `membership-scope.ts` (`requesterScopeByProject`) — populates the requester's `ScopedRole[]` (`member:true` + project-scoped roles) on both DTOs so member-policy conditions resolve. Drizzle mirror of Neo4j `matchProjectScopedRoles`.

### Authorization (Drizzle condition ports)
- `MemberCondition` / `MemberWithRolesCondition` — `EXISTS` over `project_members` (roles variant adds `roles && ARRAY[...]::role[]`).
- `SensitivityCondition` — collapses to `projects.sensitivity <= clearance` (denormalized column).
- `And`/`Or` aggregate `asDrizzleCondition` composition (not previously on develop).

### Wiring + cross-domain exports
- `splitDb` postgres entries on the project + project-member modules.
- `*SortColumns` / `*FilterClauses` exports added to field-region / location / user repos (consumed by Project's cross-domain sort + filter).

### Service layer
- `deleteNode → delete(id)` rename across both services + Neo4j repo `delete(id)` wrappers (Gel inherits `delete` from `RepoFor`; Drizzle uses `softDelete`). Audit `ResourceMutatedHook` firing intentionally **omitted** — the audit domain isn't on develop yet.

### DIFF
--- a/src/components/authorization/policies/conditions/sensitivity.condition.ts
+++ b/src/components/authorization/policies/conditions/sensitivity.condition.ts
@@ -165,19 +165,19 @@ const sensitivityRefForResource = (
   resource: EnhancedResource<any>,
   access: Sensitivity,
 ): SQL => {
-  const accessLit = sql.raw(`'${access}'::"sensitivity"`);
+  const accessLiteral = sql.raw(`'${access}'::"sensitivity"`);
   switch (resource.name) {
     case 'Project':
     case 'TranslationProject':
     case 'MomentumTranslationProject':
     case 'MultiplicationTranslationProject':
     case 'InternshipProject':
-      return sql`"projects"."sensitivity" <= ${accessLit}`;
+      return sql`"projects"."sensitivity" <= ${accessLiteral}`;
     case 'ProjectMember':
       return sql`(
         select "p"."sensitivity" from "projects" "p"
         where "p"."id" = "project_members"."project_id"
-      ) <= ${accessLit}`;
+      ) <= ${accessLiteral}`;
     // migration-todo: re-add a case per domain as it ports to Postgres
     // (Partnership/Budget/Engagement/Ceremony/Language/BudgetRecord read
     // sensitivity via their parent project; Partner has its own derived column —
--- a/src/components/authorization/policies/conditions/sensitivity.condition.ts
+++ b/src/components/authorization/policies/conditions/sensitivity.condition.ts
@@ -1,9 +1,15 @@
 import { type NonEmptyArray } from '@seedcompany/common';
 import { type Query } from 'cypher-query-builder';
+import { type SQL, sql } from 'drizzle-orm';
 import { inspect, type InspectOptionsStylized } from 'util';
-import { type ResourceShape, Sensitivity } from '~/common';
+import {
+  type EnhancedResource,
+  type ResourceShape,
+  Sensitivity,
+} from '~/common';
 import { matchProjectSens, rankSens } from '~/core/neo4j/query';
 import {
+  type AsDrizzleParams,
   type AsEdgeQLParams,
   type Condition,
   fqnRelativeTo,
@@ -65,6 +71,15 @@ export class SensitivityCondition<
     const ranked = sensitivityRank[this.access];
     const param = query.params.addParam(ranked, 'requiredSens');
     return `${CQL_VAR} <= ${String(param)}`;
+  }
+
+  asDrizzleCondition({ resource }: AsDrizzleParams<TResourceStatic>) {
+    // PG's sensitivity enum is declared in `Low < Medium < High` order, so
+    // `node.sensitivity <= 'access'` is a single-column compare. For Project
+    // subtypes the column lives on the row directly (denormalized). For
+    // Project-scoped children the column is on the parent project; use a
+    // correlated subquery.
+    return sensitivityRefForResource(resource, this.access);
   }
 
   setupEdgeQLContext({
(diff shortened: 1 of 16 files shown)


src/components/authorization/policies/conditions/sensitivity.condition.ts
```diff
@@ -139,3 +154,38 @@ export const withEffectiveSensitivity = <T extends object>(
...
+ * Build the `sensitivity <= access` SQL fragment for `resource`. Project rows
+ * carry the denormalized column directly; project-scoped child resources read
+ * it via a correlated subquery against `projects`. Add cases here as each
+ * domain ports to Postgres.
+ */
+const sensitivityRefForResource = (
+  resource: EnhancedResource<any>,
+  access: Sensitivity,
+): SQL => {
+  const accessLit = sql.raw(`'${access}'::"sensitivity"`);
```
bryanjnelson: "accessLit" is hard to reason...it's supposed to mean "accessLiteral".  What about "accessSql"?

src/components/authorization/policy/conditions/aggregate.condition.ts
```diff
@@ -64,6 +66,21 @@ export abstract class AggregateConditions<
     return `(${inner})`;
   }
 
+  asDrizzleCondition(params: AsDrizzleParams<TResourceStatic>): SQL {
+    const inner = this.conditions.map((c) => {
```
bryanjnelson: Prefer "cond" over "c".

src/components/project/project.drizzle.repository.ts
```diff
@@ -0,0 +1,828 @@
...
+
+/**
+ * Recompute the denormalized `projects.sensitivity` for translation projects
+ * from their engaged languages — max(language sensitivity) ?? High. Mirror of
+ * Gel's recalculateProjectSens triggers; called from the engagement/language
+ * drizzle repos whenever the inputs change (create/delete engagement,
+ * language sensitivity update). Internship projects are untouched — they
+ * read own_sensitivity.
+ */
+export const recomputeProjectSensitivity = async (
```
bryanjnelson, line 491: I'm curious as to why this is an app method here when we had it as a trigger in Gel?

bryanjnelson (changes requested):

src/components/authorization/policies/conditions/sensitivity.condition.ts
rdonigian: @bryanjnelson  This occurs in a few places, I don't mind the `Lit` suffix just because it's a bit more descriptive, how about even `accessLiteral` or `accessRaw`?

src/components/project/project.drizzle.repository.ts
rdonigian: A Gel trigger and a PG trigger aren't really the same feature. In Gel this was one declarative line per trigger, written in the schema language and type-checked with it. The PG equivalent is PL/pgSQL — a separate procedural language living in a migration file where no TS review or compiler looks — wired to ~5 events: engagement inserted, engagement soft-deleted, engagement's language changed, language sensitivity changed, language soft-deleted. Soft delete is what balloons the count: nothing is actually DELETEd in this schema, so an AFTER DELETE trigger would never fire — the "delete" paths are really UPDATE OF deleted_at, which is exactly the kind of subtlety that gets missed in trigger DDL. Since the repos are the only write path (and the cutover ETL copies the already-computed value from Neo4j), an app function gets the same guarantee in typed, testable code.

There is even more to this such as even if we wanted the trigger, we couldn't create it in this PR. A trigger is attached to the table it watches, engagements and languages, and those tables don't exist yet.

Lastly, this isn't a one-off exception, it's how the whole port treats Gel triggers. The pattern across the port: the database keeps what it can enforce declaratively (foreign keys, generated columns, constraints, the one backfill-critical trigger), and anything procedural moves into TypeScript where it's reviewed, typed, and testable.

src/components/authorization/policies/conditions/sensitivity.condition.ts
bryanjnelson: Yeah, either of those work for me!

src/components/project/project.drizzle.repository.ts
bryanjnelson: Great explanation; thanks!

I guess at some point after migration we could consider moving app logic into triggers/sprocs if it makes sense.  We wouldn't want to lose type-checking though.  This might not even be a thing...just throwing out the idea.  Regardless, that's post-mig.

bryanjnelson (approved):

PR #3816, opened 2026-07-08
[pg] Notifications → Postgres (Drizzle)

### LINKED ISSUE #3751: Notifications domain — Drizzle repo

rdonigian:
### Schema / DDL (migration 0012)
- `notification_type` pgEnum: `System | CommentViaMention`.
- `notifications` — single-table inheritance: the `type` discriminator maps to the registered strategy; subtype fields are nullable columns guarded by the `notifications_shape` CHECK (`System` ⇒ `message` NOT NULL + `comment_id` NULL; `CommentViaMention` ⇒ inverse).
- `creator_id` → `users(id) ON DELETE SET NULL`.
- `comment_id` is **FK-less plain text** — the comments table lands in a later Phase-6 recut; a `migration-todo` comment marks where the FK gets added.
- `notification_recipients` — composite PK `(notification_id, user_id)`, both FKs CASCADE; `read_at` nullable, so `unread` is computed per requesting user, never stored on the notification.
- Indexes: `created_at`, `creator_id`, `comment_id`, `notification_recipients.user_id`.

### What's added
- Strategy pattern extended for Drizzle: `saveForDrizzle` (input → row columns), `hydrateExtraForDrizzle` (row → subtype DTO fields), `recipientsForDrizzle` (recipient resolution) on the base class with defaults; both existing strategies override as needed. A `migration-todo` on the base notes the `*ForNeo4j`/`*ForGel` variants drop at Phase-7 cutover.
- Repo covers create (notification + recipient fan-out), list with unread filter/count keyed to the requesting user via `notification_recipients`, and mark-read.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,7 +85,7 @@ jobs:
           # full suite.
           - database: postgres
             test-args: >-
-              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|project-member|tool)\."
+              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|project-member|tool|notification)\."
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


bryanjnelson (approved):
Onward!

PR #3818, opened 2026-07-08
[pg] Migration-confidence suite + read-filter leak guard

rdonigian:
Two e2e suites that turn the migration's structural invariants into CI checks, plus the one migration those checks immediately flagged, and a CI tweak to the postgres test leg.

### postgres-schema.e2e
Runs against the fully-migrated ephemeral DB and asserts:
- every FK's leading column has an index (PG doesn't auto-index FKs)
- `_journal.json` idx values are contiguous and tags match the migration files on disk
- the GENERATED `projects.status` CASE agrees with the app's `stepToStatus()` mapping
- DB enum values ↔ DTO enum values for the enum pairs migrated so far

### auth-read-filter.e2e
Creates rows as admin, then asserts non-admin `list()` results don't leak rows the role's policy shouldn't grant. The engagement leg is PG-gated with a `migration-todo` until Engagement migrates.

### Migration 0013 — FK index backfill
- `locations_default_field_region_id_idx` + `locations_funding_account_id_idx` — the only two FK columns predating the index-every-FK standard, flagged by the invariant suite. Matching declarations added to the Drizzle schema.

### CI
- Both suites added to the postgres `--testPathPatterns`.
- Postgres E2E leg collapsed from 6 shards to 1: the migrated-domain subset is only a few minutes of tests, and six jobs were each paying the full ~20-min setup to run ~1.5 min of them. Neo4j job names/shards are untouched (required checks unaffected); postgres stays advisory.

### DIFF
--- a/test/postgres-schema.e2e-spec.ts
+++ b/test/postgres-schema.e2e-spec.ts
@@ -13,11 +13,11 @@ import { createTestApp, type TestApp } from './utility';
 // whole suite is skipped (the schema doesn't exist there).
 // migration-todo: drop the engine gate at Phase 7 cutover (always postgres).
 const isPostgres = process.env.DATABASE === 'postgres';
-const d = isPostgres ? describe : describe.skip;
+const describePg = isPostgres ? describe : describe.skip;
 
 const migrationsDir = resolve(process.cwd(), 'src/core/drizzle/migrations');
 
-d('Postgres schema invariants', () => {
+describePg('Postgres schema invariants', () => {
   let app: TestApp;
   let db: DrizzleDb;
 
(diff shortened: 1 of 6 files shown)


test/postgres-schema.e2e-spec.ts
```diff
@@ -0,0 +1,163 @@
...
+// These assertions introspect the live Postgres catalog + the on-disk drizzle
+// migrations, so they only make sense under DATABASE=postgres. Under neo4j the
+// whole suite is skipped (the schema doesn't exist there).
+// migration-todo: drop the engine gate at Phase 7 cutover (always postgres).
+const isPostgres = process.env.DATABASE === 'postgres';
+const d = isPostgres ? describe : describe.skip;
+
+const migrationsDir = resolve(process.cwd(), 'src/core/drizzle/migrations');
+
+d('Postgres schema invariants', () => {
```
bryanjnelson: Prefer `describe` over `d`.

bryanjnelson (changes requested):

bryanjnelson (approved):
🤩

PR #3820, opened 2026-07-09
[pg] FundingAccount department-id blocks + Budget domain (Drizzle)

### LINKED ISSUE #3741: Budget domain — Drizzle repo

rdonigian:
### Summary
Ports the Budget domain to Postgres and gives funding accounts their department-ID blocks, which together activate department-ID assignment under PG. Two commits: a Drizzle-interceptor retry fix, then the schema + repos + handler work.

### What's in scope
- `DrizzleTransactionalMutationsInterceptor` now honors `TransactionRetryInformer`: `markForRetry` was only wired to the Neo4j driver, so PG mutations never retried; the interceptor now retries up to 3× when a flagged error is found on the cause chain.
- Migration 0014: `funding_accounts.department_id_block_id` → `department_id_blocks` + FK index. The funding-account repository creates/maintains each account's block: range `[acct·10000+11, (acct+1)·10000−1]`, programs Momentum + Internship.
- Migration 0015: `budget_status` pgEnum; `budgets` (project FK CASCADE, status default Pending); `budget_records` (budget FK CASCADE, organization FK, fiscal_year, amount fields); FK indexes.
- `BudgetDrizzleRepository` + `BudgetRecordDrizzleRepository`, splitDb `postgres:` arms, and `Budget`/`BudgetRecord` cases in the member + sensitivity Drizzle auth conditions (both read sensitivity via the parent project).
- Removed the `databaseEngine === 'postgres'` early-returns from four handlers that were blocked on these tables: `create-project-default-budget`, `update-project-current-budget-status`, `sync-budget-records-to-funding-partners`, and `set-department-id`. The dept-ID JOIN chain (project → primary location → funding account → block) is now complete, so its CTE path and unique-violation retry run for real.
- Budget service: `deleteNode` → `delete(id)` rename, with matching Neo4j repo wrappers so both engines expose the same method.
- CI: `partnership` added to the postgres test pattern — its one failing test was budget-record creation, which now passes.

### Design decisions
- `budget_records` partial unique `(budget_id, organization_id, fiscal_year) WHERE deleted_at IS NULL` — DB backstop for the service's record-uniqueness check; soft-deleted records don't block recreation.
- `universal_template_file_id` stays plain text (no FK): the row is inserted before `createDefinedFile` creates the file node, so a real FK would fail the insert — same ordering rationale as `partnerships.mou_id`.
- FK indexes ship with the tables rather than in a later backfill migration.

### Validation
- type-check + lint clean.
- `partnership.e2e` fully green under `DATABASE=postgres` (13 pass / 1 pre-existing skip), including budget-record creation through the re-enabled handlers.
- `project.e2e` failures drop 10 → 5; the remaining five are boundaries of domains not yet on Postgres (Pin, Engagement/Language ×3, name collation).
- `postgres-schema.e2e` invariants green over the 0014/0015 DDL; the department-ID race-retry test passes.

### DIFF
--- a/src/core/drizzle/drizzle-transactional-mutations.interceptor.ts
+++ b/src/core/drizzle/drizzle-transactional-mutations.interceptor.ts
@@ -6,6 +6,7 @@ import {
 import { ConfigService } from '~/core/config';
 import { TransactionalMutationsInterceptor } from '~/core/database/abstract-transactional-mutations.interceptor';
 import { TransactionHooks } from '~/core/database/transaction-hooks';
+import { TransactionRetryInformer } from '~/core/database/transaction-retry.informer';
 import { DrizzleService } from './drizzle.service';
 
 @Injectable()
@@ -14,6 +15,7 @@ export class DrizzleTransactionalMutationsInterceptor extends TransactionalMutat
     txHooks: TransactionHooks,
     private readonly config: ConfigService,
     private readonly drizzle: DrizzleService,
+    private readonly retryInformer: TransactionRetryInformer,
   ) {
     super(txHooks);
   }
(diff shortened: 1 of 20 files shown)


src/core/drizzle/drizzle-transactional-mutations.interceptor.ts
```diff
@@ -26,6 +28,34 @@ export class DrizzleTransactionalMutationsInterceptor extends TransactionalMutat
...
+  private markedForRetry(error: unknown): boolean {
+    // The marked error is usually a cause of the thrown one (handlers wrap
+    // the db error in a ServerException) — walk the chain.
+    let e = error;
+    while (e instanceof Error) {
+      if (this.retryInformer.shouldRetry(e)) return true;
+      e = e.cause;
+    }
+    return false;
   }
```
bryanjnelson, line 60: This removes the never hit exception at the end:

```ts
protected async inTx<R>(fn: () => Promise<R>): Promise<R> {
  // Honor TransactionRetryInformer like the Neo4j driver does: handlers
  // (e.g. SetDepartmentId's unique-violation race) mark an error retryable
  // and expect the whole mutation to re-run. Without this loop,
  // markForRetry() is a no-op under postgres. Same semantics as Neo4j
  // retryable transactions: the mutation body may execute multiple times.
  let attemptsLeft = 3;
  while (true) {
    attemptsLeft--;
    try {
      return await this.drizzle.inTx(fn);
    } catch (error) {
      if (attemptsLeft > 0 && this.markedForRetry(error)) {
        continue;
      }
      throw error;
    }
  }
}```

bryanjnelson (changes requested):
Nice.

bryanjnelson (approved):
👍🏼

PR #3824, opened 2026-07-10
[pg] Close the auth-audit findings: readMany read filters + Partner/O…

rdonigian:
### [pg] Close the auth-audit findings: readMany read filters + Partner/Org/User condition arms

Single commit closing the confirmed findings from the Drizzle auth parity audit, plus a Neo4j 5 bug the new tests exposed. 

### What's in scope

- `Project` and `ProjectMember` `readMany` now apply the policy read filter, matching their `list` methods — previously member-gated roles could resolve non-member rows by id under `DATABASE=postgres`.
- `MemberCondition` gains Drizzle arms for resources that aren't project-scoped rows: Partner (member of any project connected via a partnership), Organization (same chain extended a hop), and User/Unavailability (member of any project). These previously threw "not configured".
- `SensitivityCondition` gains Partner + Organization arms reading their own denormalized columns.
- `MemberCondition.asCypherCondition` fix: the users list binds no `project` variable, and Neo4j 5 rejects pattern expressions that introduce one (42N29) — so member-gated roles listing users got a 500. User/Unavailability now use an anonymous start node; project-scoped resources keep the bound correlation.
- New `test/auth-read-filter.e2e-spec.ts` legs: partner/org hidden from non-members, sensitivity gating, self-only users list, project-by-id denial, and a membership flip exercising all four surfaces.

### Design decisions

- The Cypher fix restores Neo4j 4.x's "member of ANY project" semantics rather than keeping the accidental fail-closed error — it's what the policy always declared, and the Drizzle arm + tests match it.
- Every Drizzle member arm requires `inactive_at is null`, excluding replaced memberships that Neo4j's `active: true` rel still honors — deliberate tightening, matches membership-scope semantics.
- The Partner/Org arms join project liveness (`projects.deleted_at is null`) so members of a soft-deleted project don't retain partner/org visibility — Neo4j gets this via label rewrites; PG has to correlate it explicitly.
- Partner/Org sensitivity is statically High until the language port wires the real derivation — fail-closed interim, so sensitivity-gated roles see nothing rather than everything.
- The readMany filter runs over a plain id-select and survivors hydrate normally, because condition SQL references literal table names.

### Validation

- Postgres e2e: auth-read-filter, partner, organization, project-member — 4 suites, 28 passed, 0 failures.
- Neo4j e2e: auth-read-filter 8/8 (exercises the Cypher fix), user 12/12.

### DIFF
--- a/src/components/authorization/policies/conditions/member.condition.ts
+++ b/src/components/authorization/policies/conditions/member.condition.ts
@@ -7,6 +7,7 @@ import { type EnhancedResource, type ResourceShape, type Role } from '~/common';
 import { matchProjectScopedRoles } from '~/core/neo4j/query';
 import { rolesForScope, type ScopedRole, splitScope } from '../../dto/role.dto';
 import {
+  type AsCypherParams,
   type AsDrizzleParams,
   type AsEdgeQLParams,
   type Condition,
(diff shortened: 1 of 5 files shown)


bryanjnelson (approved):
🚀

PR #3828, opened 2026-07-14
Fix swapped UserUpdatedHook args killing director-removal restrictions

rdonigian:
### UserUpdatedHook arg swap

- `UserUpdatedHook`'s constructor is `(updated, previous, input)`, but `user.service.ts` passed `(user, updated, input)` — every handler saw the two states reversed since #3634 (Feb 9).
- The only consumers are the region/zone director-removal guards, whose "role was removed" check therefore computed "role was added": removing `RegionalDirector` / `FieldOperationsDirector` from an acting director sailed through, leaving regions/zones pointing at a director without the role.
- One-line arg swap + regression e2e legs on both handlers — verified the legs fail without the fix and pass with it, under both engines.

### e2e CI worker pin (completes #3826)

- #3826 merged without its final review commit, so the per-file Neo4j wipe went live while Jest still runs spec files in parallel within a shard — a finishing file's wipe deletes the shared graph under a concurrently-running one. First hit: #3827's neo4j shard 3 (webhooks.e2e rxjs delivery timeouts).
- `--maxWorkers=1` on the e2e CI invocation makes the wipe's serial precondition explicit, and `NEO4J_TEST_WIPE` tightens to an exact `'true'` check.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,7 +110,12 @@ jobs:
         run: yarn start -- --gen-schema && yarn gql-tada generate output
 
       - name: E2E Tests
-        run: yarn test:e2e --shard=${{ matrix.shard }}/${{ matrix.total || 6 }} --reporters=github-actions ${{ matrix.test-args }}
+        # --maxWorkers=1: the per-file Neo4j wipe requires spec files to run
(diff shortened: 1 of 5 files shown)


bryanjnelson (approved):
:shipit:

PR #3831, opened 2026-07-16
[pg] Auth user-liveness cluster — deleted users lose login, sessions, and password reset

rdonigian:
### Summary 

When you delete a user in Cord, they shouldn't be able to do anything anymore. On Neo4j that
happens for free: deleting a user relabels its node, so every query that looks up "a User" just
stops finding them — the database's shape enforces it. Postgres doesn't work that way: a deleted
user is just a row with a `deleted_at` timestamp, and a query only excludes them if it explicitly
says `WHERE deleted_at IS NULL`. The auth queries were written without that clause, so under
Postgres, deleting a user quietly did almost nothing:

- They could **still log in** with their password.
- If already logged in, their **session kept working forever**.
- They could run "forgot password" and **reset their way back into the account**.
- Internal role lookups still treated them as active.

None of this is live today — Neo4j is still primary — but it all becomes real at the flip, which
is why the audit ranked this the top pre-cutover fix. This PR adds the missing "and they're not
deleted" condition to every place auth looks up a user, and adds one behavior neither engine
really had: deleting a user now revokes all their sessions immediately, logging them out
everywhere. Tests delete a user and then try to log in / reuse their session / reset their
password — they fail on the old code and pass with the fix, on both databases.

### What's added

- **Liveness predicates** (`isNull(user.deletedAt)`) on: `getInfoForLogin` (deleted users could
  log in), `connectSessionToUser`, `findPasswordResetToken` + `updatePasswordViaEmailToken`
  backstop (reset-flow account takeover), `rolesForUser` (stale roles into
  sessionForUser/asUser/impersonation), and `resumeSession` (deleted users kept live sessions;
  now degrades to an anonymous session rather than erroring).
- **`doesEmailAddressExist` liveness** (found by adversarial review 7/16): Neo4j's check matches
  the `:EmailAddress` label, which the delete relabels away — so `forgotPassword` silently skips
  deleted accounts on Neo4j but was sending a real reset email + inserting a dead token row under
  PG. Same void response either way = no deletion oracle. `savePasswordResetToken` stays
  liveness-blind deliberately — unreachable for deleted emails behind this gate (commented).
- **Session revocation on delete**: `UserService.delete` fires `UserDeletedHook` →
  `DeletedUserLogsThemOutHandler` calls `auth.logoutByUser(id)`. Runs **inside the delete's
  transaction** deliberately — it's a row write, so a rollback of the delete rolls back the
  revocation (the effects-after-commit rule applies to emails, not DB writes).

### Intentional decisions (review attention here)

- Not a mono port — original work against the audit findings; no mono counterpart to diff.
- `resumeSession` for a deleted user **degrades to anonymous** instead of throwing — mirrors
  Neo4j's incidental behavior.
- The `doesEmailAddressExist` fix reverses the audit ledger's earlier "leave blind = parity"
  note, which was wrong (contradicted by the same ledger's U5 reasoning); ledger corrected.
- Handler lives in `core/authentication/handlers/` beside its `Disabling…` twin — the effect is
  auth-domain, the trigger is user-domain.

### Validation performed

- **Tested**: `authentication` + `user` e2e — **17/17 on BOTH engines**, re-run after the develop
  rebase (#3825 + #3827 base) and again after the review fix. New legs verified
  **red-without-fix on PG** (fix reverted → legs fail; reapplied → green).
- **Tested**: `yarn type-check` + `yarn eslint --max-warnings 0` on touched files, clean.
- **Reviewed**: multi-lens adversarial review (7/16) — 1 finding raised, 1 confirmed and fixed,
  0 refuted-but-shipped.
- **By reading, not testing**: Neo4j behavior unchanged — the Neo4j repository file is untouched;
  label-rewrite parity claims verified against `authentication.repository.ts:266-273`.

### DIFF
--- a/src/components/user/hooks/user-deleted.hook.ts
+++ b/src/components/user/hooks/user-deleted.hook.ts
@@ -1,0 +1,6 @@
+import { type ID } from '~/common';
+
+/** Fired after a user is (soft-)deleted, in the same transaction. */
+export class UserDeletedHook {
+  constructor(readonly id: ID<'User'>) {}
+}
--- a/src/components/user/user.service.ts
+++ b/src/components/user/user.service.ts
@@ -41,6 +41,7 @@ import {
   type EducationListInput,
   type SecuredEducationList,
 } from './education/dto';
+import { UserDeletedHook } from './hooks/user-deleted.hook';
 import { UserUpdatedHook } from './hooks/user-updated.hook';
 import { KnownLanguageRepository } from './known-language.repository';
 import { UnavailabilityService } from './unavailability';
@@ -140,6 +141,9 @@ export class UserService {
   async delete(id: ID): Promise<void> {
     const object = await this.readOne(id);
     await this.userRepo.delete(id, object);
+    // Same-transaction side effects (e.g. session revocation — a deleted
+    // user must not keep live sessions).
+    await this.hooks.run(new UserDeletedHook(object.id));
   }
 
   async list(input: UserListInput): Promise<UserListOutput> {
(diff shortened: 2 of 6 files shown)


bryanjnelson (approved):
✅ 🔒

PR #3833, opened 2026-07-21
[pg] Leaves → Postgres (Drizzle): Pins, KnownLanguages, Comments, Post — migrations 0022–0025

### LINKED ISSUE #3749: Comments domain — Drizzle repo

### LINKED ISSUE #3750: Post domain — Drizzle repo

### LINKED ISSUE #3752: Pin domain — Drizzle repo

rdonigian:
### Summary 
This moves the last four small features off the legacy graph database onto Postgres. They're independent of each other:

- **Pins** — lets a user bookmark any record (a project, language, partner, etc.). Each user sees only their own pins. On the Postgres path the per-user "is this pinned by me?" flag had been shipping hardcoded to `false`; this PR makes it real everywhere it's shown and makes "filter to just my pins" work.
- **Known languages** — the languages a user knows, with a proficiency level. These reads/writes now actually run against Postgres — previously, under the Postgres engine, they still fell through to the legacy query path.
- **Comments** — threaded comments that can be attached to many different record types (users, languages, partners, projects, engagements, progress reports).
- **Posts** — short notes attached to a record, each with a visibility setting controlling who can see it.

Technical context: final wave of the Neo4j→Postgres port. It's stacked on the report-cluster PR — **base is `report-recut`** and will retarget to `develop` automatically once that merges (chain: develop → product #3829 → report → this). Migrations 0022–0025.

### What's added

- **Pins (migration 0022)** — new `pins` table plus `PinDrizzleRepository` (is-pinned / add / remove, keyed to the current user). Two reusable pieces: `pinnedByRequester` (batch-populates the flag during hydration) and `pinnedFilter` (an EXISTS / NOT-EXISTS predicate for list filtering). The four consumer repositories — partner, user, project, language — now populate the real `pinned` value in their `readMany`/`list` and honor `filter.pinned` (they previously returned the `false` stub). Anonymous requesters resolve correctly: `pinned:true` matches nothing, `pinned:false` matches everything.
- **KnownLanguages (0023)** — new `known_languages` table + `KnownLanguageDrizzleRepository` (idempotent create via `ON CONFLICT DO NOTHING`, delete, list), routed through `splitDb` on the user module. New `language_proficiency` pgEnum.
- **Comments (0024)** — `comment_threads` + `comments` tables (hard delete, cascade). Polymorphic parent resolution via a new `resolveResourceBaseNode` helper that probes the migrated parent tables. Installs the `notifications.comment_id` foreign key (the column had landed FK-less in migration 0012).
- **Post (0025)** — `posts` table with `post_type` and a 5-value `post_shareability` pgEnum. Membership-based shareability is enforced in-repo against `project_members`. Sort map covers the type's own scalar columns.

### Notes for the reviewer (intentional choices)

- **Comment audit-event firing is not wired here.** The comment and post services are unchanged from their current form; the `ResourceMutatedHook` audit instrumentation lands with the audit domain, not this PR. Called out so it isn't flagged as missing.
- **Post `ProjectTeam` shareability value is kept verbatim** to preserve current (Neo4j) behavior. Collapsing it to `Membership` is a tagged post-migration TODO, out of scope here.
- **Every FK column has a b-tree index, added directly in the domain migration** (`known_languages_language_id_idx`, `comment_threads_creator_id_idx`, `comments_creator_id_idx`, `posts_creator_id_idx`). Pins needs none — `user_id` leads the composite PK and `resource_id` is intentionally FK-less (polymorphic).
- **Comment module wiring:** only `CommentRepository` is routed through `splitDb`; `CommentThreadRepository` stays a plain provider and the Drizzle comment repo injects the thread repo concretely — this avoids a `moduleRef`/`forwardRef` cycle (documented in the module).

### Schema / DDL

- `pins` — `(user_id → users ON DELETE CASCADE, resource_id [FK-less, polymorphic], created_at)`, composite PK `(user_id, resource_id)`.
- `known_languages` — `(user_id, language_id → languages, proficiency language_proficiency, created_at)`, PK over `(user_id, language_id, proficiency)`. New enum `language_proficiency`.
- `comment_threads` — `(id, creator_id → users, parent [polymorphic], created_at)`.
- `comments` — `(id, thread_id → comment_threads ON DELETE CASCADE, creator_id → users, body, created_at, modified_at)`, hard delete.
- `notifications.comment_id` — FK constraint `notifications_comment_id_fk → comments(id) ON DELETE CASCADE` added here (column pre-existed FK-less from 0012).
- `posts` — `(id, creator_id → users, parent [polymorphic], type post_type, shareability post_shareability, body, created_at, modified_at)`. New enums `post_type`, `post_shareability` (5 values incl. `ProjectTeam`).
- Journal indices 0022–0025, contiguous.

### Validation performed

- **Fresh-DB migration apply 0000–0025**: clean — verified the 5 new tables, 3 new enums, all four folded-in FK indexes, and the `notifications_comment_id_fk` constraint.
- **`yarn type-check` / `yarn lint`**: clean.
- **Postgres e2e** (tested): pin + comment + post + known-language + postgres-schema + user = **32/32**; partner + language green; project **50/51** (the sole failure is the pre-existing name-sort collation baseline, untouched by this PR — the pin wiring in this PR actually fixed the previously-failing Pin legs).
- **Reference-engine e2e** (Neo4j oracle, tested): pin + comment + post + known-language = **6/6** parity. These four specs are new — there was no prior e2e for either engine.

### Reviewer cross-checks

- `resolveResourceBaseNode` probe list = the union of Commentable + Postable parents (User/Language/Partner/Project/Engagement/ProgressReport); each candidate filters `deleted_at IS NULL`. The services' `verifyImplements` rejects a valid-but-wrong parent (e.g. a User handed to a Post).
- Enum↔DTO parity: `language_proficiency`, `post_type`, `post_shareability` values match their DTO enums exactly (the `postgres-schema` e2e asserts this).
- Schema↔SQL: the `schema/index.ts` definitions agree with the raw `0022`–`0025` `.sql` (columns, enums, indexes, the deferred `notifications.comment_id` reference).
- `notifications.comment_id`: schema carries the thunked cascade reference; the constraint itself is created in 0024.

### Explicitly out of scope

- Comment/Post `ResourceMutatedHook` audit firing → audit domain.
- The CommentViaMention notification path exists but stays unexercised — mention `extract()` is still a `[]` stub on both engines (unchanged here).
- Post `ProjectTeam` → `Membership` collapse — tagged post-migration TODO.
- Name-sort collation ordering (project) — standing cutover decision, not this PR.

### DIFF
--- a/test/comment.e2e-spec.ts
+++ b/test/comment.e2e-spec.ts
@@ -109,7 +109,13 @@ describe('Comment e2e', () => {
       });
       const language = await createLanguage(a);
       const { createEng } = await a.graphql.mutate(CreateLangEngForCommentDoc, {
-        input: { project: project.id, language: language.id },
+        input: {
+          project: project.id,
+          language: language.id,
+          // Match the project MOU window so progress reports land in-window.
+          startDateOverride: CalendarDate.local(2023, 1, 1).toISO(),
+          endDateOverride: CalendarDate.local(2024, 1, 1).toISO(),
+        },
       });
       const reportId = createEng.engagement.progressReports.items[0]!.id;
       expect(reportId).toBeTruthy();
--- a/test/comment.e2e-spec.ts
+++ b/test/comment.e2e-spec.ts
@@ -1,0 +1,213 @@
+import { beforeAll, describe, expect, it } from '@jest/globals';
+import { CalendarDate } from '~/common';
+import { graphql } from '~/graphql';
+import {
+  createLanguage,
+  createProject,
+  createSession,
+  createTestApp,
+  registerUser,
+  runAsAdmin,
+  type TestApp,
+} from './utility';
+
+/** Minimal block-editor document the RichText (JSONObject) scalar accepts. */
+const doc = (text: string) => ({
+  version: '1',
+  time: 1,
+  blocks: [{ id: text, type: 'paragraph', data: { text } }],
+});
+
+interface RichDoc {
+  blocks: Array<{ data: { text: string } }>;
+}
+const textOf = (value: unknown) => (value as RichDoc).blocks[0]!.data.text;
+
+describe('Comment e2e', () => {
+  let app: TestApp;
+
+  beforeAll(async () => {
+    app = await createTestApp();
+    await createSession(app);
+    await registerUser(app);
+  });
+
+  it('creates, lists, updates, and deletes comments on a commentable', async () => {
+    await runAsAdmin(app, async (a) => {
+      const language = await createLanguage(a);
+
+      // First comment creates a new thread on the language.
+      const { createComment } = await a.graphql.mutate(CreateCommentDoc, {
+        input: { resource: language.id, body: doc('First comment') },
+      });
+      const first = createComment.comment;
+      const threadId = createComment.commentThread.id;
+      expect(first.id).toBeTruthy();
+      expect(textOf(first.body.value)).toBe('First comment');
+
+      // The thread is listed on the parent, hydrated with its first comment.
+      const listed = await a.graphql.query(CommentThreadsDoc, {
+        resource: language.id,
+      });
+      expect(listed.commentThreads.total).toBe(1);
+      expect(listed.commentThreads.items[0]!.id).toBe(threadId);
+      expect(listed.commentThreads.items[0]!.firstComment.id).toBe(first.id);
+
+      // Second comment attaches to the existing thread.
+      const { createComment: secondCreated } = await a.graphql.mutate(
+        CreateCommentDoc,
+        {
+          input: {
+            resource: language.id,
+            thread: threadId,
+            body: doc('Second comment'),
+          },
+        },
+      );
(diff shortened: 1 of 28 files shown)


test/comment.e2e-spec.ts
```diff
@@ -109,7 +109,13 @@ describe('Comment e2e', () => {
       });
       const language = await createLanguage(a);
       const { createEng } = await a.graphql.mutate(CreateLangEngForCommentDoc, {
-        input: { project: project.id, language: language.id },
+        input: {
```
bryanjnelson, line 112: Nit:  Create a const for the mouStart once and then base the other 3 off of it.

bryanjnelson (approved):
End of domains!  Nice work!  🙌🏼

test/comment.e2e-spec.ts
rdonigian: I'll fold into next PR.

PR #3834, opened 2026-07-22
[pg] Audit log → Postgres (Drizzle): resource-mutation trail + role_at_time — migration 0026

### LINKED ISSUE #3753: Audit log — resource_mutations infrastructure
Audit log — resource_mutations infrastructure — create table (bigint identity PK, resource_type, resource_id, actor_id, action, changes jsonb, at), shared hook diff utility, history resolver on Resource base type, wire Phase 6 domains

### LINKED ISSUE #3754: Audit log — back-instrument
Audit log — back-instrument Phases 2–5 — add @OnHook audit handlers for all previously migrated domains in one PR

rdonigian:
### Summary 
Introduces Cord's **audit log** — an append-only record of who changed what. Every user-driven create/update/delete now writes one permanent entry — the actor, the action, which fields changed, when, and **in what role** — inside the same transaction as the change (so a rolled-back mutation leaves no orphan log row). Every record gains a `history` field exposing its own change trail.

This lands migration `0026`. It's stacked on the report + leaves work and retargets to `develop` as those merge.

### What's added
- **Audit component** — `resource_mutations` table (append-only) + `mutation_action` enum; a repository, an `AuditService`, an `@OnHook` writer, and resolvers.
- **`history` on every resource** — resolved on the base `Resource` interface (via the concrete type name), so all object types inherit it with no per-domain wiring.
- **`role_at_time`** — snapshots the actor's roles at the instant of the mutation, captured at write time from the session.
- **24 services instrumented** — each fires `ResourceMutatedHook(type, id, action[, changes])` on create/update/delete.

### Notes for the reviewer (intentional choices)
- **`role_at_time` is stored as `text[]`, not the `role` enum** — a decoupled historical snapshot. The log is append-only, so a value must be captured at write time (no backfill), and storing plain text keeps historical rows immune to future role-enum changes. Exposed in GraphQL as `[String!]!` for the same reason.
- **The audit writer is Postgres-only** — under the legacy engine it's a no-op (no partial history mid-transition), tagged `migration-todo(cutover-cleanup)`. The log begins capturing at cutover.
- **`user.delete` fires both hooks** — the existing session-revocation-on-delete side effect is preserved; the audit event is added alongside it.
- **`progress-report-media`** is instrumented at the service without altering its repository call — its Postgres repository is a later wave; the audit hook is engine-agnostic and lands now.
- **Some resources are intentionally not tracked** — system-generated, sync-only, and relationship-toggle writes don't record.

### Schema / DDL
- `mutation_action` enum: `Create | Update | Delete`.
- `resource_mutations`:
  - `id` bigserial PK · `resource_type` text · `resource_id` text (polymorphic, FK-less — spans every resource) · `action` mutation_action
  - `actor_id` text → `users(id)` **ON DELETE SET NULL** (history outlives the actor)
  - `role_at_time` text[] NOT NULL DEFAULT `'{}'`
  - `changes` jsonb (diffed fields, updates only) · `at` timestamptz DEFAULT now()
  - indexes: `(resource_type, resource_id, at)`, `(actor_id)`

### Validation performed
- `yarn type-check` / `yarn lint`: clean.
- Fresh-DB apply `0000–0026`: clean — verified the table, enum, `role_at_time text[]`, `actor_id` set-null FK, and both indexes.
- **Postgres e2e (audit-log): 7/7** — history records create+update across partnership/project/organization/language-engagement/product; `history` exposed via the Resource interface; actor resolved; `changes` diff captured; `role_at_time` carries the actor's roles; and the audit row rolls back with a failed transaction.
- Note: the generated GraphQL schema types must be regenerated (`--gen-schema` + gql-tada) since `history`/`role_at_time` are new fields — done in CI's setup step.

### Reviewer cross-checks
- `history` uses the concrete GraphQL type name (not `__typename`) to match stored rows.
- `role_at_time` is captured in `AuditService.record` from the session, alongside the actor.
- The write is inside the mutation's transaction (atomicity proven by the rollback test).
- FK column `actor_id` has a full b-tree index.

### Explicitly out of scope
- ProgressReportMedia's Postgres repository (separate wave) — only its audit hook lands here.
- Scoped-per-resource roles in `role_at_time` (v1 captures global roles; scoped is a possible follow-up).
- Removing the Postgres-only guard (a cutover-cleanup item).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
             shard: 1
             total: 1
             test-args: >-
-              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post)\."
+              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log)\."
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -1,6 +1,7 @@
 import { Module } from '@nestjs/common';
 import process from 'node:process';
 import { AdminModule } from './components/admin/admin.module';
+import { AuditModule } from './components/audit/audit.module';
 import { AuthorizationModule } from './components/authorization/authorization.module';
 import { BudgetModule } from './components/budget/budget.module';
 import { CeremonyModule } from './components/ceremony/ceremony.module';
@@ -58,6 +59,7 @@ if (process.env.NODE_ENV !== 'production') {
     AuthorizationModule,
     BudgetModule,
     CeremonyModule,
+    AuditModule,
     CommentModule,
     EthnoArtModule,
     FileModule,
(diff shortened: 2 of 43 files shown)


bryanjnelson (dismissed):
🤩 🙌🏼

bryanjnelson (approved):
Most excellent addition!  🙌🏼

PR #3836, opened 2026-07-23
[pg] Search → Postgres (Drizzle): UNION-ALL ILIKE global search

### LINKED ISSUE #3758: Search domain — Drizzle repo

rdonigian:
### Summary 

Global search is the box that searches everything at once — projects, people, languages, organizations, files. Today it runs on the old database, which happens to store every piece of text in one uniform place, so a single index covers everything.

Postgres has no such place. Text lives in whatever column it belongs to. So this rebuilds global search as one query per searchable table, combined into a single result — each table contributing the columns worth matching against.

What a user sees is unchanged: same box, same results, same permission rules about who can see what.

Sits in the Postgres migration series; nothing here touches the old database's behavior. Branch `search-recut`, based on `develop`.

### What's added

- A Postgres search implementation combining per-table case-insensitive substring matches into one result set, routed through the existing engine switch so nothing above the repository changes.
- Each matched column reports which DTO field it came from, which is what lets the service apply per-field read permissions exactly as before.
- Searchable types with no meaningful text column — Partner, the Product family, the PeriodicReport family — still match on an exact id, preserving that path from the old implementation.
- A synthetic base-node result shape, so the database-agnostic search service is untouched.
- Seven end-to-end tests covering substring matching, type restriction, exact-id lookup, the three id-only types, and the empty-result case. They run against both engines unmodified.
- Search added to the Postgres CI pattern.

### No schema changes

Search adds no tables, columns, indexes or enums, so there is no migration in this PR. Worth stating explicitly since every sibling PR in this series has one.

### Intentional behavior differences from the old implementation

Both are documented in the repository file; these are the two things to weigh.

- **Result ranking.** The old database returns hits pre-ranked by relevance, so its 100-row ceiling takes the best 100. There is no equivalent score for a combined per-table query, so results come back newest-first with id as a tie-break. This only diverges once a query matches more than 100 rows — below that, both return the same set. Deterministic ordering was chosen over an arbitrary subset, which is what an unordered limit produces. Adopting full-text vectors would close the gap if search quality becomes a complaint.
- **Text search coverage.** The old implementation matched any stored value, including on types with no human-readable name. Those types now match by exact id only. Partner remains reachable by text through its organization's name, which the service already resolves.

### Validation performed

- Search end-to-end on **Postgres: 7 passed**. Same spec on **Neo4j: 7 passed** — same tests, both engines, no engine-specific branches in the test.
- Schema invariants spec on Postgres: passed alongside search, 21 tests total.
- Full unit suite: 190 passed, 3 todo.
- Rebase verified by content, not by commit identity: pre- and post-rebase patches over `src` and `test` are byte-identical, so the only change is the CI list resolution.

### Reviewer cross-checks

- Every publicly searchable type has a branch — 21 types, worth confirming against the searchable map in the search DTO rather than trusting the list.
- Project and product subtype filters use the database's own enum values, which differ from the GraphQL type names (`MomentumTranslation` versus `MomentumTranslationProject`, `DirectScripture` versus `DirectScriptureProduct`). Both mappings are in the repository file.
- The ethnologue branch depends on the service adding that type when a language search is requested. If that ever changes, the branch silently stops contributing.
- Table and column names are interpolated as raw SQL. All are hardcoded in this file, never user input — the user's query itself is parameterized and its wildcards escaped.
- One table in the set has no soft-delete column and is handled explicitly; the rest filter it.

### Explicitly out of scope

- Relevance ranking via full-text vectors.
- Text search over the id-only types.
- Any change to the old database's search path, the search service, or the GraphQL contract.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
             shard: 1
             total: 1
             test-args: >-
-              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log|progress-report-media)\."
+              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log|progress-report-media|search)\."
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
--- a/src/components/search/search.drizzle.repository.ts
+++ b/src/components/search/search.drizzle.repository.ts
@@ -1,0 +1,313 @@
+import { Injectable } from '@nestjs/common';
+import { sql, type SQL } from 'drizzle-orm';
+import { DateTime } from 'luxon';
+import { type Merge } from 'type-fest';
+import { type ID } from '~/common';
+import { DrizzleService } from '~/core/drizzle';
+import { escapeLikePattern } from '~/core/drizzle/like';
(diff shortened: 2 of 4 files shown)


bryanjnelson (approved):
🏁 🙌🏼

PR #3838, opened 2026-07-30
[pg] Audit Discovery - Live-query invalidation for the Postgres repositories

rdonigian:
### Summary

Some pages in the app update themselves. You have a project open, someone else changes it, and your view refreshes without you doing anything. That works because the server announces "this record changed, re-fetch it" every time something is written.

Under Postgres, the server had stopped announcing.

The two older database layers both announce automatically from a shared base class, so every domain got it for free. The Postgres layer's equivalent base class never did. Twelve of the app's queries depend on those announcements, and they are the detail pages of nearly every domain — project overview, engagement, partner, language, location, user, budget, product, both field region/zone pages, and both progress report views. After the cutover, editing anything on those pages would have left the open page showing stale data until the user manually refreshed.

There were 18 hand-written announcements scattered across about 13 of 55 services, and those still fire. But they mostly announce a *parent* record rather than the one that changed, so any domain that hadn't hand-rolled its own call lost the behavior entirely.

This restores it, and adds a test that stops it from silently happening again.

### What's changed

- **The shared base** (`DrizzleDtoRepository`) now announces changes on `updateColumns()` and `softDelete()`. `liveQueryStore` is property-injected rather than added to the constructor, so no subclass has to thread it through `super()` — the same pattern both other bases already use. That covers 23 of the 42 Postgres repositories.
- **`product`** and **`prompt-variant-response`** hand-roll their own writes and never reach those helpers, so they announce for themselves.
- **A structural guard** (unit test, no database) enumerates every Postgres repository that writes and fails when one neither routes through the base helpers nor announces for itself, unless it's listed with a written reason.

Two lines of actual behavior change in the base; the rest is the two repositories, tests, and documentation.

### How the remaining 19 repositories were assessed

The standard applied was **parity with the existing engines, not perfection** — a gap that exists identically on Neo4j is a pre-existing product issue, not something the migration introduced. Each write path was compared against its Neo4j counterpart, and only three came back as real gaps:

- `product` — Neo4j announces on all three property-update methods (via `db.updateProperties`, which announces at `database.service.ts:340`) and on delete. **Fixed.**
- `prompt-variant-response` — Neo4j announces on delete only. **Fixed.**
- `file` — Neo4j announces on update (`file.repository.ts:517`) and delete (`:569`). **Not fixed here**; see follow-up.

The other 16 need nothing. Six have Neo4j counterparts that don't announce either (`pin`, `product-progress`, `progress-summary`, `comment-thread`, `known-language`, `project-workflow`), three are create-only or bootstrap paths where nothing can be subscribed yet (`notification`, `system-agent`, `admin`), and the rest have no write path at all.

### Two details worth a reviewer's attention

**Products announce under their concrete subtype, not the interface.** The store keys on `${resource.name}:${id}`, and the Neo4j arm passes `DirectScriptureProduct` / `DerivativeScriptureProduct` / `OtherProduct`. A generic `Product:` key would emit something nothing subscribes to — an announcement that exists and does nothing. That failure mode is worse than no announcement at all, because the call sits right there in the diff and reads as correct. There's an assertion pinning the concrete key *and* asserting the interface key is absent.

**Announcing from inside the mutation transaction is deliberate and safe.** Both other bases announce before their write, which looks wrong under Postgres' real transactions — the obvious worry being a live query re-executing against uncommitted state, or announcing a change that then rolls back. Neither happens: `LiveQueryStoreImpl.doInvalidate` defers the work to `TransactionHooks.afterCommit`, and `DrizzleTransactionalMutationsInterceptor` runs those hooks via the `runAndClear()` it inherits. No new deferral machinery was needed; the plumbing existed and was simply unused from this base. The reasoning is in the class docblock so the placement doesn't get "corrected" later.

Also deliberate: the `updateColumns()` call sits **after** the existing no-op guard. An update whose change set came out empty must not wake every live query watching that resource.

### The structural guard, and why it's here

This class of bug was invisible for as long as it was because the symptom — a detail page that quietly stops refreshing — shows up in no functional test and in no code review. Without a guard, the next repository added silently reintroduces it.

Three assertions, and the second two matter as much as the first:

- No unexplained offenders.
- **No stale exemptions** — an exemption for a repository that now announces is misleading documentation, so removing it is mandatory.
- A count floor, so a broken directory walk can't make the whole suite pass vacuously.

### Validation performed

Verified by running, including the negative cases — a guard that has never been observed failing isn't a guard:

- Live-query e2e **4/4** under Postgres. Removing the two base lines makes 2 of 3 base cases fail.
- Structural guard **3/3**. Removing one exemption makes it fail and name the offending file.
- The update test carries a control assertion that the row actually changed, so a zero-announcement result can only mean the announcement is missing — not that the mutation silently didn't run.
- No regressions under Postgres: product 14/15, periodic-report + organization 16/22 (all skips pre-existing).
- Neo4j leg: live-query + product **18/19** (1 pre-existing skip).
- Full unit project **193/196** (3 todo), run the way CI runs it.
- `yarn type-check` and `yarn lint` both exit 0.
- **CI wiring checked on both sides:** the e2e spec is added to the Postgres test-path pattern, which enumerates specs by name — a new file would otherwise never run. The guard needs no wiring; the `Unit` job runs `yarn test` unfiltered.

### Reviewer cross-checks

- That property injection on an abstract base resolves for every subclass. It does, and the spec fails loudly if it doesn't, but it's the least conventional line in the diff.
- That the identifier shape `[resource, id]` matches what the other two bases emit, so keys collide correctly across engines instead of forming parallel namespaces.
- That the `updateColumns()` announcement is after, not before, the empty-change-set early return.
- Whether any of the 23 base-routed repositories has a *second* write path that bypasses the helpers.
- The parity calls in the guard's exemption list — each names the Neo4j file it was checked against, and each is worth a second opinion.

### Follow-up, not in this PR

- **`file` is the one real defect left.** Its Neo4j arm announces on update and delete; the Postgres arm doesn't. It's held out solely because that file is modified by a branch currently in flight, so editing it here would conflict. It's recorded in the guard's exemption list as the single entry flagged a genuine defect, and should land in that branch or immediately after it merges.
- **The six parity repositories are a real gap on all three engines** — live queries don't refresh there today either. Worth a ticket, but it isn't a migration regression and isn't cutover-blocking.
- **No announcement on create**, on any engine. List-level queries go stale when a record is created. Pre-existing; not introduced here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
             shard: 1
             total: 1
             test-args: >-
-              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log|progress-report-media|search)\."
+              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log|progress-report-media|search|live-query-invalidation)\."
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
--- a/src/components/product/product.drizzle.repository.ts
+++ b/src/components/product/product.drizzle.repository.ts
@@ -35,7 +35,9 @@ import {
   productCompletionDescriptions,
   products,
 } from '~/core/drizzle/schema';
+import { LiveQueryStore } from '~/core/live-query';
 import { type BaseNode } from '~/core/neo4j/results';
+import { type ResourceLike } from '~/core/resources';
 import { type ScopedRole } from '../authorization/dto/role.dto';
 import { requesterScopeByProject } from '../project/project-member/membership-scope';
 import {
@@ -115,6 +117,7 @@ export class ProductDrizzleRepository {
   constructor(
     private readonly drizzle: DrizzleService,
     private readonly identity: Identity,
+    private readonly liveQueryStore: LiveQueryStore,
   ) {}
 
   // Transaction-aware client — see DrizzleDtoRepository.db for why a getter.
@@ -354,10 +357,15 @@ export class ProductDrizzleRepository {
     }
   }
 
+  // Each of these invalidates with its CONCRETE subtype, matching what the Neo4j
+  // arm passes to `db.updateProperties({ type })` — the store keys on
+  // `${resource.name}:${id}`, so invalidating a generic `Product` here would
+  // emit a key nothing subscribes to and read as a silent no-op.
   async updateProperties(
     object: DirectScriptureProduct,
     changes: DbChanges<DirectScriptureProduct>,
   ) {
+    this.liveQueryStore.invalidate([DirectScriptureProduct, object.id]);
     return await this.applyChanges(object, changes);
   }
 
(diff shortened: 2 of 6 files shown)


bryanjnelson (approved):
LGTM!

PR #3839, opened 2026-08-03
[pg] Schema fixes: relax name/code uniqueness (0030) + system-agent workflow actors (0031)

rdonigian:
### Summary

Two schema problems, both found while preparing for the dry run. Each would have silently destroyed data during the load, and one is also wrong today independent of any migration.

**Postgres was enforcing five uniqueness rules that the real system doesn't have.**

Two of them — language name and display name — came from a misreading. A comment claimed they mirrored constraints in the current database; those constraints don't exist there, and never did. The other three — a location's ISO code and both ethnologue codes — were inherited from the schema this was ported from, which still declares them unique today. That declaration is the stale one, not this change.

This breaks in two independent ways:

- **It's wrong right now.** Two genuinely distinct languages can share a name, and real data contains such groups — each one separable by its ROLV code, so these are distinct languages, not accidental duplicates. Ethnologue codes are shared far more widely still. Under these rules, the next person creating a legitimately distinct language that happens to share a name is simply refused. Forever, on either database.
- **It would drop data at load, quietly.** The loader can't insert a row that violates a unique rule, so it skips it — and then everything hanging off that row goes too, because a language is required downstream. A handful of rejected languages removes orders of magnitude more rows than it appears to: engagements, ceremonies, products, periodic reports, progress and summaries all follow. Worse, reconciliation reports each table separately, so a single root cause reads as seven unrelated problems.

**Workflow events couldn't record that the system did something.**

The actor column required a real user, which encodes "every project step change was made by a logged-in person." That isn't true — most step transitions are performed by the system, not a person. Under the old shape every one of those rows fails the foreign key, so the load drops them: most of every project's step history, gone, and silently, because a skipped row isn't an error. Local development data hid this almost completely; it only surfaced against a production-shaped copy.

Reads already handled this correctly — the current database has always allowed either kind of actor, and the API already returns a person-or-system union. Only the Postgres column was behind.

### How

- **Drop the five rules; keep the three that are real.** The ROLV code stays unique because it genuinely is a language's natural key. Location *names* stay unique — only the ISO code is dropped. The one-ethnologue-row-per-language rule stays, because it's structural rather than a natural key. All five dropped rules were already scoped to non-deleted rows, so nothing about the soft-delete convention changes; those columns simply stop being keys.
- **Nothing replaces them, deliberately** — and the reason is the tables, not the columns being unused. As filters these columns are only ever matched with a "contains" search, which no b-tree can serve regardless of uniqueness. As sort keys three of them are genuinely reachable by clients today. But both tables are small, and the list path runs an unconditional count over the whole table beside every page query, so it gets scanned whatever index exists. Measured on a comparable table, removing the index moved a paged sort from an index scan to a sequential scan — a sub-millisecond difference. If one of these lists ever becomes hot, add a plain non-unique index; don't restore uniqueness.
- **Let a workflow event point at either a person or a system actor**, with a database-level rule that exactly one of the two is always set — so a row can never be ambiguous about who acted, and can never be empty. This deliberately reuses the same shape the audit log adopted for the same problem rather than inventing a second idiom.
- **Repository error handling that existed only to catch the dropped rules is removed**, since it's now unreachable.

### Judgement calls worth a reviewer's attention

- **The system-actor reference is stored as an id, where the audit log stores a name.** That's intentional, not an inconsistency. An audit row is a historical snapshot and has to survive the removal of the thing it names, so a hard reference would be wrong there. A workflow event's actor is live project history that gets read back as a hydrated object, so it needs an id — storing a name would mean a lookup on every read to rebuild exactly what a reference already gives us.
- **A known write-path gap, pre-existing and not introduced here.** The step-change notification handler looks the actor up as a person, so a system-authored transition that carries notification recipients throws — and because hooks run inside the transition's transaction, the whole transition rolls back rather than half-committing. This change only moves *where* that fails: previously the insert was rejected outright, now the insert succeeds and the handler throws instead. It strictly improves the no-recipients case, which now works. The current database fails the same flow for the same reason, so this is parity rather than a regression, and the fix is tracked as post-cutover work. The system-actored events that already exist came from a backfill that writes directly and fires no hooks, so reading them is fine.
- **Search indexing is not at parity, and this is worth stating plainly** because an earlier version of this reasoning claimed the opposite. The two language columns *are* indexed on the current database, with a text index that tokenises and folds accents. Postgres matches them with a plain substring comparison and no text index at all. Those two approaches don't return the same rows — so it's a behavioural difference in search, not just a slower one. Tracked as a follow-up: either trigram indexes, or an accepted difference that gets written down. Dropping the unique rules neither causes nor worsens it, since a unique index couldn't serve a leading-wildcard match either.

### Validation performed

- Every migration applies cleanly from scratch, and the schema-invariants spec verifies the migration journal is contiguous with strictly increasing, non-duplicated timestamps.
- Postgres end-to-end on the affected surfaces: **4 suites, 0 failed, 75 passed** — schema invariants, language, project workflow, engagement.
- The live-query structural check passes, including its stale-exemption assertion. Worth noting because that check landed after this work began and inspects every Postgres repository, three of which this touches.
- Rebased onto current develop with no content change — the patch is byte-identical to the pre-rebase version, so earlier review findings still apply as-is.

### Explicitly out of scope

- **Text-search parity** for the two language columns, described above.
- **The notification handler's person-only actor lookup**, which matches current behaviour and is tracked as post-cutover work.

### DIFF
--- a/src/components/language/ethnologue-language/ethnologue-language.drizzle.repository.ts
+++ b/src/components/language/ethnologue-language/ethnologue-language.drizzle.repository.ts
@@ -1,11 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { ilike, type SQL } from 'drizzle-orm';
 import { generateId, type ID, type UnsecuredDto } from '~/common';
-import {
-  catchUniqueViolation,
-  DrizzleDtoRepository,
-  escapeLikePattern,
-} from '~/core/drizzle';
+import { DrizzleDtoRepository, escapeLikePattern } from '~/core/drizzle';
 import { type DrizzleDb, DrizzleService } from '~/core/drizzle/drizzle.service';
 import { ethnologueLanguages } from '~/core/drizzle/schema';
 import {
@@ -15,16 +11,9 @@ import {
   type UpdateEthnologueLanguage,
 } from '../dto';
 
-const catchCodeUnique = catchUniqueViolation(
-  'ethnologue_languages_code_unique',
-  'code',
-  'EthnologueLanguage with this code already exists.',
-);
-const catchProvisionalCodeUnique = catchUniqueViolation(
-  'ethnologue_languages_provisional_code_unique',
-  'provisionalCode',
-  'EthnologueLanguage with this provisional code already exists.',
-);
+// No unique-violation mapping: neither `code` nor `provisional_code` is unique
+// (migration 0030 — languages share ethnologue codes routinely), so there is no
+// violation to map. The unique key is the ROLV code on `languages`.
 
 @Injectable()
 export class EthnologueLanguageDrizzleRepository extends DrizzleDtoRepository<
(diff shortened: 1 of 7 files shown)


bryanjnelson (approved):
LGTM

PR #3840, opened 2026-08-04
[pg] Run every e2e spec under Postgres, and fix the ordering bug it exposed

rdonigian:
### Summary

Postgres CI only ran the test files named on a hand-maintained list. Anything
missing from that list simply didn't run — and a test that never ran looks
exactly like a test that passed, so the run reported success either way.

That list was wrong in three different ways at once:

- A domain whose own migration PR never added it, so it was never covered.
- A name that didn't match what it looked like it matched. Entries were followed
  by a literal `.`, so `periodic-report` covered `periodic-report.e2e-spec.ts`
  but not `periodic-report-narrative-file.e2e-spec.ts`.
- Every spec inside a `test/` subdirectory. The pattern's shape could not match
  those at all, so an entire directory of tests was invisible.

The third one is why nobody noticed that the webhooks feature has no Postgres
implementation at all — its tests were in the group that could never run.

So Postgres now runs the whole suite. Suites that genuinely cannot pass yet say
so inside the test file and are reported as **skipped, with the reason stated**,
instead of being left off a list where they are indistinguishable from passing.

Running everything surfaced one real bug, and this PR fixes it: **listing
records sorted by name came back in a different order on Postgres than on
Neo4j.** Sorting text in Postgres uses the database's default ordering, and that
default depends on which C library the Postgres image was built against. The
alpine image has no locale-aware ordering at all, so it silently falls back to
raw byte order — every capital letter ahead of every lowercase one. The same
image name on a Debian build behaves differently again, and an OS upgrade can
shift it. The fix names an ordering explicitly in the query, so every
environment sorts identically and matches how Neo4j orders names today.

Two smaller honesty fixes ride along:

- Ten test cases that only apply to Postgres used to `return` early on Neo4j.
  Jest counts that as **passed**, so the Neo4j run claimed ten passing tests
  that never asserted anything. They now report as skipped.
- Two organization sort tests had been disabled long enough that the behaviour
  they covered was unverified. They're restored and now assert case-insensitive
  ordering in both directions.

Because the tests now honour `DATABASE` from the env files — which they
previously ignored — the README gained a short section on which database the
tests run against, and on why `POSTGRES_URL` has to be exported rather than read
from a file. Both are consequences of this change, so they're documented here
rather than left for whoever hits them.

### A previous decision this reverses

A note in `test/project.e2e-spec.ts` recorded a decision from 2026-06-12 that
Postgres would keep its own ordering and the test expectations would be updated
to match, accepting that the two databases sort names differently.

That decision was made from a measurement taken against the alpine image, and
the measurement was misleading: what was observed was not how Postgres orders
names, it was that image having no locale-aware ordering to use. A glibc-based
Postgres with the very same collation name orders names the way Neo4j does. So
the difference being accepted did not need to be accepted, and it would also
have reappeared or shifted depending on which image ran.

Both databases now assert one expectation, and the test says what to check first
if it ever fails again.

### Schema

One new migration, `0032_display_order_collation.sql`:

```sql
CREATE COLLATION display_order (
  provider = icu,
  locale = 'und-u-ka-shifted-ks-level1',
  deterministic = false
);
```

It ignores case, accents, punctuation and spaces when comparing, which is how
names are ordered today — so lists don't silently reorder as domains move over.

Two constraints on it that matter for review:

- **Use it in `ORDER BY` only, never as a column's collation.** It has to be
  non-deterministic so that strings differing only in case or accent can compare
  equal for ordering, and Postgres refuses `LIKE`/`ILIKE` against a
  non-deterministic collation. Global search runs `ILIKE` against these same
  name columns, so attaching it to a column would break search. Per-`ORDER BY`
  is allowed and is what the code does.
- It cannot be applied to every column type. `uuid` reports a string data type
  but cannot be collated, and neither can an enum type — Postgres raises
  "collations are not supported by type" for both.

### What's added

- `src/core/drizzle/order-by.ts` — `displayOrder(col)` attaches the collation
  when the column holds display text, and returns the column untouched
  otherwise. `resolveOrderBy` routes every sort through it. Three kinds of column
  are deliberately left alone:
  - dates and numbers, which keep their natural ordering;
  - columns limited to a fixed set of values, which hold codes such as a status
    or a step and should sort exactly as written rather than depend on how the
    collation treats an underscore (and enum types cannot be collated at all);
  - **primary keys**, which hold opaque generated identifiers. Every id in this
    schema is a text column, so without this they would be case-folded like
    prose. That mattered: the user list's default sort key is its id, and `id` is
    not in that list's sort map, so every user query without an explicit sort
    falls through to it. Collating it reordered the default user list away from
    what Neo4j returns — the opposite of this collation's purpose — and stopped
    the ordering being read off the primary key index, making every page a full
    scan plus a sort. Of the 25 sort-resolver call sites, that is the only
    fallback which is an identifier; the rest fall back to a name, a date or a
    number, and no sort map points at an id column.
- `src/core/drizzle/order-by.spec.ts` (new) — pins the collate/don't-collate
  decision, because nothing else can see it. A wrong decision still produces
  valid SQL and a full page of results.
- The three list queries that sort by a joined table's name (partner,
  partnership, project) build their `ORDER BY` by hand rather than through
  `resolveOrderBy`, so each now calls `displayOrder` directly. Without that they
  would order differently from every other list in the app.
- `.github/workflows/test.yml` — the Postgres file list is removed; the job runs
  the full suite on a single shard. The comment records why an allowlist is not
  the right tool here, so it doesn't get reintroduced.
- `test/setup/resolve-engine-env.ts` (new, registered first in `jest.config.ts`
  `setupFiles`) — makes `process.env.DATABASE` agree with the engine the app will
  actually boot. A test file decides which database it is testing when it loads,
  before any app starts, so it sees only real environment variables and never a
  `DATABASE=` line in `.env.local`. Left alone the two disagree, and every spec
  runs its Neo4j path against a Postgres app — which reads exactly like a
  regression. It resolves the same precedence the app uses (real environment,
  then dotenv, then the app's default), reading the same four env files in the
  same order and parsing them with **dotenv's own `parse`** — the same function
  the app's own environment service calls — so quoting, `export ` prefixes and
  inline comments resolve identically instead of via a lookalike of dotenv's
  rules. An explicitly set variable still wins, so CI is unaffected.
- `test/webhooks.e2e-spec.ts` and the three `*-changeset-aware` specs switch
  themselves off under Postgres, each stating why: webhooks has no Postgres
  implementation yet, and changesets are not being carried forward. Both carry a
  `migration-todo` naming what removes the condition.
- `test/audit-log.e2e-spec.ts` and `test/file.e2e-spec.ts` — the ten
  Postgres-only cases now use `itPostgresOnly` / `describePostgresOnly`. Guards
  that assert real Neo4j behaviour before returning were deliberately left as
  they are; only the ones that asserted nothing changed.
- `test/organization.e2e-spec.ts` — the two restored sort tests seed four records
  whose names differ only by case, then page through the full list and compare
  against the expected order. They page because list input caps `count` at 100.
- `README.md` — a "which database the tests run against" section, the
  `POSTGRES_URL` requirement with its reason, and `docker-compose` corrected to
  `docker compose` in three places. The Gel setup step also claimed Gel was "the
  current primary database", which contradicted the section directly below it
  describing a Neo4j-to-PostgreSQL migration; for a first-time setup it was
  unclear whether Gel was required, optional or historical. It is left from an
  earlier migration target and nothing new is built on it, but its generated
  client is not committed, so the project will not compile without that step —
  now said plainly, with the two sections cross-linked.

### Validation performed

Run against this branch rebased onto develop, clean tree, both engines:

- `yarn type-check` — exit 0. `yarn lint` — exit 0.
- **Postgres, full suite** — 0 failed, 320 passed, 95 skipped, 2 todo; 41 suites
  passed, 4 skipped, exit 0. Before this branch the same full run had 5 failing
  suites and 63 failing tests.
- **Neo4j, every spec file this branch changes** (8 files) — 0 failed, 120
  passed, 19 skipped, exit 0. The rest of the Neo4j suite is CI's job and was not
  run locally.
- The new `displayOrder` unit spec — 5 passed. Then the same spec with the
  primary-key exclusion removed: **exactly one case fails and the other four
  still pass.** A guard that has never failed is not evidence, so it was broken
  on purpose before being trusted.
- The engine resolver, checked in both directions because it loads before every
  spec: with `DATABASE` unset it picks the engine out of the env files and the
  Postgres-only schema spec runs its 14 tests; with `DATABASE=neo4j` set
  explicitly the explicit value wins and those same 14 report skipped.
- Each run wrote its own machine-readable results with the commit stamped, and
  the per-suite counts were checked to sum to the reported totals, rather than
  reading suite names out of console output.

Two limits of the above, stated rather than left to be assumed:

- ⚠️ **The end-to-end suite cannot see the ordering defect this fixes.** The full
  Postgres suite returns identical counts before and after the fix — same
  failures, same passes, same skips, and no suite's passing count changed. No list
  spec asserts the default user ordering. The unit spec is the only detector, and
  that is why it exists.
- The full-suite runs predate the last two commits and were not repeated for
  them. That is deliberate, not skipped work: both local runs and CI set
  `DATABASE` explicitly, which bypasses the code path the resolver commit
  changed, so the suite cannot exercise it — the two-directional check above is
  what covers it. The final commit is documentation only.

The direction of every self-skip was checked on **both** engines, since a
one-directional condition that fires the wrong way round is the failure mode:

- webhooks — 59 skipped on Postgres; 59 **passed** on Neo4j.
- the three changeset specs — all skipped on Postgres; all pass on Neo4j.
- the ten Postgres-only cases — run on Postgres (audit-log 6, file 4); report
  **skipped** on Neo4j, where they previously reported passed.
- the restored organization sort tests — pass on both.
- the project name-ordering test — **passes on both**, which is the point of the
  collation change. It fails on Postgres without it.

The full Postgres run also applies every migration to a fresh database per spec
file, so 0030 → 0031 → 0032 applying in order is exercised rather than reasoned
about.

### Reviewer cross-checks

- The collation appears only in `ORDER BY`, never on a column definition. Worth
  grepping, since attaching it to a name column would break global search rather
  than fail loudly here.
- Whether any column that *should* sort case-insensitively lost it when primary
  keys were excluded. The check that convinced me: no sort map anywhere points at
  an id column, and the user list is the only call site whose fallback is one.
- `displayOrder` can only inspect a column, so a sort written as a raw `sql`
  expression comes back uncollated with no error. The engagement sort map is the
  only place that can hit this today — its two raw expressions sort dates, so
  they are unaffected, but the two name sorts already scheduled there will have
  to name the collation inline. Called out at that call site.
- `displayOrder` skips `uuid` and enum-typed columns. Both raise a Postgres error
  if collated, so a widened list would fail at query time, not at build.
- The migration journal: indexes contiguous, `when` values strictly increasing,
  no duplicates. A duplicate `when` is skipped silently by the migrator rather
  than reported as an error.
- Every self-skip condition reads `DATABASE`. A misspelled variable makes the
  condition never fire, and the suite goes back to failing — or worse, to
  passing while asserting nothing.
- Under Neo4j, webhooks and the three changeset specs must still **run**, not
  skip. The conditions are one-directional on purpose.

### Explicitly out of scope

- **Letting the Postgres job block a merge.** It still runs with
  `continue-on-error`, so a Postgres failure reports without blocking. Flipping
  that is a one-line change and belongs in its own PR, once this one has
  reported green at least once.
- **The webhooks port.** This PR only makes its absence visible. The feature has
  to be ported before Neo4j goes away, and the test file says so.
- **Removing the changeset specs.** They stay until the changeset feature itself
  is removed.
- Re-sharding the Postgres job. One shard is right while the full suite is about
  ten minutes serial; the comment says what to change when that grows.
- **Choosing collated columns explicitly instead of by column type.** Deciding
  from a column's type is a proxy for the real question, and excluding primary
  keys patches the proxy rather than replacing it. The better design names the
  display-text columns outright. It is deliberately not done here: all 25 sort
  targets were checked and, with primary keys excluded, every text column that
  actually gets sorted is one the Neo4j path also case-folds — including the two
  that looked like exceptions, `educations.institution` and `producibles.name`.
  So no ordering difference between the engines remains, and the change would
  touch all 25 sort maps for no change in behaviour. It is tracked for after the
  migration completes, when "match the old ordering" stops being the standard and
  the question becomes which columns should read case-insensitively for a person
  scanning a list.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,11 @@ jobs:
           --health-start-period 30s
       # Needed for the postgres matrix entries; idle for the neo4j ones.
       postgres:
+        # alpine is fine BECAUSE text ordering no longer depends on the image:
+        # every text sort names the `display_order` collation explicitly
+        # (migration 0032). That makes ordering identical here, on a glibc
+        # Postgres, and across OS upgrades — so CI proves the ordering
+        # production will actually have, whatever image it runs.
         image: postgres:16-alpine
         ports:
           - 5432:5432
@@ -80,23 +85,30 @@ jobs:
           - neo4j
 #          - gel
         include:
-          # Postgres runs only the suites for domains that have a Drizzle repo
-          # on this branch — unmigrated domains fall back to Neo4j via splitDb
-          # and hit cross-engine boundaries. Each entry below MUST have a
-          # matching *.drizzle.repository.ts; add a domain here only in the same
-          # PR that lands its repo. Widen as each phase lands. migration-todo:
-          # drop the filter entirely at Phase 7 cutover so postgres runs the
-          # full suite (and re-shard it to match neo4j).
+          # Postgres runs the WHOLE suite. It used to run an opt-in allowlist of
+          # migrated domains, which is removed deliberately: an allowlist
+          # silently excludes anything nobody remembered to name, and it failed
+          # that way three separate times — a domain whose own migration PR
+          # never added it, a name that didn't match because entries are
+          # anchored to a following `.` (so `periodic-report` never covered
+          # `periodic-report-narrative-file`), and every spec under `test/`
+          # subdirectories, which the pattern's shape could never match at all.
+          # The last of those hid the fact that webhooks was never ported.
           #
-          # Single shard on purpose: the migrated-domain subset is only a few
-          # minutes of tests — six jobs paying the full setup cost for ~1.5 min
-          # of tests each was pure runner waste. Bump `total` (and add shard
-          # entries) as the subset grows.
+          # Suites that genuinely cannot pass under Postgres now switch
+          # themselves off inside the test file (`describe.skip` on a DATABASE
+          # check), so they are reported as skipped, appear in the run output,
+          # and state their reason — instead of being invisible. Currently: the
+          # three *-changeset-aware specs (changesets are not being carried
+          # forward) and webhooks (not ported yet). Add that condition to the
+          # test file — never a filter here.
+          #
+          # Single shard on purpose: the full suite is ~10 min serial, and six
+          # jobs each paying the full setup cost was pure runner waste. Bump
+          # `total` (and add shard entries) if that grows uncomfortable.
           - database: postgres
             shard: 1
             total: 1
-            test-args: >-
-              --testPathPatterns "test/(authentication|user|education|unavailability|file|location|organization|region|zone|funding-account|partner|partnership|project-member|tool|notification|postgres-schema|auth-read-filter|budget|language|engagement|engagement-workflow|ceremony|product|film|story|project-workflow|periodic-report|rev79|pin|known-language|comment|post|audit-log|progress-report-media|search|live-query-invalidation)\."
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
@@ -120,7 +132,7 @@ jobs:
         # V8's default 4GB heap cap late in the shard (all-shards OOM,
         # 2026-07-14). Recycling the worker between files when it bloats keeps
         # serial semantics with bounded memory.
-        run: yarn test:e2e --maxWorkers=1 --workerIdleMemoryLimit=1500MB --shard=${{ matrix.shard }}/${{ matrix.total || 6 }} --reporters=github-actions ${{ matrix.test-args }}
+        run: yarn test:e2e --maxWorkers=1 --workerIdleMemoryLimit=1500MB --shard=${{ matrix.shard }}/${{ matrix.total || 6 }} --reporters=github-actions
         env:
           NEO4J_VERSION: ${{ matrix.neo4j-version }}
           DATABASE: ${{ matrix.database }}
--- a/README.md
+++ b/README.md
@@ -18,13 +18,16 @@ Bible translation project management API.
 1. Copy `.env.local.example` to `.env.local` and fill in any required values
 1. Start the databases:
     ```bash
-    docker-compose up -d db postgres
+    docker compose up -d db postgres
     ```
-1. Setup a Gel instance (the current primary database):
+1. Setup a Gel instance. Gel is not the primary database — see [Database](#database)
+   below — but a number of repositories still use it and its generated client is
+   not committed to the repo, so the project will not compile without this step:
     ```bash
     gel project init
     yarn gel:gen
     ```
+   Re-run `yarn gel:gen` after any change to `dbschema/`.
 
 ## Database
 
@@ -39,9 +42,14 @@ The `DATABASE` env var controls which is active for each domain:
 Both services must be running locally regardless of which mode is active:
 
 ```bash
-docker-compose up -d db        # Neo4j
-docker-compose up -d postgres  # PostgreSQL
+docker compose up -d db        # Neo4j
+docker compose up -d postgres  # PostgreSQL
 ```
+
+Gel is a third database in the tree, left from an earlier migration target. It is
+neither the destination nor the primary store and nothing new is being built on
+it, but it has not been removed yet — so setting it up is still required in order
+to build. See [Setup](#setup) above.
 
 PostgreSQL migrations run automatically on startup when `DATABASE=postgres`.
 To generate a new migration after a schema change:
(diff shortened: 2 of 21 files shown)


bryanjnelson (approved):
👍🏼

PR #3842, opened 2026-08-04
[pg] Engine-aware transactions and boot, and the transaction-scoping bugs that surfaced

rdonigian:
### Summary

Shared plumbing still assumed Neo4j regardless of which database was configured.
This makes it engine-aware, and fixes the transaction-scoping problems that came
to light once transactions started being routed to the engine actually in use.

Four things were reaching for Neo4j unconditionally:

- The Neo4j indexer and auto-migrate ran at boot on every engine, doing work
  against a database the app was not using.
- `@Transactional()` opened its transaction on Neo4j's connection, so a decorated
  method's writes went to Neo4j even when the domain it touched lives elsewhere.
- The Postgres repositories emitted a bare `{ id }` where `parent` is read as a
  graph node. That is not a wrong value but a crash: the resolver reads
  `.labels` and `.properties.id` off it. Four repositories shipped it, and no test
  on any migrated entity ever selected `parent`, which is how it survived three
  changes.
- Asking a project for its change requests threw under Postgres, because the
  repository behind it is Neo4j-only by design — changesets are not being carried
  forward.

Routing transactions to the active engine then exposed two scoping bugs, both
fixed here. Neither is visible in a passing test, which is most of why they are
worth describing carefully.

**A nested transaction started a second, unrelated one.** Repositories never take
a transaction parameter — the service keeps the transaction in an async-local
store and every repository reads a client getter that returns it when set. A
nested call ignored that store and began again from the pool, so it got another
connection and an independent transaction. `updateProject` reaches this today,
because the mutation interceptor has already opened a transaction by the time the
decorated method runs. Consequences: an inner write survived an outer rollback, so
a failed mutation no longer undid everything it did; two pool connections were
held per nested call, which exhausts the pool and hangs rather than failing; and
the retry budget nested, so one retryable error could re-run the method body and
every hook it fires many times over. Neo4j never had any of this — its equivalent
has always continued an open transaction — so the fix restores the behaviour the
other engines already had rather than inventing one.

**Work that outlived its transaction was silently allowed.** An async context
created inside a transaction keeps seeing it after it has settled and its
connection has gone back to the pool, so "a transaction is in the store" did not
mean "there is a transaction to use". Work started inside a transaction and not
awaited lands there. It did not error — the query ran and returned rows, outside
the transaction it believed it was in. For a read that is merely wrong; for a
write it would persist data the caller thought was transactional and that no
rollback would undo. Liveness is now recorded with the transaction and both
entry points refuse a settled one, naming the mistake and pointing at the
supported way to run work after a commit.

### What's added

- A `TransactionRunner` that is the single place knowing how each engine begins a
  transaction. The per-engine mutation interceptors and `@Transactional()` both
  delegate to it, so no caller hardcodes an engine. Carries a `migration-todo` for
  dropping the non-Postgres arms at cutover.
- Transaction continuation in the Drizzle and Gel paths, matching Neo4j. The guard
  sits in three places because they are reached differently: the Drizzle service
  itself, which also covers the admin repository calling it directly; Gel's
  equivalent, which had the identical defect; and the runner, which returns before
  its retry loop — a guard further down would fix the transaction but leave the
  retries nested.
- Liveness tracking on the stored transaction, so a settled one is refused by both
  the client getter and the nested-transaction check rather than quietly used.
- `ChangesetAware.parent` now emits a typed reference from the Postgres
  repositories, with the base declaration widened to accept either shape and the
  resolver normalizing both. The six subclass re-declarations that narrowed it
  back are widened to match — type-only, no runtime or schema effect, since each
  field's GraphQL type comes from its own `@Field()` decorator. `ProgressReport`
  and `Post` keep the narrow form deliberately: different lineage, and their
  resolvers do read the graph-node properties.
- A Postgres `ProjectChangeRequest` repository that answers reads as empty and
  refuses writes, so asking a project for its change requests succeeds instead of
  throwing. Neo4j is untouched.
- Boot skips for the Neo4j indexer and auto-migrate when the engine is not Neo4j.
- `test/postgres-transactions.e2e-spec.ts`, covering both scoping fixes, plus
  `parent` coverage on partnership — the first test anywhere to select that field.

### Validation performed

- Type-check and lint clean.
- **Neo4j, 8 mutation-heavy specs** (project, engagement, partnership, budget,
  user, all three changeset-aware): 96 passed, 0 failed. **Byte-identical before
  and after the transaction work** — which is the point, since production runs
  Neo4j and the transaction runner touches its interceptor and decorator.
- **Neo4j, webhooks**: 59 passed, unchanged.
- **Postgres, full suite**: the same five failing suites as develop, no new
  failures and none fixed, with the added tests passing. Run three times across
  the branch's development; the failing set never moved.
- Every fix has a negative control, because none of these are visible in a
  passing test:
  - Revert the parent shape → the new partnership test fails with the exact
    TypeError the fix removes, and nothing else fails.
  - Remove transaction continuation → all three scoping tests fail, in three
    different ways: different backend connections, the nested read cannot see the
    outer's uncommitted row, and the inner write survives the rollback.
  - Never mark a transaction settled → both escaped-work tests fail while the
    three continuation tests still pass, so they are not restating each other.
- The transaction fix was also reviewed independently against the source by a
  separate model, which confirmed the semantics, the race-freedom of the liveness
  check, that hooks stay inside the correct transaction, and that the test's use
  of the backend connection id is a sound proxy in this context.

Stated plainly rather than left to be assumed: **under Postgres, webhooks goes
from 51 to 53 failing tests.** Two tests that passed there now do not. Webhooks has
no Postgres implementation, the suite is already almost entirely failing on that
engine, and **Neo4j is completely unaffected at 59 of 59** — but it is a real
behaviour change caused by transactions no longer being opened on Neo4j for a
Neo4j-only domain, so it belongs here rather than in a footnote.

### Reviewer cross-checks

- The transaction continuation is in all three places, and the runner's returns
  before the retry loop rather than inside it. A guard in only one place leaves
  either the second transaction or the nested retry budget.
- Nothing relied on a nested transaction being isolated. Where isolation is
  genuinely wanted the pattern is a savepoint — calling `.transaction()` on the
  current client — as the partnership repository does so that losing a uniqueness
  race cannot poison the surrounding mutation. Continuation by default and
  savepoints where asked for are two halves of one design; worth confirming that
  reading holds.
- The parent shape is correct for each of the five repositories, not just the
  tested one. Budget's and Engagement's type derivations are textually identical
  to the proven partnership one; BudgetRecord's fixed typename is a distinct path.
- `ProgressReport` and `Post` must keep the narrow `parent` declaration.
- The `ProjectChangeRequest` stub cannot be reached under Neo4j, and its refusals
  surface as handled errors rather than raw ones.
- Boot skips are tagged with a `migration-todo` naming what to delete at cutover.

### Explicitly out of scope

- **Per-repository `parent` tests for Budget, BudgetRecord and Engagement.** Only
  partnership is covered. Those fields are reachable and untested on both engines —
  the same condition that let the original bug ship. Tracked rather than fixed
  here to keep this from growing further.
- **A dead `parent` key on the Ceremony repository**, along with the extra column
  fetched to build it. Ceremony exposes no `parent` field, so nothing can read it.
  Harmless, and left with the rest of the cleanup.
- **The connection pool has no explicit maximum and no connection timeout.** Not
  introduced here, and not in this diff, but it is what turns any future
  connection-holding bug into an unbounded wait instead of a loud failure. Worth
  its own change.
- **The `parent { id }`-only selection throws** on these types, for a reason
  unrelated to shape: the id-only shortcut omits a field the Project interface's
  custom type resolver needs. Identical on both engines, so pre-existing. Recorded
  as a skipped test with the explanation instead of being left to be rediscovered.
- Webhooks itself, which needs porting before Neo4j goes away.

### DIFF
--- a/src/components/budget/budget-record.drizzle.repository.ts
+++ b/src/components/budget/budget-record.drizzle.repository.ts
@@ -195,7 +195,7 @@ export class BudgetRecordDrizzleRepository extends DrizzleDtoRepository<
       preApprovedAmount: row.preApprovedAmount,
       status: row.budget.status,
       sensitivity: row.budget.project.sensitivity,
-      parent: { id: row.budget.id },
+      parent: { id: row.budget.id, __typename: 'Budget' },
       // PCR is excluded; resolver navigation marker stays undefined.
       changeset: undefined,
       canDelete: true,
--- a/src/components/budget/budget.drizzle.repository.ts
+++ b/src/components/budget/budget.drizzle.repository.ts
@@ -17,7 +17,7 @@ import {
   type SortMap,
 } from '~/core/drizzle';
 import { DrizzleService } from '~/core/drizzle/drizzle.service';
-import { budgets } from '~/core/drizzle/schema';
+import { budgets, type projects } from '~/core/drizzle/schema';
 import { type ScopedRole } from '../authorization/dto/role.dto';
 import { PolicyExecutor } from '../authorization/policy/executor/policy-executor';
 import { type FileId } from '../file/dto';
@@ -32,7 +32,10 @@ import {
 } from './dto';
 
 type BudgetRow = typeof budgets.$inferSelect & {
-  project?: { id: ID<'Project'>; sensitivity: string } | null;
+  project?: Pick<
+    typeof projects.$inferSelect,
+    'id' | 'sensitivity' | 'type'
+  > | null;
 };
 
 @Injectable()
@@ -91,7 +94,8 @@ export class BudgetDrizzleRepository extends DrizzleDtoRepository<
     const rows = await this.db.query.budgets.findMany({
       where: (b) => and(inArray(b.id, [...ids]), isNull(b.deletedAt)),
       with: {
-        project: { columns: { id: true, sensitivity: true } },
+        // `type` feeds the parent ref's __typename (`${type}Project`).
+        project: { columns: { id: true, sensitivity: true, type: true } },
       },
     });
     const scopeByProject = await requesterScopeByProject(
@@ -191,7 +195,10 @@ export class BudgetDrizzleRepository extends DrizzleDtoRepository<
         : null,
       // Assembled by the service via listRecords.
       records: [],
-      parent: { id: row.project.id },
+      parent: {
+        id: row.project.id,
+        __typename: `${row.project.type}Project`,
+      },
       // PCR is excluded; resolver navigation marker stays undefined.
       changeset: undefined,
       canDelete: true,
--- a/src/components/budget/budget.drizzle.repository.ts
+++ b/src/components/budget/budget.drizzle.repository.ts
@@ -184,15 +184,20 @@ export class BudgetDrizzleRepository extends DrizzleDtoRepository<
         `Budget ${row.id} has no parent project row — FK invariant violated`,
       );
     }
-    const dto: unknown = {
+    // Not laundered through `unknown` before the cast below: that stops
+    // TypeScript comparing this object to the DTO at all, which is how a field
+    // can go missing or take the wrong shape unnoticed. The direct cast still
+    // allows the service-layer overlays (canDelete, scope) and the fields the
+    // service fills in later.
+    const dto = {
       id: row.id,
       __typename: 'Budget',
       createdAt: DateTime.fromJSDate(row.createdAt),
       status: row.status,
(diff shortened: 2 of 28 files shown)


bryanjnelson (approved):
Big finds here!  🙌🏼

PR #3843, opened 2026-08-05
[pg] Fill in Project.marketingRegion under Postgres, and stop hiding DTO mismatches

rdonigian:
### Summary

A project's marketing region isn't stored — it's worked out when read: use the region someone set explicitly as an override, and otherwise inherit the one the project's marketing location points at. The Neo4j repository does exactly that. The Postgres one returns the override and stops, so any project relying on the inherited value answers with nothing.

Nothing errors when this happens. The field just comes back empty, which is why it survived the original port and two reviews.

The reason the compiler didn't catch it is the second half of this change. The repository built its result object, parked it in a variable typed as "could be anything", and then asserted it was a Project. That middle step throws away the real shape, so TypeScript never compares the object against what a Project is supposed to contain — a missing field costs nothing. Removing it makes this class of mistake a compile error instead of empty data, which is worth more than the single fix.

Two smaller things surfaced once that checking was back on, both in the prompt-response repository:

- The column holding a prompt's id was described as ordinary text, so it read back as an untyped string. It now carries the same id type the DTO uses.
- Which response variants are valid depends on the subtype the repository was built for, and that genuinely can't be proven from a text column. That one field is now narrowed deliberately, in place, rather than switching off checking for everything around it.

### Schema / DDL

No migration. The only schema-file change is a TypeScript type on an existing column (`prompt_variant_responses.prompt`), which changes neither the stored data nor the generated SQL.

### Validation performed

- Type-check clean, lint clean.
- New end-to-end test covering both halves of the rule: with no override the region is inherited from the marketing location, and setting an override replaces it.
- That test passes on Neo4j, which is the parity baseline, and on Postgres with this fix.
- **Negative check:** with the fix removed, the same test fails on Postgres and the assertion shows the field arriving as nothing. Removing the field is now also a compile error, so the mistake can't be made silently twice.
- Full project spec family, both engines: Postgres 43 passed / 0 failed, Neo4j 46 passed / 0 failed. The difference is the change-request suite, which is deliberately skipped under Postgres. Every project read goes through the query this touches, which is why the whole family was run and not just the new test.

### Reviewer cross-checks

- The inherited region is looked up only for live locations. Neo4j reaches it through an active relationship, so a deleted location contributes nothing there either — worth confirming that reading matches yours.
- The lookup is batched once per read rather than done per project, following the pattern this method already uses for memberships, workflow timestamps and pins.
- Precedence: an override always wins, including when the marketing location also carries a default.
- The narrowed variant field is the only place checking is bypassed, and only for that one field.

### Explicitly out of scope

- Seventeen other Postgres repositories still park their result object in an "anything" variable the same way. Sixteen of them need nothing but the deletion — that's measured, not assumed. Separate change, and it wants a quiet tree.
- One mismatch remains in the budget repository: the database allows its template file to be absent while the DTO claims the field is always present. Equally untrue on Neo4j, so it isn't a migration regression.

### DIFF
--- a/test/project.e2e-spec.ts
+++ b/test/project.e2e-spec.ts
@@ -928,62 +928,80 @@ describe('Project e2e', () => {
     });
   });
 
-  it('takes marketingRegion from the marketing location, and lets the override win', async () => {
-    await runAsAdmin(app, async () => {
-      const inheritedRegion = await createLocation(app);
+  describe('marketingRegion', () => {
+    const readMarketingRegion = async (id: ID) => {
+      const result = await app.graphql.query(
+        graphql(`
+          query projectMarketingRegion($id: ID!) {
+            project(id: $id) {
+              marketingRegion {
+                canRead
+                value {
+                  id
+                }
+              }
+            }
+          }
+        `),
+        { id },
+      );
+      return result.project.marketingRegion;
+    };
+
+    /** A project whose marketing location carries a default region. */
+    const createProjectWithMarketingLocation = async () => {
+      const locationDefaultRegion = await createLocation(app);
       const marketingLocation = await createLocation(app, {
-        defaultMarketingRegion: inheritedRegion.id,
+        defaultMarketingRegion: locationDefaultRegion.id,
       });
       const project = await createProject(app, {
         marketingLocation: marketingLocation.id,
         fieldRegion: fieldRegion.id,
       });
+      return { project, locationDefaultRegion };
+    };
 
-      const readMarketingRegion = async () => {
-        const result = await app.graphql.query(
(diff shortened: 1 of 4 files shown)


test/project.e2e-spec.ts
```diff
@@ -928,6 +928,65 @@ describe('Project e2e', () => {
     });
   });
 
+  it('takes marketingRegion from the marketing location, and lets the override win', async () => {
```
bryanjnelson: I think this test would be better split into two separate tests.

bryanjnelson (changes requested):

bryanjnelson (approved):
👊

PR #3847, opened 2026-08-06
[pg] Port the last four Postgres domains — migrations 0033–0036

rdonigian:
### Summary

Four things the app stores in Neo4j had no Postgres equivalent yet, so under Postgres they simply returned nothing:

- which partnership is responsible for producing each medium on a language engagement
- which resources use which tools, and since when
- why a progress report differs from plan — the canned reasons plus free prose
- the history of a progress report's status changes

The last one is the easiest to lose quietly: a report's *current* status is a column that already carried over, so the field looks right and only the history panel is empty.

Periodic reports also now soft-delete rather than disappearing, which is what Neo4j has always done — its delete relabels a node instead of removing it. That moves the dedup guarantee off the primary key and onto a partial unique index over live rows, so a deleted-then-revived reporting interval gets a fresh id instead of colliding with the old one.

The rest of the branch is review fixes, several worth attention because they were wrong in ways that read as working:

- After-commit work queued by an attempt that **rolled back** could still run when a later retry committed.
- The resource-id resolver only knew six families of table, so `tools` and similar answered nothing for other types. It now registers every table behind a Resource-implementing type, and resolves them in one statement instead of a query per table.
- Two workflow-event read paths returned rows the reader isn't allowed to see. Fixing the project one meant giving the shared transition condition a SQL form, since it previously existed only for Neo4j.
- Name-sort collation was inferred from column type, which folded punctuation and case on nine columns Neo4j leaves alone. It's now an explicit list, because the folding comes from `@NameField` in the app, not from the database.
- A liveness filter used a correlated `EXISTS` inside the relational query builder, which aliases the table and so silently matched nothing. It's a plain join now.

Deleting a tool now also deletes its usages, and `ToolUsage.creator` became nullable so a removed creator no longer breaks the page. **That last one is a schema change cord-field needs to know about** — `Actor!` became `Actor`.

### Text ordering: staying on the alpine Postgres image, and why

The final commit reverts an image swap and rewrites the explanation on both sides of it, because the old comment and its replacement were each wrong in a different direction.

Text sorts split into two groups. Name columns name the `display_order` collation explicitly — an ICU collation applied per sort clause — so they order the same on any image. Everything else falls back to the database default: two address fields, a post body, two product descriptions, an ISO code, a PMC entity code, a department id, an unavailability description. Neo4j orders those by raw code points, and UTF-8 byte order *is* code-point order, so musl — which has no locale-aware comparison and compares bytes — reproduces Neo4j exactly. A glibc image case-folds them and ignores leading spaces: nicer for a person reading a list, but not what the other engine does. Parity wins.

The comment being replaced claimed alpine was safe because *every* text sort names the collation. That stopped being true when the collation was narrowed to name columns, which is the set Neo4j actually folds.

One consequence is recorded as a `migration-todo` in both the compose file and CI: those other columns still depend on the C library, so a glibc production server would order them differently. Emitting `collate "C"` for non-name sorts removes the dependence entirely. Left undone deliberately — it changes ordering on glibc, so it's a decision rather than a cleanup.

**`docker-compose.yml` and `.github/workflows/test.yml` are comment-only changes against the base.** Both pin the same image the base does; only the reasoning above changed.

### Schema / DDL

- **0033** `partnership_producing_mediums` — composite primary key `(engagement_id, medium)`, both columns real foreign keys with `ON DELETE CASCADE`. No `deleted_at`: this is a pure current assignment, and the audit trail covers "who changed this".
- **0034** `tool_usages` — soft-deleted, so uniqueness is `UNIQUE INDEX … WHERE deleted_at IS NULL` on (container, tool) rather than an inline constraint. `container_id` deliberately carries no foreign key, since a usage may attach to any resource; `container_type` stores the concrete typename alongside it.
- **0035** `periodic_reports` gains `deleted_at`, plus `periodic_reports_live_interval_unique` as a partial unique index over live rows.
- **0036** `progress_report_variance_explanations` — `report_id` *is* the primary key, since a report can hold at most one. And `progress_report_workflow_events`, with indexes on `(report_id, at)` and `who`.
- **0019 and 0032 are edited, comments only, no SQL.** Safe to re-apply against an existing database: the migrator decides what to run by comparing timestamps and never compares the stored file hash. 0019's header claimed deletion was hard and dedup came from the primary key; 0035 made both untrue, so it now says so rather than misleading the next reader.
- **0035 also drops an index it originally created.** `periodic_reports_deleted_at_idx` was justified as "every read path filters on this," which isn't the case — liveness is never the only filter, reads narrow by parent id and usually type and date window, and the existing parent-id indexes already lead those. A bare index on the column couldn't be chosen anyway, since almost every row is live, and nothing asks for the selective direction. It was also the only bare `deleted_at` index among the soft-deleted tables. Since this migration is new in this branch, removing it now means no database outside local development ever had it.

### Validation performed

- Type-check and lint clean.
- Migration journal checked mechanically: 37 entries, indexes contiguous, timestamps strictly increasing, no duplicates. A duplicate timestamp is a silent skip in the migrator, so it's worth its own check. Also verified in both directions that every journal entry has a matching `.sql` file and no `.sql` file is missing from the journal.
- Fresh-database apply of all 37 migrations, exercised by the schema-invariant suite: 28 tests passed across the schema, periodic-report and narrative-file specs, 0 failed, 0 skipped. That's the check that matters given two migration files were edited.
- Full unit suite: 18 suites, 211 passed, 0 failed — covering the unit specs this branch changes.
- Tool-usage container parity, both engines: 5 passed on Postgres, 5 on Neo4j. Includes the deliberate divergence — under Postgres a usage whose creator was deleted survives and reports no creator; under Neo4j it drops out. Both sides of that case assert; neither returns early, which would pass while checking nothing.
- The full dual-engine e2e suite passed before this branch was rebased onto its current base and has not been re-run in full since. The rebase is verifiably content-neutral — every commit's patch is unchanged — so the unexercised surface is interaction with what the base added, not this branch's own changes.

### Reviewer cross-checks

- Every typename-to-stored-value mapping in the resource-id resolver, particularly that project and engagement rows still carry both their concrete and family labels, and that liveness is applied for tables that have `deleted_at` and correctly skipped for those that don't.
- The explicit name-collation list in both directions: every entry backed by a `@NameField`, and no `@NameField`-backed sort column missing. Nine columns deliberately stop being collated — and per the section above, that set is exactly why the image family matters.
- That the caller's own conditions and the auth filter end up in **one** `where` — chaining a second one in Drizzle replaces the clause rather than adding to it, a bug that was made and fixed inside this branch.
- The new SQL form of the shared transition condition denies rather than widens when it cannot resolve, and a null transition cannot match.

### Explicitly out of scope

- `workflowEvents` has no end-to-end test on either engine, which is why the read-filter gap survived earlier reviews. Worth adding, not here.
- Four cases in the live-query invalidation spec return early on Neo4j and so pass while asserting nothing. Tracked separately.
- The project workflow read path applies the filter now, but a general "which table holds this resource" map still doesn't exist; conditions reach their column through an explicitly passed column instead. A deliberate smaller step.
- One permission grant reads more project workflow events than intended, because a condition was written before the action it was meant to restrict rather than after. Identical on both engines, so not a migration regression — tracked for after the cutover.


bryanjnelson (dismissed):
😮‍💨 🤝

bryanjnelson (approved):

PR #3848, opened 2026-08-12
Don't run after-commit work queued by an attempt that rolled back

rdonigian:
### Summary

A mutation that fails and then succeeds on a retry could leave behind a
notification for the change from its *failed* attempt — one that never
actually happened. This fixes that, so a retry can no longer cause a
duplicate or phantom notification. It's not related to the Postgres work;
it's a pre-existing gap in shared code that both databases run today.

### What's in scope

- Discard any queued after-commit work (notifications, broadcasts) as soon
  as a mutation attempt fails, before its rollback callbacks run
- Add a `clear()` method to the callback manager to drop queued callbacks
  without running them

### Design decisions

- The clear only happens on the failure path — a committing attempt's own
  callbacks still have to run, so this can't be an unconditional clear that
  silences every broadcast
- Cleared before the rollback callbacks run, so a rollback callback that
  throws can't leave stale commit work behind either

### Validation

- tsc and lint: clean
- New unit spec, since an e2e test can't easily force a mutation to fail
  and then succeed on retry: asserts a rolled-back attempt's callback never
  runs, a committing attempt's still does, and nothing runs when every
  attempt fails
- 3 files changed, 152 insertions

### DIFF
--- a/src/core/database/abstract-transactional-mutations.interceptor.ts
+++ b/src/core/database/abstract-transactional-mutations.interceptor.ts
@@ -32,6 +32,19 @@ export abstract class TransactionalMutationsInterceptor implements NestIntercept
         try {
           return await lastValueFrom(next.handle());
         } catch (e) {
+          // Throw away anything this attempt queued to run after a commit.
+          //
+          // A failed mutation may be retried, and a retry reuses the same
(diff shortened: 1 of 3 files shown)


bryanjnelson (approved):
LGTM

PR #3849, opened 2026-08-13
[pg] Webhooks Domain

rdonigian:
### Summary

This adds a working Postgres version of the webhooks feature — the last piece of the app that didn't have one yet. Webhooks let a user get a callback whenever something they're subscribed to happens (e.g., a project being created). It previously only worked against the old database (Neo4j); now it works the same way against the new one (Postgres) too.

A few things worth knowing going in:
- **No permission system change needed.** Webhooks don't have a separate access-control layer — security works entirely by only ever showing or touching a user's *own* webhooks. That's baked directly into every query here, matching how the old database already enforced it.
- **Two unrelated test-setup bugs got fixed along the way.** They were found while turning on this feature's test file for Postgres, but they affect *any* test that asks for a temporary Postgres database and custom config at the same time — not just this one. See "What's added" below.
- **One dormant edge case was found and fixed during review.** If a user's webhook secret is ever missing (which shouldn't currently happen — nothing deletes it separately from the webhook), the Postgres version would have shown that webhook with a blank secret instead of just hiding it, unlike the old database. Fixed to match; not reachable in production today.

Branch `webhooks-pg`, rebased directly onto `develop`. This closes out the full database migration for repositories — every domain now has a Postgres implementation.

### What's added

- **Migration 0037** — four new tables: `webhooks`, `webhook_executors`, `broadcast_channels`, `webhook_channel_observations`. Schema below.
- Two new Postgres repositories (webhook management + webhook channel tracking), wired in through the existing routing layer so the old database keeps working until Postgres is switched on.
- Owner-scoped access control: every read/write filters to the logged-in user's own webhooks, directly in the repository.
- Removed the test skip that had kept the entire webhooks test file from running under Postgres — it now runs unconditionally on both databases in CI.
- Fixed two bugs in the shared test app-setup helper, found while removing that skip:
  - It always created a brand-new temporary Postgres database, even when a caller had explicitly asked to reuse another one already in use — silently creating and connecting to the wrong database.
  - Custom test configuration was silently getting dropped whenever a temporary Postgres database was also requested, because both were registered under the same internal slot and only the last one registered actually took effect.
- Fixed the dormant secret-fallback bug mentioned above: a webhook whose secret record is missing is now excluded from results, matching the old database, instead of showing up with a blank secret.
- Rerouted two tests that used to simulate "a deploy just happened" through old-database-only internals that don't exist under Postgres — they now call the Postgres migration step directly instead.
- Widened an existing structural check (that every Postgres repository properly notifies the app when its data changes) to also look in one more source folder it had been silently skipping — this exact set of files lived there and would have gone unchecked otherwise.

### Schema (migration 0037)

```sql
CREATE TABLE "webhook_executors" (
  "user_id"    text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "secret"     text NOT NULL,
  "created_at" timestamptz NOT NULL DEFAULT now(),
  "updated_at" timestamptz NOT NULL DEFAULT now(),
  CONSTRAINT "webhook_executors_pkey" PRIMARY KEY ("user_id")
);

CREATE TABLE "webhooks" (
  "id"           text NOT NULL,
  "owner_id"     text NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "key"          text NOT NULL,
  "name"         text NOT NULL,
  "subscription" text NOT NULL,
  "variables"    jsonb,
  "url"          text NOT NULL,
  "metadata"     jsonb,
  "valid"        boolean NOT NULL DEFAULT true,
  "created_at"   timestamptz NOT NULL DEFAULT now(),
  "modified_at"  timestamptz NOT NULL DEFAULT now(),
  CONSTRAINT "webhooks_pkey" PRIMARY KEY ("id")
);
CREATE UNIQUE INDEX "webhooks_owner_key_unique" ON "webhooks" ("owner_id", "key");

CREATE TABLE "broadcast_channels" (
  "name"       text NOT NULL,
  "created_at" timestamptz NOT NULL DEFAULT now(),
  CONSTRAINT "broadcast_channels_pkey" PRIMARY KEY ("name")
);

CREATE TABLE "webhook_channel_observations" (
  "webhook_id"    text NOT NULL REFERENCES "webhooks"("id") ON DELETE CASCADE,
  "channel_name"  text NOT NULL REFERENCES "broadcast_channels"("name"),
  "evaluated_at"  timestamptz NOT NULL DEFAULT now(),
  CONSTRAINT "webhook_channel_observations_pkey" PRIMARY KEY ("webhook_id", "channel_name")
);
CREATE INDEX "webhook_channel_observations_channel_name_idx"
  ON "webhook_channel_observations" ("channel_name");
```

No backfill or data migration needed — production currently has zero rows across all the tables this replaces, so this is a clean, empty start.

### Validation performed

- `yarn type-check` — clean.
- `yarn lint` — clean.
- Webhooks end-to-end test suite, run twice under Postgres: **59 passed / 2 pre-existing todo / 0 failed**, identically both times. Covers save, list, read-by-key, delete, secret rotation, and the multi-owner lookup path used by the channel repository.
- Both runs also hit one unrelated, reproducible issue at the very end of the suite: a connection error to the old database during final cleanup, after every actual test had already passed. The old database's own logs show no errors, restarts, or crashes the whole time, and it responds normally immediately after. This looks like a pre-existing test-setup quirk — likely a connection that's never used during an all-Postgres run going stale before cleanup tries to close it — unrelated to anything changed in this PR. Flagged separately, not fixed here.
- Not re-run against the old database in this session; not expected to be affected, since the only touched test logic (the two rerouted tests) only takes its new path when Postgres is the active database.

### Reviewer cross-checks

- Confirm both repository registrations in the webhooks module list only a Postgres implementation (no third database arm) — correct, since this domain has no separate future migration target.
- Confirm `webhooks_owner_key_unique` (`owner_id` + `key`) is the right stand-in for the old database's implicit per-user uniqueness.
- Confirm migration 0037's position in the migration list doesn't collide with anything else (checked at rebase time — no duplicates found).
- Confirm the two test-setup fixes don't change behavior for any other existing test that relies on both a custom config *and* a temporary Postgres database at once — only that combination is affected, and it previously silently dropped the custom config.

### Explicitly out of scope

- The old database's own migration-bookkeeping system is untouched — it stays old-database-only by design and is removed later, once the whole migration is complete.
- The pre-existing cleanup-time connection quirk noted under Validation is not fixed in this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,8 +117,9 @@ jobs:
           # check), so they are reported as skipped, appear in the run output,
           # and state their reason — instead of being invisible. Currently: the
           # three *-changeset-aware specs (changesets are not being carried
-          # forward) and webhooks (not ported yet). Add that condition to the
-          # test file — never a filter here.
+          # forward). Add that condition to the test file — never a filter
+          # here. (Webhooks used to be listed here too — ported 2026-08-05,
+          # its spec now runs unconditionally on both engines.)
           #
           # Single shard on purpose: the full suite is ~10 min serial, and six
           # jobs each paying the full setup cost was pure runner waste. Bump
--- a/src/core/drizzle/live-query-invalidation.spec.ts
+++ b/src/core/drizzle/live-query-invalidation.spec.ts
@@ -31,6 +31,13 @@ import { join, posix, relative, sep } from 'node:path';
  *   `src/components/audit/resource-mutation.repository.ts` is the one such file
  *   today (append-only audit trail, nothing subscribes to it, so nothing to
  *   invalidate). Renaming it to match the convention is tracked separately.
+ *
+ * The walk covers `src/components` AND `src/core`: Postgres repos aren't only
+ * under `src/components` (`src/core/authentication/authentication.drizzle.repository.ts`
+ * and the webhooks repos below live under `src/core`), and scoping the walk to
+ * `src/components` alone would silently exempt every one of them from this
+ * check — the exact same blind spot that let the whole Webhooks domain go
+ * unported with no tracker row in the first place.
  */
 
 /**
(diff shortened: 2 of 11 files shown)


bryanjnelson (changes requested):
One nit.

bryanjnelson (approved):
Let us go!

PR #3850, opened 2026-08-13
[pg] Wire cross-domain project/engagement filters, port ProgressReport, and close the UAT-found authz gaps

rdonigian:
### Summary

This PR wires up a batch of project/engagement search filters that used to just throw an error under the new database, adds a Postgres version of the progress-report list query (the last major query that didn't have one), and fixes a sizable batch of real bugs found while testing all of that against a copy of production data.

The bugs fixed here matter because several of them are security- or data-integrity-shaped, not just "a filter doesn't work yet":

- **A filter meant to show "my projects" was showing everyone's.** Under the new database, the check that should have limited results to the current user's own projects was missing entirely. For most users this was hidden by a separate permission check elsewhere, but for a few roles with broader read access (like Financial Analyst), it meant seeing far more projects than they should have.
- **Anyone signed in could delete any user account.** The new database's user-deletion code had no permission check at all — the old database's equivalent code did.
- **Deleting a project didn't hide its data everywhere.** Ceremonies and project memberships belonging to a deleted project kept showing up in full, including sensitive membership details, instead of disappearing along with the project.
- **A brand-new "who changed what" audit trail had no access control at all.** Anyone could read any resource's full change history. It's now restricted to administrators as a stopgap until proper per-resource permissions are built.
- **A missing reference (like a deleted director) used to crash the whole record instead of just leaving that one field blank.**
- Several smaller data-correctness gaps: a filter that silently ignored one of its inputs, a file search where typing a literal `%` matched everything instead of nothing extra, deleted files' associated data still showing up, and a couple of places where other users watching a page in real time wouldn't see something get deleted.

On top of all that, during this session I ran an automated adversarial code review plus hands-on testing against the rebased branch, which caught one more real bug and two smaller gaps — all described in "What's added" below.

Branch `pg-wire-cross-domain-filters`, rebased directly onto `develop` (it was waiting on a separate, larger PR to merge first — that landed earlier today). No database schema changes; this is all application code.

### What's added

**Filters and reads wired up (previously threw "not supported" errors under the new database):**
- Engagement's project sub-filter, and Project's partner/partnership/primary-partnership and "only multiple engagements" filters — all now work now that the domains they depend on have landed.
- Project and Engagement's "uses Rev79 tool" filter, and the underlying `usesRev79` field, which always read as false before. Fixed to also catch tool usage recorded through an engagement, not just directly on the project (a real gap it also missed before this fix).
- The standalone progress-report list query, ported to the new database for the first time — it had never been wired up at all, so it silently kept reading the old database regardless of configuration. Read/create/update/delete for the underlying record are unaffected (they already went through a shared, already-ported code path); this port only covers the list/search query.

**Real bugs found and fixed (via a structured test pass against a copy of production data):**
- "My projects" / "projects I'm a member of" filter was missing its current-user check entirely (see summary above). Now scoped correctly, and fails safe (returns nothing) rather than fails open if the current user can't be determined for some reason.
- User deletion had no permission check under the new database.
- Ceremonies and project memberships kept resolving with full detail after their parent project or engagement was deleted.
- The new resource change-history feature had no access control; now restricted to administrators.
- A missing single reference (e.g., a deleted director) crashed the entire parent object instead of leaving just that field empty.
- Two shared permission-check files had no logic at all for the progress-report type, which would have crashed the query for most roles the moment real data hit it.
- Partner and Organization records weren't correctly recognizing when the current user was actually a member, causing an internal error instead of correctly evaluating permission for legitimate members.
- A region filter silently ignored one of its inputs.
- Saving a new version of a media file dropped its alt-text/caption instead of carrying it forward.
- File search by name didn't escape a literal `%`, so it matched far more than intended.
- Reading extraction results for a deleted file still worked when it shouldn't have.
- Deleting a comment or a post didn't notify other users viewing the same page in real time.

**Found and fixed in this session, after rebasing:**
- **Partnership "primary" flag bug** (found by hands-on testing): marking a second or third partnership as the primary one for a project silently failed to actually take over — the old primary stayed marked as primary too, because of an ordering bug (the code tried to grab "primary" before releasing it from the old holder, which failed, and that failure was wrongly treated as an unrelated rare edge case instead of the normal request it was). Fixed by releasing the old primary first, matching how editing an existing partnership into primary already worked correctly. Applied identically across all three ways this data can be stored, though only the new database actually had the bug.
- **A related filter gap** (found by automated review): the "my projects" fix above wasn't carried through to a few places where that same filter can be nested inside another one — e.g., asking for engagements belonging to "my projects." Those nested cases were silently returning wrong results instead of a visible error. Fixed by passing the current user through everywhere it was missing.
- **A sorting gap** (found by automated review, low severity): a handful of progress-report sort options that work on the old database aren't implemented yet on the new one, and silently default to a different sort instead of saying so. Documented as a known, deferred gap, matching how an identical gap is already documented elsewhere in the codebase.
- **Test coverage for the new admin-only history restriction:** the tests exercising that feature were still running as a non-admin and started failing once the restriction was added. Updated them to read history as an admin (without changing the main test user's role, since other parts of those same tests depend on it), and added a new test proving a non-admin is actually rejected — that protection had no test coverage in either direction before.

### Validation performed

- Type-check and lint: clean throughout.
- Every touched test file re-run in isolation under the new database after all fixes above:
  - `partnership.e2e-spec.ts` — 14/14 passing (previously had 1 failure — the primary-flag bug).
  - `engagement.e2e-spec.ts` — 32/33 passing (1 pre-existing, unrelated skip).
  - `project.e2e-spec.ts` — 30/30 passing.
  - `progress-report-media.e2e-spec.ts` — 11/11 passing.
  - `audit-log.e2e-spec.ts` — 13/13 passing (previously had 6 failures from the new admin-only restriction; 1 new test added).
  - `organization.e2e-spec.ts` — 10/14 passing (4 pre-existing, unrelated skips).
- All of the above runs hit one recurring, unrelated issue at the very tail end of the process: a database connection error during final test-runner shutdown, after every real test had already passed. It doesn't reproduce on shorter runs and looks like a pre-existing quirk in the shared test setup, not something this PR causes — flagged separately, not fixed here.
- Not re-run against the old database engine in this session. The two places here with different logic per engine (the progress-report routing, and the partnership primary-flag fix) were checked by hand for the old-database path — the old database has no equivalent bug (it has no database-level uniqueness constraint to conflict with, so the ordering bug this PR fixes doesn't apply there) — but no automated old-engine test run backs that up beyond the hand-check.

### Reviewer cross-checks

- The "my projects" scoping fix and its nested-filter follow-up fail closed (return nothing) rather than fail open whenever the current user can't be determined — worth double-checking there's no code path that calls these filter functions in a context with no logged-in user where "return nothing" would be the wrong behavior (e.g. a background job).
- The administrator-only restriction on resource history is explicitly called out as an interim measure, not a final permissions design — confirm that framing matches expectations before this merges, since it's a real behavior change for anyone currently relying on non-admin access to that data (unlikely, since it's not exposed in any UI yet).
- The partnership primary-flag fix changes what `create()` returns (an extra field) on all three storage implementations — confirm nothing outside this PR was relying on the old return shape.

### Explicitly out of scope

- No database schema changes.
- The progress-report sort gap (engagement/PnP/summary-based sorting) is documented but not implemented — matches an already-accepted gap elsewhere, not addressed here.
- Full old-database-engine re-validation of this session's four newest commits wasn't run; only reasoned through by hand (see Validation above).
- The recurring test-teardown connection quirk noted under Validation is not fixed in this PR.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,6 +16,15 @@ concurrency:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    env:
+      # Type-aware rules and `tsc` itself both build a full TypeScript program
+      # in memory, and that's grown enough (dozens of new domains from the
+      # Postgres migration) to hit Node's default ~4GB heap ceiling on
+      # ubuntu-latest runners — an OOM crash mid-step, not a real problem with
+      # the code. Runners have far more RAM than that default; just raise the
+      # ceiling for every step in this job, not only whichever one crashed
+      # most recently.
+      NODE_OPTIONS: '--max-old-space-size=8192'
 
     steps:
       - uses: actions/checkout@v4
--- a/src/common/functions/secured-mapper.ts
+++ b/src/common/functions/secured-mapper.ts
@@ -15,8 +15,15 @@ export async function mapSecuredValue<T extends Secured<any>, S>(
   if (!rest.canRead || value == null) {
     return rest;
   }
-  const mapped = await mapper(value);
-  return { ...rest, value: mapped };
+  try {
+    const mapped = await mapper(value);
+    return { ...rest, value: mapped };
+  } catch (e) {
+    if (e instanceof NotFoundException) {
+      return rest;
+    }
+    throw e;
+  }
 }
 
 /**
--- a/src/components/audit/resource.resolver.ts
+++ b/src/components/audit/resource.resolver.ts
@@ -1,7 +1,8 @@
 import { Info, Parent, ResolveField, Resolver } from '@nestjs/graphql';
 import { stripIndent } from 'common-tags';
 import { type GraphQLResolveInfo } from 'graphql';
-import { ListArg, Resource } from '~/common';
+import { ListArg, Resource, UnauthorizedException } from '~/common';
+import { Identity } from '~/core/authentication';
 import { AuditService } from './audit.service';
 import {
   ResourceMutationList,
@@ -16,7 +17,10 @@ import {
  */
 @Resolver(Resource)
 export class ResourceHistoryResolver {
-  constructor(private readonly audit: AuditService) {}
+  constructor(
+    private readonly audit: AuditService,
+    private readonly identity: Identity,
+  ) {}
 
   @ResolveField(() => ResourceMutationList, {
     description: stripIndent`
@@ -33,6 +37,16 @@ export class ResourceHistoryResolver {
     @Info() info: GraphQLResolveInfo,
     @ListArg(ResourceMutationListInput) input: ResourceMutationListInput,
   ): Promise<ResourceMutationList> {
+    // Interim gate: this field isn't behind any per-resource/per-field policy
+    // yet (unlike the rest of the schema, which goes through Privileges), and
+    // it's not exposed in any UI. Restricting it to admins keeps it from being
+    // a wide-open read of every resource's mutation history — the finer-grained
+    // permission model is deferred to a post-migration pass.
+    if (!this.identity.isAdmin) {
+      throw new UnauthorizedException(
+        'Only administrators can view resource history',
+      );
+    }
     // `info.parentType.name` is the concrete GraphQL type GraphQL already
     // resolved this object to (e.g. MomentumTranslationProject), which is what
     // the firing service records under. NOT `resource.__typename` — for the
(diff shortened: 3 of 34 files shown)


bryanjnelson (approved):
LGTM!

PR #3851, opened 2026-08-14
Default the users list to sorting by full name instead of id

rdonigian:
### Summary 

The user directory (the `users` list query) was sorting by an internal id — effectively random order, since that id has nothing to do with when someone was created or their name. Every other list in the app that has a name field (languages, organizations, locations, projects, etc.) already defaults to sorting by name; the user list was the one exception. Anyone browsing the full list of users would see them in a meaningless order and have to sort client-side or hunt for the person they wanted.

The fix flips the default to sort by full name (first name, then last name as a tiebreaker). The one wrinkle: this app is mid-migration between three different database implementations, and only the current one had "sort by name" actually built for the user list — the two newer implementations would have silently kept sorting by id even after the default changed, since they fall back silently instead of erroring on an unsupported sort. This PR builds that sorting logic into both of the newer implementations too, so the fix holds regardless of which database is active.

Branch `user-list-default-sort-by-name`, based directly on `develop`. Single commit, no dependencies on other in-flight work.

### What's added

- Changed the default sort key for the users list from `id` to `fullName`.
- Implemented full-name sorting (first name, then last name) on the two newer database implementations, matching what the current, live database already does.
- Exposed one small type from the shared data-access layer that was needed to express a two-column sort order — a one-line addition following the existing pattern for that layer's exports.
- Corrected a comment in the shared sorting helper that specifically referenced "the user list defaults to sorting by id," which is no longer true anywhere in the app.
- Added an end-to-end test: creates a few users with names deliberately out of alphabetical order and confirms the list comes back correctly sorted.

### Validation performed

- `yarn lint` — clean, zero warnings
- `yarn type-check` — clean
- `test/user.e2e-spec.ts` run against the current live database — 13/13 passed, including the new sort test
- `test/user.e2e-spec.ts` run against Postgres (one of the two newer implementations) — 13/13 passed, including the new sort test
- The third implementation (Gel) was verified by type-checking and by matching an existing, already-reviewed pattern elsewhere in the codebase (see below) — it isn't reachable in any environment currently in use, so there was no live database to test it against.

### Reviewer cross-checks

- The Gel-side change is currently unreachable code — nothing in this codebase is configured to route user list queries there today. It's included anyway because it's a real gap and this is the natural place to close it; worth confirming you agree that's the right call rather than leaving it for whenever Gel is actually wired up.
- The multi-column sort applies the same ascending/descending direction to both first and last name together, rather than letting them be independent. That matches the current live database's existing behavior.

### Explicitly out of scope

- There's a broader, separate issue: right now, requesting an unsupported or unrecognized sort key anywhere in the app silently falls back to sorting by id instead of returning an error. This PR only changes the User list's *default* value — it doesn't change that silent-fallback behavior, which affects many other lists beyond just users and deserves its own decision before being changed everywhere.
- This sorts by a person's real first/last name, not their display name (some users have a different display name set). That matches what the current live database already does — changing which name is used for sorting would be a separate, deliberate decision, not a byproduct of this fix.

### DIFF
--- a/test/user.e2e-spec.ts
+++ b/test/user.e2e-spec.ts
@@ -266,9 +266,24 @@ describe('User e2e', () => {
 
   it('lists users sorted by full name by default', async () => {
     const prefix = 'ZzzSortTest' + (await generateId());
-    await createPerson(app, { realFirstName: `${prefix}_Charlie` });
-    await createPerson(app, { realFirstName: `${prefix}_Alice` });
-    await createPerson(app, { realFirstName: `${prefix}_Bob` });
+    await createPerson(app, {
+      realFirstName: `${prefix}_Charlie`,
+      realLastName: 'Smith',
+    });
+    await createPerson(app, {
+      realFirstName: `${prefix}_Alice`,
+      realLastName: 'Smith',
+    });
+    // Same first name, different last names — verifies fullName sorting
+    // breaks ties on last name instead of stopping at first name.
+    await createPerson(app, {
+      realFirstName: `${prefix}_Bob`,
+      realLastName: 'Young',
+    });
+    await createPerson(app, {
+      realFirstName: `${prefix}_Bob`,
+      realLastName: 'Adams',
+    });
 
     const { users } = await app.graphql.query(
       graphql(`
@@ -278,6 +293,9 @@ describe('User e2e', () => {
               realFirstName {
                 value
               }
+              realLastName {
+                value
+              }
             }
           }
         }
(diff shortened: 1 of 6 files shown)


test/user.e2e-spec.ts
```diff
@@ -210,6 +210,34 @@ describe('User e2e', () => {
...
+      { prefix },
+    );
+
+    expect(users.items.map((user) => user.realFirstName.value)).toEqual([
+      `${prefix}_Alice`,
+      `${prefix}_Bob`,
+      `${prefix}_Charlie`,
+    ]);
+  });
+
```
bryanjnelson, line 317: There's a gap here in test coverage.  The test is trying to affirm that `fullName` is working correctly as a sort, but the data that is set here is only `realFirstName`.  There either needs to be another test to set `realLastName` and validate that, or more likely `realLastName` just needs added in here to be more comprehensive.

bryanjnelson (changes requested):

bryanjnelson (dismissed):
👍🏼

bryanjnelson (approved):

PR #3854, opened 2026-08-14
Implement Engagement name/engagedName/startDate/endDate filters on Postgres

rdonigian:
### Summary

The engagement list screen lets people search and narrow results — by typing a name, or picking a date range. Four of those search options were completely broken on the new database: searching by name, searching by the specific language or person on the engagement, and filtering by start or end date. Any attempt to use them just failed outright with a "not implemented" error.

An engagement is either tied to a language (translation work) or a person (an internship). The "name" search is supposed to match the project's name, OR the language's name, OR the person's name — whichever applies. There's also a narrower version that only matches the language or person, deliberately leaving the project name out, for when someone wants "the engagement with this specific language" without also pulling in every engagement whose project happens to share that word.

The date filters are trickier because an engagement's start and end dates are often not set directly — they fall back to the dates of the project it belongs to. Filtering "engagements starting after such-and-such date" has to check the engagement's own date first and fall back to the project's date, or it would wrongly skip engagements that never got their own override.

This branches off `develop` (1 commit, ~230 lines) and updates `engagement.drizzle.repository.ts`.

### What's in scope

- `name` and `engagedName` filters, matching against project/language/intern name text.
- `startDate`/`endDate` range filters, using the same project-date fallback that engagement sorting already relies on — filtering and sorting now agree on what an engagement's "real" date is, instead of risking two different answers.

### Design decisions

- The date fallback logic is pulled out into a shared piece of code used by both sorting and filtering, rather than duplicating it — they need to answer the exact same question ("what's this engagement's actual date?") and drifting apart would produce inconsistent results between the two.
- The name searches don't rely on the data being joined together in one query — the engagement, project, language, and person all live in separate lookups. Each name filter runs a small lookup first (which projects/languages/people match this text?) and then matches engagements against those results.

### Validation

- `yarn type-check` and `yarn lint` both clean.
- `test/engagement.e2e-spec.ts` passes in full against Postgres (35/36, the 1 skip pre-existing and unrelated), including 4 new tests covering each filter.

### DIFF
--- a/src/components/engagement/engagement.drizzle.repository.ts
+++ b/src/components/engagement/engagement.drizzle.repository.ts
@@ -1,9 +1,23 @@
 import { Injectable } from '@nestjs/common';
-import { and, eq, inArray, isNull, sql, type SQL } from 'drizzle-orm';
+import {
+  and,
+  eq,
+  gt,
+  gte,
+  ilike,
+  inArray,
+  isNull,
+  lt,
+  lte,
+  or,
+  sql,
+  type SQL,
+} from 'drizzle-orm';
 import { difference } from 'lodash';
 import { DateTime } from 'luxon';
 import {
   CalendarDate,
+  type DateFilter,
   DuplicateException,
   EnhancedResource,
   generateId,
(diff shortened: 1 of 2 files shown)


brentkulwicki (dismissed):
LGTM

bryanjnelson (approved):
👍🏼

PR #2910, opened 2023-10-06
Added Short Label to prompts and removed prompts for  progress report…

atGit2021:
… highlights

┆Issue is synchronized with this [Monday item](https://seed-company-squad.monday.com/boards/3451697530/pulses/5290576887) by [Unito](https://www.unito.io)

PR #3869, opened 2026-08-26
Add smoke coverage for the 25 untested mutations

rdonigian:
Comparing the 137 fields on the Mutation type against every mutation used anywhere in test/ — as a GraphQL document or through a test/utility helper — left 30 with no test at all. Every one sits in a domain that already has a Drizzle repository, so they were ported code with nothing looking at them, and that is precisely where the last two defects turned up: a stored column Postgres never maintained, and a resolver returning a hardcoded null.

These assert only that each mutation runs and returns the right shape. That is deliberate — the job is to find mutations that are unwired or broken on one engine, not to pin down their semantics. Where this file turns something up, the real test belongs in that domain's own spec.

It earned its keep immediately. Run against develop, setProjectTypeFinancialApprover passes on Neo4j and fails on Postgres, because the financial-approver domain has no Drizzle repository here yet — the port is sitting on pg-financial-approvers, unmerged. That one test runs on Neo4j and is held back on Postgres, tagged migration-todo to switch back to a plain `it` when the port lands. Nothing else differs between the engines.

Two of the thirty are reachable and simply not done: updateMediaMetadata needs the existing image-upload flow, and reextractPnpProgress needs a PnP spreadsheet fixture that does not exist anywhere in test/. Three more are out of scope and say why in the file header, rather than being dropped quietly.

Two things the pass established along the way, both recorded where they matter: the highlight prompt list is empty on purpose — PR #2910 removed it — so those four mutations cannot be reached by design, which the file now pins rather than skips; and producible deletes ARE reachable by an Administrator, contradicting a comment on the delete-tests branch that has since been corrected.

### DIFF
--- a/test/mutation-smoke.e2e-spec.ts
+++ b/test/mutation-smoke.e2e-spec.ts
@@ -1,0 +1,765 @@
+import { faker } from '@faker-js/faker';
+import { beforeAll, describe, expect, it } from '@jest/globals';
+import { CalendarDate, type ID, Role } from '~/common';
+import { graphql } from '~/graphql';
+import {
+  createLanguage,
+  createProject,
+  createSession,
+  createTestApp,
+  createTool,
+  fragments,
+  registerUser,
+  runAsAdmin,
+  type TestApp,
+  type TestUser,
+} from './utility';
+
+/**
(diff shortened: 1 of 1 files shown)


bryanjnelson (approved):
I ❤️ this one very hard.

PR #3874, opened 2026-08-31
Add a read-only maintenance mode

rdonigian:
> [!IMPORTANT]
> Setting **`READ_ONLY=true`** freezes the data while keeping the API up. People can still **sign in and read everything**, but anything that would change data — from any account, including admins — is refused with a clear *"CORD is temporarily read-only while maintenance is underway. Viewing data still works, but changes cannot be saved right now."* error (code `ReadOnlyMode`, HTTP 200). Built for the database cutover's freeze window; reusable for any future maintenance window. Merging this changes nothing on its own — the flag defaults off.

```mermaid
flowchart LR
  OP(["GraphQL operation"]) --> KIND{"kind?"}
  KIND -->|"query /<br/>subscription"| RUN(["runs normally"])
  KIND -->|mutation| WHO{"only login /<br/>logout fields?"}
  WHO -->|yes| RUN
  WHO ==>|"anything else"| REF(["refused with<br/>ReadOnlyMode"])
```

*Every operation — HTTP and WebSocket share the same pipeline — passes this decision while the flag is on. With the flag off, the check steps aside entirely.*

### How the request freeze works

- One plugin sits ahead of GraphQL execution and judges the fields an operation would **actually execute**, using the executor's own field-collection logic. A write hidden behind a fragment spread is caught; a write field disabled by `@skip`/`@include` doesn't count against the request, so a login that carries directive-disabled fields still goes through.
- **Signing in and out stays available on purpose.** Sessions are not part of the frozen data — they don't survive the maintenance the freeze exists for — and nobody can read anything without signing in. Password changes, resets, and registration are *not* exempt: those write real user data, which would be silently lost when the frozen database is replaced.
- The refusal is a normal GraphQL error (`codes: [ReadOnlyMode, ServiceUnavailable, Transient, Server]`) with a user-facing message, so existing clients surface it wherever they already show mutation errors — no client change required.

### Boot-time writes are frozen too

- The app normally writes at startup: Neo4j index creation and migrations, root user / default organization sync, and Postgres migrations at boot. With `READ_ONLY=true` all of these are skipped, regardless of their individual flags — so restarting the frozen instance cannot drift the database underneath a maintenance operation reading it.
- All three engine variants of the root-object sync now honor the same `DB_ROOT_OBJECTS_SYNC` switch; previously only the Neo4j one checked it.
- Dev-machine nicety that fell out of this: whether root objects sync at boot in development is now judged by the **active engine's own URL locality** (Postgres checks the Postgres URL) instead of always Neo4j's — so developing against Postgres with a remote Neo4j URL configured no longer skips local Postgres bootstrap.

<details>

- e2e `read-only-mode.e2e-spec`: **3/3 on Neo4j and 3/3 on Postgres** — queries answer, a data mutation is refused with `ReadOnlyMode`, signing out and back in works while frozen.
- Unit **13/13** on the plugin: allowlist, mixed operations, fragment spreads, inline fragments, multi-operation documents, `__typename`-only mutations, and the three directive cases (`@skip`, variable-driven `@include`, a whole fragment disabled by `@skip`).
- Lint 0 warnings · type-check 0 errors.
- Verified live against a real Neo4j with the actual env var (not a test override): boot logs the warning, login as the root admin works, the same root admin's `createOrganization` is refused, and a full before/after node count reconciles exactly — the only new rows are the session tokens the freeze deliberately allows, proving the boot itself wrote nothing.
- The refusal's HTTP status was measured (200 with the error in the GraphQL response), so uptime monitoring won't page during an intentional maintenance window.

</details>

<details>
<summary><strong>Accepted limitations and follow-ups</strong></summary>

- **Queued background jobs are outside the freeze.** A BullMQ job already in Redis when the flag flips can still run and write. The only queue user today is webhook delivery, and production has zero registered webhooks (measured), so this is currently unreachable; the config doc carries the "let the queue drain first" guidance for when webhooks gain a consumer.
- **No UI banner.** Clients learn about the freeze when a save fails with the friendly message. A maintenance flag the UI could read up front (e.g. on the `session` query) is a possible follow-up if the freeze window wants better UX.
- The WebSocket transport shares the enforced pipeline (verified by code trace and covered at the unit level) but isn't exercised by the e2e suite.

</details>

### DIFF
--- a/src/common/exceptions/index.ts
+++ b/src/common/exceptions/index.ts
@@ -8,6 +8,7 @@ export * from './not-implemented.exception';
 export * from './unauthenticated.exception';
 export * from './unauthorized.exception';
 export * from './service-unavailable.exception';
+export * from './read-only-mode.exception';
 export * from './invalid-id-for-type.exception';
 export * from './creation-failed.exception';
 export * from './date-override-conflict.exception';
--- a/src/common/exceptions/read-only-mode.exception.ts
+++ b/src/common/exceptions/read-only-mode.exception.ts
@@ -1,0 +1,17 @@
+import { ServiceUnavailableException } from './service-unavailable.exception';
+
+/**
+ * A write was attempted while the API is in read-only maintenance mode.
+ * @see ReadOnlyModePlugin in ~/core/graphql for what enforces this.
+ */
+export class ReadOnlyModeException extends ServiceUnavailableException {
+  constructor(message?: string, previous?: Error) {
+    super(
+      message ??
+        'CORD is temporarily read-only while maintenance is underway. ' +
+          'Viewing data still works, but changes cannot be saved right now. ' +
+          'Please try again later.',
+      previous,
+    );
+  }
+}
--- a/src/components/admin/admin.drizzle.service.ts
+++ b/src/components/admin/admin.drizzle.service.ts
@@ -25,6 +25,9 @@ export class AdminDrizzleService implements OnApplicationBootstrap {
   }
 
   async onApplicationBootstrap(): Promise<void> {
+    // Same switch the Neo4j AdminService honors — also how read-only
+    // maintenance mode keeps boot from writing.
+    if (!this.config.dbRootObjectsSync) return;
     const finishing = this.repo.finishing(async () => {
       await this.setupRootUser();
       await this.setupDefaultOrg();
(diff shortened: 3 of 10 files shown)


brentkulwicki (approved):
I could take or leave the code rabbit suggestion, so I'll leave that up to you. But LGTM

bryanjnelson (commented):
These tests make me feel good about this @rdonigian!

PR #3882, opened 2026-09-08
Collapse the 43 Postgres migrations into one genesis migration

rdonigian:
> [!IMPORTANT]
> The **43 hand-written Postgres migration files are replaced by a single `0000_genesis.sql`** holding the entire schema. **Databases that are already up to date are not touched** — they skip the new file entirely and keep running. Only a brand-new database reads it, and it lands on a schema **byte-identical** to what the 43 files produced.
>
> The one thing to know before deploying: a database that was left **part-way** through the old sequence can no longer be brought up to date, because the individual files it still needed are gone. It used to boot onto a half-built schema and report success; it now **refuses to boot and says it must be rebuilt**.

```mermaid
flowchart TD
    Boot["App boots, migrator runs"] --> Q{"Highest timestamp recorded<br/>in the migrations table?"}

    Q -->|"nothing recorded<br/>(brand-new database)"| Apply["Apply 0000_genesis.sql<br/>→ full 67-table schema"]
    Q -.->|"= retired 0042<br/>(every deployed database)"| Skip["Skip it — genesis is older,<br/>so nothing re-runs"]
    Q ==>|"strictly between genesis<br/>and retired 0042"| Refuse["Refuse to boot:<br/>stranded, must be rebuilt"]

    style Refuse stroke-width:3px
```

*Solid = a fresh database's path. Dashed = the unchanged path every deployed database takes. Thick = the new guard this PR adds, replacing a silent half-built boot.*

### Why one file, and how it was built

Generated from a `pg_dump --schema-only` of a database built by applying all 43 migrations in order — deliberately **not** by concatenating them. Concatenation would preserve dead history: `0041` replaces the trigger function `0010` defines, and `0030`/`0042` relax constraints set up earlier. Dumping the end state keeps only the definitions that actually survive.

Two classes of line had to come out of the dump, and both matter:

- **psql meta-commands.** `pg_dump` 16 emits `\restrict` / `\unrestrict`, which are not SQL. The migrator sends the file straight to the server, which would reject them.
- **The session preamble.** `SET statement_timeout`, and especially `set_config('search_path', '')`. Migrations run on the application's *pooled* connection, so anything set there would outlive the migration and quietly reconfigure that connection for the rest of the process's life. Dropping the `search_path` line is safe precisely because `pg_dump` emitted every name schema-qualified under it.

The generator ships beside the file as `build-genesis.mjs`, carrying the full re-baseline and verification recipe. A generated 4,500-line baseline that nobody can re-derive isn't reviewable, and the file itself says to add a normal numbered migration rather than hand-edit it. Nothing imports the generator; the migrator only ever reads the `.sql` named by the journal.

### Why deployed databases are unaffected

Drizzle picks what to run by **comparing timestamps, never file hashes**: it reads the single highest `created_at` from its bookkeeping table and applies any journal entry whose `when` exceeds it. The recorded hash is written but never compared, so swapping 43 files for one is invisible to a database that already ran them.

So the journal entry keeps the **original `0000` timestamp**. Any database that ran the old sequence sits above it and skips the file; a brand-new database has nothing recorded and applies it. Raising the timestamp instead would be actively wrong — it would make already-migrated databases try to re-run genesis over their existing schema.

### Two hazards that split creates

Fresh and deployed databases no longer converge on the same high-water mark, and that opens a window between them. Both ends are now closed.

**A stranded database no longer boots quietly.** Its mark sits above genesis, so it skips the new file too — and the files it still needed no longer exist. The migrator applied nothing, succeeded, logged "migrations complete", and let the app start against a schema missing columns, which surfaced later as confusing runtime errors. It now fails immediately and names the real problem. Both bounds are exclusive, so the two healthy states pass: a database that finished the old sequence records exactly the retired end, and one built from genesis records exactly genesis.

> [!WARNING]
> A database not fully through the old sequence **cannot be migrated after this deploys** — it has to be rebuilt. Every environment that auto-migrates is already past the end of the sequence and is unaffected, but a database parked behind read-only maintenance mode, a restored older snapshot, or a developer's stale local database will now refuse to boot instead of coming up half-built. That is the intended behavior; the fix is to rebuild, not to work around the guard.

**A future migration can no longer go missing in production.** Any new migration whose `when` lands inside that window would apply on fresh databases and be **silently skipped on every deployed one** — no error on either side, green in CI, missing column in production. Two things used to prevent this and neither does any more: the journal's last entry used to be the floor, so a low timestamp broke the "strictly increasing" rule; and a fresh database replayed all 43 and ended at the same mark as a deployed one, so both behaved identically and CI would show the gap.

> [!WARNING]
> Every migration added after this one must carry a `when` **later than the retired `0042`'s timestamp**, not merely later than genesis's. `Date.now()` satisfies this; a hand-picked or carried-over date may not. The journal test now fails on a violation, and the documented steps for adding a migration were corrected to say so.

### An unrelated race this surfaced

Adding two queries at boot made a pre-existing test-harness race far more likely: a suite began failing in teardown with `FATAL 57P01 — terminating connection`, *after* all of its assertions had passed. An app that shares another app's ephemeral database gets no cleanup of its own, and apps were closed in creation order — so the owner force-dropped the database out from under the sharer's still-open pool. Teardown now unwinds newest-first, which is always safe because a sharer can only be created after the owner it borrows from.

### Schema

**No schema change.** The effective DDL is identical to what the 43 migrations produced — same 67 tables, 195 indexes, 40 enum types, 100 foreign keys, 17 check constraints, the one generated column, the trigger and its function, and the non-deterministic ICU collation. This was verified by comparison, not by inspection (see Validation).

What does change is migration numbering. The journal is now a single entry:

```sql
-- meta/_journal.json, one entry
--   idx: 0, tag: '0000_genesis', when: 1745625600000  (the original 0000's)
```

So the next migration is `0001_*`, and its `when` must exceed the retired `0042`'s timestamp — the two rules the journal test now enforces alongside contiguous indexes. Any in-flight branch carrying a higher-numbered migration will need its file and journal index renumbered when it rebases.

Postgres-only by construction: the migrator returns early unless the active engine is Postgres, so nothing here runs on the Neo4j path.

<details>

**Equivalence — the central claim, verified by comparison rather than reading.** A database built from the 43 migrations and a database built from `0000_genesis.sql` were both dumped with identical `pg_dump` flags; the two dumps are **byte-identical at 115,770 bytes**. To confirm that comparison actually discriminates rather than comparing two empty things, a defect was planted (one `CREATE INDEX` removed) and it produced an 8-line diff. Re-proved after every subsequent edit to the file.

**Full e2e from scratch, Postgres engine** — `success: true`, **47 suites passed, 0 failed, 0 runtime errors**; **475 tests passed, 0 failed**, 25 skipped, 2 todo, of 502. Every spec file creates its own empty database and builds it from genesis on boot, so the new file is exercised ~49 times per run. Confirmed live mid-run that a sampled ephemeral database held **exactly one** bookkeeping row, carrying genesis's timestamp, against a full 67-table schema.

The 25 skipped tests and 4 skipped suites are pre-existing and unrelated: the four changeset-aware suites are gated off on Postgres (three by engine, one unconditionally with a standing TODO), plus scattered skips in other files. No suite ran empty — checked specifically, since an engine-gated suite that returns early reports as passed and would hide absent coverage. No test file was modified by this PR.

**Unit tests** — 19/19 suites, 225 passed, 0 failed, 3 skipped.

**`type-check`** clean. **`lint`** clean at zero warnings.

**The stranded-database guard** was exercised against the real class (not a reimplementation) across 8 seeded database states, all as expected: no bookkeeping table, table present but empty, exactly genesis's timestamp, exactly the retired end, and a timestamp past the end all boot; three interior values — mid-sequence, one step above genesis, and one millisecond below the end — all refuse. The two exclusive bounds are the cases that matter, and both pass.

**The journal floor assertion** was checked in both directions: it passes on the real journal, and planting the realistic mistake (a migration carrying a mid-2025 timestamp) makes it fail.

**The teardown fix** took the affected suite from failing roughly 1 run in 3 to **6/6 green** in isolation, and the final full run shows zero `57P01` errors, zero unhandled errors, and zero suites failing to run.

**Not run:** the e2e suite on the Neo4j engine. The migrator returns early unless the active engine is Postgres, so this change cannot affect that path — but that is reasoning, not a measurement.

</details>

<details>
<summary><strong>Accepted limitations and follow-ups</strong></summary>

**Migration history leaves the working tree.** That is the point of the change, and the 43 files remain in git history. Anything that needs the old per-file detail reads it from there.

**A database that ran only the original `0000` is indistinguishable from a genesis database** by timestamp — both record the same value — so the guard cannot catch that one case. Telling them apart would need the recorded hash. That population is empty in practice.

**The guard is transitional.** It only protects against databases built before this change, and is tagged in code for removal once no environment can predate it.

**Genesis is applied as 420 separate statements, not one.** Keeping the statement-breakpoint markers costs about 97 ms on a fresh database versus the 43 files (measured, interleaved runs) and buys a 305-byte error naming the offending statement instead of one quoting the whole 116 KB file. Deployed databases skip the file, so they pay nothing. This was a deliberate choice for diagnosability over speed.

**Pre-existing, deliberately not fixed here — the connection pool has no error handler.** The Postgres pool is created without an `'error'` listener, and node-postgres crashes the process when an idle client errors. In production that means a failover, an administrative restart, or an idle-connection timeout could take the API down rather than reconnect. It is unrelated to this change and wants its own fix.

**Pre-existing, deliberately not fixed here — the build's migration asset glob points at a directory that no longer exists,** left over from an old rename, so the build copies no migration files into the compiled output. This is harmless only because the migrator resolves the migrations folder from the working directory at runtime, which means the deployed image must continue to ship the source tree. Worth correcting, but not as part of this change.

**A comment in the test harness overstates what it does.** Two places say they share a database with the suite's app but only pass the Neo4j connection, so on Postgres each quietly gets its own database — losing the intended speedup without breaking correctness. Left alone: changing it would alter test isolation and belongs in its own change.

</details>


bryanjnelson (approved):
✅  1 < 42

PR #3884, opened 2026-09-11
Implement the list sort keys Postgres was ignoring, and put blanks last

rdonigian:
> [!IMPORTANT]
> Column headers across the app sorted the list by **a different column**. An
> unrecognised sort key didn't fail — Postgres applied the direction to the
> list's fallback column, so the list visibly reordered while the column you
> clicked stayed out of order. This answers every sort key the grids actually
> send, and fixes a separate ordering difference that affected **every list
> sorted descending**.

```mermaid
flowchart LR
  UI["column header<br/>sort: engagement.project.primaryLocation.name"]
  UI -.-> PR["progress report<br/>answers start, end, status, …"]
  PR -->|"engagement.*"| ENG["engagement<br/>answers status, dates, milestone, …"]
  ENG -->|"project.*"| PROJ["project<br/>answers name, step, status, …"]
  PROJ -->|"primaryLocation.*"| LOC["location<br/>answers name, type"]
```

*Solid arrows are the delegations this adds: each domain answers the keys it owns
and passes the rest of the key on, which is how one click reaches three tables.*

> [!WARNING]
> Two things change visibly at deploy:
> - **Every list sorted descending** on a nullable column moves its blank rows
>   from the top to the bottom. Intended — it's what makes both engines agree.
> - The engagements grid's **Language / Intern** column now orders by the name it
>   displays rather than the language's stored `name`.
>
> And one thing must not be switched on yet: the Tools grid's **Usages** column
> has to stay unsortable until the cutover, or it returns an empty grid on Neo4j.

<details>
<summary><strong>What now sorts</strong></summary>

The engagement grids' default `project.name`, plus their `project.*`,
`language.*` and `currentProgressReportDue.*` keys; the languages grid's
`ethnologue.*`, `population`, `usesAIAssistance` and `signLanguageCode`; the
dashboard widgets' `engagement.*` and summary-period keys; `title` and `status`
on users; `aiBased` and a new `usageCount` on tools; the engagement counts on
projects.

Each reproduces what Neo4j answers, quirks included, so nothing reorders at the
cutover:

- The project engagement count counts **language** engagements only, so an
  internship project sorts as zero while reporting a non-zero total of its own.
- `population` coalesces a language's override with its ethnologue value — the
  number the field itself returns. A language with neither sorts blank rather
  than dropping out of the list, as it does on Neo4j.
- The two concatenated engagement name sorts are **not** case-folded, while
  `project.name` is. A name is folded only where a sort transformer exists for
  the field being sorted, and those two are sort-only keys with no field behind
  them. Measured on languages named `Zebra` and `apple`: `Zebra | apple` for the
  name keys, `apple | Zebra` for `project.name`.

Cross-domain keys resolve through one expression per domain, which is what lets
`engagement.project.primaryLocation.name` work without every domain knowing
about the others.

</details>

<details>
<summary><strong>Blanks last in both directions</strong></summary>

Postgres puts blank values last ascending but **first** descending; Neo4j puts
them last either way. Measured on `Language.registryOfLanguageVarietiesCode`,
three rows where one code is blank: `22222 | 11111 | blank` on Neo4j against
`blank | 22222 | 11111` on Postgres.

Fixed in one place, with every hand-rolled sort site converted to go through it.
This is the change with app-wide reach, so it's where review attention is best
spent.

</details>

<details>
<summary><strong>Three deliberate differences from Neo4j</strong></summary>

- **Language / Intern** sorts by the language's display name, the value the
  column shows. Sorting by the stored `name` put rows in positions the visible
  text couldn't explain — on a scrubbed copy the two differ for every language,
  which is what made the column look unsorted.
- **Tool usage lists** get a defined order (tool name, then oldest first) where
  they had none, so a chip can't change position after an unrelated edit.
- **`usageCount`** is Postgres-only: Neo4j answers a sort key with no stored
  property behind it by returning an empty list, so there's nothing to match.

</details>

<details>

A new `test/list-sorts.e2e-spec.ts` covers the cross-domain keys and the blank
ordering: **10 tests, passing on both engines**. Every fixture set is labelled so
alphabetical order *disagrees* with the order under test — a silent fallback to
the list's default column would otherwise pass the assertions without the sort
working at all.

- `test/project.e2e-spec.ts` — 36 on both engines, including the engagement count
  sort that started this.
- `test/tool.e2e-spec.ts` + `test/tool-usage-container-parity.e2e-spec.ts` — 17 on
  Postgres, 16 plus **1 skipped** on Neo4j. The skip is `usageCount`, reported as
  skipped rather than passing against behavior that doesn't exist there.
- Regression sweep across the twelve suites touching these lists: 177 passed, 4
  skipped on Postgres; 102 passed on Neo4j.
- Lint and type-check clean.

The engine difference in Language / Intern is pinned with a per-engine expected
answer, and the `usageCount` test checks the order **and** that the displayed
totals read 0, 1, 2 in that order — tying the sort to the population the column
shows.

Implemented with no order asserted, rather than implied: the summary-period sorts
(progress summaries come from PnP extraction and no fixture can create them —
checked only for valid SQL that loses no rows) and `currentProgressReportDue.*`.

</details>

<details>
<summary><strong>Limitations and follow-ups</strong></summary>

- **Correct, not fast.** Cross-table keys compute their value with a correlated
  subquery per row. On a production-sized copy the engagement grid's default sort
  takes 82 ms where a join takes 12 ms, and Language / Intern 253 ms against
  10 ms. Subqueries compose for the nested keys, can't duplicate rows, and need no
  new query builder; converting the single-row hops to joins later changes no
  semantics and is worth doing for those two first. Either way none of these keys
  is indexable — the value lives in another table — and Neo4j also evaluates its
  sorters per row, so this is not slower than today.
- **`Tool.filteredUsages` accepts a sort argument and ignores it.** Nothing
  queries that field; wiring it is three layers and belongs with whatever builds
  a real usages grid.
- **`Pinned`, `Member`, `defaultFieldRegion` and the PnP `Errors` column** sort on
  neither engine and need a frontend change; handed off separately.
- **The users list has the same defect as Language / Intern** — sorts by the real
  name, displays the display name. Untouched here; the two should move together.
- **The read comparison doesn't exercise any of these keys**, so it won't report
  the intentional Language / Intern difference. If a case for that key is added,
  it needs a known-difference rule at the same time.

</details>


bryanjnelson (approved):
:shipit:

PR #3889, opened 2026-09-15
Advance modifiedAt when a comment or post is edited on Postgres

rdonigian:
> [!IMPORTANT]
> Editing a comment or post now advances its **"last modified"** timestamp on Postgres, matching the old engine. Before this, the field froze at the creation time forever — every edit was invisible, timestamp-wise — and the audit trail recorded a `modifiedAt` change on each edit that was never actually persisted.

```mermaid
flowchart LR
  E["edit a comment / post"] --> G["shared change-builder<br/>injects modifiedAt: now"]
  G --> N["old engine: writes every<br/>field it is handed ✓"]
  G --> P["Postgres repo: copies<br/>fields by name"]
  P -. "modifiedAt was not on the list" .-> F["frozen at creation"]
```

_Solid is the data flow; dashed is the dropped field this change fixes._

> [!WARNING]
> The comment repository used to carry an in-code comment asserting the freeze **was** old-engine parity, on purpose. That was mistaken — the timestamp bump happens one layer above the repositories, in the shared change-builder, so it is invisible from either repository's code. Don't "restore parity" by removing the pass-through; the divergence was reproduced on both engines and the old engine does advance the field.

<details>
<summary><strong>What was wrong</strong></summary>

The shared change-builder (`getActualChanges`) injects `modifiedAt: now` into every non-empty change set for any resource that declares the field — upstream of both database engines. The old engine's repositories persist whatever the change set contains, so the injected timestamp lands without ever being named in repository code. The Postgres repositories enumerate their columns explicitly (`body` for comments; `type`, `shareability`, `body` for posts) and silently dropped it.

Both update methods now pass the injected timestamp through. No schema change — the `modified_at` columns always existed; they just were never written after creation.

</details>

<details>

`yarn lint` and `yarn type-check` clean. The comment and post e2e suites pass on **both** engines (3/3 each); their update flows now assert `modifiedAt` advances past its creation value, an expectation that runs identically on both engines since the fixed behavior is parity.

Re-planting the omission — dropping the two pass-through lines — turns exactly the two new assertions red on Postgres and nothing else, and re-greens on revert.

</details>

<details>
<summary><strong>Limitations</strong></summary>

Sorting comment lists by `modifiedAt` is still unwired on Postgres (the key is silently ignored while the old engine sorts by it) — a separate, already-recorded finding, not addressed here.

</details>


bryanjnelson (approved):
LGTM

PR #3933, opened 2026-09-22
Add `yarn pg:restore` to load a scrubbed dump into local Postgres

rdonigian:
> [!IMPORTANT]
> Loading the **scrubbed production dump** into local Postgres is now one command: `yarn pg:restore <path-to-dump>`. The script **drops and recreates the local `cord` database** before restoring, so it works on a fresh checkout and for reloading a newer dump later, where the old hand-typed `pg_restore` failed once the app had created tables. Local development only; nothing here runs in production.

<details>
<summary><strong>What the script does and why</strong></summary>

The database name and user come from the compose file, so the dump path is the only input. Before restoring it starts the postgres service if needed and waits for it to accept connections.

The `pg_restore` flags are unchanged from the README's previous manual command.
</details>

<details>

- Shell syntax check passes; yarn hands the path argument through.
- No argument prints a usage line and exits 2. A missing file names it and exits 2.
- The connection guard was read against the local database (0 sessions) but the drop, recreate, and restore path was not exercised end to end in this change; the `pg_restore` invocation is the README's existing command verbatim.
</details>


brentkulwicki:
My big takeaway here is I think the logic here is good, but we need to wire this up with infra to run the dump-creation logic inside an ECS task and store the output in S3, rather than this being a manual/local step.

☝️ Does this sound good to you? I can work on that mechanism today.

PR #3935, opened 2026-09-22
Restore global search parity with Neo4j after the Postgres cutover

rdonigian:
> [!IMPORTANT]
> Global search stopped finding most things when production moved to Postgres, and this restores it.
> Neo4j indexed **every stored string**, so a project's **department ID**, a partner's **PMC entity
> code**, a location's **ISO code** and a dozen other fields were all searchable. The Postgres
> implementation listed only `name`-like columns, so those queries returned **"No results found"** —
> which is how this was reported. Matching also regressed in ways that had nothing to do with which
> columns were listed: a person's **full name** found nothing, because the first and last name live in
> separate columns and neither contains `"First Last"`. Same for words typed out of order, hyphenated
> names typed with a space, and **accented names** searched without the accents.
>
> Search now matches the way it used to, and orders results by **relevance** instead of by date.

```mermaid
flowchart LR
  Q["query<br/><i>Fn Ln</i>"] --> W["split into words<br/><b>Fn</b> · <b>Ln</b>"]
  R["row's searchable columns"] --> H["haystack<br/><code>unaccent(concat_ws(' ', …))</code><br/><i>once per scanned row</i>"]
  W --> M{"any word<br/>in the haystack?"}
  H --> M
  M -->|no| X["not a hit"]
  M -->|yes| S["score<br/>1000 exact id<br/>+100 whole query contiguous<br/>+10 per word matched"]
  S --> O["order by score, then newest, then id<br/>keep the best 100"]
  O -.-> P["read-permission check<br/>on the field that matched"]
  style X stroke-dasharray: 4 3
```

*Solid is what this PR introduces; the dashed step downstream is unchanged. Previously the whole query
string was compared against each column on its own, which is why `Fn Ln` matched neither the first-name
column nor the last-name one.*

> [!WARNING]
> - **The migration must run before the new code serves traffic.** Every search comparison calls
>   `unaccent()`, which does not exist until the migration enables the extension. Migrations run at
>   startup, so a normal deploy is fine; a code-only rollout would make every search fail.
> - **Enabling the extension needs `rds_superuser`.** It is on the RDS allowlist. If the role lacks it
>   the migration throws and the app will not boot, so this fails loudly at deploy rather than quietly
>   at runtime.
> - **Results change for everyone at deploy.** Searches return more rows than yesterday and in a
>   different order. That is the point, but support should not be surprised by it.

<details>
<summary><strong>What changed</strong></summary>

Each row is reduced to one "haystack": its searchable columns joined with a space and accent-stripped
once, built in a lateral join so it is computed a single time per scanned row rather than once per
column per word. The query is split on whitespace and a row matches if **any** word appears in that
haystack, or if the id matches exactly.

Words are OR'd rather than AND'd because that is what Neo4j did — searching a name plus a typo still
finds the name. That is only tolerable because results are ranked now; without ranking the junk would
be indistinguishable from the real matches.

Matching against one combined value is what lets a query span two columns, and it is why the accent
folding is affordable. It does cost per-column attribution, which the read-permission check depends on,
so the columns are still tested individually — but in the SELECT list, where the work is proportional
to hits rather than to table size.

Fields restored: organization address; partner PMC entity code and address; language pronunciation, ROD
code, sign-language code, least-of-these reason and tags; project department ID, Rev79 ID and tags;
location ISO alpha-3; user phone, title and about; product title, description, completion description,
placeholder description and unspecified scripture book; periodic report skipped reason. The two `tags`
columns are arrays, so they are flattened to text rather than compared directly.

</details>

<details>
<summary><strong>Ranking, and why there is no index</strong></summary>

Neo4j returned hits already ordered by relevance, so its 100-row ceiling took the best 100. Postgres had
no score, so the ceiling was taking the 100 newest. The score here reproduces the **shape** of the
Lucene query the old code built, not its arithmetic: an exact id wins, the whole query matched
contiguously beats the same words scattered, and more words beat fewer. Ties fall back to newest and
then id so the order is fully deterministic. There is deliberately no inverse-document-frequency term
and no field-length normalization; add them only if ranking quality is ever the actual complaint.

Generated `tsvector` columns with GIN indexes were the obvious alternative and were dropped on
measurement. Against a copy of production data:

- the text-searched tables hold about **18,000 rows** in total, the largest being projects at **5,312**
- the query the search page actually sends runs in **38 ms**
- products (**81,057** rows) adds **153 ms** and periodic reports (**211,987**) adds **47 ms**, and only
  when those types are requested — the search page never asks for them

Generated columns and indexes across thirteen tables would have been a large migration against live
production rows for no measurable gain. If it is ever needed, `pg_trgm` GIN supports substring matching
directly, but it cannot be applied over `unaccent()` without an immutable wrapper that asserts the
dictionary will never change, and the index silently rots if it does. That trap is written into the
migration.

</details>

<details>
<summary><strong>Schema</strong></summary>

One statement, no table, column, index or constraint:

```sql
CREATE EXTENSION IF NOT EXISTS "unaccent";
```

It takes no lock on any existing relation and rewrites no rows, so it is safe against live production
regardless of table size. The extension installs into `public`, which is already on the search path.

Numbered `0002`, journal `when` of `1790035200000`. `IF NOT EXISTS` is load-bearing beyond idempotence:
anyone who enabled the extension by hand while testing would otherwise hit
`ERROR: extension "unaccent" already exists` on their next boot.

`unaccent` normalizes Latin script only. It folds diacritics, the Nordic slashed and ringed letters, the
German sharp s and the underdot and macron marks used in language documentation. Cyrillic, Japanese and
other scripts pass through untouched.

</details>

<details>
<summary><strong>Deliberate differences from the old behavior</strong></summary>

Three places where matching Neo4j exactly would have been worse, each easy to reverse:

- **Enum values are not searched.** Neo4j indexed project step, status and type, so `"Active"` returned
  every active project. That noise would crowd real matches out of the 100-row ceiling. A user's
  timezone is excluded for the same reason: `"America"` would match thousands of people.
- **Substrings inside a word match.** Lucene has no leading wildcard, so Neo4j could only match whole
  words and prefixes. This finds strictly more, never less.
- **A whitespace-only query returns nothing.** Neo4j returned six arbitrary rows, because its trailing
  wildcard matched everything and the ceiling truncated the result.

The last two are asserted only under Postgres, so the Neo4j test job does not fail on behavior it never
had.

Known limitation: the 100-row ceiling is still applied before the read-permission check, so a requester
who can read very little can still see fewer than their requested count. That predates this change.

</details>

<details>

The search spec grew from 8 cases to 35 and runs on both engines. Every field and behavior case was
confirmed to pass on Neo4j **before** being added, so each one documents a real regression rather than a
new aspiration.

- Postgres: **35 of 35 pass**
- Neo4j: **33 pass, 2 skipped** — the two Postgres-only improvements above
- Negative control: **25 of the 35 fail** against the current `develop` implementation
- Unit tests **225 pass**; language, partner and location specs **34 pass** as a migration smoke test
- `eslint` zero warnings, `type-check` exit 0

Checked against real production data rather than only fixtures: searching a genuine department ID
returns the one project, scored 110, with the matched field correctly attributed to `departmentId` —
which is what the read-permission check consumes.

One caveat on the green check: the Postgres end-to-end job is currently `continue-on-error`, so it is
advisory and only the Neo4j job blocks a merge. Worth knowing when weighing how much the CI result says
about a Postgres-only change. Flipping that is a one-line change and belongs in its own PR.

</details>


bryanjnelson (approved):
Very nice.

